PR #1089, opened 2022-10-07
Store StableID only in model store tags

### LINKED ISSUE #1085: optimize serialized format of models
The model store can shave a few bytes off each model by not storing the StableID in the serialized JSON and just placing it in the model store tags. The StableID is already present in the tags, the system just needs to keep from serializing the StableID

ashmrtn:
## Description

Not sure if this is too aggressive of an optimization. If it is we can just close this PR

Stable ID is used as a lookup key which means the model store stores it in its tags. Due to that it doesn't necessarily need to be serialized into the JSON of the struct as well. Code that reads the model will repopulate the field properly prior to returning the read data to the caller.

As this change does not modify the external behavior of the models, is contained completely in the model store, and the old behavior is benign (i.e. deserializing a model that has the ID stored in it changes nothing) it's alright to not bump the model version if we don't want to

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1085

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ashmrtn:
closing for now. Can revisit/reopen if we want this optimization

Issue #902, opened 2022-09-19
Load Testing


ryanfkeepers:
# Overview
E2E testing currently utilizes a small and highly curated (if still external and living) m365 account.  It should be assumed to operate on an (effectively) single user who has a minimal number of items (emails, events, contacts, files, etc) and folders, and for each of those to contain a small quantity of data.  Likewise, each test is expected to restrict the data they touch as much as possible.  These expectations are in place to ensure that our CI test pipeline can be ran quickly and efficiently.

These conditions do not represent real world environments.  Corso will be expected to succeed at long-running processes that process large quantities of data, within organizations that contain large numbers of users.  In order to maintain assurance about success under those conditions, a suite of tests involving large-scale operations is required.

# Expectations
- [x] Create a new suite of tests in which the datasets involve a large organizations whose users have many items, and/or very large items.
- [x] This new suite of tests will not run at the same cadence as our regular CI pipeline.  A nightly or other kind of rolling test would be preferred.
- [ ] Failures in these tests should alert the team, and be treated as production bugs.
- [ ] An alternative azure account may be preferred to keep standard CI expectations protected from the larger scale of this suite (and vice versa).
- [x] Test results should initially be (manually) tracked for review: https://www.notion.so/alcion/1ced2fe56ba34a34b4b76bb71de5095a?v=993cae9d617a41f196c00b68f698245a
- [x] Build a factory generator for producing mass amounts of data.  Also to be used for rolling additions/changes to data (to simulate change over time.)
- [x] Hook up an open telemetry recorder (suggested: aws xray) to gather spans/metrics during load tests.  Should also be compliant with external sdk user's otel aggregations.

PR #1111, opened 2022-10-10
Add AWS X-Ray support

ryanfkeepers:
## Description

Add aws xray observability tooling, plus ways to quickly and
easily run the tools/load tests.  Next steps are 1/ running the
daemon container alongside the load tests within github, and
2/ creating client credentials in aws's xray service so that
gathered trace info is propagated from the daemon proxy
to the xray service.

## Type of change

- [x] :robot: Test

## Issue(s)

* #902

## Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -10,6 +10,7 @@ import (
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
+	D "github.com/alcionai/corso/src/internal/diagnostics"
 	"github.com/alcionai/corso/src/internal/observe"
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/control"
@@ -95,7 +96,8 @@ func restoreCollection(
 	deets *details.Details,
 	errUpdater func(string, error),
 ) (support.CollectionMetrics, bool) {
-	defer trace.StartRegion(ctx, "gc:oneDrive:restoreCollection").End()
+	ctx, end := D.Span(ctx, "gc:oneDrive:restoreCollection", D.Label("path", dc.FullPath()))
+	defer end()
 
 	var (
 		metrics    = support.CollectionMetrics{}
@@ -228,7 +230,8 @@ func restoreItem(
 	driveID, parentFolderID string,
 	copyBuffer []byte,
 ) (*details.OneDriveInfo, error) {
-	defer trace.StartRegion(ctx, "gc:oneDrive:restoreItem").End()
+	ctx, end := D.Span(ctx, "gc:oneDrive:restoreItem", D.Label("item_uuid", itemData.UUID()))
+	defer end()
 
 	itemName := itemData.UUID()
 	trace.Log(ctx, "gc:oneDrive:restoreItem", itemName)
--- a/src/internal/diagnostics/diagnostics.go
+++ b/src/internal/diagnostics/diagnostics.go
@@ -107,10 +107,15 @@ func Label(k string, v any) extender {
 
 // Adds a Span to the given context.  Spans may be extended with indexes
 // for filtering and grouping, or with labels for contextual info.
-func Span(ctx context.Context, name string, ext ...extender) (context.Context, func()) {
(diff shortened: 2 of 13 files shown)


src/internal/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,160 @@
...
+Currently using AWS x-ray for observability:
+https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-annotations
+
+runtime/trace is also collected for load_test metrics gathering.
+*/
+
+// Initialize trace and span collection and emission.
+// Should be called as an initialization
+func InitCollector() error {
+	// TODO: ignore configuration if observability tooling (such as the x-ray daemon)
```
vkamra, line 28: Just to make sure I understand how things get initialized - is this TODO needed here or is this controlled by whoever calls `InitCollector`?

src/internal/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,160 @@
...
+	cfg := xray.Config{
+		DaemonAddr:     "0.0.0.0:2000",
+		ServiceVersion: "3.3.5",
+	}
+
+	if err := xray.Configure(cfg); err != nil {
+		return errors.Wrap(err, "initializing observability tooling")
+	}
+
+	xray.SetLogger(xraylog.NewDefaultLogger(os.Stderr, xraylog.LogLevelInfo))
```
vkamra, line 39: TODO: Let's plugin our logger here eventually https://github.com/aws/aws-xray-sdk-go#documentation

src/internal/connector/onedrive/restore.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
+	D "github.com/alcionai/corso/src/internal/diagnostics"
```
vkamra, line 13: Is there precedent for using an import alias like `D`? I haven't seen this before.

src/internal/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return nil
+}
+
+// Start kicks off a parent segment for tracking.  Start should only be called
+// internally, and only once per corso execution.  SDK users may provide contexts
+// with existing segments rather calling Start.
+func Start(ctx context.Context, name string) (context.Context, func()) {
```
vkamra, line 47: Nit - document what the caller should do with the return values.

src/internal/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,160 @@
...
+func Label(k string, v any) extender {
+	return metadata{k, v}
+}
+
+// Adds a Span to the given context.  Spans may be extended with indexes
+// for filtering and grouping, or with labels for contextual info.
+// Named variable returns are necessary here to prevent nil responses
+// during panic handling.
+func Span(ctx context.Context, name string, ext ...extender) (_ctx context.Context, _fn func()) {
+	// spans created without an existing parent segment in the ctx will panic.
```
vkamra, line 113: Instead of the recover - Why not add a check and make this a no-op if there's no parent segment in ctx?

vkamra (approved):
Some questions but otherwise looks good

src/internal/diagnostics/diagnostics.go
ryanfkeepers: Ah, I *think* I had that wrong.  IIRC, sdk consumers should be initializing and running their own otel collector, and shouldn't need to call this initializer.  The comment here imagined this running outside of the load tests (maybe by building it into the deployment, or some such) much like our metrics.  Doesn't seem like a point of concern at this time.

src/internal/connector/onedrive/restore.go
ryanfkeepers: Something I've personally done before when a common library has a longer name.  No strong feelings about it.

src/internal/diagnostics/diagnostics.go
ryanfkeepers: Ah, meant to do that and forgot.  Thanks for the reminder.

PR #1331, opened 2022-10-25
Replace src/ README

ntolia:
## Description

The src README is being included with our binary artifacts. We should either prevent that
or replace it with something more relevant to an end-user.

## Type of change

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,20 +1,3 @@
-# SRC Directory
+# Corso documentation
 
-## /pkg
-API and Components which are exposed for external usage.
-
-* /pkg/repository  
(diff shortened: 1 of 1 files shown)


ntolia:
I don't have strong opinions on this patch. I think the cleaner thing is to NOT distribute the README but...

ryanfkeepers (approved):

ryanfkeepers:
Would it be better to remove this file altogether?

ntolia:
> Would it be better to remove this file altogether?

I am happy to remove it. I don't know if it will break anything though. @meain?

ntolia:
Closing in favor of https://github.com/alcionai/corso/pull/1337

PR #1337, opened 2022-10-25
Do not add README to the released archives

vkamra:
## Description

Overrides default goreleaser behavior and skips adding any files to the archive other than the binary

See https://goreleaser.com/customization/archive/#packaging-only-the-binaries

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor


src/.goreleaser.yaml
```diff
@@ -28,6 +28,8 @@ archives:
     format_overrides:
       - goos: windows
         format: zip
+  - files:
```
meain, line 31:
```suggestion
    files:
```
A small typo

meain (approved):

Issue #1216, opened 2022-10-19
Selectors: match folder by parentRef entry


ryanfkeepers:
## Issue
Much like how items can be selected in `details` and `restore` commands by either the itemID or the shortRef value, folders should be accessible by either the folder name or the parentRef property.

## Acceptance criteria
Within the selector reduction path processor, add `parentRef` comparison for non-root, non-leaf categories.  Comparison should use a logical _or_ to match on either the folder name (current handling) or full equality on the parentRef.

ashmrtn:
Filling in some more details: from out of band discussion with some other folks it sounds like giving a folder reference should behave like giving a folder name in the sense that it should do prefix matching on the path the folder represents

PR #1341, opened 2022-10-26
Check if a string matches shortref pattern

ashmrtn:
## Description

Function to see if a given string matches the pattern of a shortref. Can help infer if something is a shortref or not.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1216 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/path/path.go
+++ b/src/pkg/path/path.go
@@ -39,6 +39,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


src/pkg/path/path.go
```diff
@@ -96,6 +99,25 @@ type Path interface {
 	ToBuilder() *Builder
 }
 
+// MaybeShortRef takes a stirng and returns true if this string could possibly
+// be a ShortRef. Otherwise it returns false. A return value of true does not
+// absolutely mean the string is a ShortRef returned by some item, only that it
+// fits the pattern ShortRefs use.
```
ryanfkeepers, line 105:
```suggestion
// MaybeShortRef takes a string and returns true if this string complies with
// the ShortRef format. Otherwise it returns false. A return value of true does not
// guarantee that the string is a ShortRef, nor that it represents an existing item.
```

ryanfkeepers (approved):

PR #1342, opened 2022-10-26
Function to get only folders from backup details

ashmrtn:
## Description

Mirror of Items() function but returns only folders entries.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :computer: CI/Deployment
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1216 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #1343, opened 2022-10-26
1216 exchange folder refs

ashmrtn:
## Description

Translate exchange folder ShortRef into a folder path that can be used in a selector.

#1341 is cherry-picked into this branch as the first commit so it compiles, but will be removed once it merges

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1216 

merge after:
* #1342 
* #1341 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/exchange.go
+++ b/src/pkg/selectors/exchange.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strconv"
 
+	"github.com/pkg/errors"
+
 	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/filters"
@@ -638,6 +640,15 @@ func (s ExchangeScope) setDefaults() {
 // ---------------------------------------------------------------------------
 // Backup Details Filtering
 // ---------------------------------------------------------------------------
+
+func (s exchange) FolderRefToPath(ref string, deets *details.Details) (string, error) {
+	p, err := folderRefToPath(ref, deets)
+	if err != nil {
+		return "", errors.Wrap(err, "resolving Exchange folder ref")
+	}
+
+	return p.Folder(), nil
+}
 
 // Reduce filters the entries in a details struct to only those that match the
 // inclusions, filters, and exclusions in the selector.
--- a/src/pkg/selectors/selectors.go
+++ b/src/pkg/selectors/selectors.go
@@ -31,6 +31,8 @@ var serviceToPathType = map[service]path.ServiceType{
 var (
 	ErrorBadSelectorCast = errors.New("wrong selector service type")
 	ErrorNoMatchingItems = errors.New("no items match the specified selectors")
+	ErrorNotShortRef     = errors.New("string doesn't match ShortRef pattern")
+	ErrorNotFound        = errors.New("no matching item found")
 )
 
 const (
@@ -117,6 +119,33 @@ func (s Selector) String() string {
 	}
 
 	return string(bs)
+}
+
+// folderRefToPath takes a string that may be a ShortRef and returns a path.Path
+// if it matches a folder in deets. If ref does not match the pattern for
+// a ShortRef or it does not match any ShortRef in deets returns an error.
+func folderRefToPath(ref string, deets *details.Details) (path.Path, error) {
+	if !path.MaybeShortRef(ref) {
+		return nil, ErrorNotShortRef
+	}
+
+	for _, folder := range deets.Folders() {
+		if folder.ShortRef != ref {
+			continue
+		}
+
+		p, err := path.FromDataLayerPath(folder.RepoRef, false)
+		if err != nil {
+			// Either we were given a ref for something further up the hierarchy than
+			// the folder itself or the path is malformed (i.e. bad service/category
+			// pair etc).
+			return nil, errors.Wrapf(err, "resolving folder ShortRef %s", ref)
+		}
+
+		return p, nil
+	}
+
+	return nil, ErrorNotFound
 }
 
 // appendScopes iterates through each scope in the list of scope slices,
--- a/src/pkg/path/path.go
+++ b/src/pkg/path/path.go
@@ -39,6 +39,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -57,6 +58,8 @@ var charactersToEscape = map[rune]struct{}{
 	pathSeparator:   {},
 	escapeCharacter: {},
 }
+
+var shortRefRegex = regexp.MustCompile("^[a-fA-F0-9]*$")
 
 var errMissingSegment = errors.New("missing required path element")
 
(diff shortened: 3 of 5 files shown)


src/pkg/selectors/selectors.go
```diff
@@ -119,6 +121,33 @@ func (s Selector) String() string {
 	return string(bs)
 }
 
+// folderRefToPath takes a string that may be a ShortRef and returns a path.Path
+// if it matches a folder in deets. If ref does not match the pattern for
+// a ShortRef or it does not match any ShortRef in deets returns an error.
+func folderRefToPath(ref string, deets *details.Details) (path.Path, error) {
```
ryanfkeepers, line 127: Not sure I agree on the return values here.  Two reasons: 1/ an error implies that the caller either has some control over the input (and thus can fix it) or is required to back out of processing when the error arises.  The former isn't true (whichever controller calls this func is just feeding it end-user data), and the latter isn't wanted (no reason to back out of the operation here).  I think we can drop the error case, as it doesn't aid processing here.  But then we'll also want point 2.

2/ Returning a string, instead of a path, is more resilient for this case.  Consider that end users can only provide one of two inputs: a short ref, or an actual folder name/path.  If the input is not a ref, it must be an actual name/path.  By returning a string you can return either the matched path.Folder(), or the original string input, and the caller should be able to use the result without knowledge of the presence or absence of transformation (primarily because we're only doing prefix matching, which is okay).

There is a corner case here, which is: what if the user provides a valid shortRef string that matches no details entries?  Currently this returns a "NotFound" error.  In that case, according to the above suggestion, we'd return the original input again (the short ref), and it will (probably) match nothing in the reduction step, causing that selector to be ignored.  This is the same handling that occurs if the user provides, say, a typo ("inboz"), or other non-matching input.  By deferring to the reducer in this way, it also resolves another corner case in the current implementation: that an input can pass maybeShortRef without being a short ref, but will currently produce a "not found" error.

src/pkg/selectors/exchange.go
```diff
@@ -639,6 +641,15 @@ func (s ExchangeScope) setDefaults() {
 // Backup Details Filtering
 // ---------------------------------------------------------------------------
 
+func (s exchange) FolderRefToPath(ref string, deets *details.Details) (string, error) {
```
ryanfkeepers, line 644: I know yesterday we talked about the controller packages that call into selectors.Reduce call this package.  But in thinking about it further that approach is beginning to smell a little.  It requires the caller to fetch details before building the scopes that reduce the details, which feels off.

At the start of the Reduce call there's already a set of pre-processing being done to massage inputs into a more manageable state.  I think you could run a map over the scopes at that stage to transform all the folder-level scopes as needed.  Would only require a single pass over the details folder entries (o(n*m) for needing to compare each folder scope to a detail entry, but 'm' is pretty well guaranteed to be small), and then those changes would cascade onto the rest of the reduce.  Thoughts?

src/pkg/selectors/selectors.go
ashmrtn: > 1/ an error implies that the caller either has some control over the input (and thus can fix it) or is required to back out of processing when the error arises

I think an error can help shed light on the specific sort of failure that occurred, not just imply caller control or change in execution flow. However, given that you point out a corner case later on, I think it would be alright to have a single return value if that fit better overall

> 2/ Returning a string, instead of a path

Yea, I thought about returning a string too. This function is internal to the package though so the exported functions that call it can choose to munge the return values. The reason it's returning `path.Builder` instead of `string` is that OneDrive will need to remove the first three elements of the path. That's best done with some functions from the `path` package because it's already split the path. Returning a string would require re-parsing to the `path` package to remove leading elements

src/pkg/selectors/exchange.go
ashmrtn: Doing all this in a cleaner fashion sounds great. I know I picked this issue up, but I'm not sure I actually have the knowledge of the selectors package to figure out the part of the code to modify based off the description you gave. You'll either have to point me to where the code needs modified (hopefully it's more of lift'n'shift than anything) or, if you have bandwidth, you can finish this PR series/issue off

src/pkg/selectors/selectors.go
ryanfkeepers: > OneDrive will need to remove the first three elements of the path
Hmm, fair, fair.  Will be an interesting condition to comply with.

src/pkg/selectors/exchange.go
ryanfkeepers: > point me to where the code needs modified

Happily!  Around about [here](https://github.com/alcionai/corso/blob/main/src/pkg/selectors/scopes.go#L237) is where my mind is at.  For each set of scopes (excludes, filters, includes), we'd do a pre-pass like:
```go
for _, ent := range deets.Folders() {
  for _, s := range s.Excludes|Filers|Includes {
    scat := s.Category()
    if !scat.isLeaf() && scat != scat.RootCat() {
      // check folderRefToPath
      // if is ref to path, update (a copy of?) the scope to filter on the found path
    }
  }
}
```
Thoughts?

ashmrtn: hmmm thanks for the pointer.

Looking at the referenced code, a couple of thoughts:
* we still need to iterate through details 2x, once to collect the folders and once to collect the items. This would be true for any solution of this form and is something I failed to consider when writing the original proposal for it (🙄). Given that, maybe the strawman would be good enough for now? I really don't know
* embedding the resolution of folder ShortRefs in the reduce operation may make it difficult to determine what went wrong if nothing matches. For instance, it's hard to say if there was a typo in the ref as they're more or less random hex strings. On the other hand, given that this may not be available as CLI behavior it's probably less of a concern, though still a point about debuggability overall

ryanfkeepers: > we still need to iterate through details 2x

```go
func (dm DetailsModel) FoldersItems() ([]*detailsEntry, []*detailsEntry)
```

> difficult to determine what went wrong if nothing matches

I think that is an unfortunately present problem independent of data sourcing.  We can only do so much to protect against typos.  It's whether we're opening ourselves up to an inability to distinguish a typo from a legitimate bug that that would concern me more.  And, I think a more centralized and standardized handling is the best defense against that possibility.

> maybe the strawman would be good enough for now? I really don't know

Neither do I.  But I want to try something (that's hopefully a bit of fun) to help us find out.  We don't often have the breathing room to compare and contrast approaches.  Would you be interested in each of us taking on the problem in each way: the strawman for you, the pre-processer for me?  Have a quick code jam and see which PR we like better?

ashmrtn: >Would you be interested in each of us taking on the problem in each way: the strawman for you, the pre-processer for me?

Sure, seems like it could be interesting. I'll spin up a bit of code with the strawman

ryanfkeepers: CODE JAM TIME

ashmrtn: alright, there's now have #1349 to compare against

PR #1355, opened 2022-10-27
GC: ContainerCache: IsNewCache Placement Fix

### LINKED ISSUE #1201: establish*RestoreLocation `isNewCache` check placement
## Issue
When restoring to an existing folder, the `isNewCache` check within the `establish*RestoreLocation` funcs occurs after an attempted folder creation.  This can lead to unnecessary errors when the folder already exists.

## Acceptance Criteria
Move the `isNewCache` handling above folder creation.

dadams39:
Rearranged `isNewCache` to be able to ensure that the cache is created prior to the first cache inspection. This will allow for restoring to an existing folder. 

## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* closes  #1201<issue>

## Test Plan

- [x] :zap: Unit test


ryanfkeepers:
Changes look good at face value.  Looks like it caused some test failures though.

dadams39:
> Changes look good at face value. Looks like it caused some test failures though.
Failures resolved locally with [1ff1b3b](https://github.com/alcionai/corso/pull/1355/commits/1ff1b3b3546adef466a3f23860ebcf930751599e)

src/internal/connector/exchange/service_restore.go
```diff
@@ -522,6 +522,15 @@ func establishMailRestoreLocation(
 	folderID := rootFolderAlias
 	pb := path.Builder{}
 
+	// Only populate the cache if we actually had to create it. Since we set
+	// newCache to false in this we'll only try to populate it once per function
+	// call even if we make a new cache.
+	if isNewCache {
+		if err := mfc.Populate(ctx, folderID, ""); err != nil {
```
ashmrtn, line 529: Just a minor note, feel free to override me on this. This will cause the system to fetch all folders for mail (and the other types) as it's using the default root folder. This is fine, but not quite as efficient if all we really want is the folders from the top-level restore folder down

However, if the root folder isn't used then there's a bit of a chicken and egg problem. That's because the folder ID for the root of the cache is required to populate the cache. However, we won't know the ID of the top-level restore folder until we either create it or check if it exists

Again, this is just a question of efficiency, so it can be ignored if that's best in the long-run. I think this solution is probably a bit better suited for the case where we allow restoring to an existing folder though

dadams39: I had a similar thought. This would enable `RestoreToExistingLocation` which in essence is anywhere in the Ecosystem. We would be creating a larger lookup table as a tradeoff. @ryanfkeepers Do you have any feelings on this?

ryanfkeepers: As a sort of middle ground, is there a way we could restrict that population to only a certain category of folder?  "populate, starting from this folder (the root), only mail|contact|event-system folders"?

dadams39: I wouldn't call that a middle ground. The search trees generated by the populate function are dynamic.
In the case of Restore, the original intention was that `Corso` would restore to a "new" folder at the beginning of every restore operation. The previous optimization was to use a `do-while` style of approach that ensured that `Corso` would not find the folder within the cache as it was not already populated. 

From that space, the Restore folder would be used as the base of the  Search tree. All following queries would search only the "restore portion" of the search space. For Contacts, the restore folder is created the first time within the cache and all following restores are sent to that one folder. 

As far as population, in general, Events already ignores the rootID and base due to its unique structure.

dadams39:
After vote, this PR is to be closed until a different restore policy is implemented.

Issue #1133, opened 2022-10-11
Selector minor QoL improvements


ryanfkeepers:
## Issue
Aggregating some minor quality-of-life functions that could be useful for consumers of selectors.  The goal is minimize consumer code duplication (ex, writing the same interpreter wrappers within every graph connector service package), without bloating the selectors package either.

## Wants
- [x] `sel.Reduce()` func that can be called from a `selectors.Selector`, rather than the service instance.  This prevents operations and other packages from needing to cast the selector to a service instance just for this call.
- [x] expand `scope.Match()` to support any-match on a variadic input.  See: https://www.notion.so/alcion/Recursive-Folder-Discovery-for-Testing-f7cf3d12a50f4ed9995069e903987393#16ec3334e35b4bb6a73e58acf67d5708
- [x] `scope.PathCategory()` func that returns a path category based on the scope leaf type.

PR #1356, opened 2022-10-27
Add qol generic selector Reduce func

ryanfkeepers:
## Type of change

- [x] :sunflower: Feature
- [x] :hamster: Trivial/Minor

## Issue(s)

* #1133

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/pkg/selectors/selectors.go
+++ b/src/pkg/selectors/selectors.go
@@ -97,19 +97,6 @@ func newSelector(s service) Selector {
 	}
 }
 
-// Any returns the set matching any value.
-func Any() []string {
-	return []string{AnyTgt}
-}
-
-// None returns the set matching None of the values.
-// This is primarily a fallback for empty values.  Adding None()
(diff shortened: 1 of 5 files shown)


src/pkg/selectors/selectors.go
```diff
@@ -190,6 +177,36 @@ func (s Selector) PathService() path.ServiceType {
 	return serviceToPathType[s.Service]
 }
 
+type reducer interface {
+	Reduce(ctx context.Context, deets *details.Details) *details.Details
+}
```
ashmrtn, line 182: I believe this is already defined towards the top of the file as an exported interface

ryanfkeepers: haha, right you are, thanks for the reminder.  I kicked one of those out.

ashmrtn (approved):
just some updates in `selector_reduce_test.go` to use `reducer`

Issue #547, opened 2022-08-15
Add support for kopia-assisted incremental backups


ashmrtn:
Getting incremental backups in corso requires a few changes to kopia itself. This tracks all the sub-issues required to get incremental backups enabled.

- [x] #512 
- [x] #513
- [x] #514 
- [x] #515 
- [x] #615 
- [x] #621
- [x] #622
- [x] #1404
- [ ] Delay fetching item data from Graph API until it's clear kopia will be uploading the item

ashmrtn:
closing for now as there are proposals for other changes that would affect how and when item information is fetched

PR #1405, opened 2022-10-31
Set file mod time in KopiaWrapper

### LINKED ISSUE #621: modify corso's kopia wrapper code to set mod time on `StreamingFile`s
This will allow kopia to skip uploading files if their modified time matches that of a previous backup

ashmrtn:
## Description

Set the mod time of uploaded files to either the mod time of the item (if it has one) or the current time (if it does not have one). Also add tests to check caching in kopia works properly.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #621 

part of:
* #547 

merge after:
* #1427 
* #1430

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/data/data_collection.go
+++ b/src/internal/data/data_collection.go
@@ -2,6 +2,7 @@ package data
 
 import (
 	"io"
+	"time"
 
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/path"
@@ -45,6 +46,11 @@ type StreamInfo interface {
 // information about the Stream
 type StreamSize interface {
 	Size() int64
+}
+
+// StreamModTime is used to provide the modified time of the stream's data.
+type StreamModTime interface {
+	ModTime() time.Time
 }
 
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/internal/data/data_collection.go
```diff
@@ -47,6 +48,11 @@ type StreamSize interface {
 	Size() int64
 }
 
+// StreamModTime is used to provide the modified time of the stream's data.
+type StreamModTime interface {
```
ryanfkeepers, line 52: Nit: slightly more in keeping with go interface naming convention.
```suggestion
// ModTimer is used to provide the modified time of the stream's data.
type ModTimer interface {
```

ryanfkeepers (approved):

src/internal/data/data_collection.go
ashmrtn: hmmm actually now that I look at it more, can we hold off on this. I wanted to keep it consistent with some of the other interfaces defined in the file (`StreamSize`, `StreamInfo`). I think either we should change all of them in a different patch or leave them all consistency

PR #1408, opened 2022-10-31
Tag kopia snapshots for later lookups

ashmrtn:
## Description

Add tags to each kopia snapshot that include all service/category pairs in the snapshot and all resource owners in the snapshot. This allows future snapshots to lookup existing snapshots by those tags so they can be fed into the snapshot function. Feeding previous snapshots into the snapshot function enables kopia to detect previously uploaded files and skip uploading the data again

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1404 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -389,26 +389,34 @@ func newTreeMap() *treeMap {
 	}
 }
 
-// inflateDirTree returns an fs.Directory tree rooted at the oldest common
-// ancestor of the streams and uses virtualfs.StaticDirectory for internal nodes
-// in the hierarchy. Leaf nodes are virtualfs.StreamingDirectory with the given
-// DataCollections.
+// inflateDirTree returns a set of tags representing all the resource owners and
+// service/categories in the snapshot and a fs.Directory tree rooted at the
+// oldest common ancestor of the streams. All nodes are
+// virtualfs.StreamingDirectory with the given DataCollections if there is one
+// for that node. Tags can be used in future backups to fetch old snapshots for
+// caching reasons.
 func inflateDirTree(
 	ctx context.Context,
 	collections []data.Collection,
 	progress *corsoProgress,
-) (fs.Directory, error) {
+) (fs.Directory, map[string]struct{}, map[string]struct{}, error) {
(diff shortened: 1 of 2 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -497,22 +505,48 @@ func (w Wrapper) BackupCollections(
...
-	s, err := w.makeSnapshotWithRoot(ctx, dirTree, progress)
+	s, err := w.makeSnapshotWithRoot(ctx, dirTree, serviceCats, resourceOwners, progress)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	return s, progress.deets, nil
 }
 
+func serviceCatTag(s path.ServiceType, c path.CategoryType) string {
```
ryanfkeepers, line 521: api here seems unnecessarily one step distanced.  Any reason to to take `serviceCatTag(path)`?  That would be consistent with all of your invocations.

src/internal/kopia/wrapper.go
```diff
@@ -389,26 +389,34 @@ func newTreeMap() *treeMap {
...
+// oldest common ancestor of the streams. All nodes are
+// virtualfs.StreamingDirectory with the given DataCollections if there is one
+// for that node. Tags can be used in future backups to fetch old snapshots for
+// caching reasons.
 func inflateDirTree(
 	ctx context.Context,
 	collections []data.Collection,
 	progress *corsoProgress,
-) (fs.Directory, error) {
+) (fs.Directory, map[string]struct{}, map[string]struct{}, error) {
```
ryanfkeepers, line 402: This feels a little cluttered.  Thoughts on passing around an internal struct instead?  Ex:
```go
type ownersCats struct {
  resourceOwners map[string]struct{}
  serviceCats map[string]struct{}
}
```
```suggestion
) (fs.Directory, ownersCats, error) {
```

ryanfkeepers (approved):

PR #1423, opened 2022-11-02
Always validate building docs in CI for PRs

meain:
## Description

We can probably remove the requirement on `Publish-Docs` as we have move the check to `Docs-Linting`. Having this check in `Publish-Docs` does not run for docs only change as the previous steps(Test-Suite) are skipped and as a result this gets skipped.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/ci.yml
```diff
@@ -97,6 +121,23 @@ jobs:
...
+        env:
+          CORSO_VERSION: ${{ needs.SetEnv.outputs.version }}
+        run: |
+          cd docs &&
+            npm ci &&
+            npm run build
+
+      - uses: actions/upload-artifact@master
+        # uncomment next line to disable creating docs artifacts on PR runs
+        # if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'
```
vkamra, line 134: @meain - should we do uncomment?

vkamra (approved):

.github/workflows/ci.yml
meain: I was kinda undecided on that. As it does not get pushed anywhere for PRs, I thought having this in the artifacts will let us validate the docs that was built in CI in case we wanted to.

vkamra: I see - let's keep it

meain:
@vkamra Now that we moved the build step to `Docs-Linting` we can probably remove the requirement on `Publish-Docs` for merge. *PS: It won't cause any issues even if we leave it there as it will be skipped which is treated as success.*

ntolia:
I believe we need to revert or fix this. See https://github.com/alcionai/corso/actions/runs/3383428890/jobs/5619320099

Issue #1224, opened 2022-10-19
Filters - add PathContains and PathPrefix comparators


ryanfkeepers:
## Issue
The standard `Prefix` and `Contains` filter comparators are too coarse-grained for usable functionality when comparing path structures.  This causes implementations throughout the app to mediate between the state of data and the usage of the filters or scopes that are compared to said state.  These adhoc wrappers are inconsistent and are likely to increase corner case issues, rather than resolve them.

## Proposal
Introduce two new filter comparators: PathContains and PathPrefix.  These comparators would differ from other filers in two meaningful ways:

1.  They would rely on `pkg/path` during comparison.  Ex: given `pathContains("foo").Matches("bar/foo/baz")`, the filter would call pkg/path to split apart the provided path, and compare each element within to the target "foo".  This would be in place of a standard `strings.Contains(path, "foo")`.
2. They would accept a list of strings in their constructor, rather than a single string, and internally run an any-match on each value in the constructor during comparison.  Ex: given `pathPrefix(["Inbox", "Inbox/foo"]).Matches("Inbox/foo/bar")`, the comparison would return true if either of the target value match the provided path.

These introductions would improved ease of use in the following conditions:
1. Reducing folders in selectors could be easily and reliably differentiated between prefix and contains matchers.
2. Reducing folders in selectors would no longer need to extract the filter target to build a reverse filter by splitting on that value.
3. Backup creation could compare against `qp.Scope.Matches(folderType, somePathString)` without needing to transform, index, or otherwise massage the path input.

ryanfkeepers:
See conversation in https://github.com/alcionai/corso/issues/1146 for behavioral requirements.

PR #1425, opened 2022-11-02
add path prefix/contains filters

ryanfkeepers:
## Description

Adds two specialized filters: path prefix and path
contains.  These filters differ from prior additions
in two ways: 1/ they accept a slice of inputs and
perform an implicit any-match when compared,
2/ they have particular rules about matching
completely within delimited elements.

In the next PR, these filters will replace the
standard prefix and contains comparisons that
are currently used in folder selection.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1224

## Test Plan

- [x] :green_heart: E2E


src/pkg/filters/filters.go
```diff
@@ -24,21 +27,48 @@ const (
...
+	TargetPathContains
 )
 
 func norm(s string) string {
 	return strings.ToLower(s)
 }
 
+// normPathElem ensures the string is:
+// 1. prefixed with a single os path delimiter (ex: `/`)
+// 2. suffixed with a single os path delimiter (ex: `/`)
```
ashmrtn, line 44: nit: saying `os path delimiter` makes it sound platform specific. This *should not* be platform specific, it should always be `/`
```suggestion
// 1. prefixed with a single path delimiter (ex: `/`)
// 2. suffixed with a single path delimiter (ex: `/`)
```

src/pkg/filters/filters.go
```diff
@@ -24,21 +27,48 @@ const (
...
+		return s
+	}
+
+	if s[0] != path.PathSeparator {
+		s = string(path.PathSeparator) + s
+	}
+
+	if s[len(s)-1] != path.PathSeparator {
+		s += string(path.PathSeparator)
+	}
```
ashmrtn, line 59: this may not always do the "right" thing. Things get tricky if the input has an escaped `/` at the end (i.e. `foo\/`). In that case, the last character is the path separator, but it's not acting as a path separator. However, just checking the last 2 characters is also insufficient because there could be any number of escaped `\` characters as in `foo\\\\\/`

It would probably be easiest to just do `s = path.TrimTrailingSlash(s)+path.PathSeparator`

Also, trailing `/` is currently being trimmed off by the CLI [here](https://github.com/alcionai/corso/blob/b5a274600818a075a48f862c7430769fcebad1fc/src/cli/utils/flags.go#L62) ([PR](https://github.com/alcionai/corso/pull/1175)). If we're always adding a `/` here we could probably ditch the CLI code (but do keep the CLI tests in `cli/backup/(onedrive|exchange)_test.go`)

src/pkg/filters/filters.go
```diff
@@ -133,9 +163,102 @@ func NotPrefix(target string) Filter {
...
 
+// PathPrefix creates a filter where Compare(v) is true if
+// target.Prefix(v) &&
+// split(target)[i].Equals(split(v)[i]) for _all_ i in 0..len(target)-1
+// ex: target "/foo/bar" returns true for input "/foo/bar/baz",
+// but false for "/foo/barbaz"
+//
+// Unlike single-target filters, this filter accepts a
+// slice of targets, and implicitly handles comparisons as
+// an `any-match` check (ie: Compare will return true if any
```
ashmrtn, line 174: nit: maybe be less confusing to use `logical OR` instead of `any-match` since `any` is often used for `*` in selectors

src/pkg/filters/filters.go
```diff
@@ -133,9 +163,102 @@ func NotPrefix(target string) Filter {
...
+func PathContains(targets []string) Filter {
+	tgts := make([]string, len(targets))
+	for i := range targets {
+		tgts[i] = normPathElem(targets[i])
+	}
+
+	return newSliceFilter(TargetPathContains, tgts, false)
+}
+
+// PathContains creates a filter where Compare(v) is true if
```
ashmrtn, line 225:
```suggestion
// NotPathContains creates a filter where Compare(v) is true if
```

src/pkg/filters/filters.go
```diff
@@ -215,18 +365,67 @@ func prefixed(target, input string) bool {
 	return strings.HasPrefix(input, target)
 }
 
+// true if target is an _element complete_ prefix match
+// on the input.  Element complete means we do not
+// succeed on partial elementt matches (ex: "/foo" does
```
ashmrtn, line 370:
```suggestion
// succeed on partial element matches (ex: "/foo" does
```

src/pkg/filters/filters.go
```diff
@@ -215,18 +365,67 @@ func prefixed(target, input string) bool {
...
+// Basic principle: For any normalized, well formed target
+// and input (ex: values like "/a/b/c/"), the target is a
+// prefix if the regex `^target` matches the input.
+func pathPrefix(target, input string) bool {
+	re, err := regexp.Compile("^" + target)
+	if err != nil {
+		return false
+	}
+
+	return re.MatchString(normPathElem(input))
```
ashmrtn, line 388: since the regex pattern is fixed to the norm-ed CLI input with the exception of `^` can't this be simplified as `strings.HasPrefix(normPathElem(input), target)`?

src/pkg/filters/filters.go
```diff
@@ -215,18 +365,67 @@ func prefixed(target, input string) bool {
...
+// Basic principle: For any normalized, well formed target
+// and input (ex: values like "/a/b/c/"), the input contains
+// the target if the regex `target` matches the input.
+func pathContains(target, input string) bool {
+	re, err := regexp.Compile(target)
+	if err != nil {
+		return false
+	}
+
+	return re.MatchString(normPathElem(input))
```
ashmrtn, line 412: this seems like it simplifies to `strings.Contains(normPathElem(input), target)`?

ashmrtn (approved):
little confused by the use of regex for path prefix/contains as it seems like `strings` package functions should have equivalent behavior. Unless we plan on some day supporting regex matches? That seems like it could be addressed when we get to that point instead of adding it now?

src/pkg/filters/filters.go
ryanfkeepers: Fair point, fixed up the language a little.

src/pkg/filters/filters.go
ryanfkeepers: tt strikes again!

src/pkg/filters/filters.go
ryanfkeepers: I suppose it does.

src/pkg/filters/filters.go
ryanfkeepers: Good call, sticking with the trim-then-add.

Per the cli behavior, I'm fine keeping that in-tact.  Cleaning inputs is good for downstream handling, and filters will want to ensure things are normalized either way.

ashmrtn: heh yea, it was a bit of a trip figuring out all the edge cases for trim when writing the path package 🙃

PR #1427, opened 2022-11-02
Find previous kopia snapshots for given set of tags

ashmrtn:
## Description

Helper functions that allow finding the most recent complete
(and maybe incomplete) snapshots for a given set of tags. The
returned snapshots can later be used to do more efficient kopia
snapshots by allowing kopia to determine it's already uploaded
a file and skip uploading it again.

The number of most recent snapshots to return can be tuned (right
now it returns 1 most recent), but may cause more memory usage
during backups.

Kopia currently has some oddities when getting snapshot
manifests via tags:
* tag values cannot be empty or the comparison returns always
  true, selecting all snapshot manifests
* kopia does not currently tag snapshot manifests made during
  checkpoints. A patch to upstream kopia is needed to fix this.
  All other manifests that match the tags will be returned though

Added code is not currently connected to any backup logic.

Viewing by commit may help as there was a tad of lift'n'shift code
movement for the sake of organization

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1404 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+const (
+	// Kopia does not do comparisons properly for empty tags right now so add some
+	// placeholder value to them.
+	defaultTagValue = "0"
+	// Kopia CLI prefixes all user tags with "tag:"[1]. Maintaining this will
```
ryanfkeepers, line 19: Nit: easier reading.
```suggestion
	defaultTagValue = "0"

	// Kopia CLI prefixes all user tags with "tag:"[1]. Maintaining this will
```

src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+type ownersCats struct {
+	resourceOwners map[string]struct{}
+	serviceCats    map[string]struct{}
+}
+
+func serviceCatTag(p path.Path) string {
+	return p.Service().String() + p.Category().String()
+}
+
+func makeTagPair(k string) (string, string) {
```
ryanfkeepers, line 44: Nit: "pair" can sound like you're producing two tags (ie: two keyed entries or some similar) from the one input.  Largely in the case that a reader doesn't already assume a "tag" to imply a k:v map entry.
```suggestion
func makeTagKV(k string) (string, string) {
```

src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+	return -1
+}
+
+// findRecentManifests searches through mans and returns the most recent
+// complete manifest (if one exists) and maybe the most recent incomplete
+// manifest. If the newest incomplete manifest is more recent than the newest
+// complete manifest then adds it to the returned list. Otherwise no incomplete
+// manifest is returned. Returns nil if there are no complete or complete
+// manifests in mans.
+func findRecentManifests(
```
ryanfkeepers, line 100: Nit: the comment does a good job explaining the behavior.  Still, the function name itself is easily misinterpreted.
```suggestion
func manifestsSinceLastComplete(
```

src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Manifests should maintain the sort order of the original IDs that were used
+	// to fetch the data, but just in case sort oldest to newest.
+	mans = snapshot.SortByTime(mans, false)
+
+	for i := len(mans) - 1; i >= 0; i-- {
+		m := mans[i]
+
+		if len(m.IncompleteReason) > 0 {
+			if !foundIncomplete {
```
ryanfkeepers, line 116: No possibility of having multiple incomplete snapshots?  Or is there a conflict that occurs if this includes multiple incompletes?

src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+	for i := lastCompleteIdx + 1; i < len(metas); i++ {
+		ids = append(ids, metas[i].ID)
+	}
+
+	mans, err := sm.LoadSnapshots(ctx, ids)
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching previous manifests")
+	}
+
+	return findRecentManifests(mans), nil
```
ryanfkeepers, line 180: on line 160, you return `nil` if the most recent snapshot is complete; ie, not returning a manifest.  Here, you're returning (potentially) the most recent complete manifest plus the most recent incomplete one.  Feels like a bit of a discrepancy.  What is the reason to not return the recent completed manifest on 160?

src/internal/kopia/snapshot_manager.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func fetchPrevSnapshotManifests(
```
ryanfkeepers, line 183: This doesn't seem to be getting used, except in tests.  I assume it'll roll forward into some future case.  For now, a comment would help relate the use case.

ryanfkeepers (approved):

src/internal/kopia/snapshot_manager.go
ashmrtn: it's possible, but unclear how much having multiple would help. There's a bit of tuning we can do with that going forward. Essentially, more snapshots returned from this whole set of functions is more chances to match against previously uploaded data. However, it also means that kopia will pull more directory manifests from remote storage and have more state in memory (potentially one directory manifest for each manifest returned)

Also if a backup is always building off of a previous one (or the changes between files in a kopia backup are minimal, excluding new files) having multiple previous snapshots doesn't help much because the most recent one will cover most things already

src/internal/kopia/snapshot_manager.go
ashmrtn: ah, so there's two parts to this function:
1. see if we've already got a cached version of the manifest in `foundMans`
2. fetch newer manifests if we don't have the most recent manifest cached

Line 160 handles the case where we had a cached manifest _and_ it was the most recent. If we don't hit that then this will fetch all manifests newer than the cached one we found and try to find a complete manifest and/or an incomplete manifest

src/internal/kopia/snapshot_manager.go
ryanfkeepers: > Also if a backup is always building off of a previous one (or the changes between files in a kopia backup are minimal, excluding new files) having multiple previous snapshots doesn't help much because the most recent one will cover most things already

Ah, I was curious if it had that sort of rolling progress.  Good to know.

src/internal/kopia/snapshot_manager.go
ryanfkeepers: > we had a cached manifest and it was the most recent

So in this case the caller is expected to already know, of the cached manifests that it has, which is the most recent?

ashmrtn: mmm I suppose? I would view this more as a helper function that's only going to be called by a single caller (the one that's calling it now). All it needs to do is return any currently uncached manifests to the caller. The caller will add the returned manifests to its set and eventually those are all passed to kopia which doesn't really care if it's only the most recent manifests or not

Given that the manifests are being added to a map anyway, I could return the cached manifest on line 160, but it doesn't really matter one way or the other in the long run

ryanfkeepers: okedoke, that makes sense

PR #1434, opened 2022-11-03
Revert "Always validate building docs in CI for PRs"

ntolia:
Reverts alcionai/corso#1423


meain:
I think it would be better to enable building cli docs instead https://github.com/alcionai/corso/pull/1435

Issue #1366, opened 2022-10-28
Expose more information about the backed up items


rajatkateja:
In addition to the already present information in `ItemInfo, it would be useful to have the following information about the backed up items. 

Across all items:

- createdDateTime 
- modifiedDateTime
- accessedDateTime
- size (i.e., the actual size of the item)
- storedSize (i.e., how much space does this item take in the backup)

For OneDriveItem type
- owner

For ExchangeEvent type:
- eventDuration

For the common information, happy to discuss any nuances, e.g., some that I can think of: 
- for received emails, created and modified date times be same as received date time. same for accessed date time (unless there is some way to know when a user opens an email)
- for drafts, there can be separate created and modified date times? Although I wonder what is the received date time for drafts?
- for contacts, events, not sure what the accessed date time should be
- for folders, the size and storedSize should be the aggregate of their entries.

vkamra:
@meain - should start with OneDrive (owner) (see https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/item.go#L69) for how OneDriveInfo is populated. 

This information can then be viewed via the `corso backup details onedrive --backup <id>` command

meain:
Maybe I'm missing something. I could not find the `backup details` command is hitting the function that you mentioned. I was able to track it down to https://github.com/alcionai/corso/blob/b5a274600818a075a48f862c7430769fcebad1fc/src/internal/kopia/model_store.go#L337 for loading which looks like it is doing a `json.Unmarshal` directly from kopia. I'm looking to how/where that information gets saved into kopia in the first place.

ashmrtn:
Some links/execution flow to fill in the missing pieces for storing the data:

* Kopia Wrapper is requesting the item info during a backup [here](https://github.com/alcionai/corso/blob/b5a274600818a075a48f862c7430769fcebad1fc/src/internal/kopia/wrapper.go#L338). It adds it to a temp buffer, waits for the item to be uploaded to kopia completely, and then [adds](https://github.com/alcionai/corso/blob/b5a274600818a075a48f862c7430769fcebad1fc/src/internal/kopia/wrapper.go#L197) it to the `details.Details` struct in a callback that kopia exposes

* The complete set of details is [returned](https://github.com/alcionai/corso/blob/b5a274600818a075a48f862c7430769fcebad1fc/src/internal/operations/backup.go#L161) from KopiaWrapper to the backup operation when the backup is complete

[...]

PR #1436, opened 2022-11-03
Store/Retrieve owner info for OneDrive files

meain:
## Description

Store and retrieve owner info for onedrive files.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -67,11 +67,13 @@ func driveItemReader(
 // doesn't have its size value updated as a side effect of creation,
 // and kiota drops any SetSize update.
 func driveItemInfo(di models.DriveItemable, itemSize int64) *details.OneDriveInfo {
-	email := ""
 	ed, ok := di.GetCreatedBy().GetUser().GetAdditionalData()["email"]
+
+	email := ""
 	if ok {
 		email = *ed.(*string)
 	}
+
 	return &details.OneDriveInfo{
 		ItemType: details.OneDriveItem,
 		ItemName: *di.GetName(),
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -67,12 +67,14 @@ func driveItemReader(
 // doesn't have its size value updated as a side effect of creation,
 // and kiota drops any SetSize update.
 func driveItemInfo(di models.DriveItemable, itemSize int64) *details.OneDriveInfo {
+	adata := di.GetCreatedBy().GetUser().GetAdditionalData()
 	return &details.OneDriveInfo{
 		ItemType: details.OneDriveItem,
 		ItemName: *di.GetName(),
 		Created:  *di.GetCreatedDateTime(),
 		Modified: *di.GetLastModifiedDateTime(),
 		Size:     itemSize,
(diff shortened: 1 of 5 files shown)


src/internal/connector/onedrive/item.go
```diff
@@ -67,12 +67,20 @@ func driveItemReader(
 // doesn't have its size value updated as a side effect of creation,
 // and kiota drops any SetSize update.
 func driveItemInfo(di models.DriveItemable, itemSize int64) *details.OneDriveInfo {
+	ed, ok := di.GetCreatedBy().GetUser().GetAdditionalData()["email"]
+
+	email := ""
+	if ok {
+		email = *ed.(*string)
+	}
+
```
dadams39, line 76: I believe this is incorrect. Perhaps:
```go
	user  := di.GetCreatedByUser()
	if user != nil{
             
                //  Need to verify that principal name is not empty
		ptr := user.GetUserPrincipalName()
                if ptr != nil{
                          email = *ptr
                          }
	         }
```

src/pkg/backup/details/details.go
```diff
@@ -367,21 +367,22 @@ type OneDriveInfo struct {
...
+	Owner      string    `json:"owner,omitempty"`
 	Created    time.Time `json:"created,omitempty"`
 	Modified   time.Time `json:"modified,omitempty"`
 }
 
 // Headers returns the human-readable names of properties in a OneDriveInfo
 // for printing out to a terminal in a columnar display.
 func (i OneDriveInfo) Headers() []string {
-	return []string{"ItemName", "ParentPath", "Size", "Created", "Modified"}
+	return []string{"ItemName", "ParentPath", "Size", "Owner", "Created", "Modified"}
```
ryanfkeepers, line 378: Does this change need to be reflected in any docs?

vkamra: I think we're good - I don't see any reference to details anywhere.

vkamra (approved):
@meain - should be good to merge after the changes @dadams39 suggested are made. Approving so that we can move forward after that.

src/pkg/backup/details/details.go
gmatev: The doc examples are for exchange so this is ok

src/internal/connector/onedrive/item.go
meain: Let me look into it more, but `GetCreateByUser` was always `nil` when I tested it.

meain: `GetCreateByUser` seems to be always `nil` but we have data in `GetCreatedBy`. I have made this change locally, but thought I would check once  again with you before pushing it upstream.
/cc @vkamra

vkamra: Let's move forward with this. Why do we need to use `GetAdditonalData` though? What does `GetUser()` return?

meain: `GetUser` only has `DisplayName`.

PR #1441, opened 2022-11-03
Add a make clean step for the website

ntolia:
## Description

`make clean` will now be executed before a Docker image is built. While it technically has no impact on the Docker image, it will ensure that there are no stale Node modules lying around when a build is executed after the creation of a new image.

While running a Docker build without a cache is an additional option, it does not prevent stale local node_modules directories and suffers from the overhead of having to pull down a large Ubuntu base image every time.

## Type of change

- [x] :bug: Bugfix

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -1,9 +1,9 @@
-.PHONY: buildimage dev shell build publish sync
+.PHONY: buildimage dev shell build publish sync clean
 
 WEBC := docker run --rm -it --init -p 5050:3000 -p 5051:3001 \
 		-v ${PWD}:/usr/src/website corso/website:latest
 
-buildimage:
+buildimage: clean
 	docker build -t "corso/website:latest" .
 
 dev:
@@ -23,3 +23,6 @@ publish: build
 sync:
 	aws s3 sync /usr/src/website/dist/ s3://corsobackup.io/ --exclude ".git/*" --delete
 	aws cloudfront create-invalidation --distribution-id E1W9NGI9YTVZ1A --paths "/*"
+
+clean:
+	rm -rf node_modules


website/Makefile
```diff
@@ -1,9 +1,9 @@
-.PHONY: buildimage dev shell build publish sync
+.PHONY: buildimage dev shell build publish sync clean
 
 WEBC := docker run --rm -it --init -p 5050:3000 -p 5051:3001 \
 		-v ${PWD}:/usr/src/website corso/website:latest
 
-buildimage:
+buildimage: clean
 	docker build -t "corso/website:latest" .
```
gmatev, line 7:
```suggestion
	docker build --no-cache -t "corso/website:latest" .
```
It is longer but this is not a common step. I've ran into issues where some state around the modules was not properly rebuilt even if I nuked the modules like `clean` would do.

gmatev:
@ntolia saw the full description now. I don't know if `clean` alone is enough. I think some of the structures of modules are installed gets persisted in the image. I've ran into issues in the past with rebuild even if I deleted node_modules

ntolia:
> @ntolia saw the full description now. I don't know if `clean` alone is enough. I think some of the structures of modules are installed gets persisted in the image. I've ran into issues in the past with rebuild even if I deleted node_modules

For that corner-case, I would recommend doing a `docker rmi <image-id> && docker system prune -f`. I mean we can do this (it is only 80MB for Ubuntu) in a separate PR but, ideally, CI will just generate "build" images stored in GHR every time the Dockerfile or package.json changes and/or on releases.

Finally, I think both changes are orthogonal and I don't object to a separate PR for the no-cache Docker build option if you feel we should have it.

meain:
The `node_modules` consumed by gulp is not the same as the one we are deleting. It is outside the website dir within the container [ref](https://github.com/alcionai/corso/blob/d2965e0b09a5f10992da51d12627fbcb1b026768/website/Dockerfile#L27). But this change should be good on its own.

ntolia:
> The `node_modules` consumed by gulp is not the same as the one we are deleting. It is outside the website dir within the container [ref](https://github.com/alcionai/corso/blob/d2965e0b09a5f10992da51d12627fbcb1b026768/website/Dockerfile#L27). But this change should be good on its own.

There really shouldn't be node_modules in the website directory but it can get accidentally left behind when we upgrade or add new packages manually. I believe we should merge this and work on the other issue in parallel.

meain (approved):

PR #1442, opened 2022-11-03
fix span reporting null check

ryanfkeepers:
## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :muscle: Manual


ashmrtn (approved):
I hate it when I do this

ryanfkeepers:
> I hate it when I do this

Don't we all.  And yet, we keep doing it... 🤔

PR #1443, opened 2022-11-03
applies folder matching behavior to cli input

ryanfkeepers:
## Description

Allows users to specify whether they want to
select folders by prefix or search behavior.
Search/contains behavior is the default case,
with prefix being an optional deviation if the
folder input is prepended with a '/' character.

Also, propagates the PrefixMatch setting to
all integration tests that rely on selecting only
the default folder in exchange.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1224

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/utils/utils.go
+++ b/src/cli/utils/utils.go
@@ -8,7 +8,9 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
+	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
 // common flag names
@@ -63,4 +65,35 @@ func AddCommand(parent, c *cobra.Command) (*cobra.Command, *pflag.FlagSet) {
 	c.Flags().SortFlags = false
 
 	return c, c.Flags()
+}
+
+// separates the provided folders into two sets: folders that use a pathContains
+// comparison (the default), and folders that use a pathPrefix comparison.
+// Any element beginning with a path.PathSeparator (ie: '/') is moved to the prefix
+// comparison set.  If folders is nil, returns only containsFolders with the any matcher.
+func splitFoldersIntoContainsAndPrefix(folders []string) ([]string, []string) {
+	var (
+		containsFolders = []string{}
+		prefixFolders   = []string{}
+	)
+
+	if len(folders) == 0 {
+		return selectors.Any(), nil
+	}
+
+	// separate folder selection inputs by behavior.
+	// any input beginning with a '/' character acts as a prefix match.
+	for _, f := range folders {
+		if len(f) == 0 {
+			continue
+		}
+
+		if f[0] == path.PathSeparator {
+			prefixFolders = append(prefixFolders, f)
+		} else {
+			containsFolders = append(containsFolders, f)
+		}
+	}
+
+	return containsFolders, prefixFolders
 }
--- a/src/cli/utils/utils_test.go
+++ b/src/cli/utils/utils_test.go
@@ -1,4 +1,4 @@
-package utils_test
+package utils
 
 import (
 	"testing"
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 
-	"github.com/alcionai/corso/src/cli/utils"
+	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
 type CliUtilsSuite struct {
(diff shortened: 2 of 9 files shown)


src/cli/utils/utils_test.go
```diff
@@ -33,6 +33,52 @@ func (suite *CliUtilsSuite) TestRequireProps() {
...
+			input:   []string{"/a", "b", "/c"},
+			expectC: []string{"b"},
+			expectP: []string{"/a", "/c"},
+		},
+	}
+	for _, test := range table {
+		suite.T().Run(test.name, func(t *testing.T) {
+			c, p := splitFoldersIntoContainsAndPrefix(test.input)
+			assert.Equal(t, test.expectC, c, "contains set")
+			assert.Equal(t, test.expectP, p, "prefix set")
```
ashmrtn, line 81: nit: should probably use `ElementsMatch` as it's not sensitive to order
```suggestion
			assert.ElementsMatch(t, test.expectC, c, "contains set")
			assert.ElementsMatch(t, test.expectP, p, "prefix set")
```

src/cli/backup/exchange_integration_test.go
```diff
@@ -291,13 +291,16 @@ func (suite *PreparedBackupExchangeIntegrationSuite) SetupSuite() {
 
 		switch set {
 		case email:
-			scopes = sel.MailFolders([]string{suite.m365UserID}, []string{exchange.DefaultMailFolder})
+			scopes = sel.MailFolders([]string{suite.m365UserID}, []string{exchange.DefaultMailFolder}, selectors.PrefixMatch())
```
ashmrtn, line 294: hmmm so actually looking at this, we might need exact matches in cases like these. I believe GraphConnector is currently doing exact matches with the folder names (that's likely why we need to pass the full path of a subfolder in to get it to back it up), but changing to a prefix match is not equivalent in this case. Here we actually do just want to get the folder mentioned, but none of its subfolders. If we start pulling in subfolders via prefix matching I worry we may eventually time out backing up contacts because technically all contact folders are a child of the root contact folder

We do have some path munging in there for contacts folders I think (iirc they're flattened, which may need revisiting...?), but it's probably best not to rely on something like that

src/cli/utils/utils.go
```diff
@@ -64,3 +66,30 @@ func AddCommand(parent, c *cobra.Command) (*cobra.Command, *pflag.FlagSet) {
 
 	return c, c.Flags()
 }
+
+// separates the provided folders into two sets: folders that use a pathContains
+// comparison (the default), and folders that use a pathPrefix comparison.
+// Any element beginning with a path.PathSeparator (ie: '/') is moved to the prefix
+// comparison set.  If folders is nil, returns only containsFolders with any any matcher.
```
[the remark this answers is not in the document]
ryanfkeepers: You guessed correctly!

src/cli/utils/utils.go
```diff
@@ -64,3 +66,30 @@ func AddCommand(parent, c *cobra.Command) (*cobra.Command, *pflag.FlagSet) {
...
+	)
+
+	if len(folders) == 0 {
+		return selectors.Any(), nil
+	}
+
+	// separate folder selection inputs by behavior.
+	// any input beginning with a '/' character acts as a prefix match.
+	for _, f := range folders {
+		if len(f) > 0 && f[0] == path.PathSeparator {
```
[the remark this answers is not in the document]
ryanfkeepers: I suppose none of our users will intend to select on an empty string.  Mostly likely case is that this would happen due to a scripting failure, yeah?  `--foo-type a,,c`.  Skipping that value here is sensible, but also hides the issue from the user (potentially with unintended consequences).  I think I'll put this suggestion through, but also raise a ticket for pre-validation on these inputs as a separate step.

src/cli/backup/exchange_integration_test.go
ryanfkeepers: I don't think it's a cause for concern.  In researching these changes the folders we compare are all the first children of the root.  For example, here's the iteration for contacts for the noted test (first word is the category, second is the directory).  Same behavior goes for email and events.
```
contacts Contacts
contacts Corso_Restore_03-Nov-2022_20-39-32.251826
contacts Corso_Restore_03-Nov-2022_20-45-25.710348
contacts Corso_Restore_03-Nov-2022_21-09-07.307926
contacts Corso_Restore_03-Nov-2022_21-10-26.431697
contacts Corso_Restore_03-Nov-2022_21-05-14.498555
contacts Corso_Restore_03-Nov-2022_21-05-23.622714
contacts TrialSubFolder
contacts Corso_Restore_03-Nov-2022_20-37-32.867436
contacts TrialFolder
```

ashmrtn (approved):

PR #1446, opened 2022-11-04
Add EndTime to Exchange events

meain:
## Description

Add EndTime to Exchange events. Also fixes a testcase around start time.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/event_test.go
```diff
@@ -27,7 +31,7 @@ func TestEventSuite(t *testing.T) {
 // can be properly retrieved from a models.Eventable object
 func (suite *EventSuite) TestEventInfo() {
 	initial := time.Now()
-	now := common.FormatTime(initial)
+	now := initial.Format(timeFormat)
```
meain, line 34: Necessary because of https://github.com/alcionai/corso/blob/d2965e0b09a5f10992da51d12627fbcb1b026768/src/internal/connector/exchange/event.go#L40

meain:
@vkamra We had a discussion on slack around adding both end time and duration assuming that it would help for recurring events. The event duration for recurring events will be same as the one when calculated with the end date mentioned here and so I have not added it as it will be just duplicated info. But if we want to add the recurrence end date, we can add in `evt.GetRecurrence().GetRange().GetEndDate()` as a separate field.

meain:
@gmatev You had mentioned a possible doc change for exchange in https://github.com/alcionai/corso/pull/1436#discussion_r1013558408 but I could not find any places where we have to update this information.

src/internal/connector/exchange/event_test.go
ryanfkeepers: This interpretation should get added to `common/time.go` parsing, instead of distributing its handling across multiple files.

ryanfkeepers (approved):
Approved for functionality, but wouldn't mind seeing the time handling move into common, as suggested.  (a follow-up pr is also acceptable)

meain:
@ryanfkeepers I've moved the format to common and replaced the calls with `common.FormatTimeWith`. I didn't think creating a new util func was necessary, at least as of now. Let me know if you think we should do it and I can update that in a followup PR.

PR #1452, opened 2022-11-04
Add created and modified time to all

meain:
## Description

Adds Created and Modified time to all objects. Need to go after https://github.com/alcionai/corso/pull/1446 as it is built on top of that. Keeping in draft until then.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/contact_test.go
```diff
@@ -19,25 +20,39 @@ func TestContactSuite(t *testing.T) {
...
 	tests := []struct {
 		name         string
 		contactAndRP func() (models.Contactable, *details.ExchangeInfo)
 	}{
 		{
 			name: "Empty Contact",
 			contactAndRP: func() (models.Contactable, *details.ExchangeInfo) {
-				i := &details.ExchangeInfo{ItemType: details.ExchangeContact}
-				return models.NewContact(), i
+				var contact = models.NewContact()
```
ryanfkeepers, line 32: Nit: avoid `var` unless initializing zero-valued variables.
```suggestion
				contact := models.NewContact()
```

src/pkg/backup/details/details.go
```diff
@@ -302,20 +302,22 @@ type ExchangeInfo struct {
...
+	Modified    time.Time `json:"modified,omitempty"`
 }
 
 // Headers returns the human-readable names of properties in an ExchangeInfo
 // for printing out to a terminal in a columnar display.
 func (i ExchangeInfo) Headers() []string {
 	switch i.ItemType {
 	case ExchangeEvent:
-		return []string{"Organizer", "Subject", "Starts", "Ends", "Recurring"}
+		return []string{"Organizer", "Subject", "Starts", "Ends", "Recurring", "Created", "Modified"}
```
ryanfkeepers, line 314: Long term question: if/as the list of info for each data type continues to grow we'll be more likely to overextend the terminal line width and have a fairly ugly output.  Let's keep these values in for now, but also keep in mind that we might want to reduce the display later on.

ryanfkeepers (approved):

meain:
I'm looking to why the CI is failing. I checked the backup created separately and they do seem to have the created and modified times.

meain:
Seems like we are not storing information about the created time and last modified time within the backup for exchange files. I was able to track down creating a new object to be restored to [here](https://github.com/alcionai/corso/blob/490917782296b545d86ed55796a9a7714c0fcd25/src/internal/connector/exchange/service_restore.go#L165), but we don't seem to have the created and modified time available here. Same seems to be the case for other exchange items as well. The `corso details` command is able to show that info as that is pulled from the details metadata and not from the actual backup. I'm probably missing something obvious, but could someone point me to what I might be missing here.
/cc @ryanfkeepers

meain:
I'm merging this as the previous backups also was not aware of created/modified time and the old backup don't seem to have them. But let me know if someone has any pointers on how I should approach the restoring them properly.
Ref: https://github.com/alcionai/corso/pull/1452#issuecomment-1306902141

PR #1466, opened 2022-11-05
Corso Terminal demo

gmatev:
## Description

Adds Corso CLI demo to website. Includes instructions on how to use Terminalizer to generate the demo. 

Can and should investigate a better maintained project in the future. 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


gmatev:
@ntolia This is placement that I liked on the site but if you have a better suggestion let me know. 

One issue is that the GIF after many attempts to minimize it with various techniques is still 40MB which is not ideal.

ntolia:
40MB is a killer. I replaced it with an MP4. LMK what you think about the other PR based on this one. 

I don't know why the autoplay isn't working but I wanted to get this out there.

ntolia:
> 40MB is a killer. I replaced it with an MP4. LMK what you think about the other PR based on this one.
> 
> I don't know why the autoplay isn't working but I wanted to get this out there.

Also, if you can ship me the uncompressed GIF, the final video might be even better without a size increase.

ntolia (approved):

meain:
Just found out this PR. Should we consider not adding the mp4 files to git? After a few versions in, we might end up making the git object store pretty big. One option is git lfs, or we can generate that on the fly when publishing the website.

gmatev:
Generating on the fly will be pretty challenging, at least for now. There are a few manual steps described in the readme as part of the rendering process that will be tough to automate.

In my opinion, we can leave it as is for now. It will not change frequently and the mp4 is more reasonably sized.

Not sure if there is a way to cleanup the history and remove the GIF.




On Nov 9, 2022 05:28, Abin Simon ***@***.***> wrote:

Just found out this PR. Should we consider not adding the mp4 files to git? After a few versions in, we might end up making the git object store pretty big. One option is git lfs, or we can generate that on the fly when publishing the website.

meain:
Once commited to git, there is no way to clean it up. One pattern that I have seen folks do for content like this is to host in a different branch or upload it to an issue and then use the link GitHub generates in the project(if we don't want to host it separately).

ntolia:
There shouldn't be a GIF in git. The file never made it to the commit as it was a single commit (926ee7c6cd690ec8cb7db97e5009213414fec49a).
________________________________

ntolia:
Ah, let's figure it out if this file does see many revisions.

Niraj
________________________________

PR #1492, opened 2022-11-09
Update docs with new fields

meain:
## Description

Related: #569. Creating a separate PR as that one is already in queue. Keeping it in draft until that is merged.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/docs/.markdownlint.json
+++ b/docs/.markdownlint.json
@@ -1,6 +1,8 @@
 {
     "MD013": {
-        "line_length": 120
+        "line_length": 120,
+        "code_blocks": false,
+        "tables": false
     },
     "no-inline-html": {
         "allowed_elements": [
@@ -15,5 +17,5 @@
             "details",
             "summary"
         ]
-    },
+    }
 }
--- a/docs/docs/quickstart.md
+++ b/docs/docs/quickstart.md
@@ -237,9 +237,9 @@ The output from the command above should display a list of any matching emails. 
 of the email you would like to use for testing restore.
 
 ```text
-  ID            Sender                 Subject                                  Received
-  360bf6840396  phish@contoso.info     Re: Request for Apple/Amazon gift cards  2022-10-18T02:27:47Z
-  84dbad89b9f5  ravi@cohovineyard.com  Come join us!                            2022-10-19T06:12:08Z
+  ID            Sender                 Subject                                  Received              Created               Modified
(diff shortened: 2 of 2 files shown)


docs/.markdownlint.json
```diff
@@ -1,6 +1,8 @@
 {
     "MD013": {
-        "line_length": 120
+        "line_length": 120,
+        "code_blocks": false,
+        "tables": false
```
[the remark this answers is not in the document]
gmatev: If I understand correctly, this disables these overall. Is that necessary?

I thought we already had other way to handle code blocks and for lines, it really should only be an issue for links and that was working ok before.

docs/docs/quickstart.md
```diff
@@ -237,9 +237,9 @@ The output from the command above should display a list of any matching emails.
 of the email you would like to use for testing restore.
 
 ```text
-  ID            Sender                 Subject                                  Received
-  360bf6840396  phish@contoso.info     Re: Request for Apple/Amazon gift cards  2022-10-18T02:27:47Z
-  84dbad89b9f5  ravi@cohovineyard.com  Come join us!                            2022-10-19T06:12:08Z
+  ID            Sender                 Subject                                  Received              Created               Modified
```
gmatev, line 240: I don't think we should be displaying these fields (at least not be default) since it increases the chances of lines wrapping and the created and modified, while useful in the manifest, are not that meaningful to an end user to warrant sacrificing the horizontal screen real estate.

We can file a separate enhancement that would be for a 'wide' display option

gmatev (changes requested):

docs/.markdownlint.json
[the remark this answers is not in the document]
meain: Yes, this would disable it for all code_blocks and tables, but keeps the warning for normal lines.

docs/docs/quickstart.md
meain: Makes sense. I'll drop showing created/modified field for exchange items.

meain: Without the Created and Modified, there is no need for this PR. I'll close this as well.

meain:
https://github.com/alcionai/corso/pull/1492#discussion_r1018795411

Issue #1361, opened 2022-10-28
Output `--version` information to stdout instead of stderr


meain:
Noticed while working on https://github.com/alcionai/corso/pull/1360 that version is printed to stderr  instead of stdout.
https://github.com/alcionai/corso/blob/2a2af20f94db66095bd4315d07d7c714e9e1f205/src/cli/cli.go#L43

> Should we introduce a new `print.<something>`?

PR #1503, opened 2022-11-14
Print version info to stdout instead of stderr

meain:
## Description

Previously we were printing the version information with `corso --version` to stderr. This changes it to stdout.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/1361

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/print/print.go
```diff
@@ -89,6 +89,39 @@ func err(w io.Writer, s ...any) {
...
 
+// Print prints the params to cobra's output writer (stdOut by default)
+// if s is nil, prints nothing.
+func Print(ctx context.Context, s ...any) {
+	print(getRootCmd(ctx).OutOrStdout(), s...)
+}
+
+// print is the testable core of Print()
+//revive:disable:redefines-builtin-id
+func print(w io.Writer, s ...any) {
```
ryanfkeepers, line 100: Lets not redefine the builtin here (I recognize we were already doing that; should've renamed it then, too).  It's an unexported func, it doesn't need to comply with the other patterns exactly.

src/cli/print/print.go
```diff
@@ -89,6 +89,39 @@ func err(w io.Writer, s ...any) {
 	fmt.Fprint(w, msg...)
 }
 
+// Print prints the params to cobra's output writer (stdOut by default)
+// if s is nil, prints nothing.
+func Print(ctx context.Context, s ...any) {
```
ryanfkeepers, line 94: Not sure I'm a fan of the stutter (print.Print).  Not sure I have a good alternative, either.  What do you think of `Out` instead?
```suggestion
func Out(ctx context.Context, s ...any) {
```

meain:
@ryanfkeepers I have renamed it to `Out`, let me know if you have any other suggestions.

ryanfkeepers (approved):

PR #1504, opened 2022-11-14
fix: integration tests running on PRs from forks

uncledru:
## Description

* Allows integration tests to run via `repository_dispatch` events when PRs are opened from forks 

Credit: https://github.com/imjohnbo/ok-to-test

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [X] :computer: CI/Deployment
- [ ] :hamster: Trivial/Minor

## Issue(s)

* ALC-635

## Test Plan

- [X] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/ci.yml
```diff
@@ -137,11 +143,15 @@ jobs:
...
+  Test-Suite-Trusted:
     needs: [Precheck, Checkout]
     environment: Testing
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main' || needs.precheck.outputs.srcfileschanged == 'true'
+    if:
+      (startsWith(github.ref, 'refs/tags/') ||
+      github.ref == 'refs/heads/main' ||
+      needs.precheck.outputs.srcfileschanged == 'true') &&
+      github.event.pull_request.head.repo.full_name == github.repository
```
uncledru, line 154: This will only run when we have a non-fork PR

.github/workflows/ci.yml
```diff
@@ -173,7 +183,70 @@ jobs:
...
+      # Check out merge commit
+      - name: Fork based /ok-to-test checkout
+        uses: actions/checkout@v2
+        with:
+          ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
+
+      - name: Setup Golang with cache
+        uses: magnetikonline/action-golang-cache@v3
+        with:
+          go-version-file: src/go.mod
```
uncledru, line 227: @meain little confused here. Looks like we have the `Setup Golang with cache` step in the `Testing-Suite` jobs but also a step inside of `Checkout` to setup a golang cache. Do we need both?

uncledru:
Will need to merge and create a dummy PR to verify things are setup correctly.

ashmrtn:
We should discuss security implications of running CI for forks. There's a chance external users could try to leak secrets or other sensitive info if we always enable CI when a PR from a fork is made with no extra intervention/checks
@vkamra

.github/workflows/ci.yml
meain: To be frank I had blindly copied it from the previous implementation by @ryanfkeepers . I think we will need to retain it here, don't think it gets transferred over between jobs.

.github/workflows/ci.yml
```diff
@@ -132,16 +137,18 @@ jobs:
           name: docs
           path: docs/build
 
-
   # ----------------------------------------------------------------------------------------------------
   # --- Integration and Unit Testing -------------------------------------------------------------------
   # ----------------------------------------------------------------------------------------------------
 
-  Test-Suite:
+  Test-Suite-Trusted:
```
meain, line 144: Be default, for a PR from a fork, both will be in a skipped state which is treated as "success" for merge check by GitHub (in a case where we add both `Test-Sute-Trusted` and `Test-Sute-Fork` to the checks). This might cause aviator to merge even when we have not got the CI to pass.

There are two options that I can think to deal with this.

#### Option 1

We can have a single Test-Suite job with a step inside which will check if `!(github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'repository_dispatch')` and fail with an error message, something like "Need member approval to run this job.". This way, we can keep the current flow.

### Option 2

Have two jobs but have another job which just runs and returns true if one of these is a success. This way we can target that job to the mergeability check. One caveat is that we will have to make both of the Test-Suite branches fail/succeed (not skip) so that the dependent job can run.

.github/workflows/ci.yml
```diff
@@ -194,6 +201,98 @@ jobs:
...
+              ref
+            });
+            const check = checks.check_runs.filter(c => c.name === process.env.job);
+            const { data: result } = await github.rest.checks.update({
+              ...context.repo,
+              check_run_id: check[0].id,
+              status: 'completed',
+              conclusion: process.env.conclusion
+            });
+            return result;
```
meain, line 294: I don't know if this is feasible, but if yes, should we consider commenting the link to actual run on the PR? I have found https://github.com/marketplace/actions/sticky-pull-request-comment to be useful if we are going that route.

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  ok-to-test:
+    runs-on: ubuntu-latest
+    # Only run for PRs, not issue comments
+    if: ${{ github.event.issue.pull_request }}
+    steps:
+      # Generate a GitHub App installation access token from an App ID and private key
+      # To create a new GitHub App:
+      #   https://developer.github.com/apps/building-github-apps/creating-a-github-app/
+      # See app.yml for an example app manifest
```
meain, line 17: Just a nit, we can probably remove this line.

.github/workflows/ci.yml
uncledru: >  This might cause aviator to merge even when we have not got the CI to pass.

We shouldn't be auto-merging (aviator) PRs from forks right? 🤔 

Even if all tests pass...

meain: Yeah, I guess we shouldn't be auto merging. Now that you mention it, I can see an issue where we add the `mergequeue` label, somebody can push a new commit, on top of which aviator might merge master and it will cause it to run making the non-verified code run. That said, I think having this check could be helpful in even preventing us from accidentally merging it in.

uncledru: I agree. I can update the rule on the main branch to require the `Test-Suite-Fork` be required? So if it's in a skipped state and cannot be merged (can only do this after it is merged into `main`).

@ntolia can you confirm if aviators merge rules prevent it from auto-merging PRs from forks. I do not have access.

meain: @uncledru Aviator does not seem to differentiate between PRs from branches/vs forks. It just uses the GitHub merability check to decide if a PR should go in or not.

uncledru: Do you have access to `aviator`? Maybe we can add [merge rules](https://docs.aviator.co/reference/merge-rules) to prevent PRs from forks being auto-merged?

.github/workflows/ci.yml
ryanfkeepers: Each job runs in an isolated VM.  Caches you retrieved in initial jobs will not carry over to the later ones.  Similar to why we need to call the checkout action at the top of each job. 

On the topic, a note on the difference between the two: `./.github/actions/go-setup-cache` ensures we retrieve any modules missing from the cache, and will update the cache afterward (thus we only call it once, at the beginning).  `magnetikonline/action-golang-cache@v3` will read available caches, but won't update them.

.github/workflows/ci.yml
meain: Aviator is currently set to use the same rules as what GitHub uses to check for mergeability.

.github/workflows/ci.yml
meain: We recently merged a PR which enables separate caches for each step. We might wanna incorporate that changes here.
ref: https://github.com/alcionai/corso/pull/1532

uncledru: I see it in the PR here: https://github.com/alcionai/corso/blob/749dd6c1abc1f3818f2612de6ad97084b5bb2129/.github/workflows/ci.yml#L154

But it's not in main: https://github.com/alcionai/corso/blob/main/.github/workflows/ci.yml#L153

Am I missing something?

.github/workflows/ci.yml
uncledru: If you click "customize" does it give any options that could help us out here?

.github/workflows/ci.yml
meain: We can skip that. Another follow up PR removed it the next day 🤣 .
https://github.com/alcionai/corso/pull/1534

.github/workflows/ci.yml
meain: We can specify which jobs we want to check by adding more jobs to block on. As of now this list is synced with what is checked in GitHub. I don't have the access to GitHub settings but if you can update the list, it will update them here as well.

uncledru: Done: 

image

uncledru: I will remove the `Test-Suite` once this is merged.

uncledru: @meain I had to revert the required checks since it was stopping corso PRs from being mergable. After this is merged (approval still needed) - we can configure the repo and aviator to respect these new checks.

meain: The changes looks good. My concern here was weather we should consider making `Test-Sute-Fork` fail when it has not been approved so as to not accidentally merge it. As of not it is skipped when the PR is ran without approval and GitHub treats skipped and successful test in the same way.

uncledru: 🤔 if we make it a required check in GitHub Settings - it should be prevent the PR being merged even if its skipped. 

After the `ok-to-test` comment is added - Actions should update the check. 

Do you have a working example of what you're talking about? (allowing the PR to merge even if a required check is skipped)

uncledru: Never mind... https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

image

Any ideas/suggestions on making Test-Sute-Fork fail when it has not been approved?

uncledru: I'm thinking we can remove the `if` condition on the `Test-Suite-Fork` job and have it always run. 

The first step in the job could be a bash script that runs the same logic checks as the current `if`:

```yaml
    if: github.event_name == 'repository_dispatch' &&
      github.event.client_payload.slash_command.sha != '' &&
      contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.sha)
```

If this resolves to `false`... We exit the job with `exit 1` and it'll get registered as a failure which will prevent merging. 

```bash
exit 1
```

After a developer comments `ok-to-test` it will run again but this time via `workflow_dispatch`... the logical check above will evaluate to `true` and we won't exit early. 

Thoughts?

meain: Yup! That was more or less what I had in mind as well.

.github/workflows/ci.yml
```diff
@@ -194,6 +201,111 @@ jobs:
           if-no-files-found: error
           retention-days: 14
 
+  Test-Suite-Fork:
+    needs: [Precheck]
+    environment: Testing
+    if: ((startsWith(github.ref, 'refs/tags/') ||
+      github.ref == 'refs/heads/main' ||
+      needs.precheck.outputs.srcfileschanged == 'true') &&
+      github.event.pull_request.head.repo.full_name == github.repository) != true # skip when trusted workflow runs
```
uncledru, line 210: @meain let me know if you can think of a cleaner way to check this 🤔

.github/workflows/ci.yml
```diff
@@ -194,6 +201,111 @@ jobs:
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: src
+    steps:
+      - name: Fail check
+        if: ${{ (github.event_name == 'repository_dispatch' &&
+          github.event.client_payload.slash_command.sha != '' &&
+          contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.sha)) != true }}
+        run: exit 1
```
uncledru, line 220: @meain here we check if its NOT a repo_dispatch event - exit with an error

.github/workflows/ci.yml
meain: I think we can just check `needs.precheck.outputs.srcfileschanged == 'true' && github.event.pull_request.head.repo.full_name != github.repository` as we don't have to run this for `main` or tags.

.github/workflows/ci.yml
meain: Can we just simplify the check to `github.event_name != 'repository_dispatch'`? Also, maybe we can add a message in the `run` section so as to give a hint to contributors as to what is happening. Something like below:
```suggestion
        run: |
        	echo "Workflow requires approval from a maintainer to run. It will be automatically rerun on approval."
        	exit 1
```

meain (approved):

meain:
@uncledru You can add `mergequeue` label if you think this is good to go.

uncledru:
@meain it seems the `Test-Suite` job won't run on these changes which prevents me from merging 🤔

meain:
I think we will have to remove `Test-Sute` job and instead add the new jobs to "require" check on GH. Just notify folks on Slack about rebasing the changes from `main`.

Issue #1506, opened 2022-11-14
GC: Backup: SharePoint Libraries (aka drives) collection population


ryanfkeepers:
Populate backup collections for SharePoint libraries.

PR #1508, opened 2022-11-14
Adds the foundation of sharepoint data coll

ryanfkeepers:
## Description

boilerplate to graph connector.  To minimize
changes, some dependencies or half-baked
solutions have been listed as TODOs for follow-
up.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -89,18 +92,23 @@ func verifyBackupInputs(sel selectors.Selector, mapOfUsers map[string]string) er
...
-// to the GraphConnector struct.
-func (gc *GraphConnector) createCollections(
+// determines the type of collections that are retrieved.
+func (gc *GraphConnector) createExchangeCollections(
 	ctx context.Context,
 	scope selectors.ExchangeScope,
 ) ([]*exchange.Collection, error) {
-	var errs *multierror.Error
+	var (
+		errs           *multierror.Error
```
dadams39, line 107: For standardization, var errs should be changed to `error` in line with the other vars in the file.

src/internal/connector/data_collections.go
```diff
@@ -229,3 +241,109 @@ func (gc *GraphConnector) OneDriveDataCollections(
...
+			FailFast:    gc.failFast,
+			Credentials: gc.credentials,
+		}
+
+		itemCategory := qp.Scope.Category().PathType()
+
+		foldersComplete, closer := observe.MessageWithCompletion(fmt.Sprintf("∙ %s - %s:", itemCategory.String(), site))
+		defer closer()
+		defer close(foldersComplete)
+
```
dadams39, line 281: Is it possible that these should be deferred with one command rather than 2?

src/internal/connector/data_collections.go
```diff
@@ -229,3 +241,109 @@ func (gc *GraphConnector) OneDriveDataCollections(
...
+		resolver, err := exchange.PopulateExchangeContainerResolver(
```
dadams39, line 282: Unsure if this is the correct set-up. List and Drive items are different when it comes to paths. Need to ensure that the resolvers are also different for these two items.

src/internal/connector/graph_connector.go
```diff
@@ -191,6 +198,79 @@ func (gc *GraphConnector) GetUsersIds() []string {
 	return buildFromMap(false, gc.Users)
 }
 
+// setTenantSites queries the M365 to identify the sites in the
+// workspace. The sitets field is updated during this method
+// iff the return value is nil
+func (gc *GraphConnector) setTenantSites(ctx context.Context) error {
+	// TODO
+	gc.Sites = map[string]string{}
```
dadams39, line 206: Why would `Sites` become a visible field for GC?

ryanfkeepers: Sites are the acting Resource Owner for SharePoint application handling.  We don't catalog data by users, so there's not much point in retrieving them.  I'd say a more generic means of handling this would be to swap both Users and Sites for a `ResourceOwner` map and pivot on the type as needed, but I know ark also uses this to retrieve users, so I don't want to futz with that at this moment.

src/internal/connector/data_collections.go
ryanfkeepers: That came up during design of the observer package.  Long and short is we have some technical reasons to keep them separate.  Though we can always try to clean it up in a later refactor if/when desired.

src/internal/connector/data_collections.go
ryanfkeepers: Err handling in this file is all over the place.  I'll fix it up how you've proposed in a follow-up PR, to isolate the changes from this one.

src/internal/connector/graph_connector.go
```diff
@@ -97,6 +98,12 @@ func NewGraphConnector(ctx context.Context, acct account.Account) (*GraphConnect
 		return nil, errors.Wrap(err, "retrieving tenant user list")
 	}
 
+	// TODO: users or sites, one or the other, not both.
+	err = gc.setTenantSites(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "retrieveing tenant site list")
+	}
+
 	return &gc, nil
```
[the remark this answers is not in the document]
ryanfkeepers: I don't quite follow.  Are you looking for validation on the retrieved sites, or on something else?

src/internal/connector/graph_connector.go
```diff
@@ -191,6 +198,79 @@ func (gc *GraphConnector) GetUsersIds() []string {
...
+}
+
+// GetSites returns the siteIDs of sharepoint sites within tenant.
+func (gc *GraphConnector) GetSites() []string {
+	return buildFromMap(true, gc.Sites)
+}
+
+// GetSiteIds returns the M365 id for the user
+func (gc *GraphConnector) GetSiteIds() []string {
+	return buildFromMap(false, gc.Sites)
```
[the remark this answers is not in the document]
ryanfkeepers: I don't believe we're tracking the user at all (outside of details.Info).  The site is its own resource owner.

src/internal/tester/resource_owners.go
```diff
@@ -57,3 +57,14 @@ func LoadTestM365OrgUsers(t *testing.T) []string {
 
 	return strings.Split(users, ",")
 }
+
+// M365SiteID returns a siteID string representing the m365SiteID described
+// by either the env var CORSO_M365_TEST_SITE_ID, the corso_test.toml config
+// file or the default value (in that order of priority).  The default is a
+// last-attempt fallback that will only work on alcion's testing org.
+func M365SiteID(t *testing.T) string {
```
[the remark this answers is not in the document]
ryanfkeepers: I figure if our exchange/onedrive integration tests identify a user as the primary resource owner, and we're swapping the user for a site for sharepoint, then we would want to allow the caller to identify the site they're using as resource owner for the test suites in the same way.  Then again, I haven't gone through the motions of figuring out how many different sites can exist in an org as independent resource owners.  I'm assuming multiple.

dadams39 (approved):

PR #1511, opened 2022-11-14
reduce load tests, fix purge creation time

ryanfkeepers:
## Description

Removes the individual user load test set to
focus on backup-restore of the standard org.
Fixes the folder-creation time value provided to
the purge script from load tests.

## Type of change

- [x] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #902

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/load_test.yml
+++ b/.github/workflows/load_test.yml
@@ -3,8 +3,8 @@ on:
   schedule:
     # every day at 01:59 (01:59am) UTC
     # - cron: "59 1 * * *"
-    # temp, for testing: every 6 hours
-    - cron: "0 */6 * * *"
+    # temp, for testing: every 4 hours
+    - cron: "0 */4 * * *"
 
 permissions:
   # required to retrieve AWS credentials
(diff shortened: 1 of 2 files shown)


.github/workflows/load_test.yml
```diff
@@ -3,8 +3,8 @@ on:
   schedule:
     # every day at 01:59 (01:59am) UTC
     # - cron: "59 1 * * *"
-    # temp, for testing: every 6 hours
-    - cron: "0 */6 * * *"
+    # temp, for testing: every 4 hours
+    - cron: "0 */4 * * *"
```
vkamra, line 7: Is the load test using CI credentials? I believe it is and this will add to flakiness related to the [per-app/mailbox limit](https://learn.microsoft.com/en-us/answers/questions/734143/mailboxconcurrency-limit-throttling-exception.html).

We should create a separate app for load tests (@ryanfkeepers - I can do this if you want)

vkamra (approved):

.github/workflows/load_test.yml
ryanfkeepers: Ah, good point.  I can take care of it, no worries.  Will hold on merging this PR until I can get those references in place.

ryanfkeepers: @vkamra New App is up and configured, Load Testing has been given a separate environment in github for secrets, and those changes have been implemented in this PR.

Issue #1486, opened 2022-11-09
GC: exchange container cache causes flakey tests after purge runs


ashmrtn:
Container resolver should not return errors if it fails to pull data. Current failures are causing tests to flake out. Instead it should follow best-effort and try to get as much data as it can

PR #1518, opened 2022-11-15
GC. `EventCalendarResolver` Populate() refactor

dadams39:
## Description
Iterator Removal from Event Resolver.
The iterator overhead contributes to the complexities of the GC backup pipeline. Issue 1486's goal is to remove errors that are populated from non-error scenarios (e.g. container no longer exists).  
- Stage 1: Reduce the number of error clauses in `Populate()` function
- Stage 2: Update the M365ID retrieval process per resolver

PR introduces Iterate approach used in `delta` functionality to populate the event calendar resolver.  Reduces the amount of calls and variable creation during the populate function. The increase in efficiency will reduce the amount of time `concurrent` Mailbox queries are being made (application throttling).

## Type of change

- [x] :sunflower: Feature
- [x] :hamster: Trivial/Minor

## Issue(s)
Related to #1486 

## Test Plan

- [x] :zap: Unit test


src/internal/connector/exchange/event_calendar_cache.go
```diff
@@ -38,68 +37,54 @@ func (ecc *eventCalendarCache) Populate(
...
-
-	cb := IterativeCollectCalendarContainers(
-		directories,
-		"",
-		errUpdater,
-	)
+	for {
+		resp, err := builder.Get(ctx, options)
+		if err != nil {
+			return errors.Wrap(err, support.ConnectorStackErrorTrace(err))
```
ashmrtn, line 49: this may be something we need to revisit later but I'm not completely sure. The code overall looks fine, but I'm wondering how this will function if we get some transient error in the request. In that case, we may want to take some special action instead of returning an error immediately. However, that is something that can be handled in a different PR (and after we have more insight into how errors from calls like this appear if we don't already know since we've been using msgraph iterators thus far)

ashmrtn (approved):
could you please update the description with more information about why this change would be useful? I'm guessing it's so we have easier access to stuff like next/delta links and a bit more control over retries for pages of data if needed, but I could be thinking about it the wrong way

src/internal/connector/exchange/event_calendar_cache.go
dadams39: @ashmrtn updates have been made to the description. Your approach to the delta operation iteration inspired this approach. It was discussed that the iterator framework carries a significant overhead to lines of code. Thus, prior to moving into the eventID section described as `Stage 2` needed to ensure that the approach to iteration aligns with your thoughts on what the overall framework will look like.

As for abstraction, the `Get()` functions called by the builders and the builders themselves make it difficult to abstract it to an `exchange.Iterator`. However, doing this inline for the resolvers is not a lot of overhead.

Issue #1520, opened 2022-11-15
Remove Purge.go script


ryanfkeepers:
Now that we have pwsh scripts in place to remove test-generated user data, we no longer need the golang scripts.  There's a not-insubstantial amount of code in graph_connector propping up the purge action.  If it wasn't for that I'd just leave the purge script as it is.  But if the code is dead we might as well remove the maintenance of it, too.

PR #1521, opened 2022-11-15
remove unused purge code

ryanfkeepers:
## Description

Now that pwsh scripts do the purging, we can
get rid of the faultier graphapi exchange purge
code, along with the functions supporting it.

## Type of change

- [x] :robot: Test

## Issue(s)

* #1520

## Test Plan

- [x] :muscle: Manual


src/cmd/purge/purge.go
```diff
@@ -31,24 +27,6 @@ var purgeCmd = &cobra.Command{
...
-	RunE:  handleCalendarFolderPurge,
-}
-
-var contactsCmd = &cobra.Command{
-	Use:   "contacts",
-	Short: "Purges contacts folders",
-	RunE:  handleContactsFolderPurge,
-}
-
 var oneDriveCmd = &cobra.Command{
```
ryanfkeepers, line 30: @ashmrtn I left in your onedrive purge commands (though I'm not sure we're calling them atm) since we don't have pwsh coverage for those yet.  I'm hoping that if we investigate those controls we can get an actual hard delete, instead of needing to follow up in the UI manually.

dadams39 (approved):

PR #1524, opened 2022-11-15
Limit container resolver recursion depth

### LINKED ISSUE #1519: harden path resolution by exiting if too many calls are made
The container cache uses recursive calls to resolve folder names. However, there is not check for the recursion depth. Some sort of tracker should be added to make sure we don't recurse too many times, especially now that we're expanding the logic of fetching containers to be more resilient to errors. The check should be able to report an error even if there's a loop in the containers (i.e. `a -> b -> c -> a`). The check does not have to report the loop, it should just notice it's recursed too many times

ashmrtn:
## Description

Now that the container resolver may not have all folders in-memory (we're relaxing the limits on when we return an error), add a recursion depth limit. Limit aligns with currently supported M365 service folder hierarchy depth limits

Exchange limitations for [mail folders](https://learn.microsoft.com/en-us/office365/servicedescriptions/exchange-online-service-description/exchange-online-limits#mailbox-folder-limits) (search "Maximum folder hierarchy depth")
[OneDrive](https://support.microsoft.com/en-us/office/restrictions-and-limitations-in-onedrive-and-sharepoint-64883a5d-228e-48f5-b3d2-eb39e07630fa) limit on path length (search "file name and path lengths")

Minor refactoring of test code to make it easier to spin up tests for multiple cases

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1519

merge after
* #1523 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/container_resolver.go
```diff
@@ -24,6 +29,18 @@ func (cr *containerResolver) IDToPath(
...
+	return cr.idToPath(ctx, folderID, 0)
+}
+
+func (cr *containerResolver) idToPath(
+	ctx context.Context,
+	folderID string,
+	depth int,
+) (*path.Builder, error) {
+	if depth >= maxIterations {
+		return nil, errors.New("path contains cycle or is too tall")
```
ryanfkeepers, line 41: Is a cyclical path possible?

ryanfkeepers (approved):

src/internal/connector/exchange/container_resolver.go
ashmrtn: I don't think it should be possible in M365, but this code is generic and just relies on following parentID references. Due to that, it's possible there could be a misconfiguration somewhere that could lead to a cycle and this could would try to resolve it. It's just highly unlikely to encounter that

PR #1526, opened 2022-11-15
split Corso backups by resource owner, service, and category

ashmrtn:
## Description

This is a draft PR to get feedback on the approach. It is not expected to pass tests right now.

Split backups that span multiple (resource owner, service, category)
tuples into discrete corso backups, one per tuple. This shards the
backup details and allows for better parallelism in the future if we
want it. Could also be useful for incremental backups.

Root cause of many issues revolves around the 1:1 mapping we've had
from operation to output data. This pattern changes that though as
there's now subtasks in the backup operation that have notable
results.

Some things that don't fit well with the current patterns in the code:
* same selector saved in each backup making it difficult to determine
  what the backup contains from the summary information
* number of items read by GC is aggregate for all backups so is no
  longer persisted
* many backup IDs now need to be reported to caller
* status handling for the operation vs. individual backups

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #<issue>

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -25,11 +25,15 @@ import (
 	"github.com/alcionai/corso/src/pkg/store"
 )
 
+var (
+	errBackupNotStarted = errors.New("errors prevented the operation from processing")
+)
+
 // BackupOperation wraps an operation with backup-specific props.
 type BackupOperation struct {
 	operation
 
-	Results   BackupResults      `json:"results"`
+	BackupIDs []model.StableID
 	Selectors selectors.Selector `json:"selectors"`
 	Version   string             `json:"version"`
 
@@ -42,6 +46,7 @@ type BackupResults struct {
 	stats.ReadWrites
 	stats.StartAndEndTime
 	BackupID model.StableID `json:"backupID"`
+	Status   opStatus
 }
 
 // NewBackupOperation constructs and validates a backup operation.
@@ -88,40 +93,8 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	ctx, end := D.Span(ctx, "operations:backup:run")
 	defer end()
 
-	var (
-		opStats       backupStats
-		backupDetails *details.Details
-		startTime     = time.Now()
-	)
-
-	op.Results.BackupID = model.StableID(uuid.NewString())
-
-	op.bus.Event(
-		ctx,
-		events.BackupStart,
-		map[string]any{
-			events.StartTime: startTime,
-			events.Service:   op.Selectors.Service.String(),
-			events.BackupID:  op.Results.BackupID,
-		},
-	)
-
-	// persist operation results to the model store on exit
-	defer func() {
-		// wait for the progress display to clean up
-		observe.Complete()
-
-		err = op.persistResults(startTime, &opStats)
-		if err != nil {
-			return
-		}
-
-		err = op.createBackupModels(ctx, opStats.k.SnapshotID, backupDetails)
-		if err != nil {
-			// todo: we're not persisting this yet, except for the error shown to the user.
-			opStats.writeErr = err
-		}
-	}()
+	// wait for the progress display to clean up
+	defer observe.Complete()
 
 	complete, closer := observe.MessageWithCompletion("Connecting to M365:")
 	defer closer()
@@ -131,9 +104,12 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	gc, err := connector.NewGraphConnector(ctx, op.account)
 	if err != nil {
 		err = errors.Wrap(err, "connecting to graph api")
-		opStats.readErr = err
 
-		return err
+		// Errors here don't cause anything to be persisted.
+		return multierror.Append(
+			errBackupNotStarted,
+			err,
+		)
 	}
 	complete <- struct{}{}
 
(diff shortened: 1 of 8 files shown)


src/internal/operations/backup.go
```diff
@@ -144,73 +120,161 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		err           error
+		opStats       backupStats
+		backupDetails *details.Details
+		results       = &BackupResults{}
+		startTime     = time.Now()
+		hadData       = len(cs) > 0
+	)
+
+	// persist operation results to the model store on exit
+	// TODO(ashmrtn): Find some error handling method for this.
```
ryanfkeepers, line 193: How to return errors from a defer: https://go.dev/play/p/hrufTHNbzqJ

ryanfkeepers:
The approach looks sound so far.  Some thoughts around current patterns:

> same selector saved in each backup making it difficult to determine what the backup contains from the summary information

Should be easy to make a func like `selector.FilterScopes(user, category) => selector`.  I can make an issue for it, if wanted.

> number of items read by GC is aggregate for all backups so is no longer persisted

I was thinking on this earlier when handling metrics.  Was thinking about a proposal to shift metrics into the collection.  That way reads/writes/errors could be aggregated within the scope in which they occurred, and then aggregated further still as needed by higher level callers.  I'm sure there's some rough corners to figure out, I honestly haven't thought it all the way through, but it might be an option here.

> many backup IDs now need to be reported to caller

This fits in line with the expectations and needs of the app behavior.  We'll have to make sure our sdk users are prepared for this, as it'll be a breaking change, but otherwise there's no worry.

> status handling for the operation vs. individual backups

Can we get by with a quick hack like: 1/ BackupOperation.Results turns into a map[backupID]BackupResults, 2/ add a status property to BackupResults, correlate that with collection group status 3/ record main operation status as one of Processing, Complete, No Data, Failed, where nodata and failed imply "all collections"?

ashmrtn:
> > same selector saved in each backup making it difficult to determine what the backup contains from the summary information
> 
> Should be easy to make a func like `selector.FilterScopes(user, category) => selector`. I can make an issue for it, if wanted.

Question: what would GC be handed? We have a bit of a chicken and egg problem here cause some tuples can only be named after discovery is done (i.e. `--users "*"`). It's possible to extract all the needed information from the collections if needed -- that's how grouping is done currently. The only question is how to turn what we have into something we can store in the `BackupModel` in a reasonable fashion

> > number of items read by GC is aggregate for all backups so is no longer persisted
> 
> I was thinking on this earlier when handling metrics. Was thinking about a proposal to shift metrics into the collection. That way reads/writes/errors could be aggregated within the scope in which they occurred, and then aggregated further still as needed by higher level callers. I'm sure there's some rough corners to figure out, I honestly haven't thought it all the way through, but it might be an option here.

That sounds reasonable. The only other solution I can think of is to have GC return results in batches that correspond to the groups we form based on the tuple. Sending results in batches creates an implicit requirement though so I think going by collection would be better. The only unknown in the (rough) plan is how to handle stuff like failures fetching folders or resource owners which don't map directly to a single collection

Eventually we may end up with the implicit dependency if we want to start pipelining things by backing up one tuple's data while GC is discovering the next tuple's data, but having that be the smallest in scope possible would be good

> > many backup IDs now need to be reported to caller
> 
> This fits in line with the expectations and needs of the app behavior. We'll have to make sure our sdk users are prepared for this, as it'll be a breaking change, but otherwise there's no worry.

sounds good

> > status handling for the operation vs. individual backups
> 
> Can we get by with a quick hack like: 1/ BackupOperation.Results turns into a map[backupID]BackupResults, 2/ add a status property to BackupResults, correlate that with collection group status 3/ record main operation status as one of Processing, Complete, No Data, Failed, where nodata and failed imply "all collections"?

That's sort of what's currently there, except the individual results aren't directly returned to the caller and the overall operation status is never persisted. To be more concrete, each BackupModel for a tuple has its own status, mostly based off how the previous version of the code functioned. The only difference is right now we're unable to tell if any single backup had "NoData" just by checking if GC didn't pull anything. We instead check the number of collections passed.

The statuses are aggregated [here](https://github.com/alcionai/corso/pull/1526/files#diff-f088d3a5f56348ddaf6fbf485c2aeaa6d8b40a860de014f7084db6daf0753829R153) for the overall operation with the final status decided [here](https://github.com/alcionai/corso/pull/1526/files#diff-f088d3a5f56348ddaf6fbf485c2aeaa6d8b40a860de014f7084db6daf0753829R165), but there's no overarching model we have to persist that status. I'm not sure if we need to though

ryanfkeepers:
> Question: what would GC be handed? We have a bit of a chicken and egg problem here cause some tuples can only be named after discovery is done (i.e. --users "*"). It's possible to extract all the needed information from the collections if needed -- that's how grouping is done currently. The only question is how to turn what we have into something we can store in the BackupModel in a reasonable fashion

IMO, we'd hand it the complete selector.  Extracting the user param from the tuple is what i had in mind, as this transformation would really only be relevant for, like you're hoping, backupModel storage.

> The only unknown in the (rough) plan is how to handle stuff like failures fetching folders or resource owners which don't map directly to a single collection

I think mapping failures is the rough patch either way.  Ah well, still doable.   

> The only difference is right now we're unable to tell if any single backup had "NoData" just by checking if GC didn't pull anything. We instead check the number of collections passed.

👍 okay, we'll carry on from there.  Sounds like we're on the right vector.  Just need to supplement it a bit.

ashmrtn:
closing as we've decided to implement this closer to the CLI layer making these changes unneeded

PR #1531, opened 2022-11-15
migrate containerCache to graph pkg

ryanfkeepers:
## Description

ContainerCache is a generic resource that can
be moved to the graph package.  Service-
specific code is tied to the `populate` func,
which remains in each service/category.  This
change is a prepatory refactoring to reduce
boilerplate for sharepoint implementation.

## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* #1506

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/cache_container.go
+++ b/src/internal/connector/exchange/cache_container.go
@@ -2,65 +2,7 @@ package exchange
 
 import (
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
-	"github.com/pkg/errors"
-
-	"github.com/alcionai/corso/src/internal/connector/graph"
-	"github.com/alcionai/corso/src/pkg/path"
 )
-
-// checkIDAndName is a helper function to ensure that
-// the ID and name pointers are set prior to being called.
-func checkIDAndName(c graph.Container) error {
-	idPtr := c.GetId()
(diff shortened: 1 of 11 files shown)


ryanfkeepers:
@dadams39 Wellp, now that I've gone through the work and moved on to the next part, I'm not sure I did anything useful here.  If it'll help you work with lists (not sure if they're going to use a cache resolver or not), or if we have plans to use cache resolvers for onedrive (not as far as I"m aware), we can keep the changes.  Otherwise might as well toss it.  Thoughts?

dadams39:
> @dadams39 Wellp, now that I've gone through the work and moved on to the next part, I'm not sure I did anything useful here. If it'll help you work with lists (not sure if they're going to use a cache resolver or not), or if we have plans to use cache resolvers for onedrive (not as far as I"m aware), we can keep the changes. Otherwise might as well toss it. Thoughts?

This is more a question for @ashmrtn 
There has to be a `path` strategy for SharePoint as it relates to OneDrive and Lists. My early thought is that there isn't going to be a need for a resolver for these types at this time. This may change when we come to the `restore` portion of the workflow.

ryanfkeepers:
> My early thought is that there isn't going to be a need for a resolver for these types at this time

Invoking `yagni`.  Dropping the PR for now.

Issue #790, opened 2022-09-07
CI cache thrashing


ryanfkeepers:
# Problem
The golang modcache frequently thrashes.  CI tests should be runnable in 2-4 minutes.  Due to cache issues, `go test ./...` incurs a package download of an additional 15-20 minutes.  These delays are not reported as part of the test run itself, but can be observed in the CI action by enabling timestamps and reviewing the go test step.

# Solution
Reliably utilize github's go cache.  Ideally, this should be locally testable with nektos/act.

PR #1532, opened 2022-11-16
Reduce CI test times

vkamra:
## Description

The bulk of the time being spent in the `Test-Suite` job was compiling the test binaries.
This is related to the size of the MS graph library.

The impact of this was even more obvious because the Go build cache didn't seem to be getting used
in our CI tests.

The issue was that the cache is being used (and primed) in multiple places and if it happens to
get populated during a build step that doesn't compile the test binaries - then the cache would
never get updated with those and we would keep building from scratch.

The fix is to add a cache suffix so that the testsuite cache and lint cache use different locations.

Going forward - another improvement would be to create a single cache and populate it up-front
in the workflow with all Go artifacts (i.e. compile src and test up front).

This change reduces `Test-Suite` time from ~32 minutes to ~14 minutes

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #790 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph_connector_test.go
```diff
@@ -154,15 +154,16 @@ func (suite *GraphConnectorIntegrationSuite) TestExchangeDataCollection() {
...
-		{
-			name: suite.user + " Events",
-			getSelector: func(t *testing.T) selectors.Selector {
-				sel := selectors.NewExchangeBackup()
-				sel.Include(sel.EventCalendars([]string{suite.user}, []string{exchange.DefaultCalendar}, selectors.PrefixMatch()))
-
-				return sel.Selector
-			},
-		},
+		// {
```
vkamra, line 157: The changes to the test are to disable "Event restore" tests since we're debugging an issue with our test tenant calendars + Graph API.

vkamra:
@meain @ryanfkeepers @dadams39 @ashmrtn - test changes are so that we can unblock CI while we debug the Graph API issues in our test tenant.

meain (approved):

PR #1536, opened 2022-11-16
generic drive retrieval for sharepoint

ryanfkeepers:
## Description

Adapts the graph onedrive library to handle
access to drive data across both onedrive and
sharepoint services.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -247,58 +261,42 @@ func (gc *GraphConnector) OneDriveDataCollections(
...
-		// if err != nil {
-		// 	return nil, errors.Wrap(err, "filling collections")
-		// }
-
-		foldersComplete <- struct{}{}
-
-		for _, collection := range collections {
-			gc.incrementAwaitingMessages()
+		switch category {
+		case path.FilesCategory: // TODO: better category for drives
```
ashmrtn, line 279: can you give a sketch of what you'd like to use as the category instead? Also rather unfortunately, as the category is persisted as part of the reporef of an item changing it will be a breaking change

src/internal/connector/onedrive/collections.go
```diff
@@ -14,94 +14,89 @@ import (
...
-// specified user
+type driveSource int
+
+const (
+	unknownDriveSource = iota
+	OneDriveSource
+	SharePointSource
+)
+
+type isAnyMatcher interface {
```
ashmrtn, line 27: this is a bit nitpicky, but `isAnyMatcher` seems like a bit of a misnomer to me. It makes me think the interface only implements `IsAny()`

src/internal/connector/onedrive/collections.go
```diff
@@ -14,94 +14,89 @@ import (
 	"github.com/alcionai/corso/src/internal/observe"
 	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
-	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
-// Collections is used to retrieve OneDrive data for a
-// specified user
+type driveSource int
```
ashmrtn, line 19: I'm curious abouth the choice of using a const to switch execution flow instead of defining an interface and passing a concrete instance of a struct that implements the interface. It looks like right now the main things this const is used for is getting the path of the item and getting the drives. The sharepoint and onedrive versions of each of these functions take the same arguments as well so it seems like an interface would work no? If using the interface, the calling pattern would be slightly reversed I guess where the caller would make the instance of the implementation and pass it in?

src/internal/connector/sharepoint/libraries_test.go
```diff
@@ -0,0 +1,313 @@
+package sharepoint_test
```
ashmrtn, line 1: not sure how I feel about duplicating so much code for tests. It seems like it would be difficult to keep them in sync which could lead to uneven testing in the long run

src/internal/connector/data_collections.go
```diff
@@ -247,58 +261,42 @@ func (gc *GraphConnector) OneDriveDataCollections(
...
-		errs  *multierror.Error
-		sites = scope.Get(selectors.SharePointSite)
-		colls = make([]*sharepoint.Collection, 0)
+		errs        *multierror.Error
+		sites       = scope.Get(selectors.SharePointSite)
+		category    = scope.Category().PathType()
+		collections = make([]data.Collection, 0)
 	)
 
 	// Create collection of ExchangeDataCollection
```
ashmrtn, line 272: nit: not working with exchange here

src/internal/connector/sharepoint/libraries.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/internal/connector/onedrive"
+	"github.com/alcionai/corso/src/internal/connector/support"
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/selectors"
+)
+
+func CollectLibraries(
```
ashmrtn, line 14: I'm confused by the purpose of this function. It seems like the calling function (`createSharePointCollections`) already iterates over all the sites in the scope. Won't that just result in fetching everything 2x? Also, since this is pretty much a mirror image of getting OneDrive data, why is it not in the `connector/data_collections.go` file so it is easier to see that it should be kept mostly in sync with the OneDrive call?

src/internal/connector/data_collections.go
ryanfkeepers: Ah, I'm using terminology loosely here.  `drives` in this case mean sharePoint drives, specifically.  The todo is not meant to reflect upon oneDrive terminology/consts.

The categorical division in sharepoint is currently Libraries and Sites.  Libraries contain the drive-based data, Sites contain the webItem data.  Very likely, the todo will be to swap `FilesCategory` for `LibrariesCategory`.

src/internal/connector/sharepoint/libraries_test.go
ryanfkeepers: I'm happy to consolidate them in a follow-up change.  How's that sound?

src/internal/connector/onedrive/collections.go
ryanfkeepers: Great question.  What I like about enumeration is that the code (which are ultimately minor pivots in behavior) remained a bit more consolidated.  Given that there's little actual difference between the processes, it felt like a lighter touch.  But no strong feelings here.  I can swap it out for interfaces if that's preferred.

src/internal/connector/onedrive/drive.go
```diff
@@ -67,7 +67,33 @@ const (
...
+		return userDrives(ctx, service, resourceOwner)
+	case SharePointSource:
+		return siteDrives(ctx, service, resourceOwner)
+	default:
+		return nil, errors.Errorf("unrecognized drive data source")
+	}
+}
+
+func siteDrives(ctx context.Context, service graph.Service, site string) ([]models.Driveable, error) {
+	r, err := service.Client().SitesById(site).Drives().Get(ctx, nil)
```
vkamra, line 87: OneDrive sites (personal sites) will also show up here - for Sharepoint backup, we should filter those out.

src/internal/connector/sharepoint/libraries.go
ryanfkeepers: > Won't that just result in fetching everything 2x?

Nope.  Though I can see how the code makes that confusing.  At this point we've separated the scopes into `DiscreteScopes(SiteIDs...)`, and so createSharePointCollections iteration of scope site IDs should only produce a single site.  Still, feels weird to range over that twice, so I'll just send in a single siteID instead.

> why is it not in the data_collections.go

Well, now that you mention it, I realized I went halfway towards the solution I actually wanted.  It felt a bit strange to me that we have a package per service, but also a graph connector file dedicated to per-service processing.  I wanted to break that pattern with sharepoint, and apparently forgot about that goal through the implementation, leaving half of the code in data_collections anyway.

sharepoint/libraries.go has been swapped for sharepoint/data_collections.go, and the handling moved into that space.  If we're concerned about code re-use (I'm not particularly worried here, but to address it), the ideal would be to have onedrive make the same refactor, and from there to choose an appropriate point for sharepoint to turn to onedrive for the creation/population of collections.

src/internal/connector/onedrive/collections.go
vkamra: It seems like we have a combination of both approaches here. We're passing in an enum to switch execution flow for path computation (`GetCanonicalPath`) but an interface to pass in the matcher.

src/internal/connector/onedrive/collections.go
vkamra: maybe just `folderMatcher` or `scopeMatcher` ?

src/internal/connector/onedrive/drive.go
ryanfkeepers: Yeap, that's planned for a next step.

src/internal/connector/onedrive/collections.go
ryanfkeepers: `folderMatcher` sounds good, since we're packing the folder category into the match assumption.

src/internal/connector/sharepoint/data_collections_test.go
```diff
@@ -0,0 +1,313 @@
...
+func (suite *SharePointLibrariesIntegrationSuite) SetupSuite() {
+	_, err := tester.GetRequiredEnvSls(
+		tester.AWSStorageCredEnvs,
+		tester.M365AcctCredEnvs,
+	)
+
+	require.NoError(suite.T(), err)
+}
+
+func (suite *SharePointLibrariesIntegrationSuite) TestUpdateCollections() {
```
vkamra, line 82: Should remove this test completely (I saw the discussion in the previous PR). This is all tested under onedrive already. Recommend we just stick a unit test for the sharepoint matcher in this test and add a unit test for GetCanonicalPath in drive (with a SharepointSource). Other coverage should be through an integration test here for  `sharepoint.DataCollections` that we add here.

src/internal/connector/onedrive/collections.go
ryanfkeepers: I suppose I don't see that combination as an anti-pattern.  But it sounds like he preference is for an interface, so I'll swap out for that.

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -0,0 +1,158 @@
...
+		con.IncrementAwaitingMessages()
+	}
+
+	return collections, errs
+}
+
+// createSharePointCollections - utility function that retrieves M365
+// IDs through Microsoft Graph API. The selectors.SharePointScope
+// determines the type of collections that are retrieved.
+func createSharePointCollections(
```
vkamra, line 75: @ryanfkeepers - thoughts on implementing this similar to what `OneDrive` does? i.e. a `NewCollections()` and `Get` method? The logic in `DataCollections` above which has more GC specifics would move into `gc.SharePointCollections()`

I don't like for example exposing the internals of `con.IncrementAwaitingMessages()` here - that's very GC specific.

src/internal/connector/onedrive/collections.go
vkamra: I don't think it's an anti pattern - so go with what seems right and easy to read. We're in nit territory here 😃

ryanfkeepers: Going to hold off on the interface for now.  Since onedrive is generating NewCollections from connector, not inside onedrive, the supply of interface-compliant structs is getting a bit messy.  I can revisit this later, as it might be easier with some of the other proposed refactors.

src/internal/connector/sharepoint/data_collections.go
vkamra: Just to clarify - my suggestion for now is to move what is in `DataCollections` in this file to `func (gc *GraphConnector) SharePointDataCollections` and export `createSharePointCollections` as perhaps `sharepoint.NewCollections().Get(...)` (I'm hoping this is a pattern we can replicate for all the services)

RE: the duplication in `collectLibraries` - not worried about that for now - we can refactor later if needed.

ryanfkeepers: > Just to clarify...

Ahah, this is funny, because I just did _the very opposite_ in the [commit before you looked at the PR](https://github.com/alcionai/corso/pull/1536/commits/9f5f6dbceca0a94c7410955f3d44d323883ef19c).  My read on the `connector/data_collections.go` file is that we're packing in more service-specific concerns, and exercising a minimum of common gc-specific concerns (`incrementAwait` is one of the few).  And as we add cross-service interfaces and such, those service footprints grow larger while gc concerns remain static.

In all, it feels like this ownership is better housed within each service package, colocated with the other interfaces and functions used by that service.  But, this is not a strongly held opinion.  If you prefer to keep things in the collections package, I can stick with that.

vkamra: > My read on the connector/data_collections.go file is that we're packing in more service-specific concerns, and exercising a minimum of common gc-specific concerns (incrementAwait is one of the few)

Ah - I'm with you on that one. Service specific stuff under `sharepoint` and the move GC logic out into `func (gc *GraphConnector) SharePointDataCollections`

src/internal/connector/graph_connector.go
```diff
@@ -336,7 +336,7 @@ func (gc *GraphConnector) PrintableStatus() string {
 	return gc.status.String()
 }
 
-func (gc *GraphConnector) incrementAwaitingMessages() {
+func (gc *GraphConnector) IncrementAwaitingMessages() {
```
vkamra, line 339:
```suggestion
func (gc *GraphConnector) incrementAwaitingMessages() {
```

vkamra (approved):
LGTM. @ashmrtn - since you had previous comments - you should take another look to make sure this works.

+2 to unblock - we can iterate on changes.

src/internal/connector/data_collections_test.go
```diff
@@ -180,24 +181,27 @@ func (suite *ConnectorDataCollectionIntegrationSuite) TestSharePointDataCollecti
...
 		suite.T().Run(test.name, func(t *testing.T) {
-			_, err := connector.SharePointDataCollections(ctx, test.getSelector(t))
+			collection, err := sharepoint.DataCollections(
+				ctx,
+				test.getSelector(t),
+				connector.GetSiteIds(),
+				connector.credentials.AzureTenantID,
+				connector)
 			require.NoError(t, err)
+			assert.Equal(t, 1, len(collection))
```
ashmrtn, line 191: should this check that it's equal to the length of the number of site IDs? Or are we getting only a single collection for all sites?
```suggestion
			assert.Len(t, collection, len(connector.GetSiteIds())
```

src/internal/connector/onedrive/collections.go
```diff
@@ -14,94 +14,89 @@ import (
...
 type Collections struct {
-	tenant string
-	user   string
-	scope  selectors.OneDriveScope
-	// collectionMap allows lookup of the data.Collection
-	// for a OneDrive folder
-	collectionMap map[string]data.Collection
+	Tenant        string
+	ResourceOwner string
+	Source        driveSource
```
ashmrtn, line 37: nit: I'd prefer to make these unexported if they aren't used outside of the package. If they're left exported other things may start using them and then it may become harder to make changes in the future

ashmrtn (approved):

src/internal/connector/data_collections_test.go
ryanfkeepers: I think it's mocked, but this is a good change nonetheless.

src/internal/connector/onedrive/collections.go
ryanfkeepers: Ah, good catch, I must have redacted the change that used these externally.

PR #1540, opened 2022-11-16
adds tenant site lookup on new graph conn

ryanfkeepers:
## Description

Adds a lookup for all tenant site ids when a new
graph connector is created.  Also adds an enum
to flag which resource set (users, sites, or all)
that the connector should initialize.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector.go
+++ b/src/internal/connector/graph_connector.go
@@ -5,6 +5,7 @@ package connector
 import (
 	"context"
 	"runtime/trace"
+	"strings"
 	"sync"
 
 	"github.com/microsoft/kiota-abstractions-go/serialization"
@@ -217,6 +218,8 @@ func (gc *GraphConnector) setTenantSites(ctx context.Context) error {
 	return nil
 }
 
+var errKnownSkippableCase = errors.New("case is known and skippable")
+
 // Transforms an interface{} into a key,value pair representing
 // siteName:siteID.
 func identifySite(item any) (string, string, error) {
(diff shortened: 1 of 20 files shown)


src/internal/connector/graph_connector.go
```diff
@@ -73,7 +74,16 @@ func (gs graphService) ErrPolicy() bool {
...
+type resource int
+
+const (
+	UnknownResource resource = iota
+	AllResources
+	Users
+	Sites
+)
+
+func NewGraphConnector(ctx context.Context, acct account.Account, r resource) (*GraphConnector, error) {
```
vkamra, line 86: Can we also optimize so that if a specific user or site is specified - we don't enumerate *all* users/sites?

ryanfkeepers: Tracked as a separate issue here: https://github.com/alcionai/corso/issues/1547

ryanfkeepers:
@vkamra Can I kindly get a final review on this change?

src/internal/connector/sharepoint/queries.go
```diff
@@ -0,0 +1,21 @@
+package sharepoint
```
vkamra, line 1: I like the idea of consolidating queries. Thoughts on defining a SharePoint query interface and stashing that in Graph Connector? We could use that to start mocking these queries.

vkamra (approved):

vkamra:
Added mergequeue label since it was ready to go...

src/internal/connector/sharepoint/queries.go
ryanfkeepers: I'll look into it, sure.

ryanfkeepers:
> Added mergequeue label since it was ready to go...

Much appreciated

PR #1543, opened 2022-11-17
Support generating different blog and support sites

ntolia:
## Description

We are, by default, generating docs. However, if an environment variable (BLOG) is defined, only the blog site will be generated.

## Type of change


docs/docusaurus.config.js
```diff
@@ -124,8 +111,8 @@ const config = {
             title: 'More',
             items: [
               {
-                label: 'Blog',
-                to: '/blog',
+                label: (process.env.BLOG) ? 'Docs' : 'Blog',
+                href: (process.env.BLOG) ? 'https://docs.corsobackup.io/' : 'https://blog.corsobackup.io/',
```
meain, line 115: I think this one is revesed

ntolia: I think this is fine. If it is the blog site, we want a link to docs. If it is a docs site, we want a link to the blog. Does that make sense?

meain: Ahh, I misread it.

Issue #1528, opened 2022-11-15
CI issues


ryanfkeepers:
Aggregate issue for tweaks to address CI timeout issues.

PR #1546, opened 2022-11-17
fix failing test in onedrive folder create/delete

ryanfkeepers:
## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #1528

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/drive_test.go
+++ b/src/internal/connector/onedrive/drive_test.go
@@ -22,6 +22,7 @@ func TestOneDriveDriveSuite(t *testing.T) {
 	if err := tester.RunOnAny(
 		tester.CorsoCITests,
 		tester.CorsoOneDriveTests,
+		"flomp",
 	); err != nil {
 		t.Skip(err)
 	}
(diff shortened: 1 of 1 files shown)


src/internal/connector/onedrive/drive_test.go
```diff
@@ -49,22 +47,22 @@ func (suite *OneDriveSuite) TestCreateGetDeleteFolder() {
...
-		assert.NoError(t, DeleteItem(ctx, gs, driveID, folderIDs[0]))
-	}()
+	for i := 0; i < 2; i++ {
+		name := "Corso_Folder_Test_" + common.FormatNow(common.SimpleTimeTesting)
+		folderElements := []string{name}
+		id, err := createRestoreFolders(ctx, gs, driveID, folderElements)
+		require.NoError(t, err)
 
-	folderName2 := "Corso_Folder_Test_" + common.FormatNow(common.SimpleTimeTesting)
-	folderElements = append(folderElements, folderName2)
```
ryanfkeepers, line 62: Reducing folderElements to a single folder is the change that 'fixes' the tests.  We could, if wanted, bring this back by updating line 98 to check for a pre-determined length (count of the expected restored folders).  Doesn't quite explain why this only bugs out for particular users (AdeleV).

ryanfkeepers: @vkamra Unless you want us to keep it, I'm going to close this ticket.  Now that we're unable to repro the error, I'm leaning towards the assumption that it was a temporarily bad data state for the given test user.

vkamra: Let's close this and I'm going to look to see if I can repro locally to confirm the Delta behavior we discussed

PR #1554, opened 2022-11-18
filter out personal sites

ryanfkeepers:
## Description

Adds a server-side filter that removes all
personal (aka: oneDrive) sites from the drive
retrieval.

The approach is not my preferred method.  I'd rather
have used a filter in the query parameters.  In fact, I
have a functioning query in postman:
`$filter=NOT contains(webUrl, 'sharepoint.com/personal/')`
But I can't seem to get the graph client query to build
when I add that to the query params.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :green_heart: E2E


src/internal/connector/graph_connector.go
```diff
@@ -220,6 +220,8 @@ func (gc *GraphConnector) setTenantSites(ctx context.Context) error {
 
 var errKnownSkippableCase = errors.New("case is known and skippable")
 
+const personalSitePath = "sharepoint.com/personal/"
```
ashmrtn, line 223: this is a bit more of a long-term question, but do we have any way to know if this is a locale independent way to specify personal sites?

src/internal/connector/graph_connector_test.go
```diff
@@ -66,7 +66,7 @@ func (suite *GraphConnectorIntegrationSuite) TestSetTenantUsers() {
 
 	newConnector.graphService = *service
 
-	suite.Equal(len(newConnector.Users), 0)
+	suite.Equal(0, len(newConnector.Users))
```
ashmrtn, line 69: nit
```suggestion
	suite.Empty(newConnector.Users)
```

src/internal/connector/graph_connector.go
```diff
@@ -237,6 +239,12 @@ func identifySite(item any) (string, string, error) {
 		return "", "", errors.Errorf("no name for Site: %s", *m.GetId())
 	}
 
+	// personal (ie: oneDrive) sites have to be filtered out server-side.
+	url := m.GetWebUrl()
+	if url != nil && strings.Contains(*url, personalSitePath) {
```
ashmrtn, line 244: I'm not familiar with how sharepoint site URLs are formatted, but is it possible for users to generate a site with a URL that would have `sharepoint.com/personal` in it but not be a personal site?

ashmrtn (approved):

src/internal/connector/graph_connector.go
ryanfkeepers: Good question. Unfortunately not; the [site](https://learn.microsoft.com/en-us/graph/api/resources/site?view=graph-rest-1.0) sttruct is pretty sparse.  The next closest data point we have is `siteCollection.hostname`, where personal sites all have the domain `<tenant>-my.sharepoint.com`.  But, then again, that could be subject to locale as well.

src/internal/connector/graph_connector.go
ryanfkeepers: Looks like the slashes get stripped from the generated url.

image

src/internal/connector/graph_connector_test.go
ryanfkeepers: All this well curated fanciness to keep track of.

PR #1555, opened 2022-11-19
migrate sharepoint cat from file to library

ryanfkeepers:
## Description

Adds a LibraryCategory in paths, and changes
the current sharepoint code to use that cat
instead of the onedrive files cat.  This is mostly to keep sharepoint library drives separate from
onedrive and sharepoint site list items.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :zap: Unit test


src/pkg/path/path.go
```diff
@@ -10,29 +10,44 @@
 //
 // Examples of paths splitting by elements and canonicalization with escaping:
 // 1.
-//   input path: `this/is/a/path`
```
ryanfkeepers, line 13: auto-fmt changes.

ashmrtn (approved):

PR #1575, opened 2022-11-21
add sharepoint to testVerifyBackupinputs

ryanfkeepers:
## Description

adds sharepoint to testVerifyBackupInputs, with
some additional refactors to selectors for either
convenience or case coverage.

❗ This branch needs to be merged before [issue-1506-collects](https://github.com/alcionai/corso/pull/1536)
can merge. ❗ 

## Type of change

- [x] :sunflower: Feature
- [x] :bug: Bugfix

## Issue(s)

* #1506

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -29,7 +29,7 @@ func (gc *GraphConnector) DataCollections(ctx context.Context, sels selectors.Se
 	ctx, end := D.Span(ctx, "gc:dataCollections", D.Index("service", sels.Service.String()))
 	defer end()
 
-	err := verifyBackupInputs(sels, gc.Users)
+	err := verifyBackupInputs(sels, gc.GetUsersIds(), gc.GetSiteIds())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


src/internal/connector/data_collections.go
```diff
@@ -51,50 +51,48 @@ func (gc *GraphConnector) DataCollections(ctx context.Context, sels selectors.Se
...
+			return fmt.Errorf("included resource owner %s not found within tenant", ro)
+		}
+	}
 
-	for k := range mapOfUsers {
-		normUsers[strings.ToLower(k)] = struct{}{}
+	for _, ro := range resourceOwners.Excludes {
+		if _, ok := normROs[strings.ToLower(ro)]; !ok {
+			return fmt.Errorf("excluded resource owner %s not found within tenant", ro)
+		}
```
ashmrtn, line 90: this may be a little more strict than we need it tbh. If we're excluding the resource owner I don't really think it makes a difference if we find it or not. The only thing that matters is that we don't include it in the backup.

For example, this could be a set of archived users that no longer need backed up from people who have left the company. After a while those users may be deleted, but whatever is triggering the backups should still work and ideally I would think it would be alright to not fail if the script triggering the backups wasn't updated immediately to reflect the fact that users were deleted

It may be good to check with someone more in tune with UX than me though heh

src/internal/connector/graph_connector_disconnected_test.go
```diff
@@ -256,11 +262,29 @@ func (suite *DisconnectedGraphConnectorSuite) TestVerifyBackupInputs() {
...
+		},
+		{
+			name:       "invalid sites",
+			checkError: assert.Error,
+			getSelector: func(t *testing.T) selectors.Selector {
+				sel := selectors.NewSharePointBackup()
+				sel.Include(sel.Sites([]string{"fnords.smarfs.brawnhilda"}))
+				return sel.Selector
+			},
+		},
```
ashmrtn, line 282: I'm a bit split on whether we should have some sort of test for filters as well. Now that `verifyBackupInputs` is comparing against them it seems like we should test that at least a bit

ashmrtn (approved):

src/internal/connector/graph_connector_disconnected_test.go
ryanfkeepers: 👍 no harm in it, unit tests are easy.

PR #1577, opened 2022-11-21
add sharepoint backup to the cli

ryanfkeepers:
## Description

Adds boilerplate backup (create, list, delete) cli
commands for sharepoint.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1506

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test


src/cli/backup/exchange.go
```diff
@@ -30,6 +30,7 @@ import (
 var (
 	backupID     string
 	exchangeData []string
+	site         []string
```
ashmrtn, line 33: nit: it feels a little odd to have this in the exchange file. OneDrive seems to have it's own set of globals in it's file as well

ashmrtn (approved):

src/cli/backup/exchange.go
ryanfkeepers: Ah, good call.

Issue #1569, opened 2022-11-21
`--no-progress-bar` option is not respected by `MessageWithCompletion`


meain:
The `--no-progress-bar` option, while it disables progressbars for individual/category progressbar, it does not disable for messages like `Connecting to M365` or `Restoring data`.

PR #1578, opened 2022-11-21
observe always hides when told

ryanfkeepers:
## Description

Fixes a bug where not all observer progress bars
were staying hidden if the 'hide' configuration
was flagged.

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/1569

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/observe/observe.go
+++ b/src/internal/observe/observe.go
@@ -76,6 +76,10 @@ type config struct {
 	doNotDisplay bool
 }
 
+func (c config) hidden() bool {
+	return c.doNotDisplay || writer == nil
+}
+
 // SeedWriter adds default writer to the observe package.
 // Uses a noop writer until seeded.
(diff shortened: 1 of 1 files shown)


meain:
Should we consider showing a one time message for event but without the progress indicator? We could consider that in a follow up as well though.

PS: I have updated the issue link in PR desc

meain (approved):

ryanfkeepers:
>Should we consider showing a one time message for event but without the progress indicator? We could consider that in a follow up as well though

@meain Unfortunately, progress bars and logs fight for display.  Depending on race conditions and write order, it is possible for the progress bars to erase logs as they're written to the console.  Primarily, the 'no-progress' flag is to shut down the display entirely in case we need to record the logs instead.  Keeping around one-time messages will likely keep clobbering things.

Issue #581, opened 2022-08-17
Use msgraphgocore http client to download onedrive file


vkamra:
We're currently using the default http client which should be configured with the correct timeout/retry handlers. The `msgraphgocore` default client seems suitable here.

meain:
Am I missing something or is https://github.com/alcionai/corso/pull/1484 the only change needed for this(apart from maybe generating the url when we start the download)?

meain:
On a related note, we are already getting the URL for downloading at the time of actual fetch. Ref: https://github.com/alcionai/corso/blob/274fba8708085f4ef29c9db4f7144b7e9d2204e8/src/internal/connector/onedrive/item.go#L37

meain:
On a related note...

> Graph Download URL expiry - should/can we defer creation of the upload/download URL to when we start the operation

We are already creating the download URL at the time of fetch. https://github.com/alcionai/corso/blob/274fba8708085f4ef29c9db4f7144b7e9d2204e8/src/internal/connector/onedrive/item.go#L37

vkamra:
> We are already creating the download URL at the time of fetch.

So the code in item.go initializes the reader (and the download URL) but it is only returned and used via the data collection to kopia [here](https://github.com/alcionai/corso/blob/274fba8708085f4ef29c9db4f7144b7e9d2204e8/src/internal/connector/onedrive/collection.go#L100). 

My concern is that for a large hierarchy - there can be a delay. I'm not sure yet how long these URLs are valid.

meain:
We can change it to return just the URL and change `ToReader` to create a `ReadCloser` from the URL. But that will require a project wide change to the type `itemReaderFunc`.
https://github.com/alcionai/corso/blob/990e5514d22d590aa8a70e6d484730a7fac0b18f/src/internal/connector/onedrive/collection.go#L44-L49

We can have it be a URL, but that would break the nice abstraction that we have as a `ReadCloser`. We will also have to update what `Item` will look like.
https://github.com/alcionai/corso/blob/990e5514d22d590aa8a70e6d484730a7fac0b18f/src/internal/connector/onedrive/collection.go#L88-L92

[...]

Issue #1595, opened 2022-11-23
Speed up initial OneDrive/Exchange backup


meain:
Looking at backup perf of kopia and individual downloads looks like almost all of the slowdown is coming from this for loop being synchronous. Parallelizing this seems to speed up backup a lot.
https://github.com/alcionai/corso/blob/9ad1b4d481b631671966a553bf9dabc1339b25f8/src/internal/connector/onedrive/collection.go#L143

Ran into a few stability issues where the number of backed up files where less than the actual number of files for random runs. One of the major reported issues is the one below. Looking into this and potential other issues.
```
failed to get item 01GIMTAFVULBCVRFGN6BA3VCIUDUJC6Y2O: Get "https://graph.microsoft.com/v1.0/drives/b%2154S...8PlFxA/items/01GI...Y2O": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

meain:
I've got best case scenario to be a speedup from ~5m to <20s, but most of the time we seem to  be slowed down by MS API.

```
Connecting to repository:      5s done
Connecting to repository:      5s done
Connecting to M365:      1s done
Discovering items to backup:      2s done
Discovered 1007 items to backup
Backing up data:      8s done
  Started At            ID                                    Status                Selectors
  2022-11-28T08:18:32Z  1e297743-609e-4d16-a3c4-1a5afaa7fac6  Completed (0 errors)  JohannaL@8qzvrj.onmicrosoft.com

real    0m20.306s
user    0m36.820s
sys     0m3.789s
```

meain:
Closing this issue for now as we are mostly bottle necked by the graph api throttling and how fast we can download/upload things.

PR #1607, opened 2022-11-28
Parallelize OneDrive queries to get URL

meain:
## Description

This improves the initial backup speed for OneDrive. OneDrive backup was mostly slow when we had a lot of tiny files. Case where we had mostly large files was pretty much the best case scenario and we were throttled by purely how fast we can get the files from MS and how fast kopia can process and upload it. But of small files, we were slowed by the loop which was taking quite a bit of time to fetch the download urls. We have now parallelized the query for getting the download URL. Under best case scenarios, I was able to speed it up to under 20s from ~4-5m starting point. That said, MS graph api still seems to throttle us and when that happen we still go back to around ~2m for worst case scenario. I've added 3 retries as some requests were failing when we continuously making many requests.

This should also take care of the issue of url expiring mentioned in https://github.com/alcionai/corso/issues/581 as we are only prefetching a few urls ahed of time.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* https://github.com/alcionai/corso/issues/1595
* fixes https://github.com/alcionai/corso/issues/581

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,8 +9,7 @@ BAD_LINT_MSG := "Missing golangci-lint version $(WANTED_LINT_VERSION). Visit $(I
 .PHONY: check-lint check-lint-version lint load-test
 
 build:
-	go build -o corso -ldflags \
-	"-X 'github.com/alcionai/corso/src/cli/version.Version=$(shell git describe --exact-match --tags $(git rev-parse HEAD) 2>/dev/null || echo unreleased)-$(shell git rev-parse --short HEAD)'"
+	go build -o corso
 
 lint: check-lint-version
 	golangci-lint run
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,5 +1,5 @@
 # This must match the version defined in .github/workflows/lint.yaml.
-WANTED_LINT_VERSION := 1.45.2
+WANTED_LINT_VERSION := 1.50.1
 LINT_VERSION := $(shell golangci-lint version | cut -d' ' -f4)
 HAS_LINT := $(shell which golangci-lint)
 
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -23,11 +23,11 @@ const (
 	collectionChannelBufferSize = 50
 
 	// TODO: Tune this later along with collectionChannelBufferSize
-	urlPrefetchChannelBufferSize = 50
+	urlPrefetchChannelBufferSize = 25
 
 	// Max number of retries to get doc from M365
 	// Seems to timeout at times because of multiple requests
-	maxRetries = 5
+	maxRetries = 4 // 1 + 3 retries
 )
 
 var (
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -123,6 +123,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 		byteCount int64
 		itemsRead int64
 		wg        sync.WaitGroup
+		m         sync.Mutex
 	)
 
 	// Retrieve the OneDrive folder path to set later in
@@ -145,7 +146,9 @@ func (oc *Collection) populateItems(ctx context.Context) {
 	defer close(semaphoreCh)
 
 	errUpdater := func(id string, err error) {
+		m.Lock()
 		errs = support.WrapAndAppend(id, err, errs)
+		m.Unlock()
 	}
 
 	for _, itemID := range oc.driveItemIDs {
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -182,10 +182,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 			}
 
 			if err != nil {
-				if err != nil {
-					errUpdater(itemID, err)
-				}
-
+				errUpdater(itemID, err)
 				return
 			}
 
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -122,8 +122,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 		errs      error
 		byteCount int64
 		itemsRead int64
-
-		wg sync.WaitGroup
+		wg        sync.WaitGroup
 	)
 
 	// Retrieve the OneDrive folder path to set later in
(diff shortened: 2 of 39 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -129,37 +142,92 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
 
+	limitCh := make(chan struct{}, urlPrefetchChannelBufferSize)
+	defer close(limitCh)
+
+	type uerr struct {
+		itemID string
+		err    error
+	}
+
+	errCh := make(chan uerr, len(oc.driveItemIDs))
```
ryanfkeepers, line 153: I think the extra channel here adds some unnecessary complication.  Consider an alternative, that follows our other pattern of error aggregation with an anonymous func:
```go
var errs error
errUpdater := func(id string, err error) {
	errs = support.WrapAndAppend(id, err, errs)
}
```
This pattern removes the possibility of accidentally blocking on the error reporting channel (a common mistake when handling channels, which could appear in a future change, even if it's currently safe).  It would also reduce code by removing the need for the `ffail` counter, the `uerr` struct, and the loop at the end of the func.

Error aggregation on 189 could be reduced to:
```go
if err != nil {
  errUpdater(id, err)
}
```
Fast fail checks then on 158 turn into:
```go
if oc.service.ErrPolicy() && errs != nil {
```

src/internal/connector/onedrive/collection.go
```diff
@@ -129,37 +142,92 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
-			errs = support.WrapAndAppendf(itemID, err, errs)
+		if ffail > 0 {
+			break
+		}
+		limitCh <- struct{}{}
+
+		wg.Add(1)
+
+		go func(itemID string) {
+			defer func() { <-limitCh }()
```
ryanfkeepers, line 166: Due to the goroutine there's a possibility of a panic here if limitCh has been closed as a result of early exit on fast-fail.  A panic recovery would be helpful.  You can additionally protect against this by setting `limitCh` to `nil` on close, and having this defer first check `if limitCh != nil`.

src/internal/connector/onedrive/collection.go
```diff
@@ -129,37 +142,92 @@ func (oc *Collection) populateItems(ctx context.Context) {
 	defer colCloser()
 	defer close(folderProgress)
 
+	limitCh := make(chan struct{}, urlPrefetchChannelBufferSize)
```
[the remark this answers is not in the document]
meain: Renaming to `semaphoreCh`

src/internal/connector/onedrive/collection.go
```diff
@@ -129,37 +142,92 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+	defer close(limitCh)
+
+	type uerr struct {
+		itemID string
+		err    error
+	}
+
+	errCh := make(chan uerr, len(oc.driveItemIDs))
+	defer close(errCh)
```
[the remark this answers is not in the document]
meain: I'll retain them just to be explicit about it.

src/internal/connector/onedrive/collection.go
meain: Since we have a `wg.Wait()` below before we close, I don't think we will run into this. That said, I have flipped the order of the defer statement so that `wg.Done()` happens after reading from the channel.

src/internal/connector/onedrive/collection.go
meain: Maybe I am missing something but would this be concurrent safe? Since we are not likely to hit many errors, I can add add locks and update them in place as another alternative.

src/Makefile
```diff
@@ -9,8 +9,7 @@ BAD_LINT_MSG := "Missing golangci-lint version $(WANTED_LINT_VERSION). Visit $(I
 .PHONY: check-lint check-lint-version lint load-test
 
 build:
-	go build -o corso -ldflags \
-	"-X 'github.com/alcionai/corso/src/cli/version.Version=$(shell git describe --exact-match --tags $(git rev-parse HEAD) 2>/dev/null || echo unreleased)-$(shell git rev-parse --short HEAD)'"
+	go build -o corso
```
meain, line 12: When building locally, there is no point in pinning a version plus more often than not, the git state will be dirty.

src/internal/connector/onedrive/collection.go
ryanfkeepers: Add mutexes as needed, sure.  In place vs calling an async func is just a matter of cleanliness/code reduction.

src/Makefile
ryanfkeepers: The change is fine, but let's keep separate concerns split into multiple PRs in the future, please.

src/internal/connector/onedrive/collection.go
```diff
@@ -129,37 +141,76 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
-			continue
-		}
-		// Item read successfully, add to collection
-		itemsRead++
-		// byteCount iteration
-		byteCount += itemInfo.Size
+			if err != nil {
+				if err != nil {
+					errUpdater(itemID, err)
+				}
```
ryanfkeepers, line 187:
```suggestion
			if err != nil {
				errUpdater(itemID, err)
```

ryanfkeepers (approved):
Couple more minor things to clean up (as a nit, also wouldn't mind seeing the suggested simplification in retry counter from the other PR, too), but otherwise looks good.

In reference to the document about maximum concurrent requests that you linked in the other PR, I feel like this behavior could possibly run into oneDrive's 5-per-user concurrency limit.  Have you run tests to try and surface that condition specifically yet?

src/internal/connector/onedrive/collection.go
meain: Woopsie

meain:
@ryanfkeepers The concurrency limit of 4 was only for Exchange and not OneDrive. There does not seem to be any such limits for OneDrive.

src/internal/connector/onedrive/collection.go
meain: I've added locks within `errUpdater`

ryanfkeepers:
> The concurrency limit of 4 was only for Exchange and not OneDrive. There does not seem to be any such limits for OneDrive.

Doh!  I was thinking of the OneNote constraints, not OneDrive.

Issue #362, opened 2022-07-19
GC: Add routines for large operations


dadams39:
Related to Issue #360 and #356
Large requests can exceed the amount of time set aside for testing and general customer satisfaction. Need to add the ability to implement a divide-and-conquer abstraction on large tasks.

- [x] Add a cap for the number of allowable thread workers
- [ ] Ensure that all Collections retrieval operations are parallelized as much as possible to reduce the  are given separate workers to prevent deadlocking when possible

PR #1608, opened 2022-11-28
Speed up Exchange backup by parallelizing url fetch

meain:
## Description

From rough numbers we can speedup an account with ~3000 emails, ~1000 contacts and ~1000 events from ~18m to <3m.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* https://github.com/alcionai/corso/issues/362
* https://github.com/alcionai/corso/issues/1595
* https://github.com/alcionai/corso/pull/1607

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -152,7 +152,7 @@ func (col *Collection) populateByOptionIdentifier(
 		err  error
 	}
 
-	errCh := make(chan uerr)
+	errCh := make(chan uerr, len(col.jobs))
 	defer close(errCh)
 
 	ffail := int64(0)
@@ -179,7 +179,9 @@ func (col *Collection) populateByOptionIdentifier(
 					break
 				}
 				// TODO: Tweak sleep times
-				time.Sleep(time.Duration(3*(i+1)) * time.Second)
+				if i != numberOfRetries-1 {
+					time.Sleep(time.Duration(3*(i+1)) * time.Second)
+				}
 			}
 
 			if err != nil {
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -122,6 +122,8 @@ func (col *Collection) populateByOptionIdentifier(
 		success    int64
 		totalBytes int64
 
+		wg sync.WaitGroup
+
 		user = col.user
 	)
 
@@ -144,12 +146,12 @@ func (col *Collection) populateByOptionIdentifier(
 
 	limitCh := make(chan struct{}, urlPrefetchChannelBufferSize)
 	defer close(limitCh)
-	var wg sync.WaitGroup
 
 	type uerr struct {
 		user string
 		err  error
 	}
+
 	errCh := make(chan uerr)
 	defer close(errCh)
 
@@ -161,6 +163,7 @@ func (col *Collection) populateByOptionIdentifier(
 		limitCh <- struct{}{}
 
 		wg.Add(1)
+
 		go func(identifier string) {
 			defer func() { <-limitCh }()
 			defer wg.Done()
@@ -169,6 +172,7 @@ func (col *Collection) populateByOptionIdentifier(
 				response absser.Parsable
 				err      error
 			)
+
 			for i := 0; i < numberOfRetries; i++ {
 				response, err = query(ctx, col.service, user, identifier)
 				if err == nil {
@@ -177,9 +181,9 @@ func (col *Collection) populateByOptionIdentifier(
 				// TODO: Tweak sleep times
 				time.Sleep(time.Duration(3*(i+1)) * time.Second)
 			}
+
 			if err != nil {
 				errCh <- uerr{user, err}
-				// errs = support.WrapAndAppendf(user, err, errs)
 
 				if col.service.ErrPolicy() {
 					atomic.AddInt64(&ffail, 1)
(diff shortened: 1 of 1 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -139,33 +144,83 @@ func (col *Collection) populateByOptionIdentifier(
...
+				err      error
+			)
+
+			for i := 0; i < numberOfRetries; i++ {
+				response, err = query(ctx, col.service, user, identifier)
+				if err == nil {
+					break
+				}
+				// TODO: Tweak sleep times
+				if i != numberOfRetries-1 {
```
ryanfkeepers, line 182: nit: bit of a weird comparison.  I'd rather the code make more sense, even if the const seems off by one.
```suggestion
				if i >= numberOfRetries {
```

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -139,33 +144,83 @@ func (col *Collection) populateByOptionIdentifier(
 		return
 	}
 
+	limitCh := make(chan struct{}, urlPrefetchChannelBufferSize)
+	defer close(limitCh)
```
dadams39, line 148: Could this be added to the defer on  Line 135?

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -139,33 +144,83 @@ func (col *Collection) populateByOptionIdentifier(
...
+		if ffail > 0 {
+			break
+		}
+		limitCh <- struct{}{}
+
+		wg.Add(1)
+
+		go func(identifier string) {
+			defer func() { <-limitCh }()
+			defer wg.Done()
```
dadams39, line 169: These can be grouped. 

Additionally, a comment would be helpful.

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -139,33 +144,83 @@ func (col *Collection) populateByOptionIdentifier(
...
+
+			var (
+				response absser.Parsable
+				err      error
+			)
```
dadams39, line 174: There are several different code declarations. Could it be simplified into the top of the function?

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -31,8 +34,9 @@ var (
 )
 
 const (
-	collectionChannelBufferSize = 1000
-	numberOfRetries             = 4
+	collectionChannelBufferSize  = 1000
+	numberOfRetries              = 4
+	urlPrefetchChannelBufferSize = 50
```
vkamra, line 39: @meain - see https://learn.microsoft.com/en-us/graph/throttling-limits#outlook-service-limits. It seems like we shouldn't have > 4 concurrent requests.

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -139,33 +144,83 @@ func (col *Collection) populateByOptionIdentifier(
...
+			)
+
+			for i := 0; i < numberOfRetries; i++ {
+				response, err = query(ctx, col.service, user, identifier)
+				if err == nil {
+					break
+				}
+				// TODO: Tweak sleep times
+				if i != numberOfRetries-1 {
+					time.Sleep(time.Duration(3*(i+1)) * time.Second)
```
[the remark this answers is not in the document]
meain: IIRC the timeouts are reset every 10 seconds for GC, so my idea was to accumulate ~10s in total after all the retries. But let me see if I can tweak these numbers after going through how the throttling works.

src/internal/connector/exchange/exchange_data_collection.go
meain: I don't think `i >= numberOfRetries` will work. I think I'll probably change the code to the following to make it more explicit though:

```go
for i := 1; i <= numberOfRetries; i++ {
    ...
    if i < numberOfRetries {
        time.Sleep(time.Duration(3*(i+1)) * time.Second)
    }
}
```

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -115,19 +119,20 @@ func (col *Collection) populateByOptionIdentifier(
 ) {
 	var (
 		errs       error
-		success    int
+		success    int64
```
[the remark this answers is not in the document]
meain: I had to switch it to either `int32` or `int64` as I needed to do an [atomic increment](https://github.com/alcionai/corso/blob/b1b351b449f046a0282d88b2989403505b59313d/src/internal/connector/exchange/exchange_data_collection.go#L192). And since `int` will be `int64` on most platforms we will be deploying, I thought I would go with `int64`.

src/internal/connector/exchange/exchange_data_collection.go
meain: I thought it would be better to couple it with the `make` as the defer above is just for the progressbar.

src/internal/connector/exchange/exchange_data_collection.go
meain: These are the only declarations in the func called by goroutines.

src/internal/connector/exchange/exchange_data_collection.go
meain: Any specific reason for grouping them? Let me add comments to it though.

meain: I've added comments to the channel creation.

src/internal/connector/exchange/exchange_data_collection.go
ryanfkeepers: Hah, good catch.  My brain was backwards.

ryanfkeepers (approved):

src/internal/connector/exchange/exchange_data_collection.go
meain: The reason for the conditional was so that it would not sleep even with an error in the last run. If we skip that, it sleeps at the end of `n-1th` run before exiting out of the loop.

Issue #1602, opened 2022-11-23
Improve userId and repoId semantics for Rudderstack events


gmatev:
**Context:** Currently, `repoId` in telemetry events is a hash of the M365 id that Corso is deployed against. RepoId is also used to identify what constitutes a "user" in our telemetry. As a result, no matter how many Corso deployments use the same M365 domain, they will be reported

**Proposal**: 
* Move to a model where `repoId` actually represents an ID of the specific repository that Corso is connected to 
* Use the same ID as `userId` in Identify calls. Identify calls require a name so use `userId` as the name as well. 
* Introduce `m365_tenant_hash` as user Trait in the Identify call

gmatev:
This is also complete.

PR #1619, opened 2022-11-29
migrate tenant id out of event repoID

ryanfkeepers:
## Description

Moves the tenantID out of the repoID hash and
into its own value (also hashed) in the event log.

Unique user identification to follow.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1602

## Test Plan

- [x] :green_heart: E2E


src/internal/events/events.go
```diff
@@ -117,6 +118,7 @@ func (b Bus) Event(ctx context.Context, key string, data map[string]any) {
 	props := analytics.
 		NewProperties().
 		Set(repoID, b.repoID).
+		Set(tenantID, b.tenant).
```
gmatev, line 121: `tenantID` is also needed as a Trait in the Identify call

gmatev (changes requested):
`tenantID` is also needed as a Trait in the Identify call

src/internal/events/events.go
ryanfkeepers: Okay, will re-add.

gmatev:
@ryanfkeepers I think my initial comment was misleading. Having both IDs as event dimensions is good by my point was for the actual identify code. Was not sure who to leave a comment in part that has not changed. You want something like 

```
err := b.client.Enqueue(analytics.Identify{
			UserId: b.repoID,
			Traits: analytics.NewTraits().
				SetName(b.repoID),
                                 Set(tenantID, b.tenant),
		})
```

gmatev (approved):
I think this is good for this PR. We should file a ticket for a follow-up that captures the work needed to have a repoId based on an actual unique per repo identifier

Issue #1621, opened 2022-11-29
Selectors: Clone and Set Functionality


ryanfkeepers:
Needed to unblock Ashlie.  See conversation at: https://github.com/alcionai/corso/pull/1609#discussion_r1034121118

PR #1626, opened 2022-11-29
add a discrete clone method to scopes

ryanfkeepers:
## Description

Adds a method to all scopes that produces a
discrete clone of the original scope.  The result is a shallow clone of the scope, thus retaining all
original values, but is made discrete by replacing the resource owner target with a specific value.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1621

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/selectors/onedrive.go
+++ b/src/pkg/selectors/onedrive.go
@@ -459,6 +459,29 @@ func (s OneDriveScope) setDefaults() {
 	}
 }
 
+// DiscreteCopy makes a clone of the scope, then replaces the clone's user comparison
+// with only the provided user.
+func (s OneDriveScope) DiscreteCopy(user string) OneDriveScope {
+	return discreteCopy(s, user)
+}
+
+// ---------------------------------------------------------------------------
+// Backup Details Filtering
+// ---------------------------------------------------------------------------
+
+// Reduce filters the entries in a details struct to only those that match the
+// inclusions, filters, and exclusions in the selector.
+func (s oneDrive) Reduce(ctx context.Context, deets *details.Details) *details.Details {
+	return reduce[OneDriveScope](
+		ctx,
+		deets,
+		s.Selector,
+		map[path.CategoryType]oneDriveCategory{
+			path.FilesCategory: OneDriveItem,
+		},
+	)
+}
+
 // matchesInfo handles the standard behavior when comparing a scope and an oneDriveInfo
 // returns true if the scope and info match for the provided category.
 func (s OneDriveScope) matchesInfo(dii details.ItemInfo) bool {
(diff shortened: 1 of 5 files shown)


src/pkg/selectors/onedrive.go
```diff
@@ -459,6 +459,29 @@ func (s OneDriveScope) setDefaults() {
...
+	return discreteCopy(s, user)
+}
+
+// ---------------------------------------------------------------------------
+// Backup Details Filtering
+// ---------------------------------------------------------------------------
+
+// Reduce filters the entries in a details struct to only those that match the
+// inclusions, filters, and exclusions in the selector.
+func (s oneDrive) Reduce(ctx context.Context, deets *details.Details) *details.Details {
```
ryanfkeepers, line 474: Weird diff: this func wasn't changed.  What changed is that `matchesInfo()` was moved below it.

src/pkg/selectors/sharepoint.go
```diff
@@ -407,6 +407,30 @@ func (s SharePointScope) setDefaults() {
...
+// site comparison with only the provided site.
+func (s SharePointScope) DiscreteCopy(site string) SharePointScope {
+	return discreteCopy(s, site)
+}
+
+// ---------------------------------------------------------------------------
+// Backup Details Filtering
+// ---------------------------------------------------------------------------
+
+// Reduce filters the entries in a details struct to only those that match the
```
ryanfkeepers, line 420: Weird diff: this func wasn't changed.  What changed is that `matchesInfo()` was moved below it.

src/pkg/selectors/scopes_test.go
```diff
@@ -513,3 +513,20 @@ func (suite *SelectorScopesSuite) TestScopeConfig() {
...
+	)
+
+	for k, v := range orig {
+		if k != rootCatStub.String() {
+			assert.Equal(t, v.Target, clone[k].Target)
+		} else {
+			assert.Equal(t, AnyTgt, v.Target)
+			assert.Equal(t, "fnords", clone[k].Target)
+		}
+	}
```
ashmrtn, line 531: could you please also add an assertion to make sure the original has not changed as well. I know it seems obvious, but sometimes assumptions like these are what come back to get you in the end

ashmrtn (approved):

src/pkg/selectors/scopes_test.go
ryanfkeepers: The assertion on 528 is precisely that.

ashmrtn: 🤦‍♀️

PR #1633, opened 2022-11-29
Backup: SharePoint:  List content retrieval 

### LINKED ISSUE #1475: GC: Backup: SharePoint List: Populate Function
Create functions for backup Sharepoint list

dadams39:
## Description
Logic for List retrieval from M365 back store. Keeping separate from Collection creation logic as there are a lot of calls to retrieve all supporting relationship data from the backstore.  
Adds collection to 
## Type of change

- [x] :sunflower: Feature

## Issue(s)

*closes  #1475<issue>

## Test Plan

- [x] :zap: Unit test


src/internal/connector/sharepoint/list.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+
+		if resp.GetOdataNextLink() == nil {
+			break
+		}
+
+		builder = tc.NewColumnsRequestBuilder(*resp.GetOdataNextLink(), gs.Adapter())
+	}
+
+	return cs, nil
```
dadams39, line 259: These are identical. Unfortunately, the `Get()` functions are library dependent. @ashmrtn If you know a good way to combine these functions, I would greatly appreciate it.

src/internal/connector/sharepoint/collection.go
```diff
@@ -100,3 +105,63 @@ func (sc *Collection) finishPopulation(ctx context.Context, success int, totalBy
...
+
+				success++
+				sc.data <- &Item{
+					id:   *lst.GetId(),
+					data: io.NopCloser(bytes.NewReader(byteArray)),
+					info: sharePointListInfo(lst, arrayLength),
+				}
+
+				colProgress <- struct{}{}
+
```
ryanfkeepers:
```suggestion
```

src/internal/connector/sharepoint/list.go
```diff
@@ -0,0 +1,452 @@
...
+// loadLists is a utility function to populate the List object.
+// @param identifier the M365 ID that represents the SharePoint Site
+// Makes additional calls to retrieve the following relationships:
+// - Columns
+// - ContentTypes
+// - List Items
+func loadLists(
+	ctx context.Context,
+	gs graph.Service,
+	identifier string,
```
ryanfkeepers, line 37: Nit, throughout this file:
```suggestion
	id string,
```

src/internal/connector/sharepoint/list.go
```diff
@@ -0,0 +1,452 @@
...
+) ([]models.Listable, error) {
+	var (
+		prefix  = gs.Client().SitesById(identifier)
+		builder = prefix.Lists()
+		listing = make([]models.Listable, 0)
```
ryanfkeepers, line 42:
```suggestion
		results = make([]models.Listable, 0)
```

src/internal/connector/sharepoint/list.go
ryanfkeepers:
```go
// define
type columnGetter func(ctx context.Context) (models.ColumnDefinitionCollectionResponseable, error)

// swap the func param libraries for this
func fetchColumns(
	ctx context.Context,
	gs graph.Service,
	identifier string,
	get columnGetter,
) ([]models.ColumnDefinitionable, error) {

// call the func with an anonymous func
fetchColumns(
  ctx, gs, id,
  func(ctx context.Context) (models.ColumnDefinitionCollectionResponseable, error) {
    return cb.Get(ctx, nil)
  })
```

src/internal/connector/sharepoint/list.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+
+		builder = lists.NewListsRequestBuilder(*resp.GetOdataNextLink(), gs.Adapter())
+	}
+
+	return listing, nil
+}
+
+// fetchListItems utility for retrieving ListItem data and the associated relationship
+// data. All additional calls(3) return a specific object and not a collection.
```
ryanfkeepers, line 94:
```suggestion
// data. All additional calls(3) append data to the tracked items, and do not create additional collections.
```

src/internal/connector/sharepoint/list.go
```diff
@@ -0,0 +1,452 @@
...
+
+	for {
+		resp, err := builder.Get(ctx, nil)
+		if err != nil {
+			return nil, errors.Wrap(err, support.ConnectorStackErrorTrace(err))
+		}
+
+		for _, link := range resp.GetValue() {
+			links = append(links, link)
+		}
```
ryanfkeepers, line 412: Fun tip: you can append slices without iteration by using a spread operator:
```suggestion
		links = append(links, resp.GetValue()...)
```

ryanfkeepers (approved):
Code looks fine, mostly nits.

Per the review size, it could have been reduced by stubbing out the column, content type, and list item retrieval funcs, and saving each for a future PR addition.

src/internal/connector/sharepoint/collection.go
dadams39: White space removed.

src/internal/connector/sharepoint/collection.go
```diff
@@ -100,3 +105,63 @@ func (sc *Collection) finishPopulation(ctx context.Context, success int, totalBy
...
+	colProgress, closer := observe.CollectionProgress("name", sc.fullPath.Category().String(), sc.fullPath.Folder())
+	go closer()
+
+	defer func() {
+		close(colProgress)
+		sc.finishPopulation(ctx, success, totalBytes, errs)
+	}()
+
+	// sc.jobs contains query = all of the site IDs.
+	for _, identifier := range sc.jobs {
```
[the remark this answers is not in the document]
dadams39: Resolved in [1a289c0](https://github.com/alcionai/corso/pull/1633/commits/1a289c08e70c923a75064e5276e4aebddd4da124)

src/internal/connector/sharepoint/list_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+type SharePointSuite struct {
+	suite.Suite
+	userID string
+	creds  account.M365Config
+}
+
+func (suite *SharePointSuite) SetupSuite() {
+	t := suite.T()
+	suite.userID = tester.SecondaryM365UserID(suite.T())
```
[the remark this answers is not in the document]
dadams39: Resolved [82aa23b](https://github.com/alcionai/corso/pull/1633/commits/82aa23b53c8b9e9b1a42276df53f818df2b39c43)

src/internal/connector/sharepoint/list.go
ashmrtn: other option would probably be to define an interface that would allow getting the request and return an interface that implements `Get()` for some return type. The caller would instantiate and pass the struct implementation of the interface into the function that does the iteration. Something like

```
type RequestBuilder interface {
  BaseRequest(<common params>) RequestGetter
  PageRequest(link string) RequestGetter
}

type RequestGetter interface {
  Get(ctx, <params type>) (models.ColumnDefinitionCollectionResponseable, error)
}
```

That assumes the `<params type>` is the same for both. If it's not then either `any` and a type assertion/type cast would be needed in the struct implementation of the interface or generics would be needed in some way

Also `RequestGetter` is probably a bad name because it's slightly confusing but I don't know of a better one right now heh

src/internal/connector/sharepoint/list.go
dadams39: Resolved as of [6e7fdb2](https://github.com/alcionai/corso/pull/1633/commits/6e7fdb245276f7ff194e8d6e1e37f79110246330)

src/internal/connector/sharepoint/list.go
dadams39: Resolved as of [6e7fdb2](https://github.com/alcionai/corso/pull/1633/commits/6e7fdb245276f7ff194e8d6e1e37f79110246330)

src/internal/connector/sharepoint/list.go
dadams39: Resolved as of [6e7fdb2](https://github.com/alcionai/corso/pull/1633/commits/6e7fdb245276f7ff194e8d6e1e37f79110246330)

src/internal/connector/sharepoint/list.go
dadams39: The params type are not the same for both.
```go
func (*columns.ColumnsRequestBuilder).Get(ctx context.Context, requestConfiguration *columns.ColumnsRequestBuilderGetRequestConfiguration)
```
vs.
```go
func (*tc.ColumnsRequestBuilder).Get(ctx context.Context, requestConfiguration *tc.ColumnsRequestBuilderGetRequestConfiguration) (models.ColumnDefinitionCollectionResponseable, error)
```
They are two different packages. They do the same thing. However, they are not interchangeable.

dadams39: I'm going to refactor as I can, but both of the approaches don't seem to work well with the Microsoft libraries.

dadams39: Changes in [017416a](https://github.com/alcionai/corso/pull/1633/commits/017416a93c48cb0dcd88b052304e004617b91e14)
@ryanfkeepers @ashmrtn, I will let these changes sit as I work on an attempt to upgrade the version of the sdk as the lack of updates are causing some errors.

ryanfkeepers: @dadams39 I think you're going to take on more overhead by switching contexts and affecting changes otherwise.  Since it looks like the primary churn here involves querying facets of a list (columns, etc), how about first taking my suggestion to stub those parts out for this PR?  You can get the boilerplate list retrieval into place- which is less code in flight/maintenance while you handle library updates- and handle the column retrieval implementation and refinements in a follow-up PR.

Issue #1615, opened 2022-11-28
SharePoint Libraries Restore


ryanfkeepers:
## Issue
Implement restore functionality for SharePoint libraries.
- ~Restricted directories~ #2644 
- [x] Backup Details Compliance #1747 
- [x] Fix restore workflow #2489
-  ~e2e tests~ Finish in separate ticket #2525

dadams39:
Required issues for SharePoint.Libraries restore to be completed:
- #1936
- #1938 
- #2064

dadams39:
System Document libraries were excluded on 2/24/2023. 

Separate tracker for the issue: 
- https://github.com/alcionai/corso/issues/2644

PR #1636, opened 2022-11-30
add sharepoint library restore

ryanfkeepers:
## Description

adds restore boilerplate for sharepoint libraries.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1615

## Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E


src/internal/connector/onedrive/restore.go
```diff
@@ -85,10 +85,10 @@ func RestoreCollections(
 		nil
 }
 
-// restoreCollection handles restoration of an individual collection.
-// @returns Integer representing totalItems, restoredItems, and the
-// amount of bytes restored. The bool represents whether the context was cancelled
-func restoreCollection(
+// RestoreCollection handles restoration of an individual collection.
+// returns the metrics gathered during restoration.
+// returns true if the context was cancelled.
```
dadams39, line 90: This is a little ambiguous with two statements that say `returns`. If the previous was overkill, there should be a healthy middle ground that gives clarity to the function.

src/internal/connector/sharepoint/restore.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/pkg/errors"
+)
+
+// RestoreCollections will restore the specified data collections into OneDrive
+func RestoreCollections(
+	ctx context.Context,
+	service graph.Service,
+	dest control.RestoreDestination,
+	dcs []data.Collection,
+	deets *details.Details,
```
dadams39, line 22: SharePoint doesn't have a way to restore via the GraphAPI. Would need to verify that this library would work via the current APIs for a sharepoint drive.  This might require a mock collection to be added prior to this PR.

ryanfkeepers: GraphAPI doesn't have write control for _sites_, but it does interface with _driveItems_.  Since all libraries and library items are handled through the drives api, there's no blocker on using the graph api (or, more accurately, our graph-constructed api calls) for restoration.

src/internal/connector/onedrive/restore.go
ryanfkeepers: Fair, I've updated it.

ryanfkeepers:
@dadams39 This is ready for a final review.

dadams39 (approved):

PR #1642, opened 2022-11-30
adds a flag for more readable console logs

ryanfkeepers:
## Description

I was getting frustrated by the readability of logs
in the console, and wanted reduce some of my
eye strain during development.
Adds a hidden "readable-logs" flag for ease of
reading logs during development.

## Type of change

- [x] :hamster: Trivial/Minor

Logs without flag:
image

Logs with flag:
image


meain (approved):
Just a thought, should we make this the default and provide a flag to output json logs for consumption by other tools?

ryanfkeepers:
>  Just a thought, should we make this the default and provide a flag to output json logs for consumption by other tools?

If the team wants it, we can easily make that change in the future.

Issue #1622, opened 2022-11-29
Add backup data statistics to Backup model


vkamra:
This tracks adding additional statistics on the backup (logical data, etc.) to the `Backup` model and displaying via `list backups` cmd

vkamra:
@meain - the Kopia snapshot manifest has stats related to the total backup size - we should push this data into the Backup model *or* retrieve this when the backup model is enumerated.

This is what a Kopia snapshot manifest contains for example:
```
$ ./kopia manifest show c794d43ade1de712549d44132633b17b
// id: c794d43ade1de712549d44132633b17b
// length: 815
// modified: 2022-11-04 17:13:44 PDT
// label type:snapshot
// label username:corso
// label exchangecontacts:
// label exchangeemail:
// label exchangeevents:
// label hostname:corso-host
// label path:NGQ2MDMwNjAtMThkNi00NzY0LWI5YmUtNGNiNzk0ZDMyYjY5
{

  "description": "",
  "startTime": "2022-11-05T00:13:01.105603Z",
  "endTime": "2022-11-05T00:13:44.248126Z",
  "stats": {
    "totalSize": 36541799,
    "excludedTotalSize": 0,
    "fileCount": 574,
    "cachedFiles": 0,
    "nonCachedFiles": 574,
    "dirCount": 32,
    "excludedFileCount": 0,
    "excludedDirCount": 0,
    "ignoredErrorCount": 0,
    "errorCount": 0
  },
  "rootEntry": {
    "name": "NGQ2MDMwNjAtMThkNi00NzY0LWI5YmUtNGNiNzk0ZDMyYjY5",
    "type": "d",
    "mode": "0777",
    "mtime": "1754-08-30T22:43:41.128654848Z",
    "obj": "k1159f09730fb8942efe71682eb65bbad",
    "summ": {
      "size": 36541799,
      "files": 574,
      "symlinks": 0,
      "dirs": 32,
      "maxTime": "2022-11-05T00:13:44.247365Z",
      "numFailed": 0
    }
  },
  "tags": {

    "exchangecontacts": "",
    "exchangeemail": "",
    "exchangeevents": ""
  }```

meain:
I think if we can pull this info without explicitly adding to `Backup`, I'll go that route as we can avoid duplicating the data in both places.

meain:
[...]

PR #1648, opened 2022-12-01
Show backup size for list commands

meain:
## Description

Add size and stored size to cli output for `backup list`

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/1622

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/backup/backup.go
```diff
@@ -95,6 +97,8 @@ type Printable struct {
 	Status     string              `json:"status"`
 	Version    string              `json:"version"`
 	Selectors  selectors.Printable `json:"selectors"`
+	Size       int64               `json:"size"`
+	StoredSize int64               `json:"storedSize"`
```
ryanfkeepers, line 101: I'm not sure we want/need these displayed on the tabular output for the list.  Our design for that display is more directed to "help the user identify the backup they need" than "show the user all the backup metadata", and extra columns can make the former identification less clear.  Users who need access to metadata will be able to retrieve it with a `--json` output, and sdk users, like you noted, already have access within the struct.

I realize that kind-of makes this a non-issue.  Maybe it'd be good to connect with some sdk consumers and make sure there wasn't a miscommunication about what data needs to be provided.  The ask my extend beyond the sizes we already store.

src/pkg/backup/backup.go
```diff
@@ -117,6 +123,8 @@ func (b Backup) Headers() []string {
 		"ID",
 		"Status",
 		"Selectors",
+		"Size",
```
vkamra, line 126: @meain - I'm thinking that we should just call this BytesRead and BytesUploaded in the JSON output - since it's the most accurate description.

vkamra (approved):

src/pkg/backup/backup.go
meain: I've renamed them to `BytesRead` and `BytesUploaded`. I've also removed them from cli listing keeping them only in the json output as discussed.

PR #1656, opened 2022-12-01
GC: Resolver: `exchange.Mail` versioning mail bug fix

dadams39:
## Description
The changes in `msgraph-sdk-go` version 40+  causes the `Inbox` not to be returned with the `gs.Client().UsersById(user).MailFolders().Get()` call.  The solution is to manually populate Inbox with the initialization. 

This is in addition to the previous change for the GetContainerID. 

## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/service_functions.go
+++ b/src/internal/connector/exchange/service_functions.go
@@ -204,12 +204,9 @@ func pathAndMatch(
 		category,
 		false,
 	)
+	// Only true for containers without a path e.g. Root mail folder
 	if err != nil {
 		return nil, false
-	}
-
-	if dirPath == nil && category == path.EmailCategory {
-		return nil, false // Only true for root mail folder
 	}
 
 	directory = pb.String()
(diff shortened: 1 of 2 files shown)


ryanfkeepers:
@dadams39 This PR contains diffs from other PRs, which makes it difficult to separate your changes from the chaff.  Could we get rebase that isolates your changes?  Please and thank you!

src/internal/connector/exchange/service_functions.go
```diff
@@ -204,14 +204,11 @@ func pathAndMatch(
 		category,
 		false,
 	)
+	// Only true for containers without a path e.g. Root mail folder
```
ryanfkeepers, line 207: I find this comment a little confusing.  `Only true` could be interpreted as either the return value or the conditional.  Is the following change more accurate?
```suggestion
	// Containers without a path (e.g. Root mail folder) always err here.
```

ryanfkeepers (approved):

Issue #1616, opened 2022-11-28
Resolve SharePoint identifier and path design


ryanfkeepers:
## Issue
Decide if and how to handle identifying sharepoint sites.  Specifically whether, we should store and identify sites by ID, webURL, or both. 

## Internal Discussion
[Here](https://www.notion.so/alcion/Sharepoint-Site-Identification-7b33b28162c445bca34b83d9db1c3f9d)

PR #1659, opened 2022-12-01
add suffix and pathSuffix filters

ryanfkeepers:
## Description

Preparing for handling sharepoint webUrl
selectors, we will need suffix-matching filters.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1616

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/filters/filters_test.go
+++ b/src/pkg/filters/filters_test.go
@@ -206,6 +206,31 @@ func (suite *FiltersSuite) TestPrefixes() {
 	}
 }
 
+func (suite *FiltersSuite) TestSuffixes() {
+	target := "folderB"
+	f := filters.Suffix(target)
+	nf := filters.NotSuffix(target)
+
+	table := []struct {
+		name     string
+		input    string
+		expectF  assert.BoolAssertionFunc
+		expectNF assert.BoolAssertionFunc
+	}{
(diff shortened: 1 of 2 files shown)


src/pkg/filters/filters_test.go
```diff
@@ -206,6 +206,31 @@ func (suite *FiltersSuite) TestPrefixes() {
...
+	table := []struct {
+		name     string
+		input    string
+		expectF  assert.BoolAssertionFunc
+		expectNF assert.BoolAssertionFunc
+	}{
+		{"Exact match - same case", "folderB", assert.True, assert.False},
+		{"Exact match - different case", "Folderb", assert.True, assert.False},
+		{"Suffix match - same case", "folderA/folderB", assert.True, assert.False},
+		{"Suffix match - different case", "Foldera/folderb", assert.True, assert.False},
```
ashmrtn, line 223: should probably have a case to check it matches `foldera/afolderb`

ashmrtn (approved):

src/pkg/filters/filters_test.go
ryanfkeepers: The case difference is in the suffix.  "folder`B`" - "folder`b`"

PR #1671, opened 2022-12-02
add a weburl to siteid reducer in gc

ryanfkeepers:
## Description

Adds a func in graphConnector that reduces
siteIDs and webURLs into a set of siteIDs.  This
will be used by callers such as the CLI to
generate id-based selectors for sites even if
they handle webURLs as an alternative id.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1616

## Test Plan

- [x] :zap: Unit test


src/internal/connector/graph_connector_test.go
```diff
@@ -20,6 +20,103 @@ import (
...
+	suite.Suite
+}
+
+func TestGraphConnectorUnitSuite(t *testing.T) {
+	suite.Run(t, new(GraphConnectorUnitSuite))
+}
+
+func (suite *GraphConnectorUnitSuite) TestUnionSiteIDsAndWebURLs() {
+	const (
+		url1  = "www.foo.com/bar"
```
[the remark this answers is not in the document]
ryanfkeepers: done

ryanfkeepers:
@meain This is ready for a final review.

src/internal/connector/graph_connector.go
```diff
@@ -232,19 +233,55 @@ func identifySite(item any) (string, string, error) {
...
+// GetSiteIds returns the canonical site IDs in the tenant
+func (gc *GraphConnector) GetSiteIDs() []string {
 	return buildFromMap(false, gc.Sites)
 }
 
+// UnionSiteIDsAndWebURLs reduces the id and url slices into a single slice of site IDs.
+// WebURLs will run as a path-suffix style matcher.  Callers may provide partial urls, though
+// each element in the url must fully match.  Ex: the webURL value "foo" will match "www.ex.com/foo",
+// but not match "www.ex.com/foobar".
+// The returned IDs are reduced to a set of unique values.
```
dadams39, line 253: What is the unique value returned? Comment could provide an example or be made more succinct.

src/internal/connector/graph_connector.go
```diff
@@ -232,19 +233,55 @@ func identifySite(item any) (string, string, error) {
...
+			continue
+		}
+
+		idm[id] = struct{}{}
+	}
+
+	idsl := make([]string, 0, len(idm))
+	for id := range idm {
+		idsl = append(idsl, id)
+	}
```
dadams39, line 280: I believe the buildFromMap function is available to use here.

src/internal/connector/graph_connector.go
```diff
@@ -232,19 +233,55 @@ func identifySite(item any) (string, string, error) {
...
+
+	match := filters.PathSuffix(urls)
+
+	for url, id := range gc.Sites {
+		if !match.Compare(url) {
+			continue
+		}
+
+		idm[id] = struct{}{}
```
dadams39, line 274: Would it be more efficient to remove the map and append it directly to idsl?

ryanfkeepers: We'd potentially get duplicate values as a result.

src/internal/connector/graph_connector.go
ryanfkeepers: Going to leave this like it is, because I plan to move these funcs into `Discovery` as a follow-up PR.  Might migrate `buildFromMap` to the common package when I do.

meain (approved):

dadams39 (approved):

PR #1679, opened 2022-12-05
Refactor user query and return user display name

vkamra:
## Description

Move user discovery into a `discovery` package. Also use that to return user display name from `m365.Users()` call.

This will also allow us to remove `Users` from the GC struct going forward.

If this pattern works - will follow up with a similar change for `Sites()`.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/discovery/discovery.go
```diff
@@ -0,0 +1,83 @@
...
+	userSelectDisplayName   = "displayName"
+)
+
+func Users(ctx context.Context, gs graph.Service, tenantID string) ([]models.Userable, error) {
+	users := make([]models.Userable, 0)
+
+	options := &msuser.UsersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &msuser.UsersRequestBuilderGetQueryParameters{
+			Select: []string{userSelectID, userSelectPrincipalName, userSelectDisplayName},
+		},
```
dadams39, line 27: Use of the call in query_options would ensure the options are tested or if there is a change involving the use of options  in the future. 

```go
opts, err := optionsForUsers(userSelectPrincipalName, userSelectDisplayName)
```

src/pkg/services/m365/m365.go
```diff
@@ -3,41 +3,59 @@ package m365
...
 )
 
+type User struct {
+	PrincipalName string
+	ID            string
+	Name          string
+}
+
 // Users returns a list of users in the specified M365 tenant
 // TODO: Implement paging support
```
dadams39, line 21: Comment requires update

src/internal/connector/discovery/discovery.go
vkamra: @dadams39 - it wasn't clear to me how that test was helping add any more coverage. What are we testing for?

src/internal/connector/discovery/discovery.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		users = append(users, u)
+
+		return true
+	}
+
+	if err := iter.Iterate(ctx, callbackFunc); err != nil {
+		return nil, errors.Wrap(err, support.ConnectorStackErrorTrace(err))
+	}
```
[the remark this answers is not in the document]
vkamra: @dadams39 - the linked call is a for a Delta API (which PageIterator) doesn't support I believe. What is the change you're requesting and how does that help?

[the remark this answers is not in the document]
dadams39: Line 40 suggests that we are using an iterator.

dadams39 (approved):

src/internal/connector/discovery/discovery.go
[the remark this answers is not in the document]
vkamra: @dadams39 and I synced on this offline. We have a slight preference in the exchange/onedrive code for using the other style but that is related to either using `Delta` APIs or `DeltaTokens` which the page iterator doesn't support.

Since we don't have a use case for those in the user query - going to stick with this.

src/internal/connector/discovery/discovery.go
vkamra: Discussed offline. `optionsForUsers` used to be under the exchange pkg. This `Users` call pulls out that logic into `connector/discovery` which can be used as a regression test for Graph changes.

PR #1682, opened 2022-12-05
Fix staticcheck lint in CI

meain:
## Description

Missed configuring working dir for staticcheck, plus fixes for errors reported.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/cmd/getM365/getItem.go
```diff
@@ -148,6 +148,7 @@ func runDisplayM365JSON(
 
 		fmt.Println(string(array))
 
+		//lint:ignore SA4004 only expecting one item
 		return nil
```
dadams39, line 152: Why is the lint ignore necessary here?

meain: We will unconditionally exit from the loop here after the first iter and that is flagged as a warning by default under [SA4004](https://staticcheck.io/docs/checks/#SA4004) check.

meain:
@dadams39 Any other issues, if not we should be good to go here.

PR #1687, opened 2022-12-05
populate sharepoint itemInfo

ryanfkeepers:
## Description

Currently, all drive backup and restore actions
populatet a details.OneDriveInfo struct.  This
change branches that struct between one-
drive and sharepoint info, depending on the
current source.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1616

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/onedrive/item_test.go
+++ b/src/internal/connector/onedrive/item_test.go
@@ -5,13 +5,13 @@ import (
 	"context"
 	"io"
 	"testing"
-	"time"
 
 	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
+	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/tester"
 )
@@ -70,7 +70,9 @@ func (suite *ItemIntegrationSuite) SetupSuite() {
 	suite.client = msgraphsdk.NewGraphServiceClient(adapter)
 	suite.adapter = adapter
 
-	// TODO: fulfill file preconditions required for testing (expected files w/in drive)
+	// TODO: fulfill file preconditions required for testing (expected files w/in drive
+	// and guarateed drive read-write access)
+	// suite.site = tester.M365SiteID(t)
 	// spDrives, err := drives(ctx, suite, suite.site, SharePointSource)
 	// require.NoError(t, err)
 	// // Test Requirement 1: Need a drive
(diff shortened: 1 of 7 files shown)


src/internal/connector/onedrive/item_test.go
```diff
@@ -49,39 +51,53 @@ func TestItemIntegrationSuite(t *testing.T) {
...
+
+	odDrives, err := drives(ctx, suite, suite.user, OneDriveSource)
+	require.NoError(t, err)
 	// Test Requirement 1: Need a drive
-	require.Greaterf(suite.T(), len(drives), 0, "user %s does not have a drive", suite.user)
+	require.Greaterf(t, len(odDrives), 0, "user %s does not have a drive", suite.user)
 
 	// Pick the first drive
-	suite.driveID = *drives[0].GetId()
+	suite.userDriveID = *odDrives[0].GetId()
```
dadams39, line 92: There exist inconsistent test based off picking the "first" Drive. Long-term do you see a case where arbitrarily picking a drive could cause a failure in the near future?

src/internal/connector/onedrive/item_test.go
```diff
@@ -106,73 +122,90 @@ func (suite *ItemIntegrationSuite) TestItemReader() {
...
-	require.NotEmpty(suite.T(), itemInfo.ItemName)
+	require.NotNil(suite.T(), itemInfo.OneDrive)
+	require.NotEmpty(suite.T(), itemInfo.OneDrive.ItemName)
 
 	size, err := io.Copy(io.Discard, itemData)
 	require.NoError(suite.T(), err)
 	require.NotZero(suite.T(), size)
-	require.Equal(suite.T(), size, itemInfo.Size)
-	suite.T().Logf("Read %d bytes from file %s.", size, itemInfo.ItemName)
+	require.Equal(suite.T(), size, itemInfo.OneDrive.Size)
```
dadams39, line 138: Unsure if this is the best metric. I believe @ashmrtn had a few issues where byte size within the store would vary. Not sure that a variance between these numbers would necessarily be an error.

dadams39 (approved):
Mostly nits: Please review prior to merging.

src/internal/connector/onedrive/item_test.go
```diff
@@ -5,23 +5,25 @@ import (
...
 )
 
 type ItemIntegrationSuite struct {
 	suite.Suite
-	user    string
-	driveID string
-	client  *msgraphsdk.GraphServiceClient
-	adapter *msgraphsdk.GraphRequestAdapter
+	site        string
+	siteDriveID string
```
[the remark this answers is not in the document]
ryanfkeepers: Right you are.  I'll comment them out for now.

src/internal/connector/onedrive/item_test.go
ryanfkeepers: Possibly.  Depends on the conditions.  For example, user accounts should only have a single drive to begin with (their onedrive acct).  Sharepoint, on the other hand, may need additional specification.  Either way, those changes are out of the scope of this PR.

src/internal/connector/onedrive/item_test.go
ryanfkeepers: The test has been consistently successful so far.  I don't see reason to change it.

Issue #1688, opened 2022-12-05
GC: OneDrive: Drive Test is flaky.  Refactor Required


dadams39:
Test located in `src/internal/connector/onedrive/drive_test.go`
Test: `func (suite *OneDriveSuite) TestCreateGetDeleteFolder()` noted to have inconsistent behavior.
Locally, tests will pass; however, the failure that presents when deployed is as follows:
```java
drive_test.go:104: 
          	Error Trace:	/home/runner/work/corso/corso/src/internal/connector/onedrive/drive_test.go:104
          	Error:      	elements differ
          	            	
          	            	extra elements in list A:
          	            	([]interface {}) (len=2) {
          	            	 (string) (len=34) "01ZWH4M4NGSVNLKFOFTJBJZQAZVYAX4AMF",
          	            	 (string) (len=34) "01ZWH4M4JTBZ667W4GGFFJPL4NHFHK7IQI"
          	            	}
          	            	
          	            	
          	            	listA:
          	            	([]string) (len=2) {
          	            	 (string) (len=34) "01ZWH4M4NGSVNLKFOFTJBJZQAZVYAX4AMF",
          	            	 (string) (len=34) "01ZWH4M4JTBZ667W4GGFFJPL4NHFHK7IQI"
          	            	}
          	            	
          	            	
          	            	listB:
          	            	([]string) {
          	            	}
```
This suggests that `GetAllFolders()` function on L91 folders does not contain any folders that match. Tests passes locally for several developers. Need to investigate what part of the CI environment is causing the discrepancy. 

Temporary fix:
- disable test
- Long-term fix: Refactor the test to ensure local and deployed tests have the same outcome.
The 
- See [run](https://github.com/alcionai/corso/actions/runs/3621051818/jobs/6104617378)

ashmrtn:
pulling info here from another PR so it's easy to find:
>the function being tested is used in the purge script. Any issues we see in it, whether the test is enabled or not, may lead to purging more data than we intend when we run the purge script

dadams39:
Test disabled at this time.

dadams39:
Test re-enabled in #1739. 

Debrief
===
- Leading up to the version upgrade within #1479 caused the keys within the `AdditionalData` to change.  Therefore, the `Get()` command only returned the first 10 items which led to an inconsistent return. Changes within #1479 and #1739 resolved the issue.

PR #1690, opened 2022-12-05
OneDrive: TestOneDriveDriveSuite/TestCreateGetDeleteFolder Sub-Test ignored

dadams39:
## Description
OneDrive Test experienced different results locally than in CI. Test to be ignored until a consistent test can replace it. 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

related to * #1688<issue>

## Test Plan

- [x] :zap: Unit test


ryanfkeepers (approved):

ashmrtn:
what were the different results? Was the local run done on the most recent version of Corso? It would be good to have more information about this prior to disabling the test completely

ryanfkeepers:
> what were the different results? Was the local run done on the most recent version of Corso? It would be good to have more information about this prior to disabling the test completely

@ashmrtn The description here is not very accurate.  This issue is primarily around investigating resolving the delta-based flaky failures as a result of retrieving older folders.  Your prior fix to deduplicate folder names has helped, but Danny and I are concerned that we're seeing the same issue with non-duplicate folders in runs such as this: https://github.com/alcionai/corso/actions/runs/3621051818/jobs/6104617378#step:7:240

The goal here is to temporarily unblock the graph sdk update, then immediately return attention to re-enabling these tests.

ashmrtn:
Thanks for the added context

>The goal here is to temporarily unblock the graph sdk update, then immediately return attention to re-enabling these tests

This sounds reasonable, but there's one possible issue: the function being tested is used in the purge script. Any issues we see in it, whether the test is enabled or not, may lead to purging more data than we intend when we run the purge script. On the other hand, disabling the purge script will lead to eventual CI timeouts for Corso and should be avoided unless there's no other options

ryanfkeepers:
> but there's one possible issue

Sounds like something we'll need to handle on resolution of the ticket.  Would you kindly post that notice in the issue?

Issue #1685, opened 2022-12-05
persist M365 delta token


ashmrtn:
Having delta tokens is a crucial part of incremental backups that use them. However, the delta token needs to be persisted in a backup so that we can use it in the next backup of the data

Tokens should be persisted in the kopia snapshot for the backup. For now it's alright if the solution is not reusable for future restartability changes

PR #1696, opened 2022-12-05
Return interface values instead of concrete type

ashmrtn:
## Description

Having a map of interface values instead of concrete structs allows adding collection implementations via different structs. This will be useful to inject a new collection that contains metadata information like delta links.

Also refactors the code for creating collections in Exchange so that it follows golang patterns of exiting early a bit better.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* #1685

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -109,16 +109,16 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
 func (gc *GraphConnector) createExchangeCollections(
 	ctx context.Context,
 	scope selectors.ExchangeScope,
-) ([]*exchange.Collection, error) {
+) ([]data.Collection, error) {
 	var (
 		errs           *multierror.Error
 		users          = scope.Get(selectors.ExchangeUser)
-		allCollections = make([]*exchange.Collection, 0)
+		allCollections = make([]data.Collection, 0)
 	)
 
 	// Create collection of ExchangeDataCollection
 	for _, user := range users {
-		collections := make(map[string]*exchange.Collection)
+		collections := make(map[string]data.Collection)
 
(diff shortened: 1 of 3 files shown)


ryanfkeepers:
I'm generally not a fan of returning interfaces, so I want to scrutinize the approach before we go ahead with this.  Do we need to collapse everything into a single map?  Could we keep the metadata in a separate map, and provide a concrete struct for that as well?  For that matter, could it be cleaner in terms of keys/maintenance to have separate maps?

ashmrtn:
>Do we need to collapse everything into a single map?

In a sense, yes. Eventually, everything is returned to the operation as a `[]data.Collection` and then sent off to KopiaWrapper

>Could we keep the metadata in a separate map, and provide a concrete struct for that as well? For that matter, could it be cleaner in terms of keys/maintenance to have separate maps?

I think returning separate maps would make things more complicated for everything except what key to use in the map. None of the functions in the call stack actually access any of the fields specific to the concrete type. The only reason it was a concrete type to begin with is we have been using a "return parameter" instead of actually returning `map[string]data.Collection` or even just `[]data.Collection`. I didn't want to make that change in this PR as well to keep it small and because it will probably change once we implement the restartability stuff

In essence, the only extra stuff that GraphConnector needs to start handing to KopiaWrapper to persist is a `map[M365 folder ID]path` and a `map[M365 folder ID]delta token`. I was hoping to put the info for both of those as separate files in a new directory in the kopia snapshot. It doesn't seem worth it to pass that info up a few layers in the call stack instead of just making a collection in the function that has the data for it and returning that collection with all the others

Issue #1669, opened 2022-12-02
GC: Version Upgrade 40.0+


dadams39:
Breaking change at `msgraph-sdk-go v.41` . Multiple libraries from the Microsoft universe need to be aligned. In addition, it is necessary to change calls and returns as behaviors may change. 

Related
===
- #1478

dadams39:
Related:
- #1656

dadams39:
Currently, functionality passes local CI tests. 
Cannot upgrade due to #1664 and other CI failures when checked in.

Failed runs
===
- https://github.com/alcionai/corso/actions/runs/3602958472/jobs/6072767557

dadams39:
Upgrade to version `48` is a breaking change. The libraries are flattened which is a benefit to building times. However, several files will require updating.

PR #1697, opened 2022-12-06
GC: Backup: Version 48: Delta Libraries

dadams39:
## Description
Msgraph updated to version 48. Verification required to ensure that all supporting libraries/functions operate as expected

## Type of change

- [x] :sunflower: Feature

## Issue(s)

*related to  #1669 <issue>

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/delta_get.go
+++ b/src/internal/connector/exchange/delta_get.go
@@ -4,9 +4,9 @@ import (
 	"context"
 
 	abs "github.com/microsoft/kiota-abstractions-go"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
-	mscontactdelta "github.com/microsoftgraph/msgraph-sdk-go/users/item/contactfolders/item/contacts/delta"
-	msmaildelta "github.com/microsoftgraph/msgraph-sdk-go/users/item/mailfolders/item/messages/delta"
+	msuser "github.com/microsoftgraph/msgraph-sdk-go/users"
 )
 
 //nolint:lll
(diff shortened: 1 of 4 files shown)


src/internal/connector/exchange/delta_get.go
```diff
@@ -4,9 +4,9 @@ import (
 	"context"
```
ashmrtn, line 4: We can actually just delete this entire file now that we're upgrading to a version of graph SDK that supports the `select` clause on delta queries. The overall code can be made to look like other calls to graph SDK that we make

All of the code from #1643 can be removed and we can just use the graph SDK types for the query options and the regular delta call for sending requests. We should still iterate through results manually though so `service_iterators.go` will have something like
```go
builder := gs.Client().
  UsersById(user).
  MailFoldersById(directoryID).
  Messages().
  Delta()

// Lower down when we need to actually send the request
resp, err := builder.Get(ctx, options)
```

Contacts should also be updated to just use the graph SDK calls instead of the calls in this file

src/internal/connector/exchange/service_iterators.go
```diff
@@ -250,7 +250,7 @@ func FetchContactIDsFromDirectory(ctx context.Context, gs graph.Service, user, d
 
 	for {
 		// TODO(ashmrtn): Update to pass options once graph SDK dependency is updated.
```
ashmrtn, line 252: nit
```suggestion
```

src/internal/connector/exchange/query_options.go
```diff
@@ -132,16 +132,19 @@ type DeltaRequestBuilderGetRequestConfiguration struct {
 	QueryParameters *DeltaRequestBuilderGetQueryParameters
 }
```
ashmrtn, line 133: nit: this struct and the one above it should be able to be deleted now (can't comment on all of it directly so I can't make a suggestion about it)

Issue #1601, opened 2022-11-23
Version docs on every release


ntolia:
Every time we release Corso, we need to version docs. This will look something like 

```
npm run docusaurus docs:version <version>
git add versioned_docs/ versioned_sidebars/ versions.json
git commit -m "..."
```

This will allow us to keep on publishing the blog without having the "main" version of the docs change.

Docs: https://docusaurus.io/docs/versioning

Related to #1598 and #1551

meain:
Should we be storing versioned docs in the repo, or should we just generate it dynamically in the CI from the git history. We can just lookup the git tags by pulling the entire repo history and generate the versioned dirs from that.

Two  ideas that I got from [Reccomended practices](https://docusaurus.io/docs/versioning#recommended-practices) is that we can just have separate versions for major(in our case minor) releases and not patch release if we have any. We can also restrict to the last `n` (probably something like 10) release.

ntolia:
Based on what Docusaurus does, we should have versioned docs in the repo. It's the only way to do it with clean and reasonable efforts. 

And yes, we can and should cap the number of versions of docs but we can do that later once we have a lot of releases hanging around.

meain:
Here is what I have in mind for automating creating versioned docs. The release job will release with a new version created and once complete, the changed files will be submitted via a PR.

meain:
Decided to not implement this as of now, but keeping the issue open.
https://github.com/alcionai/corso/pull/1708#issuecomment-1340421998

PR #1708, opened 2022-12-06
Docusaurus setup for versioned docs

meain:
## Description

This will allow us to have versioned docs.

- Header will have something like this to change the version of docs  with it landing on the latest by default.
    
- Footer changed from "Version" to "Latest Version" as we might be on a different version.
    
- Added v0.0.3 as the first set of version.
- Once a release is done, the CI will now create a pull request to the repo to add the new versioned set of docs ([example](https://github.com/alcionai/corso/pull/1709)).

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/1601

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


ntolia:
@meain I know I asked for this but given that you have a way of pushing docs without changing versions from the last tag, do we really need this right now? Maybe we add this sometime next year?

meain:
I had a similar thought when I [initially suggested](https://github.com/alcionai/corso/issues/1601#issuecomment-1328965680) about just generating the versioned docs on the fly. The problem is that, although we display the version number as the latest tag we have no way of figuring what parts of the content is for the latest version vs what is for the next version. For example if we have a doc fix for v1.0, but have some doc changes (like addition/removal of an option) for the upcoming release, the next time we publish the website, both of these changes make it to the website under the latest version. We can hack around this by using the old history to generate the docs, but the sidebar file can cause problems.

meain:
FYI, this now means that we will have generated docs in the tree for versioned docs.

ntolia:
> I had a similar thought when I [initially suggested](https://github.com/alcionai/corso/issues/1601#issuecomment-1328965680) about just generating the versioned docs on the fly. The problem is that, although we display the version number as the latest tag we have no way of figuring what parts of the content is for the latest version vs what is for the next version. 

Yes, I agree this is a problem but I think we are safe in terms of churn and expectations here for a couple of months. Let's push this PR in in a couple of months but not pay the overhead right now.

meain:
Makes sense, I'll close the PR but keep the issue open with it removed from current dev cycle.

PR #1715, opened 2022-12-06
GC: Testing:  Inspection of models expanded to handle Odata differences.

dadams39:
## Description
file: `graph_connector_helper_test.go` 
Flag and Recipentable objects were checked for equality; however, this included the `OdataType` for the respective classes. These fields are subject to change based on upstream causes. These do not impact the ability to store items and have been excluded from the tests for these two object types. 

## Type of change

- [x] :sunflower: Feature
- [x] :bug: Bugfix

## Issue(s)

* #<issue>

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/graph_connector_helper_test.go
+++ b/src/internal/connector/graph_connector_helper_test.go
@@ -244,10 +244,9 @@ func checkMessage(
 
 	// Skip CreatedDateTime as it's tied to this specific instance of the item.
 
-	assert.Equal(t, expected.GetFlag(), got.GetFlag(), "Flag")
+	checkFlags(t, expected.GetFlag(), got.GetFlag())
 
-	assert.Equal(t, expected.GetFrom(), got.GetFrom(), "From")
-
+	checkRecipentables(t, expected.GetFrom(), got.GetFrom())
 	testEmptyOrEqual(t, expected.GetHasAttachments(), got.GetHasAttachments(), "HasAttachments")
 
 	// Skip Id as it's tied to this specific instance of the item.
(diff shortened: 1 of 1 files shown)


dadams39:
Failure on the branch is associated with Issue #1688. This will be fixed in a separate branch.

src/internal/connector/graph_connector_helper_test.go
```diff
@@ -294,6 +293,39 @@ func checkMessage(
...
+// OdataTypes are omitted as these do change between msgraph-sdk-go versions
+func checkFlags(
+	t *testing.T,
+	expected, got models.FollowupFlagable,
+) {
+	assert.Equal(t, expected.GetCompletedDateTime(), got.GetCompletedDateTime())
+	assert.Equal(t, expected.GetDueDateTime(), got.GetDueDateTime())
+	assert.Equal(t, expected.GetFlagStatus(), got.GetFlagStatus())
+	assert.Equal(t, expected.GetCompletedDateTime(), got.GetCompletedDateTime())
+	assert.Equal(t, expected.GetCompletedDateTime(), got.GetCompletedDateTime())
```
ashmrtn, line 306: should these be some other value? This field is checked 3x (once at start and 2x here)

ashmrtn (approved):

src/internal/connector/graph_connector_helper_test.go
dadams39: Resolved within [8fc5ec0](https://github.com/alcionai/corso/pull/1715/commits/8fc5ec0874d54b6db9e7c19e282ef243d64c8ff8)

PR #1716, opened 2022-12-06
generate a unique repository id on init

ryanfkeepers:
## Description

generates a unique repository ID on repo init, and retrieves it again on repo connect.  This allows
the event bus (and any other tracker) to id a repo independent of other context.

I've been unable to find any good way to retrieve a unique repository ID from kopia.  This seems like
the next quickest solution.

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #1602

## Test Plan

- [x] :muscle: Manual


ryanfkeepers:
@ashmrtn pausing here before completing the PR (tests, etc) to get a draft review/gut-check on this approach.

src/internal/events/events.go
```diff
@@ -165,8 +169,15 @@ func storageID(s storage.Storage) string {
 	return id
 }
 
-func repoHash(s storage.Storage) string {
-	return md5HashOf(storageID(s))
+// NewRepoID generates a new unique repository id hash.
+// Repo ids should only be generated once per repository,
```
ashmrtn, line 173: nit
```suggestion
// NewRepoID generates a new unique repository ID hash.
// Repo IDs should only be generated once per repository,
```

src/internal/events/events.go
```diff
@@ -165,8 +169,15 @@ func storageID(s storage.Storage) string {
...
+// and must be stored after that.
+func NewRepoID(s storage.Storage) string {
+	return repoHash(s, uuid.NewString())
+}
```
ashmrtn, line 177: this should be moved to another package. Events has nothing to do with generating and storing a repo ID

src/internal/events/events.go
```diff
@@ -165,8 +169,15 @@ func storageID(s storage.Storage) string {
...
+
+func repoHash(s storage.Storage, uid string) string {
```
ashmrtn, line 179: this should take a storageID as a parameter instead of asking storage to fetch it

src/pkg/repository/repository.go
```diff
@@ -288,3 +302,45 @@ func (r repository) DeleteBackup(ctx context.Context, id model.StableID) error {
...
+// retrieves the repository info
+func getRepoModel(ctx context.Context, ms *kopia.ModelStore) (*repositoryModel, error) {
+	rm := &repositoryModel{}
+	return rm, ms.Get(ctx, model.RepositorySchema, repositoryID, rm)
+}
+
+// should only be called following a successful init.
+func (rm repositoryModel) write(ctx context.Context, ms *kopia.ModelStore) error {
+	return ms.Put(ctx, model.RepositorySchema, &rm)
+}
```
ashmrtn, line 341: this should be inlined instead of as a separate function. With it as a function on the model it becomes confusing because it's not immediately clear where it's being written to nor does it follow the pattern for how other models are stored

src/pkg/repository/repository.go
```diff
@@ -288,3 +302,45 @@ func (r repository) DeleteBackup(ctx context.Context, id model.StableID) error {
...
+)
+
+// should only be called on init.
+func newRepoModel(repoID string) repositoryModel {
+	return repositoryModel{
+		BaseModel: model.BaseModel{
+			ID: repositoryID,
+			Tags: map[string]string{
+				repoIDTag: repoID,
+			},
```
ashmrtn, line 327: I think we should store the generated ID in `BaseModel.ID` as it normally would be stored and leverage `GetIDsForType` (see below) to lookup the model

```suggestion
			ID: repoID,
```

src/pkg/repository/repository.go
```diff
@@ -288,3 +302,45 @@ func (r repository) DeleteBackup(ctx context.Context, id model.StableID) error {
...
+		},
+	}
+}
+
+// retrieves the repository info
+func getRepoModel(ctx context.Context, ms *kopia.ModelStore) (*repositoryModel, error) {
+	rm := &repositoryModel{}
+	return rm, ms.Get(ctx, model.RepositorySchema, repositoryID, rm)
```
ashmrtn, line 335: I would prefer storing the unique generated ID in the ID field of the model and doing `GetIDsForType` to lookup the model that we need. As `GetIDsForType` returns `[]*BaseModel`, it's still possible to extract the ID without more lookups. This would also give us a way to ensure only a single `repositoryModel` is added to the model store

Since there should only be one `repositoryModel` in the model store, it should be clear which entry in the returned slice to use

src/pkg/repository/repository.go
```diff
@@ -288,3 +302,45 @@ func (r repository) DeleteBackup(ctx context.Context, id model.StableID) error {
...
+
+// should only be called following a successful init.
+func (rm repositoryModel) write(ctx context.Context, ms *kopia.ModelStore) error {
+	return ms.Put(ctx, model.RepositorySchema, &rm)
+}
+
+// retrieves the stored repository ID.
+func (rm repositoryModel) ID() string {
+	return rm.Tags[repoIDTag]
+}
```
ashmrtn, line 346: this is not required if the ID is stored in `BaseModel.ID`
```suggestion
```

ashmrtn (commented):
I'd really prefer to store fewer models overall since it would mean less for us to maintain. However, I found a [comment](https://github.com/kopia/kopia/blob/04fcedad1267f9d9ecfab27fb38c3f9aa1d6308f/repo/repository.go#L326) in the kopia repo with the following:
> // UniqueID returns unique repository ID from which many keys and secrets are derived.

That makes it seem like we should avoid sending the kopia repo's `UniqueID` across the wire so the solution here is the most reasonable

src/internal/events/events.go
ryanfkeepers: Well, I wouldn't say nothing, since it's the only thing that uses the repo ID.  But I agree on the package move.

src/pkg/repository/repository.go
ryanfkeepers: I dig it.

src/pkg/repository/repository_unexported_test.go
```diff
@@ -0,0 +1,57 @@
...
+		s        = tester.NewPrefixedS3Storage(t)
+		kopiaRef = kopia.NewConn(s)
+	)
+
+	require.NoError(t, kopiaRef.Connect(ctx))
+	defer kopiaRef.Close(ctx)
+
+	ms, err := kopia.NewModelStore(kopiaRef)
+	require.NoError(t, err)
+	require.NoError(t, newRepoModel(ctx, ms, "fnords"))
```
ashmrtn, line 52: nit: need to close the model store as well
```suggestion
	require.NoError(t, newRepoModel(ctx, ms, "fnords"))
	
	defer ms.Close(ctx)
```

ashmrtn (approved):
Please add a test in `src/pkg/repository/repository_test.go` ensuring that initializing a repo, closing it, and reopening it results in the same repoID. Since the ID is not part of the repo struct it may need to be pulled from the model store after the second time the repo is opened, but that should be sufficient to determine if the previous value was clobbered

src/pkg/repository/repository_unexported_test.go
ryanfkeepers: good catch, thanks

vkamra:
@ryanfkeepers - was debugging something and came across https://github.com/kopia/kopia/blob/master/cli/command_repository_status.go#L64. I think this would work - and if it makes sense - I can update our code to switch to this?

ashmrtn:
> was debugging something and came across https://github.com/kopia/kopia/blob/master/cli/command_repository_status.go#L64. I think this would work - and if it makes sense - I can update our code to switch to this?

See one of my [previous comments](https://github.com/alcionai/corso/pull/1716#pullrequestreview-1207538969) on the Kopia repo ID

vkamra:
> > was debugging something and came across https://github.com/kopia/kopia/blob/master/cli/command_repository_status.go#L64. I think this would work - and if it makes sense - I can update our code to switch to this?
> 
> See one of my [previous comments](https://github.com/alcionai/corso/pull/1716#pullrequestreview-1207538969) on the Kopia repo ID

Hmm - you are right. While it's not "secret" i.e. is stored in the repo plaintext -> It's probably safer not to use this.

PR #1717, opened 2022-12-06
Return delta tokens from fetch functions

ashmrtn:
## Description

When fetching Exchange items, return the resulting delta token from the query so it can be reused later.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1685 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -19,7 +19,10 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
-const nextLinkKey = "@odata.nextLink"
+const (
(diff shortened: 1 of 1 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -165,7 +168,11 @@ func IterativeCollectCalendarContainers(
 
 // FetchIDFunc collection of helper functions which return a list of strings
 // from a response.
```
ryanfkeepers, line 170: most of the comments here are now outdated

PR #1718, opened 2022-12-06
Create path categories for metadata folders

ashmrtn:
## Description

Corso needs to store some transient information between backups so that it can perform more efficient incremental backups. This transient information will live at a path `tenant-id/service/user-id/categoryMetadata`. This patch adds an initializer, category enums, and tests for these new path locations

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1685

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/path/resource_path.go
```diff
@@ -31,17 +31,33 @@ func toServiceType(service string) ServiceType {
...
+	EventsCategory                         // events
+	FilesCategory                          // files
+	ListsCategory                          // lists
+	LibrariesCategory                      // libraries
+	EmailMetadataCategory                  // emailMetadata
+	ContactsMetadataCategory               // contactsMetadata
+	EventsMetadataCategory                 // eventsMetadata
+	FilesMetadataCategory                  // filesMetadata
+	ListsMetadataCategory                  // listsMetadata
+	LibrariesMetadataCategory              // librariesMetadata
```
ashmrtn, line 60: not really sure I like having to essentially duplicate categories, but I don't really know of another elegant way to handle a "metadata" category without losing information about what category stuff was for originally. We could use a flag on the Path interface, but that feels a little clunky and error prone since the stuff in the metadata directory won't be used directly during backup and restore

ryanfkeepers: Thoughts on `{tenant}/{service}/{owner}/metadata/{category}` as the path?  Treat `metadata` as its own category keyword, and re-use he existing category enums as folders?

ashmrtn: mmm it could work though then the code that uses the metadata path has to manually determine what category the metadata path is for because it's no longer something the Path interface can tell it. I guess it would work as long as we don't nest any folders in the `metadata/{category}` hierarchy, but having that assumption makes me uneasy

ryanfkeepers: Alterrnative:  `{tenant}/{service-metadata}/{owner}/{category}`.  Promote the "metadata" evaluation up to the service level, and again re-use the category.  Still a little duplication, but should be a smaller footprint.

ashmrtn: seems like a reasonable compromise

ryanfkeepers (approved):

Issue #1594, opened 2022-11-23
Rudderstack events data improvements


gmatev:
* Ensure that Corso version is properly logged
* Backup and restore events are missing statistics about data and items written

ntolia:
Closing this issue. Will file new ones for further improvements.

PR #1721, opened 2022-12-07
Use corso version with event reporting

meain:
## Description

Add version of the current binary when reporting.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* https://github.com/alcionai/corso/issues/1594

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/version/version.go
+++ b/src/cli/version/version.go
@@ -1,0 +1,3 @@
+package version
+
+var Version = "dev"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -492,7 +492,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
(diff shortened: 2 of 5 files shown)


src/cli/version/version.go
```diff
@@ -0,0 +1,3 @@
+package version
+
+var Version = "dev"
```
meain, line 3: Moving to separate package to avoid cyclic imports in events.

ryanfkeepers: Let's move this out of `/cli` and into `/internal`, please.  Doesn't make sense for an API layer to provide configuration to internal packages.

vkamra: 👍 . @ryanfkeepers - if rest of the changes look good to you - let's approve so that @meain can merge once he's done without another review cycle.

PR #1722, opened 2022-12-07
Persist Exchange delta tokens

### LINKED ISSUE #1685: persist M365 delta token

ashmrtn:
## Description

Begin persisting Exchange delta tokens for data categories that support delta queries. Tokens are stored in a single file as a `map[M365 container ID]token` where both the container ID and token are of type `string`. The file is located in the kopia snapshot that has all the other backup data at the path `tenant-id/{service}Metadata/user/category/delta`. No information about the delta token file is stored in backup details.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1685 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -1,9 +1,7 @@
 package exchange
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -45,49 +43,6 @@ func getAdditionalDataString(
 	}
 
 	return *value
-}
-
-// makeMetadataCollection creates a metadata collection that has a file
-// containing all the delta tokens in tokens. Returns nil if the map does not
-// have any entries.
-//
-// TODO(ashmrtn): Expand this/break it out into multiple functions so that we
-// can also store map[container ID]->full container path in a file in the
-// metadata collection.
-func makeMetadataCollection(
-	tenant string,
-	user string,
-	cat path.CategoryType,
-	tokens map[string]string,
-	statusUpdater support.StatusUpdater,
-) (data.Collection, error) {
-	if len(tokens) == 0 {
-		return nil, nil
-	}
-
-	buf := &bytes.Buffer{}
-	encoder := json.NewEncoder(buf)
-
-	if err := encoder.Encode(tokens); err != nil {
-		return nil, errors.Wrap(err, "serializing delta tokens")
-	}
-
-	p, err := path.Builder{}.ToServiceCategoryMetadataPath(
-		tenant,
-		user,
-		path.ExchangeService,
-		cat,
-		false,
-	)
-	if err != nil {
-		return nil, errors.Wrap(err, "making path")
-	}
-
-	return graph.NewMetadataCollection(
-		p,
-		[]graph.MetadataItem{graph.NewMetadataItem(graph.DeltaTokenFileName, buf.Bytes())},
-		statusUpdater,
-	), nil
 }
 
 // FilterContainersAndFillCollections is a utility function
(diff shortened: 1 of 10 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -43,6 +47,49 @@ func getAdditionalDataString(
 	return *value
 }
 
+// makeMetadataCollection creates a metadata collection that has a file
+// containing all the delta tokens in tokens. Returns nil if the map does not
+// have any entries.
+//
+// TODO(ashmrtn): Expand this/break it out into multiple functions so that we
```
ryanfkeepers, line 54: There's very little here that's either delta token or even Exchange specific.  I think this would be better moved into `metadata_collection.go`.  Then have it take a path.service and a file name, and it works for everything.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -43,6 +47,49 @@ func getAdditionalDataString(
...
+// can also store map[container ID]->full container path in a file in the
+// metadata collection.
+func makeMetadataCollection(
+	tenant string,
+	user string,
```
ryanfkeepers, line 59: nit:
```suggestion
	tenant, user string,
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -113,9 +162,34 @@ func FilterContainersAndFillCollections(
...
+
+		if len(token) > 0 {
+			deltaTokens[*c.GetId()] = token
+		}
+	}
+
+	// TODO(ashmrtn): getFetchIDFunc functions should probably just return a
+	// multierror and all of the error handling should just use those so that it
+	// all ends up more consistent.
+	merrs := multierror.Append(nil, errs)
```
ryanfkeepers, line 174: While I agree with the sentiment, the most consistent thing to do here is continue to use `errs` and `support.WrapandAppend` instead of swapping to the multierr package.  Injecting these minor refactors piecemeal tends to accumulate chaff.  If we're concerned about error handling patterns, we can make an issue to change them in a consolidated effort.

src/internal/operations/backup_test.go
```diff
@@ -297,6 +390,37 @@ func (suite *BackupOpIntegrationSuite) TestBackup_Run_exchange() {
...
+			// these files will only be made for contacts and email in Exchange if any
+			// items were backed up. Events does not support delta queries.
+			if failed {
+				return
+			}
+
+			m365, err := acct.M365Config()
+			require.NoError(t, err)
+
+			for _, scope := range sel.DiscreteScopes(nil) {
```
ryanfkeepers, line 404: DiscreteScopes isn't needed here.
```suggestion
			for _, scope := range sel.Scopes() {
```

src/internal/connector/exchange/service_iterators.go
ashmrtn: if we just have it take a file name then we won't be able to have multiple separate metadata files because calling this function 2x will result in 2 separate collections with the same path but different items. I think either the function needs to assume it's getting some fixed input of data to serialize into well-known files or it should stay local to whatever code is using it to create files in the metadata collection

ryanfkeepers: Mmmkay, keep it as is for now.  We can watch for refactoring/consolidation in the future as needed.

ryanfkeepers (approved):

PR #1731, opened 2022-12-08
Expand interfaces for components used during backup

### LINKED ISSUE #1700: Update interfaces for incremental backups
Some interfaces need expanding to handle incremental backups. Included interfaces/APIs:
* data.Collection
* data.Stream
* kopia.BackupCollections

ashmrtn:
## Description

Expand interfaces for `GraphConnector.DataCollections` and `kopia.Wrapper.BackupCollections` to include parameters that will be needed during incremental backups. This patch only expands the interfaces, it does not add any extra functionality and the passed parameters are currently ignored.

In the future, passing nil for any of the new parameters should result in the current "full backup" behavior that Corso has. Passing values in these parameters should enable delta token-based incremental backups (assuming all the required data is there for the incremental backup)

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* closes #1700 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -499,8 +499,8 @@ func inflateDirTree(
 // with the data that they contain. previousSnapshots is used for incremental
 // backups and should represent the base snapshot from which metadata is sourced
 // from as well as any incomplete snapshot checkpoints that may contain more
-// recent data than the base snapshot. Passing nil for previousSnapshots means
-// denotes that a complete backup of all data should be done.
+// recent data than the base snapshot. The absence of previousSnapshots causes a
+// complete backup of all data.
 //
 // TODO(ashmrtn): Use previousSnapshots parameter.
 func (w Wrapper) BackupCollections(
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -495,8 +495,17 @@ func inflateDirTree(
 	return res, ownerCats, nil
 }
 
+// BackupCollections takes a set of collections and creates a kopia snapshot
+// with the data that they contain. previousSnapshots is used for incremental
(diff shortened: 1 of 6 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -495,8 +495,17 @@ func inflateDirTree(
 	return res, ownerCats, nil
 }
 
+// BackupCollections takes a set of collections and creates a kopia snapshot
+// with the data that they contain. previousSnapshots is used for incremental
+// backups and should represent the base snapshot from which metadata is sourced
+// from as well as any incomplete snapshot checkpoints that may contain more
+// recent data than the base snapshot. Passing nil for previousSnapshots means
+// denotes that a complete backup of all data should be done.
```
ryanfkeepers, line 503: Minor: I'm not a fan of specifying `nil` here, as that sidesteps what should be equal behavior for an empty collection.  And I wouldn't want to encourage producer funcs to return `nil, nil` as a standard.
```suggestion
// recent data than the base snapshot. Passing nil for previousSnapshots means
// denotes that a complete backup of all data should be done.
```
```suggestion
// recent data than the base snapshot. The absence of previousSnapshots
// causes a complete backup of all data.
```

ryanfkeepers (approved):

src/internal/kopia/wrapper.go
ashmrtn: fair enough. I wasn't quite sure what to put cause "nil or empty collection" sounds a bit wordy but just one or the other has the issues you pointed out 🙃

Issue #1725, opened 2022-12-07
GC should re-use delta tokens in Exchange when they're provided


ryanfkeepers:
## Issue
Now that metadata collections are becoming available for incrementals, graph needs to consume and utilize the tokens provided to it.

## Acceptance Criteria
- [x] In operations, get delta token metadata collections from kopia’s previous snapshot
- [x] pass metadata collections into GC backup
- [x] retrieve delta tokens by folder id from collections
- [x] do delta query with provided token (full retrieval if no token exists)

example of fetching a metadata file: https://github.com/alcionai/corso/pull/1722/files#diff-591549691d95b47ff2d8d0064d4816ed3f9cbc88a73d8254f6b6ad2014e3a393R208

Issue #1672, opened 2022-12-02
optimize finding the most recent snapshot(s) for incremental backups


ashmrtn:
Recent changes to upstream kopia allow for adding different tags to checkpoints than are associated with complete backups. This allows for an optimized version that requires fewer kopia reads for finding the most recent complete and incomplete backups for a given set of tags.

Work that needs to be completed to enable this:
- [x] #1629 
- [x] #1673
- [x] #1674
- [x] #1675

PR #1744, opened 2022-12-08
retrieve manifests and metadata from kopia

ryanfkeepers:
## Description

In a backup operation, begins the operation by
retrieving all backup manifests and metadata
from prior operations.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1725

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -13,6 +13,12 @@ import (
 // given endpoint. The endpoint granularity varies by service.
 const DeltaTokenFileName = "delta"
 
+// MetadataFileNames produces the standard set of filenames used to store graph
+// metadata such as delta tokens and folderID->path references.
+func MetadataFileNames() []string {
+	return []string{DeltaTokenFileName}
+}
+
 type QueryParams struct {
 	Category      path.CategoryType
 	ResourceOwner string
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -187,7 +187,7 @@ func fetchPrevManifests(
 	return manifestsSinceLastComplete(mans), nil
 }
 
-// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
+// FetchPrevSnapshotManifests returns a set of manifests for complete and maybe
 // incomplete snapshots for the given (resource owner, service, category)
 // tuples. Up to two manifests can be returned per tuple: one complete and one
 // incomplete. An incomplete manifest may be returned if it is newer than the
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -202,7 +202,7 @@ func fetchPrevManifests(
 	return manifestsSinceLastComplete(mans), nil
 }
 
-// FetchPrevSnapshotManifests returns a set of manifests for complete and maybe
+// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
 // incomplete snapshots for the given (resource owner, service, category)
 // tuples. Up to two manifests can be returned per tuple: one complete and one
 // incomplete. An incomplete manifest may be returned if it is newer than the
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -209,29 +209,34 @@ func fetchPrevManifests(
 // newest complete manifest for the tuple. Manifests are deduped such that if
 // multiple tuples match the same manifest it will only be returned once.
 // External callers can access this via wrapper.FetchPrevSnapshotManifests().
+// If tags are provided, manifests must include a superset of the k:v pairs
+// specified by those tags.  Tags should pass their raw values, and will be
+// normalized inside the func using MakeTagKV.
 func fetchPrevSnapshotManifests(
 	ctx context.Context,
 	sm snapshotManager,
 	oc *OwnersCats,
+	tags map[string]string,
 ) []*snapshot.Manifest {
 	mans := map[manifest.ID]*snapshot.Manifest{}
+	tags = normalizeTagKVs(tags)
 
 	// For each serviceCat/resource owner pair that we will be backing up, see if
 	// there's a previous incomplete snapshot and/or a previous complete snapshot
 	// we can pass in. Can be expanded to return more than the most recent
 	// snapshots, but may require more memory at runtime.
 	for serviceCat := range oc.ServiceCats {
-		serviceTagKey, serviceTagValue := MakeTagKV(serviceCat)
-
 		for resourceOwner := range oc.ResourceOwners {
-			resourceOwnerTagKey, resourceOwnerTagValue := MakeTagKV(resourceOwner)
+			allTags := normalizeTagKVs(map[string]string{
+				serviceCat:    "",
+				resourceOwner: "",
+			})
 
-			tags := map[string]string{
-				serviceTagKey:       serviceTagValue,
-				resourceOwnerTagKey: resourceOwnerTagValue,
+			for k, v := range tags {
+				allTags[k] = v
 			}
 
-			found, err := fetchPrevManifests(ctx, sm, mans, tags)
+			found, err := fetchPrevManifests(ctx, sm, mans, allTags)
 			if err != nil {
 				logger.Ctx(ctx).Warnw(
 					"fetching previous snapshot manifests for service/category/resource owner",
@@ -258,4 +263,20 @@ func fetchPrevSnapshotManifests(
 	}
 
 	return res
+}
+
+func normalizeTagKVs(tags map[string]string) map[string]string {
+	t2 := make(map[string]string, len(tags))
+
+	for k, v := range tags {
+		mk, mv := MakeTagKV(k)
+
+		if len(v) == 0 {
+			v = mv
+		}
+
+		t2[mk] = v
+	}
+
+	return t2
 }
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -44,9 +44,9 @@ type ServiceCat struct {
 }
 
 // MakeServiceCat produces the expected OwnersCats.ServiceCats key from a
-// path service and path category.
-func MakeServiceCat(s path.ServiceType, c path.CategoryType) string {
-	return serviceCatString(s, c)
+// path service and path category, as well as the ServiceCat value.
+func MakeServiceCat(s path.ServiceType, c path.CategoryType) (string, ServiceCat) {
+	return serviceCatString(s, c), ServiceCat{s, c}
 }
 
 func serviceCatTag(p path.Path) string {
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -35,11 +35,26 @@ type snapshotManager interface {
 
 type OwnersCats struct {
 	ResourceOwners map[string]struct{}
-	ServiceCats    map[string]struct{}
+	ServiceCats    map[string]ServiceCat
+}
+
+type ServiceCat struct {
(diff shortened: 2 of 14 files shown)


ryanfkeepers:
@ashmrtn This is currently missing tests.  I'm looking for a gut-check on the details before I dive into that.

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		if err := sw.Get(
+			ctx,
+			model.BackupSchema,
+			model.StableID(man.Tags[kopia.BackupIDTag]),
+			&bup,
+		); err != nil {
+			return nil, nil, err
+		}
+
+		for _, fn := range []string{graph.DeltaTokenFileName} {
```
ashmrtn, line 204: nit: eventually we should have a function in the producer that we call and it gives us a set of metadata file names (or partial paths, or maybe even complete paths). That will isolate generating the paths from consuming them, making the logic here more generic

ex.
```go
  for _, fn := range graph.GetMetadataPaths() {
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+			for ro := range oc.ResourceOwners {
+				for _, sc := range oc.ServiceCats {
+					colls, err := collectMetadata(ctx, kw, fn, tid, ro, bup.SnapshotID, sc)
+					if err != nil {
```
ashmrtn, line 208: not finding a metadata file is not necessarily an error. This could happen in cases where a backup contains only some of the services we're now backing up (in which case we may have multiple base snapshots). For example `backup create exchange --data email` and `backup create exchange --data contacts,events` followed by `backup create exchange`

As long as all of the metadata files we need are present (i.e. both delta tokens and `map[container ID]container path` _for each category_ in the future) then we can run more efficiently. If one or more are missing for a category we'll need to fallback to doing a full backup of the data

After explaining this though, it seems like we'll probably want to augment the return value of `FetchPrevSnapshotManifests` to also tell us _why_ a specific snapshot was picked. That will allow us to properly choose what metadata to extract from the snapshot. In turn, that will help avoid the case where we have two snapshots that contain metadata for the same category but we only want to use the most recent version

Ex.
```
backup create exchange --data email,contacts -> B1
backup create exchange --data email -> B2
// should use B1 to source contacts metadata, B2 to source emails metadata
backup create exchange
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+}
+
+func collectMetadata(
+	ctx context.Context,
+	kw *kopia.Wrapper,
+	fileName, tenantID, resourceOwner, snapshotID string,
+	sc kopia.ServiceCat,
+) ([]data.Collection, error) {
+	paths := []path.Path{}
+	pathsByRef := map[string][]string{}
```
ashmrtn, line 228: this was just used for testing so I could compare by collection
```suggestion
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		return nil, errors.Wrap(err, "building metadata path")
+	}
+
+	dir, err := p.Dir()
+	if err != nil {
+		return nil, errors.Wrap(err, "malformed metadata path: "+p.String())
+	}
+
+	paths = append(paths, p)
+	pathsByRef[dir.ShortRef()] = append(pathsByRef[dir.ShortRef()], fileName)
```
ashmrtn, line 248:
```suggestion
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+			true)
+	if err != nil {
+		return nil, errors.Wrap(err, "building metadata path")
+	}
+
+	dir, err := p.Dir()
+	if err != nil {
+		return nil, errors.Wrap(err, "malformed metadata path: "+p.String())
+	}
+
```
ashmrtn, line 246:
```suggestion
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		return oc
+	}
+
+	for _, sl := range [][]string{ros.Includes, ros.Filters} {
+		for _, ro := range sl {
+			oc.ResourceOwners[ro] = struct{}{}
+		}
+	}
+
+	pcs, err := sel.PathCategories()
```
ashmrtn, line 276: what does it mean if `sel.PathCategories()` returns an error? I'm not sure if it's safe in all cases to tell KW to look up snapshots with just service/categories

src/internal/operations/backup.go
ashmrtn: The "why" a snapshot was picked could probably be denoted with tags or something (ex. this resource owner and serviceCat combo), but may require an update to how we fetch snapshots since that code makes aggressive use of caching

Vaguely related, that code could also probably be simplified if we start tagging checkpoints with an additional tag. Then we wouldn't have to lookup so many manifests to determine if they were complete or not (see #1672)

src/internal/kopia/wrapper.go
```diff
@@ -262,6 +264,10 @@ type Wrapper struct {
 	c *conn
 }
 
+func (w *Wrapper) Conn() *conn {
```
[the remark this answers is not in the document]
ryanfkeepers: Totally fair.  I'll probably put that into a different PR and then rebase this one onto it.

src/internal/operations/backup.go
ryanfkeepers: It means the selector service instance (ex: ExchangeBackupSelector) doesn't comply with the interface used by this function.  Not something we should ever expect to see an error for in the wild.

ashmrtn: even though it's annoying/less efficient to deal with, it would be better to either return the error or fallback to doing a full backup (implementor's choice)

src/internal/operations/backup.go
```diff
@@ -154,11 +163,139 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+
+	m365, err := acct.M365Config()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var (
+		tid         = m365.AzureTenantID
+		oc          = selectorToOwnersCats(sel)
+		ms          = kopia.FetchPrevSnapshotManifests(ctx, kw.Conn(), &oc)
```
vkamra, line 189: I am going to introduce another snapshot to store backup details. We'll need to make sure that gets filtered out here.

ashmrtn: if we add a tag to it that somehow indicates it's for backup details it should be easy to filter out. Or perhaps don't add the same tags we do for regular backups and then it won't be returned when we search for snapshots that contain item data. It just depends on how we plan to lookup backup details snapshots really

vkamra: Well - will probably add some bookkeeping tags TBD - but I think the logic in this method should include only the relevant snapshots vs trying to filter out specific tags.

src/internal/operations/backup.go
ryanfkeepers: I'll defer to the latter, since I don't think it's valuable to return the error here.

src/internal/operations/backup.go
ashmrtn: no tags would work just fine. It's just that looking up things with tags in kopia is an "includes at least these tags" operator not an "exactly these tags" operator. So if the backup details snapshot had a superset of the tags we add to snapshots with item data then we'd need to filter those snapshots out ourselves

src/internal/kopia/wrapper.go
```diff
@@ -497,6 +503,7 @@ func (w Wrapper) BackupCollections(
 	ctx context.Context,
 	collections []data.Collection,
 	service path.ServiceType,
+	backupID string,
```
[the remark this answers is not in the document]
ryanfkeepers: > in addition to the set of snapshots passed in

Yeah, I need a way to attach the current backup's ID to the manifest, else we don't have a way (that I'm aware of) to look up the backup model from the prior manifest [here](https://github.com/alcionai/corso/pull/1744/files#diff-f088d3a5f56348ddaf6fbf485c2aeaa6d8b40a860de014f7084db6daf0753829R207).  

> we may be restarting and want to make use of what we've previously completely uploaded by having kopia-assisted dedupe kick in for those items

That depends on whether we want to re-use a backup id across retries.  I imagine we could keep updating the same backup model until the operation completes.  Would be functionally the same as rolling forward with new models: the complete manifest references a complete backup model either way.  Your preferences, here?

src/internal/operations/backup.go
ryanfkeepers: Going to update this PR to pass in multiple tags to `BackupCollections`.  One being the backupID, another as a categorical identifier to separate backups from details.  That'll get us a lookup that's sufficient for this change, and we can add tagging to details in the future as needed.

src/internal/connector/graph/service.go
```diff
@@ -13,6 +13,12 @@ import (
 // given endpoint. The endpoint granularity varies by service.
 const DeltaTokenFileName = "delta"
 
+// MetadataFileNames produces the standard set of filenames used to store graph
+// metadata such as delta tokens and folderID->path references.
+func MetadataFileNames() []string {
```
ashmrtn, line 18: I think this works for now, though in the future we should probably consider more generic use-case like "what if a services has different metadata files/paths for different categories of data"

src/internal/kopia/snapshot_manager.go
```diff
@@ -187,36 +202,41 @@ func fetchPrevManifests(
 	return manifestsSinceLastComplete(mans), nil
 }
 
-// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
+// FetchPrevSnapshotManifests returns a set of manifests for complete and maybe
```
ashmrtn, line 205:
```suggestion
// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
```

src/internal/operations/backup.go
```diff
@@ -154,11 +163,147 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		if err := sw.Get(
+			ctx,
+			model.BackupSchema,
+			model.StableID(man.Tags[kopia.TagBackupID]),
+			&bup,
+		); err != nil {
+			return nil, nil, err
+		}
+
+		for _, fn := range graph.MetadataFileNames() {
```
ashmrtn, line 216: depending on how you want to handle errors (and logging) this whole loop block could be moved into `collectMetadata` if you wanted. `collectMetadata` could then generate all paths and then make a single call to `RestoreMultipleItems`. It would be a tad more efficient in the long run because it would only load the root directory of each snapshot once instead of once for every (resource owner, service) tuple. Not that there should be many tuples, but it plays a bit better with the other logic

src/internal/kopia/wrapper.go
```diff
@@ -87,6 +93,10 @@ type Wrapper struct {
 	c *conn
 }
 
+func (w *Wrapper) Conn() *conn {
+	return w.c
+}
+
```
ashmrtn, line 99:
```suggestion
```

ashmrtn (approved):

src/internal/kopia/wrapper.go
ryanfkeepers: whoops, thanks for the reminder

src/internal/operations/backup.go
ryanfkeepers: Good call, I've shifted that arround.

Issue #1735, opened 2022-12-08
Do not use a Kopia manifest to store backup details


vkamra:
Using a kopia manifest to store backup details can blow up memory allocations in codepaths that are not optimized for large manifests and read all manifest data into memory (e.g. manifest compaction).

Instead - the current proposal is to leverage an additional "metadata snapshot" to store the backup details data. Going forward - this snapshot could potentially be used for other metadata required to enable backup restarts and incrementals.

PR #1755, opened 2022-12-09
Store backup details in a separate snapshot

vkamra:
## Description

This PR uses a separate snapshot to store backup details instead of a Corso model (i.e. a kopia manifest).

Introduces a `StreamStore` that can be leveraged to store larger metadata objects. We can also leverage this
for incrementals or restartable backups going forward.

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1735 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/go.mod
+++ b/src/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/tidwall/pretty v1.2.1
 	github.com/tomlazar/table v0.1.2
 	github.com/vbauerster/mpb/v8 v8.1.4
+	github.com/zeebo/assert v1.1.0
 	go.uber.org/zap v1.24.0
 	golang.org/x/tools v0.4.0
 	gopkg.in/resty.v1 v1.12.0
@@ -82,7 +83,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
 	github.com/microsoft/kiota-serialization-text-go v0.6.0 // indirect
 	github.com/minio/md5-simd v1.1.2 // indirect
-	github.com/minio/minio-go/v7 v7.0.39 // indirect
+	github.com/minio/minio-go/v7 v7.0.39
 	github.com/minio/sha256-simd v1.0.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
--- a/src/internal/model/model.go
+++ b/src/internal/model/model.go
@@ -17,7 +17,6 @@ const (
 	BackupOpSchema
 	RestoreOpSchema
 	BackupSchema
-	BackupDetailsSchema
 )
 
 // common tags for filtering
@@ -27,7 +26,7 @@ const (
 
 // Valid returns true if the ModelType value fits within the iota range.
 func (mt Schema) Valid() bool {
-	return mt > 0 && mt < BackupDetailsSchema+1
+	return mt > 0 && mt < BackupSchema+1
 }
 
 type Model interface {
--- a/src/internal/streamstore/streamstore.go
+++ b/src/internal/streamstore/streamstore.go
@@ -1,0 +1,194 @@
+// streamstore implements helpers to store large
+// data streams in a repository
+package streamstore
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+
+	"github.com/pkg/errors"
+
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/internal/kopia"
+	"github.com/alcionai/corso/src/internal/stats"
+	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+type streamStore struct {
+	kw       *kopia.Wrapper
+	resource string
+	tenant   string
+	service  path.ServiceType
+}
+
+func New(
+	kw *kopia.Wrapper,
+	tenant string,
+	service path.ServiceType,
+) *streamStore {
+	return &streamStore{kw: kw, resource: "metadata", tenant: tenant, service: service}
+}
+
+const detailsItemName = "details"
+
+// WriteBackupDetails persists a `details.Details`
+// object in the stream store
+func (ss *streamStore) WriteBackupDetails(
+	ctx context.Context,
+	backupDetails *details.Details,
+) (string, error) {
+	// construct the path of the container for the `details` item
+	p, err := path.Builder{}.
+		ToServiceCategoryMetadataPath(
+			ss.tenant,
+			ss.resource,
+			ss.service,
+			path.DetailsCategory,
+			false,
+		)
+	if err != nil {
+		return "", err
+	}
+
+	// TODO: We could use an io.Pipe here to avoid a double copy but that
+	// makes error handling a bit complicated
+	dbytes, err := json.Marshal(backupDetails)
+	if err != nil {
+		return "", errors.Wrap(err, "marshalling backup details")
+	}
+
+	dc := &streamCollection{
+		folderPath: p,
+		item: &streamItem{
+			name: detailsItemName,
+			data: dbytes,
+		},
+	}
+
+	backupStats, _, err := ss.kw.BackupCollections(ctx, nil, []data.Collection{dc}, ss.service)
+	if err != nil {
+		return "", nil
+	}
+
+	return backupStats.SnapshotID, nil
+}
+
+// ReadBackupDetails reads the specified details object
+// from the kopia repository
+func (ss *streamStore) ReadBackupDetails(
+	ctx context.Context,
+	detailsID string,
+) (*details.Details, error) {
+	// construct the path for the `details` item
+	detailsPath, err := path.Builder{}.
+		Append("details").
+		ToServiceCategoryMetadataPath(
+			ss.tenant,
+			ss.resource,
+			ss.service,
+			path.DetailsCategory,
+			true,
+		)
+	if err != nil {
+		return nil, err
+	}
+
+	var bc stats.ByteCounter
+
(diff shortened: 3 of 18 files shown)


src/internal/model/model.go
```diff
@@ -27,7 +26,7 @@ const (
 
 // Valid returns true if the ModelType value fits within the iota range.
 func (mt Schema) Valid() bool {
-	return mt > 0 && mt < BackupDetailsSchema+1
+	return mt > 0 && mt < BackupSchema+1
```
ashmrtn, line 29: nit: maybe we should just make a sentinel value so we don't have to keep changing the code in two places when we update the set of model schemas we have

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
...
+
+// ReadBackupDetails reads the specified details object
+// from the kopia repository
+func (ss *streamStore) ReadBackupDetails(
+	ctx context.Context,
+	detailsID string,
+) (*details.Details, error) {
+	// construct the path for the `details` item
+	detailsPath, err := path.Builder{}.
+		Append("details").
```
ashmrtn, line 87: nit
```suggestion
		Append(detailsItemName).
```

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return errors.Wrap(err, "deleting backup details failed")
+	}
+
+	return nil
+}
+
+// streamCollection is a data.Collection used to persist
+// a single data stream
+type streamCollection struct {
```
ashmrtn, line 150: if you want this could be replaced by the `graph.MetadataCollection` type (same with the struct that implements `data.Stream`). We may revisit it later if we actually begin streaming data for this, but for now they both represent Collections where all data is in-memory.

If you do use the other collection, feel free to move it to a package that makes more sense

src/pkg/path/resource_path.go
```diff
@@ -93,13 +96,16 @@ var serviceCategories = map[ServiceType]map[CategoryType]struct{}{
 		EmailCategory:    {},
 		ContactsCategory: {},
 		EventsCategory:   {},
+		DetailsCategory:  {},
```
ashmrtn, line 99: I think it would be best to make a separate validate function for metadata paths (or a separate function to get details paths, whatever makes more sense). Thus far I've been trying to make it difficult to construct a path that is incorrect. Adding details to this map will allow other packages to construct paths like `tenant-id/service/user-id/details` and add regular item data there

ashmrtn (approved):

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
...
+
+	dc := &streamCollection{
+		folderPath: p,
+		item: &streamItem{
+			name: detailsItemName,
+			data: dbytes,
+		},
+	}
+
+	backupStats, _, err := ss.kw.BackupCollections(ctx, nil, []data.Collection{dc}, ss.service)
```
ryanfkeepers, line 71: Per the discussion in the other PR about querying for specific manifests, this could embed a unique Tag for looking up (or filtering against) details snapshots.  That PR modified this func to pass in a specific backupID.  A more generic resolution to fit both cases would be to pass in a map of `Tags`.

src/pkg/store/backup.go
```diff
@@ -81,44 +79,18 @@ func (w Wrapper) GetBackups(
 
 // DeleteBackup deletes the backup and its details entry from the model store.
 func (w Wrapper) DeleteBackup(ctx context.Context, backupID model.StableID) error {
-	deets, _, err := w.GetDetailsFromBackupID(ctx, backupID)
```
ryanfkeepers, line 84: Are we okay with orphaning details snapshots when the associated backup gets deleted?

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
+// streamstore implements helpers to store large
+// data streams in a repository
+package streamstore
```
ryanfkeepers, line 3: Two nits/questions about this package:

1/ should it live in /kopia?  The two packages seem fairly coupled.

2/ imo, the name feels a little off.  Any `fooStore` sounds like you're running a separate type of store (external to kopia, which is doubly easy to interpret from the package separation), rather than wrapping an existing store.  Not entirely sure what another name would be.  `streamBlob`?  `streamWriter`?  Maybe just reverse it: `storestream`?

src/internal/streamstore/streamstore.go
ryanfkeepers: Going back to the other discussion, this suggestion is unnecessary at this time.  We can handle it in the future, when needed.

ryanfkeepers:
> be a breaking change

Good time for us to decide how we identify when to do a minor version bump as opposed to a patch bump during release.

src/internal/streamstore/streamstore.go
vkamra: 1) I thought about putting it in Kopia but the layering doesn't seem right there. I'd rather it be a consumer of Kopia (which is what it is right now - similar to model store).

2) I'd agree the name isn't perfect but I do think we should think of this as a separate store. I earlier had it as `metastore` which might be a better name. One of my thoughts as I was going through implementation was that this might be a good abstraction to plug in other external stores for this metadata. For example - if we wanted to write details directly into an external service.

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/internal/kopia"
+	"github.com/alcionai/corso/src/internal/stats"
+	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+type streamStore struct {
+	kw       *kopia.Wrapper
+	resource string
```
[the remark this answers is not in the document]
vkamra: To clarify - will remove here.

src/internal/streamstore/streamstore.go
vkamra: 😱 - yes - thank you!

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// Expect only 1 data collection
+	if len(dcs) != 1 {
+		return nil, errors.Errorf("expected 1 details data collection: %d", len(dcs))
+	}
+
+	dc := dcs[0]
+
+	for {
```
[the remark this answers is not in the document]
vkamra: They do but I believe the linter was complaining. May have been before I did a refactor though - so will confirm.

src/internal/streamstore/streamstore_test.go
```diff
@@ -0,0 +1,74 @@
+package streamstore
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"github.com/zeebo/assert"
```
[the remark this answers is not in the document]
vkamra: 🤦 - will fix.

src/pkg/store/backup.go
vkamra: Ah no - I need to wire up the call to `DeleteDetails` in the backup delete calls (in CLI).

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- *Breaking* Changed how backup details are stored in the repository to
```
vkamra, line 12: Will update this possibly in a follow up once the change in behavior is documented

src/go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/tidwall/pretty v1.2.1
 	github.com/tomlazar/table v0.1.2
 	github.com/vbauerster/mpb/v8 v8.1.4
+	github.com/zeebo/assert v1.1.0
```
ashmrtn, line 27: I believe this can also be removed since the assert package was added by mistake

src/internal/streamstore/streamstore.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// Expect only 1 data collection
+	if len(dcs) != 1 {
+		return nil, errors.Errorf("expected 1 details data collection: %d", len(dcs))
+	}
+
+	dc := dcs[0]
+
+	select {
```
ashmrtn, line 119: sorry to go back and forth on this, but I think there actually should be a for loop here. It's safe to ignore all the other files that we don't need right now, but for resource management we should drain the `Items()` channel before exiting

vkamra: Easy enough to add back but won't resources get cleaned up after we exit?

I do see a potential correctness issue if we ever decide to write more items in the collection and don't read all of them - the producer could potentially end up blocking. Is that what you're worried about?

src/internal/streamstore/streamstore.go
vkamra: Will do this in a follow up. There's a `statusUpdater` dependency in that package that makes moving more involved.

src/internal/streamstore/streamstore.go
ashmrtn: it's more memory/go routine management. The collections often use another goroutine to send items on the channel. Since there's no `Close()` function and no returned `closer` to call later, the assumption is that all items in the stream will be read. If we don't read all of them, then the goroutine and channel will stick around until Corso exits. Not the end of the world, but will consume some resources

vkamra: Addressed

src/internal/streamstore/streamstore.go
ashmrtn: ah, I forgot about that. Well, not required to consolidate code if it doesn't really fit the pattern we have already

PR #1759, opened 2022-12-09
Add a comment for updating Schema constants

### LINKED ISSUE #1757: comment how updates to model schema type should be made
Corso uses the `Schema` [type](https://github.com/alcionai/corso/blob/73877add313618c00f91e01cf8b7925025cefaf8/src/internal/model/model.go#L11) to store information about models in the model store. Since some info about this type is persisted, there should be a comment explaining how to update the type without breaking backwards compatability

ashmrtn:
## Description

Note what things need to stay the same and what things can change when updating schema constant values. This is required because some information about the constants are persisted in the model store.

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* closes #1757 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/model/model.go
+++ b/src/internal/model/model.go
@@ -11,6 +11,11 @@ type (
 	Schema   int
 )
 
+// Schema constants denote the type of model stored. The integer values of the
+// constants can be changed without issue, but the string values should remain
+// the same. If the string values are changed, additional code will be needed to
+// transform from the old value to the new value.
+//
 //go:generate go run golang.org/x/tools/cmd/stringer -type=Schema
 const (
 	UnknownSchema = Schema(iota)


src/internal/model/model.go
```diff
@@ -11,6 +11,11 @@ type (
 	Schema   int
 )
 
+// Schema constants denote the type of model stored. The integer values of the
+// constants can be changed without issue, but the string values should remain
+// the same. If the string values are changed, additional code will be needed to
+// transform from the old value to the new value.
+//
```
ryanfkeepers, line 18: Seems like we could just use string consts instead of an iota.

ryanfkeepers (approved):

src/internal/model/model.go
ashmrtn: eh, we could, but not really high on the list of things to do at the moment. Kind of sixes anyway. If golang had builtin support for transforming consts/enums to strings (apart from stringer) and back again then it would make more sense to use consts

PR #1760, opened 2022-12-09
Fix meta information for social card previews

ntolia:
## Description

This was being manuarlly overridden and was bad for SEO

## Type of change

- [x] :bug: Bugfix


serverless-mom (approved):
looks good

PR #1768, opened 2022-12-10
consume old deltas on exchange mail backup

ryanfkeepers:
## Description

When backing up exchange data, parse the
metadata collection of delta urls from prior runs
(if any exist) and pass those tokens along to the
fetch functions for re-use.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1725

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -118,7 +118,6 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
 // in case of a name change or relocation.
 // 2- deltas: folderID->deltaToken, used to look up previous delta token
 // retrievals.
-// Finally
 func parseMetadataCollections(
 	ctx context.Context,
 	colls []data.Collection,
@@ -136,7 +135,7 @@ func parseMetadataCollections(
 
 			select {
 			case <-ctx.Done():
-				return nil, nil, errors.Wrap(context.Canceled, "parsing collection metadata")
+				return nil, nil, errors.Wrap(ctx.Err(), "parsing collection metadata")
 			case item, ok := <-items:
 				if !ok {
 					breakLoop = true
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -2,7 +2,6 @@ package connector
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -33,7 +32,7 @@ import (
 func (gc *GraphConnector) DataCollections(
 	ctx context.Context,
 	sels selectors.Selector,
-	metadataCols []data.Collection,
+	metadata []data.Collection,
 ) ([]data.Collection, error) {
 	ctx, end := D.Span(ctx, "gc:dataCollections", D.Index("service", sels.Service.String()))
 	defer end()
@@ -43,14 +42,9 @@ func (gc *GraphConnector) DataCollections(
 		return nil, err
 	}
 
-	_, deltas, err := parseMetadataCollections(ctx, metadataCols)
-	if err != nil {
-		return nil, err
-	}
-
 	switch sels.Service {
 	case selectors.ServiceExchange:
-		return gc.ExchangeDataCollection(ctx, sels, deltas)
+		return gc.ExchangeDataCollection(ctx, sels, metadata)
 	case selectors.ServiceOneDrive:
 		return gc.OneDriveDataCollections(ctx, sels)
 	case selectors.ServiceSharePoint:
@@ -111,56 +105,6 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
 	}
 
 	return nil
-}
-
-// parseMetadataCollections produces two maps:
-// 1- paths: folderID->filePath, used to look up previous folder pathing
-// in case of a name change or relocation.
-// 2- deltas: folderID->deltaToken, used to look up previous delta token
-// retrievals.
-func parseMetadataCollections(
-	ctx context.Context,
-	colls []data.Collection,
-) (map[string]string, map[string]string, error) {
-	var (
-		paths  = map[string]string{}
-		deltas = map[string]string{}
-	)
-
-	for _, coll := range colls {
-		items := coll.Items()
-
-		for {
-			var breakLoop bool
-
-			select {
-			case <-ctx.Done():
-				return nil, nil, errors.Wrap(ctx.Err(), "parsing collection metadata")
-			case item, ok := <-items:
-				if !ok {
-					breakLoop = true
-					break
-				}
-
-				switch item.UUID() {
-				// case graph.PreviousPathFileName:
-				case graph.DeltaTokenFileName:
-					err := json.NewDecoder(item.ToReader()).Decode(&deltas)
-					if err != nil {
-						return nil, nil, errors.New("parsing delta token map")
-					}
-
-					breakLoop = true
-				}
-			}
-
-			if breakLoop {
-				break
-			}
-		}
-	}
-
-	return paths, deltas, nil
 }
 
 // ---------------------------------------------------------------------------
@@ -234,7 +178,7 @@ func (gc *GraphConnector) createExchangeCollections(
 func (gc *GraphConnector) ExchangeDataCollection(
 	ctx context.Context,
 	selector selectors.Selector,
-	deltas map[string]string,
+	metadata []data.Collection,
 ) ([]data.Collection, error) {
 	eb, err := selector.ToExchangeBackup()
 	if err != nil {
@@ -246,6 +190,11 @@ func (gc *GraphConnector) ExchangeDataCollection(
 		collections = []data.Collection{}
 		errs        error
 	)
+
+	_, deltas, err := exchange.ParseMetadataCollections(ctx, metadata)
+	if err != nil {
+		return nil, err
+	}
 
 	for _, scope := range scopes {
 		// Creates a map of collections based on scope
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -2,6 +2,7 @@ package connector
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -42,11 +43,14 @@ func (gc *GraphConnector) DataCollections(
 		return nil, err
 	}
 
-	// serialize metadata into maps here
+	_, deltas, err := parseMetadataCollections(ctx, metadataCols)
+	if err != nil {
+		return nil, err
+	}
 
 	switch sels.Service {
 	case selectors.ServiceExchange:
-		return gc.ExchangeDataCollection(ctx, sels)
+		return gc.ExchangeDataCollection(ctx, sels, deltas)
 	case selectors.ServiceOneDrive:
 		return gc.OneDriveDataCollections(ctx, sels)
 	case selectors.ServiceSharePoint:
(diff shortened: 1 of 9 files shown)


src/internal/connector/data_collections.go
```diff
@@ -109,6 +113,57 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
 	return nil
 }
 
+// parseMetadataCollections produces two maps:
+// 1- paths: folderID->filePath, used to look up previous folder pathing
+// in case of a name change or relocation.
+// 2- deltas: folderID->deltaToken, used to look up previous delta token
+// retrievals.
+// Finally
```
ashmrtn, line 121:
```suggestion
// Finally it offers the caller cake (which may or may not be a lie)
```

src/internal/connector/data_collections.go
```diff
@@ -109,6 +113,57 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
...
+func parseMetadataCollections(
```
ashmrtn, line 122: I'm a little hesitant to have this in this file. It makes the assumptions that
* all data categories will have the same metadata files (or if not requires extra logic)
* all data categories will deserialize their metadata into the same types here and through the entire callstack to the separate category functions

That may hold true or it may not. I'm not sure if I know enough of how things will hash out to say at this point

src/internal/connector/data_collections.go
```diff
@@ -109,6 +113,57 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
...
+
+	for _, coll := range colls {
+		items := coll.Items()
+
+		for {
+			var breakLoop bool
+
+			select {
+			case <-ctx.Done():
+				return nil, nil, errors.Wrap(context.Canceled, "parsing collection metadata")
```
ashmrtn, line 139:
```suggestion
				return nil, nil, errors.Wrap(ctx.Err(), "parsing collection metadata")
```

src/internal/connector/data_collections_test.go
```diff
@@ -302,6 +344,61 @@ func (suite *ConnectorCreateExchangeCollectionIntegrationSuite) TestMailFetch()
...
+
+			var metadata data.Collection
+
+			for _, coll := range collections {
+				if coll.FullPath().Service() == path.ExchangeMetadataService {
+					metadata = coll
+				}
+			}
+
+			require.NotNil(t, metadata, "found metadata collection")
```
ashmrtn, line 383:
```suggestion
			require.NotNil(t, metadata, "metadata collection")
```

src/internal/connector/data_collections_test.go
```diff
@@ -302,6 +344,61 @@ func (suite *ConnectorCreateExchangeCollectionIntegrationSuite) TestMailFetch()
...
+
+			_, deltas, err := parseMetadataCollections(ctx, []data.Collection{metadata})
+			require.NoError(t, err)
+
+			collections, err = gc.createExchangeCollections(ctx, test.scope, deltas)
+			require.NoError(t, err)
+
+			for _, coll := range collections {
+				if coll.FullPath().Service() != path.ExchangeMetadataService {
+					ec, ok := coll.(*exchange.Collection)
```
ashmrtn, line 393: is there a reason to specifically check this? From the perspective of all of the stuff outside of the `exchange` package the only thing that matters is that we get a `data.Collection` for everything

src/internal/connector/data_collections_test.go
```diff
@@ -302,6 +344,61 @@ func (suite *ConnectorCreateExchangeCollectionIntegrationSuite) TestMailFetch()
...
+func (suite *ConnectorCreateExchangeCollectionIntegrationSuite) TestDelta() {
+	ctx, flush := tester.NewContext()
+	defer flush()
+
+	var (
+		userID = tester.M365UserID(suite.T())
+		gc     = loadConnector(ctx, suite.T(), Users)
+	)
+
+	tests := []struct {
```
ashmrtn, line 356: should also test contacts too. Events does not have a delta endpoint

src/internal/connector/exchange/query_options.go
```diff
@@ -113,18 +113,6 @@ func CategoryToOptionIdentifier(category path.CategoryType) optionIdentifier {
 // which reduces the overall latency of complex calls
 // -----------------------------------------------------------------------
 
-// Delta requests for mail and contacts have the same parameters and config
-// structs.
-type DeltaRequestBuilderGetQueryParameters struct {
```
ashmrtn, line 118: nit: in the future this would be best done in a separate PR

src/internal/connector/exchange/service_iterators.go
```diff
@@ -307,10 +311,14 @@ func FetchContactIDsFromDirectory(
 		Contacts().
 		Delta()
 
+	if len(oldDelta) > 0 {
```
ashmrtn, line 314: this needs to remain commented out until we have all the other pieces for delta token-based incrementals in place. Otherwise we'll end up with backups that only contain the changes

src/internal/connector/exchange/service_iterators.go
```diff
@@ -367,10 +375,14 @@ func FetchMessageIDsFromDirectory(
 		Messages().
 		Delta()
 
+	if len(oldDelta) > 0 {
```
ashmrtn, line 378: same as above, this needs to remain disabled for now

src/internal/connector/data_collections.go
ryanfkeepers: 🍰

src/internal/connector/data_collections_test.go
ryanfkeepers: What I actually wanted to check is that "the query that uses a delta token should be different/empty relative to the previous query which didn't have a delta token".  It's not a very functional test at the moment (I added a TODO), but I underlying concept is useful.

My brain was burned out friday night and I had trouble coming up with a proper set of criteria.  I'll probably still punt that to a different PR, but wanted to establish a place for the test to mark that spot.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Heh, yeah, I realized that over the weekend, too.

src/internal/connector/data_collections.go
ryanfkeepers: > all data categories will have the same metadata files

This can be addressed if/when differences arise in the future via a pivot on the service/category.  Given how small the scope of usage is at this time, I'd rather start with a smaller footprint that has some hardcoded assumptions than to try and solve the world.

> all data categories will deserialize their metadata into the same types here and through the entire callstack to the separate category functions

As both the producer and consumer of its metadata, GC has the opportunity to make assumptions/assertions such as type constraints without worry of fault.  Standardizing on the broadly expected type used by (de)serialization up front will provide better foundation for future implementation.  .  After that, there's room to expand on how we interface with granular, different assumptions within categories or services.

ashmrtn: >This can be addressed if/when differences arise in the future via a pivot on the service/category.

Honestly, I'd like to see less of us manually pivoting behavior based on service/category and more of us taking advantage of language features like using interfaces and providing different concrete implementations. The generic code becomes cleaner and it becomes easier to add custom implementations because only the new implementation needs to be written instead of editing the entire call stack to add more pivot options

>As both the producer and consumer of its metadata

I think this is a bit of a red herring. GC is the top-level package for all of this stuff, but it's actually not the producer or consumer of the metadata. The producers and consumers are the individual implementations for each (service, category) pair. Tying the format to the top-level package which only contains broad abstractions seems like it's going to either lead to trying to fit things into the existing (suboptimal) design or cause more refactoring down the road.

I think that perhaps due to the current implementation that manually switches behavior based on the service/category it hasn't been as clear, but really GC (as a struct) is just an abstract entry point to reaching more specific behavior. All the subpackages of the `connector` package could be rewritten to implement a common interface, making GC just something that provides a uniform API and some setup code for other stuff. That would make it clear that GC itself is neither producing or consuming the data, just passing it to other parts of the system

src/internal/connector/data_collections_test.go
ashmrtn: I'd prefer a TODO containing this info instead of some test check that's only vaguely related. Code has a tendency to stick around longer than we expect sometimes, so it's often better to be a bit more conservative with stuff like this

src/internal/connector/data_collections.go
ryanfkeepers: > I think this is a bit of a red herring.

Good points.  This func has been moved into `exchange`, and is called here from line ~125.  We can get more idiomatic control here with some additional refactoring in this file.  But for now it should get things to meet up halfway.

src/internal/connector/data_collections_test.go
ryanfkeepers: Understood.  I think there's still remaining value, even in a just having it as a sanity test.  Not a strongly held opinion, in case you feel strongly about the todo.  I'm also feeling a little lax because I intend to fulfill the TODO her in a day or two, as soon as we've passed the previousPath and item comparison.  Revisiting the tests will be necessary to verify the changes as we tie all the work together..

ryanfkeepers:
@ashmrtn Any other reviews before moving this PR forward?

src/internal/connector/data_collections_test.go
```diff
@@ -233,6 +233,7 @@ func TestConnectorCreateExchangeCollectionIntegrationSuite(t *testing.T) {
 	if err := tester.RunOnAny(
 		tester.CorsoCITests,
 		tester.CorsoConnectorCreateExchangeCollectionTests,
+		"flomp",
```
ashmrtn, line 236:
```suggestion
```

src/internal/connector/data_collections_test.go
```diff
@@ -302,6 +303,77 @@ func (suite *ConnectorCreateExchangeCollectionIntegrationSuite) TestMailFetch()
...
+
+			for _, coll := range collections {
+				if coll.FullPath().Service() == path.ExchangeMetadataService {
+					metadata = coll
+				}
+			}
+
+			require.NotNil(t, metadata, "collections contains a metadata collection")
+
+			_, deltas, err := exchange.ParseMetadataCollections(ctx, []data.Collection{metadata})
```
ashmrtn, line 353: this is kind of sixes, but you could just call `gc.DataCollections()` 2x instead of manually parsing the metadata collection and calling `gc.createExchangeCollections()`. Whatever you prefer works

src/internal/connector/exchange/data_collections.go
```diff
@@ -0,0 +1,61 @@
...
+	var (
+		paths  = map[string]string{}
+		deltas = map[string]string{}
+	)
+
+	for _, coll := range colls {
+		items := coll.Items()
+
+		for {
+			var breakLoop bool
```
ashmrtn, line 31: this is personal preference, but if something is used for a flag I tend to explicitly set it to false just so I don't have to take the extra step to remember what the default value is
```suggestion
		breakLoop := false
```

src/internal/connector/exchange/data_collections.go
```diff
@@ -0,0 +1,61 @@
...
+
+// ParseMetadataCollections produces two maps:
+// 1- paths: folderID->filePath, used to look up previous folder pathing
+// in case of a name change or relocation.
+// 2- deltas: folderID->deltaToken, used to look up previous delta token
+// retrievals.
+func ParseMetadataCollections(
+	ctx context.Context,
+	colls []data.Collection,
+) (map[string]string, map[string]string, error) {
```
ashmrtn, line 21: I think this will work, though if we want stricter checking (or if we're worried about folder ID collisions) we may want to make these `map[path.CategoryType]map[string]string`

src/internal/kopia/snapshot_manager.go
```diff
@@ -202,7 +202,7 @@ func fetchPrevManifests(
 	return manifestsSinceLastComplete(mans), nil
 }
 
-// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
+// FetchPrevSnapshotManifests returns a set of manifests for complete and maybe
```
ashmrtn, line 205: still lost in rebase?
```suggestion
// fetchPrevSnapshotManifests returns a set of manifests for complete and maybe
```

ashmrtn (approved):
A few minor things (some of which are take or leave). Otherwise looks good

src/internal/connector/data_collections_test.go
ryanfkeepers: Poor old flompy.

src/internal/connector/data_collections_test.go
ryanfkeepers: Gonna keep it as is for now. We can look into refinements like this on the upcoming refactor.

src/internal/connector/exchange/data_collections.go
ryanfkeepers: Technically that's non-idiomatic go.  General wisdom suggests that any declaration to the zeroth value of a (non-pointer) variable should use the `var` creation.  Personally, I take that side as well.

src/internal/connector/exchange/data_collections.go
ryanfkeepers: Can't say I've seen any reason to worry over folder id collision (maybe you know of something?).  I'm not against strictness, but I think that in this case the additional complexity outweighs the value gained.

src/internal/kopia/snapshot_manager.go
ryanfkeepers: bugger.  Yes, thank you.

PR #1788, opened 2022-12-12
GC: Backup: SharePoint List Integration with Tests

### LINKED ISSUE #1403: GC: Backup: Sharepoint Data Collection. Fetch SharePoint List Data
Data Collection type required for Sharepoint. 
Will support:
- [x] Site [Lists](https://learn.microsoft.com/en-us/graph/api/resources/list?view=graph-rest-1.0)
- ~[Site](https://learn.microsoft.com/en-us/graph/api/resources/site?view=graph-rest-1.0). [Drives](https://learn.microsoft.com/en-us/graph/api/resources/drive?view=graph-rest-1.0)~

### LINKED ISSUE #1474: GC: Backup: SharePoint: List Collection
Create Sharepoint List Data Collection during Back-up Operation

### LINKED ISSUE #1795: GC: Backup: TestBackupOpIntegrationSuite to include Lists
Enable: `TestBackupOpIntegrationSuite/TestBackup_Run_sharePoint` to include `SharePoint.List` 

Tasks:
- [ ] Switch Selector L507 back to `Sites` rather than `Libraries`

dadams39:
## Description
Framework for the SharePoint backup workflow. 

### Special Instructions for Review
Ensure that the information used to build `path.Path` are in line with future PRs (@ashmrtn )
## Type of change

- [x] :sunflower: Feature

## Issue(s)

* closes #1403
* closes #1474<issue>
* closes #1795

## Test Plan
- SharePoint Integration checked:
  - `src/internal/connector/data_collections_test.go`
  - `TestSharePointDataCollection()`
- SharePoint List basics checked:  
  - `src/internal/connector/sharepoint/collection_test.go`
- SharePoint Operational Backup:
  - `src/internal/operations/backup_test.go`
  - `TestBackup_Run_sharePoint() ` verified as operational during testing
- [x] :zap: Unit test


dadams39:
WIP until the following objectives are complete:
- [ ] Test coverage shows that individual website can be backed up
- [ ] Pre-Fetch items contain all information required to be able to be used for Delta queries

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -71,17 +80,53 @@ func DataCollections(
...
+				path.ListsCategory,
+				false)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create collection path for site: %s", siteID)
+		}
+
+		collection := NewCollection(dir, serv, updater.UpdateStatus)
+		collection.AddJob(siteID)
+
+		fmt.Printf("Collection:\nPath: %v\nJobs: %v\n", collection.fullPath, collection.jobs)
```
ashmrtn, line 120: nit: should be removed prior to merging
```suggestion
```

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -71,17 +80,53 @@ func DataCollections(
...
+
+	if scope.Matches(selectors.SharePointSite, siteID) {
+		spcs := make([]data.Collection, 0)
+
+		dir, err := path.Builder{}.Append(siteID).
+			ToDataLayerSharePointPath(
+				tenantID,
+				siteID,
+				path.ListsCategory,
+				false)
```
ashmrtn, line 112: as long as we are consistent with how we name things going forward this is fine

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -71,17 +80,53 @@ func DataCollections(
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create collection path for site: %s", siteID)
+		}
+
+		collection := NewCollection(dir, serv, updater.UpdateStatus)
```
ashmrtn, line 117: it looks like lists doesn't support delta queries for items, so just doing this and leaving the implementations of functions `State()` and `PreviousPath()` for the collection is fine

ashmrtn: I stand corrected, the beta API does seem to have a delta endpoint. For now, we can leave this as is and go back and add some extra logic to implement what incrementals will need. There's no reason to squeeze that into this PR especially if we haven't yet started looking at doing incrementals for anything other than Exchange

src/internal/connector/data_collections_test.go
```diff
@@ -582,15 +596,48 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
...
+			},
+		},
+		{
+			name:     "SharePoint.Lists",
+			expected: 1,
+			sel: func(t *testing.T) selectors.Selector {
+				sel := selectors.NewSharePointBackup()
+				sel.Include(sel.Lists(
+					[]string{siteID},
+					selectors.Any(),
```
ryanfkeepers, line 627: We might want to narrow this down a bit.  Could probably be `"TODO"` for now.  In the future we'll probably want a curated site to test against, with a guaranteed default list of some kind to isolate queries against.

src/internal/connector/sharepoint/data_collections.go
ashmrtn: actually, should we be using the list name in this path? Otherwise all the items for all lists will land in the same directory and we also won't know what name each list had originally

src/internal/connector/data_collections_test.go
```diff
@@ -582,15 +596,48 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
...
+				return sel.Selector
+			},
+		},
+	}
+
+	for _, test := range tables {
+		t.Run(test.name, func(t *testing.T) {
+			cols, err := gc.DataCollections(ctx, test.sel(t), nil, control.Options{})
+			require.NoError(t, err)
+			assert.Equal(t, test.expected, len(cols))
```
ryanfkeepers, line 640: Can use the built-in `Len` checker here.
```suggestion
			assert.Len(t, cols, test.expected)
```

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -41,6 +41,7 @@ func DataCollections(
 		scopes      = b.DiscreteScopes(siteIDs)
 		collections = []data.Collection{}
 		errs        error
+		spcs        []data.Collection
```
ryanfkeepers, line 44: This should get moved to line ~55.  It's dangerous to share it across loop iterations.

src/internal/connector/sharepoint/data_collections.go
ryanfkeepers: Should be logged or removed.

src/internal/connector/data_collections_test.go
dadams39: I am unclear about this statement. Given the `siteID`, the query will return exactly one collection.  Within this collection, there are several (4 at last count) lists that are returned.

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -71,17 +80,53 @@ func DataCollections(
...
+	scope selectors.SharePointScope,
+	updater statusUpdater,
+	ctrlOpts control.Options,
+) ([]data.Collection, error) {
+	logger.Ctx(ctx).With("site", siteID).Debug("Creating SharePoint List Collections")
+
+	if scope.Matches(selectors.SharePointSite, siteID) {
+		spcs := make([]data.Collection, 0)
+
+		dir, err := path.Builder{}.Append(siteID).
```
ryanfkeepers, line 107: The site shouldn't be used as the folder.  If there isn't a list id/name that could be used here, we should at least use a static reference for "list".  Probably the path category string?

src/internal/connector/sharepoint/list.go
```diff
@@ -18,13 +18,14 @@ import (
...
-// loadLists is a utility function to populate the List object.
+// loadSiteLists is a utility function to populate a collection of SharePoint.List
+// objects associated with a given siteID.
 // @param siteID the M365 ID that represents the SharePoint Site
 // Makes additional calls to retrieve the following relationships:
 // - Columns
 // - ContentTypes
 // - List Items
-func loadLists(
+func loadSiteLists(
```
ryanfkeepers, line 28: This doesn't strike me as useful rename.  We're only handling site lists at the moment, and the func is already housed in the sharepoint package.

src/internal/connector/sharepoint/list.go
```diff
@@ -86,6 +73,42 @@ func loadLists(
 	return results, nil
 }
 
+// fetchRelationships utility function to retrieve associated relationships:
+// - Columns, ContentTypes, ListItems
+func fetchRelationships(
```
ryanfkeepers, line 78: Nit: "relationships" is a bit ambiguous.
```suggestion
func fetchListContents(
```

src/internal/connector/sharepoint/data_collections.go
dadams39: Each list would end up in the same directory with this method. Each list would be an item in that directory.  Is the desire to have a collection per list?
For example
```go
dir, err := path.Builder{}.Append(listID).ToDataLayerSharePointPath(tenantID, siteID, path.ListCategory, true)
```
Note: this would be true as this would be the item and not the directory. @ashmrtn

src/internal/connector/sharepoint/list.go
dadams39: Leaving the refactor for clarity to developers. It is true that we do not save personnel sites within `gc.SiteIDs`. This naming emphasizes that this call will only return lists that are strictly associated with a site. During prototyping, lists were added to several pages by users. Corso queries identify these lists as the possession of the user and not the site.

src/internal/connector/sharepoint/data_collections.go
dadams39: Resolved here [513b5d7](https://github.com/alcionai/corso/pull/1788/commits/513b5d7eea414a21e3cd2e5a96dd987c9caae65a)

src/internal/connector/sharepoint/list.go
dadams39: Resolved within [2f7fe31](https://github.com/alcionai/corso/pull/1788/commits/2f7fe310e3dd02babe7e8bd96cced47441bc90d3)

src/internal/connector/sharepoint/data_collections.go
dadams39: Resolved with [9b3af77](https://github.com/alcionai/corso/pull/1788/commits/9b3af77c07de84819f531f1425078cd5b5a89e3b)

src/internal/connector/sharepoint/data_collections.go
dadams39: Resolved within [54689ae](https://github.com/alcionai/corso/pull/1788/commits/54689ae7cdc60a33cce256b5cc4fd35be12282f4)

src/internal/connector/data_collections_test.go
dadams39: Resolved within [9b3af77](https://github.com/alcionai/corso/pull/1788/commits/9b3af77c07de84819f531f1425078cd5b5a89e3b)

src/internal/connector/sharepoint/data_collections.go
dadams39: Resolved [9b3af77](https://github.com/alcionai/corso/pull/1788/commits/9b3af77c07de84819f531f1425078cd5b5a89e3b)

dadams39:
All comments resolved. Awaiting review

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -71,17 +81,58 @@ func DataCollections(
...
+	if scope.Matches(selectors.SharePointSite, siteID) {
+		spcs := make([]data.Collection, 0)
+
+		listIDs, err := preFetchListIDs(ctx, serv, siteID)
+		if err != nil {
+			return nil, err
+		}
+		// Prefetch
+		for _, listID := range listIDs {
+			dir, err := path.Builder{}.Append(listID).
```
ashmrtn, line 114: do lists have meaningful names associated with them that we need to preserve? We don't _have_ to store it as the folder name if they do, but we should have a plan for how we do plan to store it if we need to

src/internal/connector/data_collections_test.go
```diff
@@ -175,14 +175,26 @@ func (suite *ConnectorDataCollectionIntegrationSuite) TestSharePointDataCollecti
 	connector := loadConnector(ctx, suite.T(), Sites)
 	tests := []struct {
 		name        string
+		expected    int
 		getSelector func(t *testing.T) selectors.Selector
```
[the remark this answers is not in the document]
dadams39: The variable is used in line 217.

src/internal/connector/sharepoint/data_collections.go
dadams39: It is unclear whether the `displayName` would be of use to the user. These are the list we are returning at present. 
- Access Requests
- Events
- SharePointHomeOrgLinks
- Documents

dadams39: Resolution within: [ae26725](https://github.com/alcionai/corso/pull/1788/commits/ae267258e36aa5c10c2fdd2280d6905e04e87154)

src/internal/connector/data_collections_test.go
```diff
@@ -607,15 +621,48 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
...
 	)
 
-	sel.Include(sel.Libraries(
-		[]string{siteID},
-		[]string{"foo"},
-		selectors.PrefixMatch(),
-	))
+	tables := []struct {
+		name     string
+		sel      func(t *testing.T) selectors.Selector
```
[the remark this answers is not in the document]
dadams39: Resolved within [927a1d7](https://github.com/alcionai/corso/pull/1788/commits/927a1d7c1f4f514e156e04168236ed76530a295d)

src/internal/connector/data_collections_test.go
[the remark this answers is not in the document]
dadams39: Resolved within [927a1d7](https://github.com/alcionai/corso/pull/1788/commits/927a1d7c1f4f514e156e04168236ed76530a295d)

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -104,22 +105,27 @@ func collectLists(
...
-				tenantID,
-				siteID,
-				path.ListsCategory,
-				false)
+		tuples, err := preFetchListIDs(ctx, serv, siteID)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to create collection path for site: %s", siteID)
+			return nil, err
 		}
+		// Prefetch
```
ryanfkeepers, line 112: Seems like the prefetch already happened?

src/internal/connector/sharepoint/collection.go
```diff
@@ -159,51 +159,48 @@ func (sc *Collection) populate(ctx context.Context) {
...
-		// Retrieve list data from M365
-		lists, err := loadSiteLists(ctx, sc.service, id)
-		objects += len(lists)
+	// Retrieve list data from M365
+	lists, err := loadSiteLists(ctx, sc.service, sc.fullPath.ResourceOwner(), sc.jobs)
+	if err != nil {
+		errs = support.WrapAndAppend(sc.fullPath.ResourceOwner(), err, errs)
+	}
 
+	objects += len(lists)
```
ryanfkeepers, line 168:
```suggestion
	count += len(lists)
```

ryanfkeepers (commented):
Mostly nits. Only really concerned about the writer close.  Otherwise it's good.

src/internal/connector/data_collections_test.go
```diff
@@ -626,14 +626,14 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
 	)
 
 	tables := []struct {
-		name     string
-		sel      func(t *testing.T) selectors.Selector
-		expected int
+		name       string
+		sel        func() selectors.Selector
+		comparator func(t assert.TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool
```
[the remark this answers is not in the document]
dadams39: Resolved with [c3b87d2](https://github.com/alcionai/corso/pull/1788/commits/c3b87d2336853bb14a74e731ee2f853bf33d44be)

src/internal/connector/data_collections_test.go
```diff
@@ -662,9 +662,9 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
 
 	for _, test := range tables {
 		t.Run(test.name, func(t *testing.T) {
-			cols, err := gc.DataCollections(ctx, test.sel(t), nil, control.Options{})
+			cols, err := gc.DataCollections(ctx, test.sel(), nil, control.Options{})
 			require.NoError(t, err)
-			assert.Equal(t, test.expected, len(cols))
+			test.comparator(t, len(cols), 0)
```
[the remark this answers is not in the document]
dadams39: Resolved with [c3b87d2](https://github.com/alcionai/corso/pull/1788/commits/c3b87d2336853bb14a74e731ee2f853bf33d44be)

src/internal/connector/sharepoint/data_collections.go
dadams39: Resolved with [02b11fd](https://github.com/alcionai/corso/pull/1788/commits/02b11fddc486ca263a35bfdfeb2c879f4bd88875)

src/internal/connector/sharepoint/list.go
```diff
@@ -11,6 +11,64 @@ import (
...
+		Select: selecting,
+	}
+	options := &mssite.SitesItemListsRequestBuilderGetRequestConfiguration{
+		QueryParameters: &queryOptions,
+	}
+
+	return options
+}
+
+func preFetchListIDs(
```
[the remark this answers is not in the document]
dadams39: Resolved with [02b11fd](https://github.com/alcionai/corso/pull/1788/commits/02b11fddc486ca263a35bfdfeb2c879f4bd88875)

src/internal/connector/sharepoint/collection.go
```diff
@@ -159,51 +159,48 @@ func (sc *Collection) populate(ctx context.Context) {
...
 
-			writer.Close()
+		byteArray, err := writer.GetSerializedContent()
+		if err != nil {
+			errs = support.WrapAndAppend(*lst.GetId(), err, errs)
+			continue
+		}
 
-			arrayLength = int64(len(byteArray))
+		writer.Close()
```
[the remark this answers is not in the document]
dadams39: Resolved within [d8232f4](https://github.com/alcionai/corso/pull/1788/commits/d8232f4c8b616fa3949f48be6e22d003346b5e71)

src/internal/connector/sharepoint/collection.go
```diff
@@ -148,52 +159,59 @@ func (sc *Collection) populate(ctx context.Context) {
...
-					t = *t1
-				}
+			colProgress <- struct{}{}
+		}
+	}
+}
 
-				totalBytes += arrayLength
+func serializeListContent(writer *kw.JsonSerializationWriter, lst models.Listable) ([]byte, error) {
+	defer writer.Close()
```
ryanfkeepers, line 204: This deferral doesn't protect from panics.  Since the writer (line 153) exists external to this func, and is not reconstructed on each iteration of the for loop, any len of lists > 1 will still cause a panic.  The `defer close()` needs to be co-located with the instance of the writer.  That means that either this func should generate a writer, rather than accept one; or the close should move out of this func.

If we're handling an iteration of items, we probably want the former behavior (creating the writer within this func).  I don't know that we'll get consistent or expected behavior from re-using a writer in the iteration, either.

ryanfkeepers (changes requested):

src/internal/connector/sharepoint/collection.go
dadams39: @ryanfkeepers We are probably disagreeing about how this function works. If a SerializationWriter is not closed, it does not cause panic. This is a helper function that I created in order to reuse the writer. This is the code [link](https://github.com/microsoft/kiota-serialization-json-go/blob/598b061421655f2d8dcc6541d4b75c0b9c1d7754/json_serialization_writer.go#L730)
```go
func (w *JsonSerializationWriter) Close() error {
	w.writer = make([]string, 0)
	return nil
}
```

For efficiency's sake, it is overkill to recreate the object in every loop.

ryanfkeepers: Mmm, fair. I disagree on the naming of `Close`, in this case, because it carries the implication that the resource will be unusable afterwards.  `Reset` would be a better func name here.

Either way, thanks for the insight.  That's enough to unblock things for me.

ryanfkeepers (approved):

PR #1790, opened 2022-12-13
move graphService to graph as Service

ryanfkeepers:
## Description

Relocates the graphService struct to graph as
the Service struct.  Replaces GC's embedded
graphService with a graph.Servicer reference.

## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* #1725

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -25,6 +25,28 @@ type QueryParams struct {
 	Credentials   account.M365Config
 }
 
+var _ Servicer = &Service{}
+
+type Service struct {
+	adapter msgraphsdk.GraphRequestAdapter
+	client  msgraphsdk.GraphServiceClient
+}
+
+func NewService(adapter *msgraphsdk.GraphRequestAdapter) Service {
+	return Service{
+		adapter: *adapter,
+		client:  *msgraphsdk.NewGraphServiceClient(adapter),
+	}
+}
+
+func (s Service) Adapter() *msgraphsdk.GraphRequestAdapter {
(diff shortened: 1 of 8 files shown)


src/internal/connector/graph/service.go
```diff
@@ -25,6 +25,28 @@ type QueryParams struct {
...
+
+type Service struct {
+	adapter msgraphsdk.GraphRequestAdapter
+	client  msgraphsdk.GraphServiceClient
+}
+
+func NewService(adapter *msgraphsdk.GraphRequestAdapter) Service {
+	return Service{
+		adapter: *adapter,
+		client:  *msgraphsdk.NewGraphServiceClient(adapter),
```
ashmrtn, line 38: if we return pointers to these anyway in `Adapther()` and `Client()` is there a reason to store a copy of them instead of a pointer to them?

ashmrtn (approved):

src/internal/connector/graph/service.go
ryanfkeepers: so, here's the question: my inclination is to make all of this a value copy.  Are there strong preferences towards passing Service as a pointer instead?

ashmrtn: mmm I think it would be good to see what the functions that consume service/adapter need, especially in the Graph SDK. If the SDK mostly uses pointers then we should return pointers so we don't have to take the address of stuff all over the place. Beyond that I don't really have a preference

Issue #1792, opened 2022-12-13
Stats update: store backup status string, human duration


ryanfkeepers:
Update the following fields:
* `status` int -> string
* `duration` add `human_duration` string; ex `1h 12m 38s`

ryanfkeepers:
Per review, dropped human_duration

PR #1793, opened 2022-12-13
update time and status fields in op events

ryanfkeepers:
## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* #1792

## Test Plan

- [x] :muscle: Manual


src/internal/events/events.go
```diff
@@ -38,6 +38,7 @@ const (
 	DataStored       = "data_stored"
 	Duration         = "duration"
 	EndTime          = "end_time"
+	HumanDuration    = "human_duration"
```
gmatev, line 41: What is the rationale of having two of these fields? Not sure I understand the difference between the two.

gmatev (changes requested):
What is the difference between `duration` and `human_duration` and why we need both?

src/internal/events/events.go
ryanfkeepers: From the slack thread you weren't able to make much immediate use of the duration value, I added the HumanDuration to give you something easily read.  However, it's not going to be easy to query against, and I assume in the future you'll want to query for data points like "count of events that took longer than X", which is why I kept the old value, too.

gmatev (approved):

Issue #1726, opened 2022-12-07
GC should accept id->path lookup collections on backup


ryanfkeepers:
## Issue
Extend the usage of metadata collections in GC backups to include building out a folder id -> path map.

## Acceptance Criteria
- [x] in operations, get path mapping metadata collections from kopia’s previous snapshot
- [x] pass collections into GC backup
- [x] serialize collections into a map of folder id -> path

Might be able to populate a ContainerResolver instead of a map.  Should probably pull this info in https://github.com/alcionai/corso/blob/14a3c2e18951481d2cb19cd8ba4b2be386d1e1ac/src/internal/connector/exchange/service_iterators.go#L116

PR #1799, opened 2022-12-13
produce previousPaths metadata

ryanfkeepers:
## Description

Adds an additional metadata collection: a folder
id to path string mapping.  This collection is
created on backup, and retrieved along with
the delta metadata on the next backup, but is
not yet parsed or utilzed downstream.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1726

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -89,56 +89,23 @@ func FilterContainersAndFillCollections(
 
 		// add the current path for the container ID to be used in the next backup
 		// as the "previous path", for reference in case of a rename or relocation.
-		if itp, err := resolver.IDToPath(ctx, cID); err != nil {
-			errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
-		} else {
-			// Some root folders have no display name. The exchange path will
-			// fail without that, so we inject a fallback here just for
-			// the previousPaths handling.
-			els := itp.Elements()
-			if len(els) == 0 || len(els[len(els)-1]) == 0 {
-				itp = itp.Append(rootFolderAlias)
-			}
-
-			p, err := itp.ToDataLayerExchangePathForCategory(
-				qp.Credentials.AzureTenantID,
-				qp.ResourceOwner,
-				qp.Category,
-				false)
-			if err != nil {
-				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
-			} else {
-				prevPaths[cID] = p.Folder()
-			}
-		}
+		prevPaths[cID] = dirPath.Folder()
 	}
 
 	if col, err := graph.MakeMetadataCollection(
 		qp.Credentials.AzureTenantID,
 		qp.ResourceOwner,
-		graph.DeltaTokenFileName,
 		path.ExchangeService,
 		qp.Category,
-		deltaURLs,
+		[]graph.MetadataCollectionEntry{
+			graph.NewMetadataEntry(graph.DeltaTokenFileName, deltaURLs),
+			graph.NewMetadataEntry(graph.PreviousPathFileName, prevPaths),
+		},
 		statusUpdater,
 	); err != nil {
-		errs = support.WrapAndAppend("making delta metadata collection", err, errs)
+		errs = support.WrapAndAppend("making metadata collection", err, errs)
 	} else if col != nil {
-		collections[graph.DeltaMetadataCollectionKey] = col
-	}
-
-	if col, err := graph.MakeMetadataCollection(
-		qp.Credentials.AzureTenantID,
-		qp.ResourceOwner,
-		graph.PreviousPathFileName,
-		path.ExchangeService,
-		qp.Category,
-		prevPaths,
-		statusUpdater,
-	); err != nil {
-		errs = support.WrapAndAppend("making paths metadata collection", err, errs)
-	} else if col != nil {
-		collections[graph.PathsMetadataCollectionKey] = col
+		collections["metadata"] = col
 	}
 
 	return errs
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -92,15 +92,20 @@ func FilterContainersAndFillCollections(
 		prevPaths[cID] = dirPath.Folder()
 	}
 
+	entries := []graph.MetadataCollectionEntry{
+		graph.NewMetadataEntry(graph.PreviousPathFileName, prevPaths),
+	}
+
+	if len(deltaURLs) > 0 {
+		entries = append(entries, graph.NewMetadataEntry(graph.DeltaTokenFileName, deltaURLs))
+	}
+
 	if col, err := graph.MakeMetadataCollection(
 		qp.Credentials.AzureTenantID,
 		qp.ResourceOwner,
 		path.ExchangeService,
 		qp.Category,
-		[]graph.MetadataCollectionEntry{
-			graph.NewMetadataEntry(graph.DeltaTokenFileName, deltaURLs),
-			graph.NewMetadataEntry(graph.PreviousPathFileName, prevPaths),
-		},
+		entries,
 		statusUpdater,
 	); err != nil {
 		errs = support.WrapAndAppend("making metadata collection", err, errs)
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -1,9 +1,7 @@
 package exchange
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -81,84 +32,110 @@ func FilterContainersAndFillCollections(
...
 		edc.jobs = append(edc.jobs, jobs...)
 
 		if len(delta) > 0 {
-			deltaURLs[dirID] = delta
+			deltaURLs[cID] = delta
+		}
+
+		// add the current path for the container ID to be used in the next backup
+		// as the "previous path", for reference in case of a rename or relocation.
+		if itp, err := resolver.IDToPath(ctx, cID); err != nil {
```
ashmrtn, line 89: I think `dirPath` from above already has what this is doing?

src/internal/connector/exchange/service_iterators.go
```diff
@@ -81,84 +32,110 @@ func FilterContainersAndFillCollections(
...
+			errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
+		} else {
+			// Some root folders have no display name. The exchange path will
+			// fail without that, so we inject a fallback here just for
+			// the previousPaths handling.
+			els := itp.Elements()
+			if len(els) == 0 || len(els[len(els)-1]) == 0 {
+				itp = itp.Append(rootFolderAlias)
```
ashmrtn, line 97: _Probably_ ok to skip this? I don't think there's any cases where we persist the root directory, the root directory has items, and we don't give the root directory a path

ashmrtn, line 97: `rootFolderAlias` refers specifically to the Exchange mail root folder. Should probably be renamed to reflect that in a different PR eventually

src/internal/connector/exchange/service_iterators.go
```diff
@@ -81,84 +32,110 @@ func FilterContainersAndFillCollections(
...
+			p, err := itp.ToDataLayerExchangePathForCategory(
+				qp.Credentials.AzureTenantID,
+				qp.ResourceOwner,
+				qp.Category,
+				false)
+			if err != nil {
+				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
+			} else {
+				prevPaths[cID] = p.Folder()
+			}
```
ashmrtn, line 109: this can also be removed because we can use `dirPath` (I think)
```suggestion
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -81,84 +32,110 @@ func FilterContainersAndFillCollections(
...
+		graph.PreviousPathFileName,
+		path.ExchangeService,
+		qp.Category,
+		prevPaths,
+		statusUpdater,
+	); err != nil {
+		errs = support.WrapAndAppend("making paths metadata collection", err, errs)
 	} else if col != nil {
-		collections[metadataKey] = col
+		collections[graph.PathsMetadataCollectionKey] = col
```
ashmrtn, line 138: this is going to cause undefined behavior in KW when it tries to consume it. We now have two collections that have the same `FullPath()`. We should instead be creating a single metadata collection with two files in it

src/internal/connector/graph/service.go
```diff
@@ -9,9 +9,23 @@ import (
...
+	PreviousPathFileName = "previouspath"
+)
+
+const (
+	// DeltaMetadataCollectionKey is the name of the collection holding the delta
+	// url lookups inside a collection map otherwise keyed by folder ids.
+	DeltaMetadataCollectionKey = "delta-metadata"
+	// PathsMetadataCollectionKey is the name of the collection holding the previous
+	// paths lookups inside a collection map otherwise keyed by folder ids.
+	PathsMetadataCollectionKey = "paths-metadata"
```
ashmrtn, line 27: this is just an artifact of needing to have some map key so we can add it to the set of collections the exchange code deals with. It could really be a single map key (since we should only be making one collection) and it should be in the exchange files not here (it's not something that will be needed outside of backup). Though, it may also be possible to use the `col.FullPath().ShortRef()` or `col.FullPath().String()` as the map key for the metadata collection

The map was used just to make sure that we only had one instance of a collection for each path when we were fetching items in a different way. It's probably not really required anymore but that code hasn't been refactored in a while

src/internal/connector/graph/service.go
```diff
@@ -9,9 +9,23 @@ import (
...
-// DeltaTokenFileName is the name of the file containing delta token(s) for a
-// given endpoint. The endpoint granularity varies by service.
-const DeltaTokenFileName = "delta"
+const (
+	// DeltaTokenFileName is the name of the file containing delta token(s) for a
+	// given endpoint. The endpoint granularity varies by service.
+	DeltaTokenFileName = "delta"
+	// PreviousPathFileName is the name of the file containing previous path(s) for a
+	// given endpoint.
+	PreviousPathFileName = "previouspath"
```
ashmrtn, line 18: This should be added to the return value of `MetadataFileNames` eventually

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Oh... yes.  Yes it does.

src/internal/connector/exchange/service_iterators.go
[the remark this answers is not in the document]
ryanfkeepers: Fwiw, it wasn't about consistency so much as getting yelled at by the builder for having zero elements.  But it's all good, because I can just use the `dirPath`!

src/internal/connector/exchange/service_iterators.go
```diff
@@ -81,84 +32,110 @@ func FilterContainersAndFillCollections(
...
+
+			p, err := itp.ToDataLayerExchangePathForCategory(
+				qp.Credentials.AzureTenantID,
+				qp.ResourceOwner,
+				qp.Category,
+				false)
+			if err != nil {
+				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
+			} else {
+				prevPaths[cID] = p.Folder()
```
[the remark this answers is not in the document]
ryanfkeepers: I was thinking about this, and that we'd probably be safer overall storing slice over the folder as a string.  Saves us from parsing collisions like this.  Something I'll change in a follow-up, to not overwhelm this PR.

src/internal/connector/graph/service.go
ryanfkeepers: whoops, good catch!

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Bother.  I knew something about that was tickling my brain.  Should be fixed now.

src/internal/connector/exchange/service_iterators.go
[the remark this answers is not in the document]
ashmrtn: mmm so there's a subtle difference here. If you just use `path.Builder` (what the `CachedContainer.Path()` returns then it can be empty. If you use a `path.Path` implementation then it will have the prefix elements (tenant, service, etc) and require at least one folder or item name element. The `ToDataLayerX()` functions create `path.Path` instances

src/internal/connector/exchange/service_iterators.go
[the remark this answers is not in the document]
ashmrtn: >safer overall storing slice over the folder as a string.

I'm not sure I understand this. Can you explain more or point me to the code when you have it if it's going to be in this PR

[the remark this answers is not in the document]
ryanfkeepers: Ah, sorry, late night brain was getting to me.  I did half a read and then misinterpreted your comment as worrying about string->path parsing.  The test is a good idea, and I'll add it in, but personally I have very few worries about json decoding/encoding. Or, at least, the golang package for it.  If graph has its own bugs there I'm not aware of them.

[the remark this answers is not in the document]
ashmrtn: mmm I don't think it would be a bug in the golang JSON parser. I'm more worried about us not properly escaping things we need to when handing data to it and that causing problems down the line. It's more of a misunderstanding of the API and assumptions than a bug

src/internal/connector/graph/metadata_collection.go
```diff
@@ -88,7 +163,7 @@ func (md MetadataCollection) Items() <-chan data.Stream {
 	return res
 }
 
-// MetadataItem is an in-memory data.Stream implementation. MetadataItem does
+// metadataItem is an in-memory data.Stream implementation. MetadataItem does
```
ashmrtn, line 166: Not sure if this was meant to be changed to unexported or not
```suggestion
// MetadataItem is an in-memory data.Stream implementation. MetadataItem does
```

src/internal/connector/graph/service.go
```diff
@@ -9,14 +9,30 @@ import (
...
 // metadata such as delta tokens and folderID->path references.
-func MetadataFileNames() []string {
-	return []string{DeltaTokenFileName}
+func AllMetadataFileNames() []string {
+	return []string{DeltaTokenFileName, PreviousPathFileName}
+}
+
+// MetadataFileNames produces the category-specific set of filenames used to
+// store graph metadata such as delta tokens and folderID->path references.
+func MetadataFileNames(cat path.CategoryType) []string {
```
ashmrtn, line 29: this feels like eventually it should live in the exchange package or something, but for now this works

ashmrtn (approved):

src/internal/connector/graph/service.go
ryanfkeepers: Sure, why not.

src/internal/connector/graph/metadata_collection.go
ryanfkeepers: Oh yeah, i went back and forth on that a few times.

src/internal/connector/exchange/service_iterators.go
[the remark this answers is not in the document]
ryanfkeepers: Topic resolved in sync discussion; appropriate tests added.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -83,85 +34,83 @@ func FilterContainersAndFillCollections(
...
-
-		jobs, delta, err := fetchFunc(ctx, edc.service, qp.ResourceOwner, dirID, oldDelta)
+		jobs, delta, err := fetchFunc(ctx, edc.service, qp.ResourceOwner, cID, oldDeltas[cID])
 		if err != nil {
-			errs = support.WrapAndAppend(
-				qp.ResourceOwner,
-				err,
-				errs,
-			)
+			errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
```
dadams39, line 81: Should a continue be added following this error?

Issue #1740, opened 2022-12-08
Add logic for merging directory hierarchies for incremental backups


ashmrtn:
Delta token-based incremental backups need to source some snapshot information from the snapshot that sourced the delta tokens used in the backup. This ticket addresses the need to merge the directory hierarchy of the "base" (previous) snapshot that sourced the tokens with the directory hierarchy that is present in the `data.Collection`s given to `kopia.Wrapper.BackupCollections`

Merging needs to be done on two levels to properly handle everything:
* directory hierarchy changes need to be accounted for. This includes adding, removing, and moving/renaming collections
* items within each directory need to be merged. This includes item updates and deletions

Items and directories that are not represented in the information sent from GraphConnector should be assumed to be unchanged and sourced from the base snapshot. Directories that have been moved/renamed should contain all their previous items unless the information from GraphConnector states otherwise

PR #1801, opened 2022-12-13
Merge directory layout for kopia incremental backups

ashmrtn:
## Description

Merge directory layouts between the passed in collections and the base snapshot(s). Also add unit tests to ensure the output kopia hierarchy looks as expected. (CLI) user observable behavior is not affected by this PR

This PR does not address:
* selecting subtrees for specific data categories in base snapshots
* not clobbering more recent info if multiple snapshots have subtrees for the same data category (ties into above)
* file deletions for services that can only report item deletions at a global level (e.x. OneDrive file deletions)

Viewing individual commits in PR may make changes easier to review

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1740

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/upload_test.go
```diff
@@ -674,8 +755,867 @@ func (suite *HierarchyBuilderUnitSuite) TestBuildDirectoryTree_Fails() {
...
+			inputCollections: func(t *testing.T) []data.Collection {
+				personal := mockconnector.NewMockExchangeCollection(personalPath, 0)
+				personal.PrevPath = personalPath
+				personal.ColState = data.DeletedState
+
+				work := mockconnector.NewMockExchangeCollection(personalPath, 0)
+				work.PrevPath = workPath
+				work.ColState = data.MovedState
+
+				return []data.Collection{personal, work}
```
ryanfkeepers, line 1253: Ordering is paramount here, yeah?  I assume that if the collection order were reversed: `{work, personal}`, then the work file would be added to personalDir first, then deleted along with the rest of the directory.  That poses some risks in terms of collection ordering; I don't recall (and maybe I'm just forgetting) that we've made any guarantees about the order of collections from backup data producers.  Do we need to sort collections by modTime for this to work as expected?

ryanfkeepers (approved):
Looks good overall.  I think there's room for cleanup, but that's ultimately a nit.  Not sure if the ordering concern is an issue.

src/internal/kopia/upload_test.go
ashmrtn: nope, ordering does not matter at all since we first create the hierarchy and then send items to kopia. We may need to be a little more concerned about ordering if we ever start building the hierarchy in tandem with sending items to kopia, but we're not doing that right now

src/internal/kopia/upload_test.go
```diff
@@ -21,6 +24,107 @@ import (
...
+func expectFileData(
+	t *testing.T,
+	ctx context.Context,
+	expected []byte,
+	f fs.StreamingFile,
+) {
+	//revive:enable:context-as-argument
+	t.Helper()
+
+	if expected == nil {
```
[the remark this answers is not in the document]
ashmrtn: actually since this function assumes the version specifier is in there it's fine to do a check on length because there should be no files of size 0 (the minimum file size is the length of the version)

Issue #1779, opened 2022-12-12
return reason why a snapshot was selected as a base for an incremental backup


ashmrtn:
Backups are split on the tuple (resource owner, service) so it's possible that snapshots in kopia contain overlapping data from different points in time. Corso needs to be able to determine the reason a certain kopia snapshot was selected so it sources the correct data and metadata from it.

For example, consider the following:
```text
backup create exchange --data email,contacts -> B1
// Should use B1 for contacts
backup create exchange --data contacts -> B2
// Should use B1 for email and B2 for contacts
backup create exchange
```

PR #1802, opened 2022-12-13
consume prevPath on next backup

ryanfkeepers:
## Description

Parses the previousPaths metadata collections
along with deltas, and hands paths down to
exchange backup collection producers.  Does
not yet scrutinize previous/current path diffs.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1726

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/data_collections.go
+++ b/src/internal/connector/exchange/data_collections.go
@@ -22,45 +22,75 @@ func MetadataFileNames(cat path.CategoryType) []string {
 	}
 }
 
-// ParseMetadataCollections produces two maps:
-// 1- paths: folderID->filePath, used to look up previous folder pathing
-// in case of a name change or relocation.
-// 2- deltas: folderID->deltaToken, used to look up previous delta token
-// retrievals.
+type CatDeltaPaths map[path.CategoryType]DeltaPaths
+
+type DeltaPaths struct {
+	deltas map[string]string
+	paths  map[string]string
+}
+
+func makeDeltaPaths() DeltaPaths {
+	return DeltaPaths{
+		deltas: map[string]string{},
+		paths:  map[string]string{},
+	}
+}
+
+// ParseMetadataCollections produces a map of structs holding delta
+// and path lookup maps.
 func ParseMetadataCollections(
 	ctx context.Context,
 	colls []data.Collection,
-) (map[string]string, map[string]string, error) {
-	var (
-		paths  = map[string]string{}
-		deltas = map[string]string{}
-	)
+) (CatDeltaPaths, error) {
+	cdp := CatDeltaPaths{
+		path.ContactsCategory: makeDeltaPaths(),
+		path.EmailCategory:    makeDeltaPaths(),
+		path.EventsCategory:   makeDeltaPaths(),
+	}
 
 	for _, coll := range colls {
-		items := coll.Items()
+		var (
+			breakLoop bool
+			items     = coll.Items()
+			category  = coll.FullPath().Category()
+		)
 
 		for {
-			var breakLoop bool
-
 			select {
 			case <-ctx.Done():
-				return nil, nil, errors.Wrap(ctx.Err(), "parsing collection metadata")
+				return nil, errors.Wrap(ctx.Err(), "parsing collection metadata")
+
 			case item, ok := <-items:
 				if !ok {
 					breakLoop = true
 					break
 				}
 
+				m := map[string]string{}
+				cdps := cdp[category]
+
+				err := json.NewDecoder(item.ToReader()).Decode(&m)
+				if err != nil {
+					return nil, errors.New("decoding metadata json")
+				}
+
 				switch item.UUID() {
-				// case graph.PreviousPathFileName:
-				case graph.DeltaTokenFileName:
-					err := json.NewDecoder(item.ToReader()).Decode(&deltas)
-					if err != nil {
-						return nil, nil, errors.New("parsing delta token map")
+				case graph.PreviousPathFileName:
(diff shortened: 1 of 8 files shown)


src/internal/connector/exchange/data_collections.go
```diff
@@ -39,15 +40,24 @@ func ParseMetadataCollections(
...
+
 				switch item.UUID() {
-				// case graph.PreviousPathFileName:
+				case graph.PreviousPathFileName:
+					err := json.NewDecoder(item.ToReader()).Decode(&m)
+					if err != nil {
+						return nil, nil, errors.New("parsing delta token map")
+					}
+
+					paths = common.UnionMaps(m, paths)
```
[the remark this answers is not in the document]
ryanfkeepers: I'm convinced.  That's been updated.

src/internal/connector/exchange/data_collections.go
```diff
@@ -16,51 +17,78 @@ import (
...
-					breakLoop = true
+					cdps.deltas = common.UnionMaps(cdps.deltas, m)
+
+				case graph.PreviousPathFileName:
+					err := json.NewDecoder(item.ToReader()).Decode(&m)
+					if err != nil {
+						return nil, errors.New("parsing previous path map")
+					}
+
+					cdps.paths = common.UnionMaps(cdps.paths, m)
```
ashmrtn, line 88: this kind of worries me because it seems like it's going to make it harder to find bugs

Under the current implementation of selectors it's technically possible for GC to be backing up multiple resource owners at the same time. This is saying that we are going to make a single large map containing all the delta URLs and path mappings for all resource owners

Under the expected updated version of selectors, we would have a 1:1 mapping between resource owner and other filters/includes, but then it should be an error to have multiple different "base" snapshots for a single (resource owner, service, category) tuple. It's possible to have multiple bases for a (resource owner, service) pair, but once the category is added it should be 1:1, otherwise we are essentially pulling data from multiple different points in time and that will lead to correctness bugs

ryanfkeepers: > Under the current implementation of selectors it's technically possible for GC to be backing up multiple resource owners at the same time.

A bit pedantic, but: although we're looking to selectors for part of this fix, this is less of a selectors issue to enforce than it is a GC/operations layer enforcement.

> This is saying that we are going to make a single large map containing all the delta URLs and path mappings for all resource owners

We're already guaranteeing that isolation from the CLI, and although the sdk api is awaiting that enforcement as well our only known sdk consumer is known to work on a per-resource-owner basis.  I don't think there's much to stress about here.

In either case, the larger concern might be memory issues.  A correctness bug would require duplicate folderIDs across resource owners.  Unlikely, at best.

ashmrtn: >A correctness bug would require duplicate folderIDs across resource owners. Unlikely, at best.

Actually there's another correctness issue: we could end up accidentally pulling in metadata from multiple base snapshots and that metadata could overlap in some places. I don't think there's currently code to handle this anywhere in the codebase (I've been working on a patch for KW to handle this when building the hierarchy). Under that condition, it would be possible to end up with two sets of delta tokens and path mappings for a category, one from each of the bases. It would be incorrect to use both and it's also incorrect to have some merge algorithm for them. We should only ever use all of one or all of the other for any given (resource owner, service, category) tuple

To make this more concrete, consider my new favorite example:
```text
// Full backup, no base or metadata
backup create exchange user1 email,contacts -> B1

// Should use B1 to source email metadata
backup create exchange user1 email -> B2

// Should use B1 to source contacts metadata and B2 to source email metadata
backup create exchange user1 email,contacts
```

If we just pull all the metadata for all the base backups and merge the results then we'll end up with some mixture of B1 and B2 for the starting point of email when making the final backup. That will be a correctness issue and I think it would be undefined behavior with the current code since the outcome is dependent on iteration order in some places as well as how the M365 hierarchy has changed

src/internal/operations/backup_test.go
```diff
@@ -229,6 +230,8 @@ func checkMetadataFilesExist(
 		for item := range col.Items() {
 			assert.Implements(t, (*data.StreamSize)(nil), item)
 
+			fmt.Printf("\n-----\nrestored %v %v\n-----\n", item.UUID(), col.FullPath())
```
ashmrtn, line 233: nit: should be removed prior to merging

ashmrtn (approved):
I would encourage some update to how we're merging metadata after it's deserialized, but otherwise looks good

src/internal/operations/backup_test.go
ryanfkeepers: 🤦

src/internal/connector/exchange/data_collections.go
ryanfkeepers: Ah, gotcha.  That makes sense.  So, for the purposes of metadata, is there ever a situation where we want the consumer to contend with this state?  Forwardly: is this a consumer concern, or should it be a producer concern?  If it's necessary to only consume a certain subset of the metadata, perhaps kopia should only return that set.

ashmrtn: > If it's necessary to only consume a certain subset of the metadata, perhaps kopia should only return that set.

Since we're using the `Restore` API in KW, implementing this would require pushing extra logic into KW and I'd like to avoid that.

It would be possible for BackupOp to weed out some of this, and maybe that's the best we can do. The code would be rather generic and would make use of the `Reason` field that was recently added to returned snapshots (#1779)

Issue #1727, opened 2022-12-07
GC should fulfill new Collection interface additions in Exchange when populating backups


ryanfkeepers:
## Issue
GC needs to populate the new previous path, folder state, and item deletion Collection interface additions.

## Acceptance Criteria
- Populate `previousPath` and `state` when backing up items in GC
- Pack this data into the collections we send to kopia (fulfill interface)
- Track item deletion (OData tag, probably) and set Deleted() flag on items.

PR #1815, opened 2022-12-14
produce coll state and item deletion

ryanfkeepers:
## Description

Fulfills the data collection interfaces to provide item deleted
flags and collection state enums.

A note on the additional data consumption: I'm making an assumption
that the @removed property will appear in that map.  This is awaiting
testing for verification, which I'll get into as a follow-up step.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1727

## Test Plan

- [x] :question: In a follow-up PR


src/internal/common/slices.go
```diff
@@ -23,13 +21,18 @@ func First(vs ...string) string {
...
 }
 
-// parseBool returns the bool value represented by the string
-// or false on error
-func ParseBool(v string) bool {
-	s, err := strconv.ParseBool(v)
-	if err != nil {
+// Equal returns true if both slices contain the same
+// elements in the same order
+func AreSameSlice[T comparable](s1, s2 []T) bool {
```
ashmrtn, line 26: most/all of this function could be replaced by `reflect.DeepEqual`. According to the [doc](https://pkg.go.dev/reflect#DeepEqual), it will return false for `DeepEqual(nil, []bytes{})` or equivalent. However, that could be worked around with something like
```go
func AreSameSlice[T comparable](s1, s2 []T) bool {
  if len(s1) == len(s2) && len(s1) == 0 {
    return true
  }

  return reflect.DeepEqual(s1, s2)
}
```

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -84,15 +92,32 @@ func NewCollection(
...
+		return data.DeletedState
+	}
+
+	if prev == nil || len(prev.String()) == 0 {
+		return data.NewState
+	}
+
+	if !common.AreSameSlice(curr.Folders(), prev.Folders()) {
+		return data.MovedState
+	}
```
ashmrtn, line 116: these are both paths, why not just compare strings? I think `curr.Folder() != prev.Folder()` or `curr.String() != prev.String()` would suffice using only basic comparisons defined in go. This would remove the dependency on slice comparison code as well

src/internal/connector/exchange/exchange_data_collection_test.go
```diff
@@ -104,3 +106,100 @@ func (suite *ExchangeDataCollectionSuite) TestExchangeCollection_AddJob() {
...
+			c := NewCollection(
+				"u",
+				test.curr, test.prev,
+				0, nil, nil, control.Options{})
+			assert.Equal(t, test.expect, c.State())
+		})
+	}
+}
+
+func (suite *ExchangeDataCollectionSuite) TestStateOf() {
```
ashmrtn, line 160: this test is not particularly useful given that `TestNewCollection_state` does the same thing, just less directly. I would be alright with removing it as it tests an unexported function that is tested by other stuff

src/internal/connector/exchange/mail_folder_cache.go
```diff
@@ -86,6 +87,12 @@ func (mc *mailFolderCache) Populate(
 		MailFolders().
 		Delta()
 
+	// TODO(rkeepers): Awaiting full integration of incremental support, else this
```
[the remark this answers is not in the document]
ashmrtn: oh, also cause I sort of forgot about this just looking at the changes: we use the container cache to resolve paths for Exchange items, so we actually always need to pull the full hierarchy into it (or at least the path from item to root for everything that has changed)

src/internal/connector/exchange/exchange_data_collection.go
ryanfkeepers: Oh... you know, because I'm daft.

ashmrtn: eh, it's easy to get caught up in the moment. I've definitely done it before. I think the common one for me is when I'm using C++ to go through a vector and delete all the object instances and then get null pointer errors cause I forgot to reset the vector 🙃

src/internal/connector/exchange/container_resolver.go
```diff
@@ -86,8 +86,16 @@ func (cr *containerResolver) PathInCache(pathString string) (string, bool) {
...
-	// populated.
+	// deleted folders bypass the other checks
+	addtl := cf.GetAdditionalData()
+	_, cf.Removed = addtl[graph.AddtlDataRemoved]
+
+	if cf.Removed {
+		cr.cache[*cf.GetId()] = &cf
+		return nil
+	}
+
```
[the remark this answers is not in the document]
ryanfkeepers: Makes sense.  I was imagining this as running the delta token from root, whereupon we'd see the @removed additional data.  But since we don't re-use that link, this change is moot.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -105,7 +102,36 @@ func FilterContainersAndFillCollections(
...
+			continue
+		}
+
+		prevPath, err := pathFromPrevString(ps)
+		if err != nil {
+			logger.Ctx(ctx).Error(err)
+			continue
+		}
+
+		edc := NewCollection(
```
ashmrtn, line 125: you could probably use an empty `MetadataCollection` here or something. I know it's kind of jank, but I'm also a bit uncomfortable with having to make a service for a collection that has no items and really won't have `Items()` called on it at all

src/internal/common/slices.go
ashmrtn, line 26: this function is no longer used and can be deleted

src/internal/connector/graph/service.go
```diff
@@ -81,9 +72,17 @@ type Displayable interface {
 	GetDisplayName() *string
 }
 
+// Additionalable represents objects that implement msgraph-sdk-go/models.entityable
+// and have the concept of an additional data map.
+type Additionalable interface {
+	Idable
+	GetAdditionalData() map[string]any
+}
+
```
ashmrtn, line 81: I don't think this is required anymore since we don't pull delta info from the container resolver
```suggestion
```

src/internal/connector/graph/service.go
```diff
@@ -81,9 +72,17 @@ type Displayable interface {
...
 type Container interface {
 	Descendable
 	Displayable
+	Additionalable
```
ashmrtn, line 85:
```suggestion
```

src/internal/connector/exchange/service_functions.go
```diff
@@ -132,6 +132,7 @@ func DeleteContactFolder(ctx context.Context, gs graph.Servicer, user, folderID
 func PopulateExchangeContainerResolver(
 	ctx context.Context,
 	qp graph.QueryParams,
+	dps DeltaPaths,
```
ashmrtn, line 135:
```suggestion
```

src/internal/connector/exchange/service_functions.go
```diff
@@ -148,6 +149,7 @@ func PopulateExchangeContainerResolver(
 		res = &mailFolderCache{
 			userID: qp.ResourceOwner,
 			gs:     service,
+			dps:    dps,
```
ashmrtn, line 152:
```suggestion
```

src/internal/connector/exchange/mail_folder_cache.go
```diff
@@ -21,6 +21,7 @@ type mailFolderCache struct {
 	*containerResolver
 	gs     graph.Servicer
 	userID string
+	dps    DeltaPaths
```
ashmrtn, line 24:
```suggestion
```

src/internal/connector/exchange/container_resolver_test.go
```diff
@@ -256,6 +261,10 @@ func (m *mockCachedContainer) SetPath(newPath *path.Builder) {
 	m.p = newPath
 }
 
+func (m *mockCachedContainer) GetAdditionalData() map[string]any {
+	return nil
+}
+
```
ashmrtn, line 267: We won't be using delta info to find deleted folders in the container resolver
```suggestion
```

src/internal/connector/exchange/container_resolver_test.go
```diff
@@ -232,6 +236,7 @@ type mockCachedContainer struct {
 	displayName  string
 	p            *path.Builder
 	expectedPath string
+	removed      bool
```
ashmrtn, line 239:
```suggestion
```

src/internal/connector/exchange/container_resolver_test.go
```diff
@@ -34,6 +34,10 @@ func (m mockContainer) GetParentFolderId() *string {
 	return m.parentID
 }
 
+func (m mockContainer) GetAdditionalData() map[string]any {
+	return nil
+}
+
```
ashmrtn, line 40:
```suggestion
```

src/internal/connector/data_collections.go
```diff
@@ -149,7 +149,7 @@ func (gc *GraphConnector) createExchangeCollections(
 		defer closer()
 		defer close(foldersComplete)
 
-		resolver, err := exchange.PopulateExchangeContainerResolver(ctx, qp)
+		resolver, err := exchange.PopulateExchangeContainerResolver(ctx, qp, dps)
```
ashmrtn, line 152:
```suggestion
		resolver, err := exchange.PopulateExchangeContainerResolver(ctx, qp)
```

ashmrtn (approved):
Mostly just minor cleanup stuff from rebase hell

src/internal/connector/exchange/container_resolver_test.go
ryanfkeepers: it's required, unfortunately.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Not sure I follow your meaning of `a service` in this case.

ashmrtn: ah, sorry for being unclear. The ExchangeCollection requires a `graph.Servicer` instance because it assumes it will need to fetch items. However, Collections that represent deleted folders really just need to have the state set and implement the proper interface. This means it's kind of useless to create `service` on line 145 and use it in this call

PR #1817, opened 2022-12-15
Clean up JS

ntolia:
## Description

Removes unused files and renames a file to better reflect its contents

- [x] :no_entry: No 

## Type of change


website/src/components/parts/FirstSection.js
```diff
@@ -1,74 +0,0 @@
-import React from "react";
```
gmatev, line 1: @ntolia did you actually mean to delete these?

gmatev: Not seeing the corresponding properly named component files. Maybe a different PR?

gmatev: Nenermind. They are already there.

gmatev (approved):

PR #1828, opened 2022-12-15
Select subtrees of data from base snapshots based on reason snapshot was picked

ashmrtn:
## Description

Pick what data to source from a base snapshot by examining the reason the snapshot was selected as a base. This helps avoid two issues:
* pulling in unwanted data when the base has a superset of what is being backed up. Example: pulling in contacts from the base when only email is being backed up
* clobbering already selected data from a different base with data in the base currently being examined. Example: two snapshots, one with contacts and emails and the other with just emails. Second snapshot is newer than the first. The email items in the first snapshot should not clobber those in the first when building the hierarchy

This PR also has the effect of, under some conditions, reducing the amount of data that is pulled from the remote kopia repo when building the hierarchy. This occurs because only the subtrees that will be used in the new backup are traversed instead of traversing the entire snapshot

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1740 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -650,6 +650,29 @@ func traverseBaseDir(
 	return nil
 }
 
+// TODO(ashmrtn): We may want to move this to BackupOp and pass in
+// (Manifest, path) to kopia.BackupCollections() instead of passing in
+// ManifestEntry. That would keep kopia from having to know anything about how
+// paths are formed. It would just encode/decode them and do basic manipulations
+// like pushing/popping elements on a path based on location in the hierarchy.
+func encodedElementsForPath(tenant string, r Reason) (*path.Builder, error) {
+	// This is hacky, but we want the path package to format the path the right
+	// way (e.x. proper order for service, category, etc), but we don't care about
+	// the folders after the prefix.
+	p, err := path.Builder{}.Append("tmp").ToDataLayerPath(
+		tenant,
+		r.ResourceOwner,
+		r.Service,
+		r.Category,
+		false,
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "building path")
+	}
+
+	return p.ToBuilder().Dir(), nil
+}
+
 func inflateBaseTree(
 	ctx context.Context,
 	loader snapshotLoader,
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -696,22 +696,48 @@ func inflateBaseTree(
 		return errors.Errorf("snapshot %s root is not a directory", snap.ID)
 	}
 
-	// TODO(ashmrtn): We should actually only traverse a subtree of the snapshot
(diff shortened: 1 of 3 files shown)


src/internal/kopia/upload.go
```diff
@@ -650,10 +649,33 @@ func traverseBaseDir(
 	return nil
 }
 
+// TODO(ashmrtn): We may want to move this to BackupOp and pass in
```
ryanfkeepers, line 652: Alternatively, I think this should be in the path package, sans the Reason struct.  Something like his:

```go
func NewBuilder(tenant, owner string, s ServiceType, c CategoryType, append string...) (*path.Builder, error) {...}
```

We do a fair bit of `path.Builder{}.Append("...").ToSomePath()` throughout Corso.  This feels like a natural QoL.  Run the "tmp">.Dir() version as shown if "append" is nil, and we've got both cases covered.

Otherwise, I agree about having operations pass in the premade paths.

src/internal/kopia/upload.go
```diff
@@ -674,22 +696,48 @@ func inflateBaseTree(
...
-		0,
-		updatedPaths,
-		&path.Builder{},
-		&path.Builder{},
-		dir,
-		roots,
-	); err != nil {
-		return errors.Wrapf(err, "traversing base snapshot %s", snap.ID)
+		// Root directory is not included in the lookup path we give kopia as we're
+		// starting from the root.
```
ryanfkeepers, line 714:
```suggestion
		// We're starting from the root already, it's already included.
```

src/internal/kopia/upload_test.go
```diff
@@ -31,30 +32,28 @@ func makePath(t *testing.T, elements []string) path.Path {
 	return p
 }
 
-// baseWithChildren returns an fs.Entry hierarchy where the first four levels
-// are the encoded values of tenant, service, user, and category respectively.
-// All items in children are made a direct descendent of the category entry.
+// baseWithChildren returns an fs.Entry hierarchy where the first len(basic)
+// levels are the encoded values of basic in order. All items in children are
+// made a direct descendent of the final entry in basic.
```
ryanfkeepers, line 37:
```suggestion
// used as the direct descendant of the final entry in basic.
```

ryanfkeepers (approved):

src/internal/kopia/upload.go
ashmrtn: I'll make a follow up PR to move making the path and passing it to BackupCollections as it adds more code and there's already a lot going on in this PR. I did make #1832 to track it though. I will probably hold off a bit on revisiting the path package for QoL stuff right now

Issue #1823, opened 2022-12-15
Backup fails with `model has no StableID`


meain:
Every backup after the very fist backup on a new bucket seems to fail with this error:

```
Error: 1 error occurred:
        * Failed to run OneDrive backup for user [J**L@8**0.onmicrosoft.com]: 2 errors occurred:
        * errors prevented the operation from processing
        * connecting to M365: model has no StableID
```

I have not been able to go in depth into the issue, but just from a git bisect, looks like it got added in https://github.com/alcionai/corso/commit/4d46847f6c2d127638c814f6fd28c4fbd0e7e4e6 . You can repro it by deleting all the backups in a bucket and then trying to back it up twice (`./corso backup create onedrive --user $CORSO_M365_TEST_USER_ID`). It will fail on the second backup.

vkamra:
@ryanfkeepers - can you triage?

ryanfkeepers:
I have half of the bug figured out: an incorrect tag key skipped pulling the backup ID from the tags.

There's another half I'm going to need additional eyes on for help.  The model store says the backup doesn't exist.  I can, however, look up the same backup by ID with a `list onedrive --id id` command.  Currently investigating the difference.

ryanfkeepers:
^ Other half fixed up.

meain:
This should now be fixed from #1830

meain:
Seems to now fail with the following error:

```
Error: 1 error occurred:
        * Failed to run OneDrive backup for user [JohannaL@8qzvrj.onmicrosoft.com]: 2 errors occurred:
        * errors prevented the operation from processing
        * connecting to M365: getting backup: getting ModelStoreID: not found
```

ryanfkeepers:
Hmm, that should have been the "second half" of the bug that got fixed.  I"ll look into it.

ryanfkeepers:
@meain I'm unable to repro that error.  Can you provide the list of steps you used?

meain:
I was hitting this error on the very first backup with `144d04756`, but somehow I am not able to repro it now.

PR #1830, opened 2022-12-15
correct backup tag key, metadata not found

ryanfkeepers:
## Description

Fixes two bugs: 1. building the correc tag key
when retrieving the prior backup id from a
manifest.  2. returning ErrNotFound when a
snapshot entry isn't found for restore.  This
can occur for metadata if the service has
a prior backup that does not contain the
expected metadata files.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #1823

## Test Plan

- [x] :muscle: Manual


src/internal/kopia/wrapper.go
```diff
@@ -277,6 +278,10 @@ func getItemStream(
 		encodeElements(itemPath.PopFront().Elements()...),
 	)
 	if err != nil {
+		if strings.Contains(err.Error(), "entry not found") {
+			err = errors.Wrap(ErrNotFound, err.Error())
+		}
```
ashmrtn, line 283: should we be using the error type in `internal/common/errors.go` here so we preserve any stack trace information the error may have?

ashmrtn (approved):

src/internal/kopia/wrapper.go
ryanfkeepers: I don't think that's necessary here.  And we can always backtrack if needed.

PR #1833, opened 2022-12-15
Have BackupOp pass subtree paths to BackupCollections

### LINKED ISSUE #1832: pass relevant subtrees to use in base from BackupOp to kopia.Wrapper
The path to the root of the subtree for each base snapshot is currently made in `kopia/upload.go`. However, this pulls in more context than we'd like about path formatting into the kopia package. It would be cleaner if the BackupOp would create the paths in question and pass them to kopia.Wrapper during backup instead of having kopia make them

ashmrtn:
## Description

Have BackupOp produce the paths for relevant subtrees in each snapshot and pass those to BackupCollections. This removes the need for code in the kopia package to call into more service/category-specific path package code, thus keeping the kopia package more generic.

As in #1828, prefix info for each subtree path is pulled from the Reason a snapshot was selected.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1832 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/operations/backup.go
```diff
@@ -93,9 +93,15 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	ctx, end := D.Span(ctx, "operations:backup:run")
 	defer end()
 
+	m365, err := op.account.M365Config()
```
ryanfkeepers, line 96: You can retrieve the ID without the m365 configuration with `op.account.ID()`.  We should probably rename that func to `TenantID()`.  But, hey, at least it's there.

ryanfkeepers (approved):

PR #1836, opened 2022-12-16
Select specific metadata from base snapshot by Reasons base snapshot was picked

### LINKED ISSUE #1829: only select relevant metadata from each base snapshot
Incremental backups can have multiple base snapshots that they source data from. However, across all of the base snapshots every Reason a snapshot was selected should be unique. In other words, if all of the Reasons for all of the snapshots are listed, there should be no duplicate Reasons. Every reason is a tuple of (resource owner, service, category).

In practice, this means that we should source metadata corresponding only to the Reason a snapshot was selected from every snapshot. This ensures we don't clobber data sourced from other base snapshots and that we have the most recent version of that data as the starting point for the new backup we're working on

ashmrtn:
## Description

Use the Reasons a snapshot was selected to retrieve only the metadata
corresponding to those reasons. This will avoid having multiple versions
of metadata for the same (resource owner, service, category) tuple as
well as pulling in more metadata than required for some backups.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1829 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -219,7 +219,7 @@ func produceManifestsAndMetadata(
 			return nil, nil, err
 		}
 
-		colls, err := collectMetadata(ctx, kw, graph.AllMetadataFileNames(), oc, tid, bup.SnapshotID)
+		colls, err := collectMetadata(ctx, kw, graph.AllMetadataFileNames(), tid, man)
 		if err != nil && !errors.Is(err, kopia.ErrNotFound) {
 			// prior metadata isn't guaranteed to exist.
 			// if it doesn't, we'll just have to do a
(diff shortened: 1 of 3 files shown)


ashmrtn:
will add some unit tests for at least getting the metadata collections soon

src/internal/operations/backup.go
```diff
@@ -233,36 +236,43 @@ func produceManifestsAndMetadata(
...
+		snapshotID string,
+		paths []path.Path,
+		bc kopia.ByteCounter,
+	) ([]data.Collection, error)
+}
+
 func collectMetadata(
 	ctx context.Context,
-	kw *kopia.Wrapper,
+	kw restorer,
```
ryanfkeepers, line 250: nit
```suggestion
	r restorer,
```

ryanfkeepers, line 250: nit
```suggestion
	r restorer,
```

ryanfkeepers (approved):

PR #1842, opened 2022-12-16
Speedup OneDrive backup by pulling urls in delta API

meain:
## Description

This brings in another huge improvement in the backup speed dropping from ~15m to ~1.5m for an account with ~5000 files. This one also drastically reduces the number of requests we have to make for the same account from 5505 to just ~55. This also means that we don't get throttled anymore and we can easily run multiple backup jobs in parallel before we hit the 1024 limit.

~The code works as of now, but I have to double check the numbers as well as fix an issue with us opening too many files and causing program to crash with 'too many open files' when we bump up the numbers. With the current numbers, it works, but I wanna double check and optimize them. Plus some code cleanup.~

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* #<issue>

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/drive.go
+++ b/src/internal/connector/onedrive/drive.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/internal/connector/support"
+	"github.com/alcionai/corso/src/pkg/logger"
 	msgraphgocore "github.com/microsoftgraph/msgraph-sdk-go-core"
 	msdrive "github.com/microsoftgraph/msgraph-sdk-go/drive"
 	msdrives "github.com/microsoftgraph/msgraph-sdk-go/drives"
@@ -12,10 +15,6 @@ import (
 	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 	"github.com/microsoftgraph/msgraph-sdk-go/sites"
 	"github.com/pkg/errors"
-
-	"github.com/alcionai/corso/src/internal/connector/graph"
-	"github.com/alcionai/corso/src/internal/connector/support"
-	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


src/internal/connector/onedrive/item.go
```diff
@@ -66,35 +69,17 @@ func oneDriveItemReader(
...
-	// These URLs are pre-authenticated and can be used to download the data using the standard
-	// http client
-	if _, found := item.GetAdditionalData()[downloadURLKey]; !found {
-		return nil, nil, errors.Errorf("file does not have a download URL. ID: %s, %#v",
-			itemID, item.GetAdditionalData())
-	}
-
-	downloadURL := item.GetAdditionalData()[downloadURLKey].(*string)
-
+	url string,
```
ryanfkeepers, line 72: Pulling @ashmrtn into this PR as well, for additional perspective on restartability.  I imagine this will handle it fine as is, since were just grabbing a super-set of the id alone.  But wanted to double-check.

src/internal/connector/onedrive/drive.go
```diff
@@ -143,9 +143,34 @@ func collectItems(
 	// https://docs.microsoft.com/en-us/graph/api/driveitem-delta?
 	// view=graph-rest-1.0&tabs=http#example-4-retrieving-delta-results-using-a-timestamp
 	builder := service.Client().DrivesById(driveID).Root().Delta()
+	pageCount := int32(999)    // max we can do is 999
+	requestFields := []string{ // TODO(meain): shrink this list to only needed ones
```
ryanfkeepers, line 147: Per TODO: I'd look at what we're persisting in the onedrive and sharepoint item info, and stick with that.

ryanfkeepers (commented):
Generally seems okay. I'm a little curious about memory concerns (it'd have to be a very, very large drive).  But nothing stands out as an immediate concern.

src/internal/connector/onedrive/item.go
ashmrtn: yea, it should be fine when we start adding restartability later on. The extra data will just need to be serialized in the ID file along with the rest. We'll need to make sure we make a service-dependent way to serialize/deserialize all that

ryanfkeepers (approved):

PR #1849, opened 2022-12-16
Do not return an error if folder was deleted

### LINKED ISSUE #1846: fix test failure output for some exchange tests
Some exchange tests use the top-level test instance for assertions run during subtests. This has the effect of stopping the subtest but the error output in the logs is confusing because the subtest just shows it was stopped due to a failure in an ancestor test.

Fix this by running assertions on the leaf test instance instead of an ancestor

ashmrtn:
## Description

When backing up Exchange, don't return an error if the folder/calendar we're trying to fetch item IDs for has been deleted.

Error codes pulled with graph explorer API

Manually tested with an Exchange mail backup

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1846 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -25,6 +25,8 @@ const (
 	errItemNotFound        = "ErrorItemNotFound"
 )
 
+var errContainerDeleted = errors.New("container deleted")
+
 func hasErrorCode(err error, code string) bool {
 	var oDataError *odataerrors.ODataError
 	if !errors.As(err, &oDataError) {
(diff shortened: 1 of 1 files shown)


dadams39 (approved):
gtg

src/internal/connector/exchange/service_iterators.go
```diff
@@ -226,6 +243,14 @@ func FetchEventIDsFromCalendar(
 	for {
 		resp, err := builder.Get(ctx, options)
 		if err != nil {
+			if hasErrorCode(err, errItemNotFound) {
+				// The folder was deleted between the time we populated the container
+				// cache and when we tried to fetch data for it. All we can do is
+				// return no jobs because we've only pulled basic info about each
+				// item.
+				return nil, "", nil
```
ryanfkeepers, line 251: This is kinda tough, but `nil, nil` returns tend to be frowned upon.  Could we return and handle/discard `errItemNotFound` or some similar in the caller?

ryanfkeepers (approved):

Issue #1858, opened 2022-12-19
Graph req/resp non-2xx middleware logging/metrics


ryanfkeepers:
## Issue
Within the graph api client, we need to add request/response logging to track the occurrence and details around non-happy communication with msoft apis.  In addition to logging, we need to track metrics on response patterns overall.

## Acceptance Criteria
* Within the graph http client middleware, log the response body and status on non-2xx status responses.
* Include relevant request metadata in the logs.
* Track overall metrics on response statuses, including metadata like call duration.

PR #1860, opened 2022-12-19
log non-2xx graph api calls

ryanfkeepers:
## Description

Adds logging for non-2xx http communication
within the graph API client.  Logging req/resp
metadata is turned off by default, and requires
either a cli flag or os env var to enable.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #1858

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/service_helper.go
+++ b/src/internal/connector/graph/service_helper.go
@@ -67,10 +67,9 @@ func (handler *LoggingMiddleware) Intercept(
 	req *nethttp.Request,
 ) (*nethttp.Response, error) {
 	var (
-		ctx            = req.Context()
-		requestDump, _ = httputil.DumpRequest(req, true)
-		resp, err      = pipeline.Next(req, middlewareIndex)
-		metadata       = []any{}
+		ctx       = req.Context()
+		resp, err = pipeline.Next(req, middlewareIndex)
+		metadata  = []any{}
 	)
 
 	if resp != nil && (resp.StatusCode/100) != 2 {
@@ -82,10 +81,6 @@ func (handler *LoggingMiddleware) Intercept(
 			"status", resp.Status,
 			"statusCode", resp.StatusCode,
 			"request", string(respDump),
-		}
-
-		if resp.StatusCode == 400 {
-			metadata = append(metadata, "requestBody", string(requestDump))
 		}
 	}
 
--- a/src/internal/connector/graph/service_helper.go
+++ b/src/internal/connector/graph/service_helper.go
@@ -1,7 +1,6 @@
 package graph
 
 import (
-	"context"
 	nethttp "net/http"
 	"net/http/httputil"
 	"os"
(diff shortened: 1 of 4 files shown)


src/internal/connector/graph/service_helper.go
```diff
@@ -51,12 +50,7 @@ func CreateAdapter(tenant, client, secret string) (*msgraphsdk.GraphRequestAdapt
 func CreateHTTPClient() *nethttp.Client {
 	clientOptions := msgraphsdk.GetDefaultClientOptions()
 	middlewares := msgraphgocore.GetDefaultMiddlewaresWithOptions(&clientOptions)
-
-	// When true, additional logging middleware support added for http request
-	if os.Getenv(logGraphRequestsEnvKey) != "" {
-		middlewares = append(middlewares, &LoggingMiddleware{})
-	}
-
+	middlewares = append(middlewares, &LoggingMiddleware{})
```
ashmrtn, line 53: I like the idea behind this, but before we enable it everywhere should we check if it has any large impacts on memory usage, runtime, etc?

ryanfkeepers: This is a standard pattern, so I'm not particularly concerned about performance impacts.  Though I recognize that I kept the request body dump around from prior debugging.  We can ditch that for now and re-add if necessary in the future.

ashmrtn (approved):

src/internal/connector/graph/service_helper.go
```diff
@@ -68,12 +62,39 @@ type LoggingMiddleware struct{}
...
+	var (
+		ctx       = req.Context()
+		resp, err = pipeline.Next(req, middlewareIndex)
+	)
+
+	if resp == nil {
+		return resp, err
+	}
+
+	if (resp.StatusCode / 100) == 2 {
```
meain, line 78: I feel like just having non 2xx along might be misleading/hard when we try to walk trough and figure out what might be going on. Should we provide an additional flag to also log 2xx requests. Something like `-v` `-vv` pattern.

ryanfkeepers: I don't think that's necessary at the moment.  In past experience its been largely unhelpful to retain happy calls.  On one hand, those successes are often captured by other info/debug logs, and on the other it's rare that a successful api call provides detail useful for investigation.

What might appeal to you, however, is that we will be adding some telemetry here to track overall call health (success rates, time, etc).  This is likely to give us much of the same insight.

ryanfkeepers:
@meain jumping from the other PR:

On the note of debug logging, I've transitioned the graph api non-2xx dumps from debug to error level logging.  That'll help reduce the need for debug logging (will still need people to flag graph calls, though).

As for identifiable data, I think only the url (ie username and item/folder ids) could expose anything that we're tracking.  We can sanitize that as needed in the future.

meain:
@ryanfkeepers Transitioning non 2xx from debug to error game me another idea of how we can possibly handle this if we are going that route. We can rename the flag from `--debug-api-calls` to `--log-api-calls`. Plus have 2xx calls to be logged as debug entries. This way, by default we will just have failed ones, but if the user enables `--log-level debug`, it gives us the complete logs.

meain:
@ryanfkeepers Having the logs printed via `Errorw` ends up printing the stacktrace which feels like unnecessary information.

PR #1867, opened 2022-12-19
Fetch isRead attribute in email delta queries

### LINKED ISSUE #1866: fetch changes to emails even if the change is just [un]marking read flag
Updating the status of the `isRead` flag in M365 by marking an email as [un]read does not cause the item to be returned by the delta query unless the `isRead` field is specifically requested. This is likely due to the fact that changes to the `isRead` flag in M365 do not update the mod time of the item in M365

ashmrtn:
## Description

Updating the isRead flag in M365 does not cause the mod time of the item to be updated. This means delta queries will not return items marked as [un]read when isRead is not one of the attributes requested in the $select clause.

Explicitly fetch isRead so that we get those items. Further changes will be required to ensure the mod time of the item is something that will not cause kopia-assisted dedupe to skip the item when only the isRead attribute has been changed.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1866 

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/service_iterators.go
```diff
@@ -421,7 +421,7 @@ func FetchMessageIDsFromDirectory(
 		deltaURL string
 	)
 
-	options, err := optionsForFolderMessagesDelta([]string{"id"})
+	options, err := optionsForFolderMessagesDelta([]string{"id", "isRead"})
```
dadams39, line 424: Just a reminder, you do not need to put `id` in the options as it is included in the buildOptions automatically.

```go
func buildOptions(options []string, optID optionIdentifier) ([]string, error) {
	var (
		allowedOptions  map[string]int
		returnedOptions = []string{"id"}
```

dadams39 (approved):
Only a nit

Issue #1780, opened 2022-12-12
logic to fallback to a full backup if a delta token has expired


ashmrtn:
Corso should detect when a delta token has expired and fallback to pulling all items for the endpoint if the token is no longer useful

PR #1882, opened 2022-12-20
handle invalid or expired deltas

ryanfkeepers:
## Description

If a delta token expires or is otherwise invalid,
the backup should fall back to the same
behavior as if the collection were new.  It must
collect the full delta of information, and create
a collection with New state.  If the previous
container moved between the two deltas,
it should be marked for deletion with a tomb-
stone collection.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1780

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -7,7 +7,6 @@ import (
 
 	multierror "github.com/hashicorp/go-multierror"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
-	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 	msuser "github.com/microsoftgraph/msgraph-sdk-go/users"
 	"github.com/pkg/errors"
 
@@ -20,22 +19,12 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
-const (
-	errEmailFolderNotFound = "ErrorSyncFolderNotFound"
-	errItemNotFound        = "ErrorItemNotFound"
-)
-
-var errContainerDeleted = errors.New("container deleted")
-
-func hasErrorCode(err error, code string) bool {
-	var oDataError *odataerrors.ODataError
-	if !errors.As(err, &oDataError) {
-		return false
-	}
-
-	return oDataError.GetError() != nil &&
-		oDataError.GetError().GetCode() != nil &&
-		*oDataError.GetError().GetCode() == code
+// carries details about delta retrieval in aggregators
+type deltaUpdate struct {
+	// the deltaLink itself
+	url string
+	// true if the old delta was marked as invalid
+	reset bool
 }
 
 // filterContainersAndFillCollections is a utility function
@@ -61,7 +50,7 @@ func filterContainersAndFillCollections(
 		currPaths = map[string]string{}
 		// copy of previousPaths.  any folder found in the resolver get
 		// deleted from this map, leaving only the deleted maps behind
-		deletedPaths = map[string]DeltaPath{}
+		tombstones = makeTombstones(dps)
 	)
 
 	getJobs, err := getFetchIDFunc(qp.Category)
@@ -84,11 +73,13 @@ func filterContainersAndFillCollections(
 
 		cID := *c.GetId()
 
-		// this folder exists (probably), do not delete it.
-		delete(deletedPaths, cID)
-
+		currPath, ok := includeContainer(qp, c, scope)
+		if currPath != nil {
+			// this path exists, do not delete it, even if it isn't
+			// included in this backup.
+			delete(tombstones, currPath.String())
+		}
 		// Only create a collection if the path matches the scope.
-		currPath, ok := pathAndMatch(qp, c, scope)
 		if !ok {
 			continue
 		}
@@ -106,19 +97,28 @@ func filterContainersAndFillCollections(
 			}
 		}
 
-		jobs, currDelta, err := getJobs(ctx, service, qp.ResourceOwner, cID, prevDelta)
+		jobs, newDelta, err := getJobs(ctx, service, qp.ResourceOwner, cID, prevDelta)
 		if err != nil {
-			// race conditions happen, the container might get
-			// deleted while this process in flight.
-			if errors.Is(err, errContainerDeleted) {
-				currPath = nil
-			} else {
+			// race conditions happen, containers might get deleted while
+			// this process is in flight.  If it was deleted, we skip it
+			// and move on, which will create a tombstone to delete the path.
+			if graph.IsErrDeletedInFlight(err) == nil {
 				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
 			}
+
+			continue
 		}
 
-		if len(currDelta) > 0 {
-			deltaURLs[cID] = currDelta
+		if len(newDelta.url) > 0 {
+			deltaURLs[cID] = newDelta.url
+		}
+
+		if newDelta.reset {
+			// the previous delta was invalid or otherwise unusable.
+			// We need to mark the collection as New, and tombstone
+			// the path to clear out any data for a clean refresh.
+			prevPath = nil
+			tombstones[currPath.String()] = struct{}{}
 		}
 
 		edc := NewCollection(
@@ -130,12 +130,8 @@ func filterContainersAndFillCollections(
 			statusUpdater,
 			ctrlOpts,
 		)
+
 		collections[cID] = &edc
-
-		if edc.State() == data.DeletedState {
-			continue
-		}
-
 		edc.jobs = append(edc.jobs, jobs...)
 
 		// add the current path for the container ID to be used in the next backup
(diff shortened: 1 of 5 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -276,65 +292,92 @@ func getFetchIDFunc(category path.CategoryType) (FetchIDFunc, error) {
 	}
 }
 
+// ---------------------------------------------------------------------------
```
ashmrtn, line 295: while these changes look like they would be quite useful for compacting code (I only skimmed them so no detailed comments. I would recommend trying to get a single interface to use instead of generics if possible), they need to be in a separate PR. Rewriting this logic is not related to reusing delta tokens or populating the proper fields on collections

src/internal/connector/graph/errors.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pkg/errors"
+)
+
+// ---------------------------------------------------------------------------
+// Error Interpretation Helpers
+// ---------------------------------------------------------------------------
+
+// ErrIs checks for graph specific errors.  Returns nil if the errors do not
+// match.  Returns errors.Wrap(graphErr, err) if the errors match.
+func ErrIs(err, graphErr error) error {
```
ashmrtn, line 17: this should be after the error definitions please

src/internal/connector/graph/errors.go
```diff
@@ -0,0 +1,92 @@
...
+	switch true {
+	// http timeouts, not retried by the graph client
+	case errors.Is(graphErr, ErrTimeout):
+		if isTimeoutErr(err) {
+			return errors.Wrap(ErrTimeout, err.Error())
```
ashmrtn, line 22: I'm not really a fan of using wrapping like this. I would suggest using the custom error package we have. This form of wrapping will lose any stack track information or other info that the error may have

ashmrtn (changes requested):

src/internal/connector/graph/errors.go
ryanfkeepers: ah, totally forgot about that package.  Truth be told I'm not a fan of the using As for sentinel errors.  Kinda wish we had a `errors.Wrap(err, sentinalErr)`.  But it'll suffice.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: For clarity: are you asking for this to split into two separate PRs, or setting expectations for the future?  I'm good with either.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -131,12 +134,14 @@ func FilterContainersAndFillCollections(
 			statusUpdater,
 			ctrlOpts,
 		)
-		collections[cID] = &edc
 
+		// in case of race conditions (something was deleted while this process
+		// was in flight), we don't want to do any work on this container.
 		if edc.State() == data.DeletedState {
```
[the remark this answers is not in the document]
ryanfkeepers: Sounds good.

src/internal/connector/exchange/service_iterators.go
ashmrtn: split this PR please

src/internal/connector/exchange/service_iterators.go
```diff
@@ -107,19 +97,32 @@ func FilterContainersAndFillCollections(
...
+		// we need to mark the collection as New.  If the path has
+		// moved since the last delta, the old path will get marked
+		// as deleted.
+		if newDelta.reset {
+			prevPath = nil
+		} else {
+			// this path exists (probably), and the delta link was
+			// not reset; do not delete it.
+			delete(deletedPaths, currPath.String())
 		}
```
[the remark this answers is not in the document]
ryanfkeepers: currPath is guaranteed non-nil at this point (following pathAndMatch), so we're safe from that.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -85,12 +73,14 @@ func FilterContainersAndFillCollections(
...
 
-		// this folder exists (probably), do not delete it.
-		delete(deletedPaths, cID)
-
 		// Only create a collection if the path matches the scope.
 		currPath, ok := pathAndMatch(qp, c, scope)
 		if !ok {
+			if currPath != nil {
+				// this path exists, do not delete it.
+				delete(deletedPaths, currPath.String())
```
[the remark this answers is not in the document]
ryanfkeepers: Sounds to me like we want to keep this block. That most accurately preserves the state of the world (these things exist, and were not modified).  You've only listed the downside to the latter, which makes it seem pretty one-sided.

I think an extension of this question is whether we should treat zero jobs (ie: a container with no changes since the last delta) with the same behavior?  Currently we both skip collections (if the scope doesn't match), and create zero-valued collections (if the scope matches, but doesn't see changes).  It's good to have similar outcomes for both, but we're probably better off to do one or the other.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: done.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -334,153 +345,203 @@ func FetchEventIDsFromCalendar(
...
+		return nil, deltaUpdate{}, errors.Wrap(err, "getting query options")
 	}
 
-	builder := gs.Client().
-		UsersById(user).
-		ContactFoldersById(directoryID).
-		Contacts().
-		Delta()
+	getIDs := func(builder *msuser.ItemContactFoldersItemContactsDeltaRequestBuilder) error {
+		for {
```
ryanfkeepers, line 375: same loop as below, just moved up to an ad-hoc func so that it can be called twice (in case of delta invalidation) without code duplication.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -62,7 +50,7 @@ func FilterContainersAndFillCollections(
 		currPaths = map[string]string{}
 		// copy of previousPaths.  any folder found in the resolver get
 		// deleted from this map, leaving only the deleted maps behind
```
ashmrtn, line 52:
```suggestion
		// deleted from this map, leaving only the deleted folders behind
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -62,7 +50,7 @@ func FilterContainersAndFillCollections(
...
-		deletedPaths = common.CopyMap(dps)
+		tombstones = makeTombstones(dps)
```
ashmrtn, line 53: ☠️

src/internal/connector/exchange/service_iterators.go
```diff
@@ -107,19 +97,28 @@ func FilterContainersAndFillCollections(
...
+		jobs, newDelta, err := getJobs(ctx, service, qp.ResourceOwner, cID, prevDelta)
 		if err != nil {
-			// race conditions happen, the container might get
-			// deleted while this process in flight.
-			if errors.Is(err, errContainerDeleted) {
-				currPath = nil
-			} else {
+			// race conditions happen, containers might get deleted while
+			// this process is in flight.  If it was deleted, we skip it
+			// and move on, which will create a tombstone to delete the path.
```
ashmrtn, line 104: >which will create a tombstone to delete the path.

We've actually already [deleted the path from the tombstones map](https://github.com/alcionai/corso/pull/1882/files#diff-67d5d9c3856fb884fe81259eefeb5d7ddb0d695582bf44db4e9ee6e447bc808dR80) so we will not mark this as deleted

src/internal/connector/exchange/service_iterators.go
```diff
@@ -131,31 +130,30 @@ func FilterContainersAndFillCollections(
...
-	// left in deletedPaths are only the previous paths that did not appear as
-	// children of the root.
-	for fID, dp := range deletedPaths {
+	// A tombstone is a collection path that needs to be marked for deletion.
+	// Tombstones can occur for a number of reasons: the delta token expired,
+	// the container was deleted in flight, or the user deleted the container
+	// between backup runs.  If events combine to both delete and write content
+	// to the same container (ex: container_1 gets deleted, then container_2
+	// gets created with the same name), it is assumed that the backup consumer
+	// processes deletions before creations, making the combined operation safe.
```
ashmrtn, line 148: >it is assumed that the backup consumer processes deletions before creations, making the combined operation safe.

It's actually a little more nuanced than that. It's that deletions apply to what is essentially the previous path of the container while creations apply to the new path. We're doing state updates now, so the previous path is really a way to find something in the previous state. While relative order of the processing of things makes it more clear, it's not strictly required for correctness as a deletion simply means "don't merge the previous version of this folder with the new version"

This all ties into the [assumptions](https://github.com/alcionai/corso/issues/1884#issuecomment-1361618183) in #1884

src/internal/connector/exchange/service_iterators.go
ashmrtn: thanks for the callout

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Ack, good catch.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Granted, but I think the handwave is sufficient for explanation at that placement.

ashmrtn (approved):

Issue #1617, opened 2022-11-28
Selectors: promote ResourceOwners to Selector


ryanfkeepers:
## Issue
Corso design has deviated in path from original Selector design, and we're beginning to see idealogical collisions between the two.  For example, in https://github.com/alcionai/corso/issues/1505 we want to transition backups to be split between resource owners (ie: moving from 1:many backup:owner towards 1:1).

Current selectors are designed for granular queries such as, "get all U1's mail, also U1 and U2's contacts".   This places resource owner identification at a per-scope level, rather than a per-selector (ie: per-backup) level, causing each selector to produce a complicated venn-diagram of owner-data pairings.  As Corso design has matured we've moved away from supporting granular and compound queries such as these.  Therefore the current design is not only complicated to reason about, it's also more complication than Corso wants to support.

## Proposal
Resource Owner identification is to be removed from the scope-level declarations.  Instead, each Selector will recognize a single, canonical list of resource owners.  IE:
```go
// this current design
s := selectors.NewExchangeBackup()
s.Include(
  s.Mail(users, folders, mailIDs),
  s.Contacts(users, folders, contactIDs)
)

// will become
s := selectors.NewExchangeBackup(users)
s.Include(
  s.Mail(folders, mailIDs),
  s.Contacts(folders, contactIDs),
)

// The new selector struct would look like this:
type Selector struct {
	Service service
        resourceOwners filters.Filter
	Excludes []scope
	Filters []scope
	Includes []scope
}
```
All scopes in the selector would implicitly apply to each resource owner.  Eg: if the resource owners contains `[Alice, Bob]`, and the scopes include `MailFolder(Inbox)`, then the selector can be interpreted as "select all mail in the inbox for alice and bob".

## Usage Changes
Backup data aggregation currently employs the following pattern:
```
1. generate a selector
2. pass the selector to a Backup Operation
3. pass the selector to GraphConnector
4. for each scope in the selector's Included scopes
5. for each resourceOwner in the scope (or, as fallback, the tenant's owner list)
6. collect all data for backup as a single set
7. run the backup
```
In the proposed change the resource owner iteration would migrate above the operation creation.  This would generate N backup operations, one for each resource owner.
```
1. generate a selector for a given set of resourceOwners (or, as fallback, the tenant's owner list)
2. for each owner in the selector.
3. pass the (owner, selector) pair to a Backup Operation
4. Pass the selector to GraphConnector
5. for each scope in the selector's Included scopes
6. collect all data for backup
[...]

PR #1887, opened 2022-12-20
require resource owner set on selector create

ryanfkeepers:
## Description

selector creation now includes a parameter for
a slice of resource owners (users or sites).  This
is step one in migrating resource owner lists out
of scopes and into the selector.  next step is to
have the selector utilize the primary list instead
of the per-scope list.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1617

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -92,10 +92,7 @@ func (gc *GraphConnector) DataCollections(
 func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) error {
 	var ids []string
 
-	resourceOwners, err := sels.ResourceOwners()
-	if err != nil {
-		return errors.Wrap(err, "invalid backup inputs")
-	}
+	resourceOwners := sels.DiscreteResourceOwners()
```
ashmrtn, line 95: how does this interact with passing "*" as the user? Or are we assuming that by this point "*" has been resolved to actual users?

ashmrtn (approved):

src/internal/connector/data_collections.go
ryanfkeepers: Right now we're still re-creating each selector within the per DiscreteScope loop.  The wildcard will cascade to a concrete value from that as expected.  Later on the wildcard will get resolved as part of selector creation/handling in a different way, but still prior to this point.

PR #1890, opened 2022-12-20
use selector owners, not scope owners

ryanfkeepers:
## Description

Migrates code away from pulling the resource
owner from each scope, and instead usees the
selector as the canon identifier of the resource
owner.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1617

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange.go
+++ b/src/cli/backup/exchange.go
@@ -284,31 +284,29 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	for _, sel := range sel.SplitByResourceOwner(users) {
 		// TODO: pass in entire selector, not individual scopes
-		for _, scope := range sel.Scopes() {
-			bo, err := r.NewBackup(ctx, sel.Selector)
-			if err != nil {
-				errs = multierror.Append(errs, errors.Wrapf(
-					err,
-					"Failed to initialize Exchange backup for user %s",
-					scope.Get(selectors.ExchangeUser),
-				))
+		bo, err := r.NewBackup(ctx, sel.Selector)
+		if err != nil {
+			errs = multierror.Append(errs, errors.Wrapf(
+				err,
+				"Failed to initialize Exchange backup for user %s",
+				sel.DiscreteOwner,
+			))
 
-				continue
-			}
-
-			err = bo.Run(ctx)
-			if err != nil {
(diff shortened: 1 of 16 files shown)


ryanfkeepers:
@ashmrtn this is now ready for review

src/internal/connector/exchange/data_collections.go
```diff
@@ -174,7 +173,8 @@ func DataCollections(
 	}
 
 	var (
-		scopes      = eb.DiscreteScopes(userPNs)
+		user        = selector.DiscreteOwner
+		scopes      = eb.DiscreteScopes([]string{user})
```
ashmrtn, line 177: I take it this will eventually be going away since we're already down to a single resource owner at this point?

ashmrtn (approved):
can you also make a follow-up PR at some point that removes `data.ResouceOwnerSet()` and just uses either a hard-coded value or some other way to determine the number of resource owners for the operation

src/internal/connector/exchange/data_collections.go
ryanfkeepers: Yep!  One thing at a time.

ryanfkeepers:
> can you also make a follow-up PR at some point that removes data.ResouceOwnerSet() and just uses either a hard-coded value or some other way to determine the number of resource owners for the operation
I believe that's already been removed in one of the later PRs.

src/internal/connector/exchange/data_collections.go
ashmrtn: sounds good. Thanks for breaking this all into chunks for review

src/cli/backup/exchange.go
```diff
@@ -284,31 +284,29 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	for _, sel := range sel.SplitByResourceOwner(users) {
```
ashmrtn, line 285: this is variable shadowing and can easily lead to errors. It should be changed to have a different variable name

ryanfkeepers: I'm pretty sure this is safe.  But no harm in changing it.  Better readability if nothing else.

PR #1891, opened 2022-12-21
Match resource owners at top of reduce

ryanfkeepers:
## Description

Checks for resource owner matches in the top
of the reduce func using the selector owners,
instead of waiting until the path match check.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1617

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
@ashmrtn this is now ready for review.

CHANGELOG.md
```diff
@@ -5,13 +5,23 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed
+
+- The selectors Reduce() process will only include details that match the DiscreteOwner, if one is specified.
+- New selector constructors will automatically set the DiscreteOwner if given a single-item slice.
 
 ### Fixed
 
 - Fixed issue where repository connect progress bar was clobbering backup/restore operation output.
+- Fixed issue where a `backup create` produced one backup record per data type.
```
ashmrtn, line 18:
```suggestion
- Fixed issue where a `backup create exchange` produced one backup record per data type.
```

src/cli/utils/exchange.go
```diff
@@ -128,30 +128,32 @@ func ValidateExchangeRestoreFlags(backupID string, opts ExchangeOpts) error {
...
 	le, lef := len(opts.Email), len(opts.EmailFolder)
 	lev, lec := len(opts.Event), len(opts.EventCalendar)
 	// either scope the request to a set of users
 	if lc+lcf+le+lef+lev+lec == 0 {
-		if len(opts.Users) == 0 {
-			opts.Users = selectors.Any()
-		}
-
-		sel.Include(sel.Users(opts.Users))
+		sel.Include(sel.Users(users))
```
ashmrtn, line 144: is this still required? We're already setting users on line 137

src/cli/utils/exchange.go
```diff
@@ -128,30 +128,32 @@ func ValidateExchangeRestoreFlags(backupID string, opts ExchangeOpts) error {
...
 
 	opts.EmailFolder = trimFolderSlash(opts.EmailFolder)
 
 	// or add selectors for each type of data
-	AddExchangeInclude(sel, opts.Users, opts.ContactFolder, opts.Contact, sel.Contacts)
-	AddExchangeInclude(sel, opts.Users, opts.EmailFolder, opts.Email, sel.Mails)
-	AddExchangeInclude(sel, opts.Users, opts.EventCalendar, opts.Event, sel.Events)
+	AddExchangeInclude(sel, users, opts.ContactFolder, opts.Contact, sel.Contacts)
+	AddExchangeInclude(sel, users, opts.EmailFolder, opts.Email, sel.Mails)
+	AddExchangeInclude(sel, users, opts.EventCalendar, opts.Event, sel.Events)
```
ashmrtn, line 154: is users still required in these? Is it planned to be removed in a future PR?

src/internal/connector/graph_connector_helper_test.go
```diff
@@ -814,20 +816,23 @@ func makeOneDriveBackupSel(
...
 	toInclude := [][]selectors.OneDriveScope{}
-	resourceOwners := []string{}
+	resourceOwners := map[string]struct{}{}
 
 	for _, d := range dests {
+		resourceOwners[d.resourceOwner] = struct{}{}
+
+		// nil owners here, we'll need to stitch this together
+		// below after the loops are complete.
 		sel := selectors.NewOneDriveBackup(nil)
```
ashmrtn, line 826: if we don't remove the users parameter from scope creation (i.e. `sel.Folders`) then it seems like it would just be better to change what the `toInclude` map holds so we don't have to play so many tricks here. This is test code so it doesn't need to be super efficient in all ways

ashmrtn:
looks like this could use some manual end-to-end testing as well. Doing `backup create exchange --user "*"` results in errors

We should also make a follow-up issue for how selectors are displayed in the output at the end if multiple users are backed up
```text
corso backup create exchange --user user1@domain.com,user2@domain.com

  Started At            ID   Status                Selectors                            
  2023-01-04T21:52:56Z  aaa  Completed (0 errors)  user1@domain.com (1 more)
  2023-01-04T21:53:43Z  bbb  Completed (0 errors)  user1@domain.com (1 more)
```

src/cli/utils/exchange.go
ryanfkeepers: Removal will happen in one of the next PRs.

ryanfkeepers:
>looks like this could use some manual end-to-end testing as well. Doing backup create exchange --user "*" results in errors

> We should also make a follow-up issue for how selectors are displayed in the output at the end if multiple users are backed up

@ashmrtn That bug exists on main.  I'll address it directly.  And the list display can be kicked to a follow-up PR as well.

ashmrtn (approved):

Issue #1800, opened 2022-12-13
merge backup details from base snapshot(s) and items in Collections


ashmrtn:
Incremental backups should still have a complete set of backup details with an entry for each item even if the item was not reported as changed/added/deleted by GraphConnector

Backup details sometimes contains information that is only available in the data stored in kopia files. This means that for unchanged items that are not reported by GraphConnector we need a way to source this information. Using the backup details of the base snapshot(s) is a good starting point as it already contains the info we need

Items that are added/updated and reported by GraphConnector are assumed to have the correct backup details

PR #1892, opened 2022-12-21
Produce map of items that need backup details entries

ashmrtn:
## Description

Generate a `map[old ShortRef]new path.Path` for every item sourced from a base snapshot during backup. Return this information at the end of BackupCollections so that callers can use it to merge backup details if desired.

This PR does a few auxiliary things as well:
* thread current and previous path through hierarchy merging
* add extra field to info tracked by corsoProgress to hold previous path
* provide current and previous paths to streamBaseItems function

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1800 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -435,6 +435,13 @@ func buildKopiaDirs(dirName string, dir *treeMap, progress *corsoProgress) (fs.D
 }
 
 type treeMap struct {
+	// path.Path representing the node's path. This is passed as a parameter to
+	// the stream item function so that even baseDir directories can properly
+	// generate the full path of items.
+	currentPath path.Path
+	// Previous path this directory may have resided at if it is sourced from a
+	// base snapshot.
+	prevPath path.Path
 	// Child directories of this directory.
 	childDirs map[string]*treeMap
 	// Reference to data pulled from the external service. Contains only items in
@@ -532,6 +539,8 @@ func inflateCollectionTree(
 		ownerCats.ResourceOwners[s.FullPath().ResourceOwner()] = struct{}{}
 
 		node.collection = s
+		node.currentPath = s.FullPath()
+		node.prevPath = s.PreviousPath()
 	}
 
 	return roots, updatedPaths, nil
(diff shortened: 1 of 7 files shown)


src/internal/kopia/upload.go
```diff
@@ -334,27 +361,43 @@ func streamBaseEntries(
...
+		// unable to find it in the old backup details because we wouldn't know what
+		// to look for.
+		prevItemPath, err := prevPath.Append(entName, true)
 		if err != nil {
-			name = dir.Name()
+			return errors.Wrap(err, "getting previous full item path for base entry")
+		}
+
+		// All items have item info in the base backup. However, we need to make
+		// sure we have enough metadata to find those entries. Ensure that by adding
```
ryanfkeepers, line 385: Nit: "ensure that..." can be read as a command, rather than a statement of fact.
```suggestion
		// sure we have enough metadata to find those entries. To do that we add
```

ryanfkeepers (approved):

src/internal/kopia/upload.go
```diff
@@ -183,6 +179,30 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 	}
 
 	cp.deets.AddFolders(folders)
+
+	// These items were sourced from a base snapshot or were cached in kopia so we
+	// never had to materialize their details in-memory.
+	if d.info == nil {
+		// TODO(ashmrtn): We should probably be returning an error here?
+		if d.prevPath == nil {
+			return
```
vkamra, line 188: Add an error log here?

vkamra (approved):

src/internal/kopia/upload.go
ashmrtn: we don't have access to a context here so unless I stash a logger in the struct instance there's not a whole lot we can do :/

ryanfkeepers: The logger will fall back to a singleton instance if given the background context.  You might miss out on some logging within tests, but for the CLI it'll still write out as expected.

ashmrtn: thinking about this more broadly, I think we should be storing errors in the struct and returning them at the end. Otherwise we could miss the fact that we failed to add some details. I'll make a follow up ticket fo this

src/internal/kopia/upload.go
```diff
@@ -183,6 +179,30 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 	}
 
 	cp.deets.AddFolders(folders)
```
vkamra, line 181: Following up on the slack discussion - where we've changed this to `AddFoldersForItem` in `main` . I think we should just defer the `AddFoldersForItem` for an item that is being sourced from the base snapshot to when it is being "merged into details"

ashmrtn: yea, that makes the most sense. A bit of a bummer cause I wanted to keep the code for building the set of directories limited to as few places as possible. Guess I'll move it to a helper function somewhere

src/internal/kopia/upload.go
ashmrtn: tracking with #1915

PR #1904, opened 2022-12-21
Functions to merge a set of backup details

ashmrtn:
## Description

Go through the provided bases, load their backup details, and check if
any of the items in them need to be merged into the details for the
current backup.

Has a small amount of logic to treat moved items as updated.

Also include all the tests

Viewing by commit may be useful

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1800 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -420,7 +420,7 @@ func mergeDetails(
 	detailsStore detailsReader,
 	mans []*kopia.ManifestEntry,
 	toMerge map[string]path.Path,
-	deets *details.Details,
+	deets *details.Builder,
 ) error {
 	// Don't bother loading any of the base details if there's nothing we need to
 	// merge.
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -220,16 +220,6 @@ func produceManifestsAndMetadata(
 			continue
 		}
 
-		// TODO(ashmrtn): Uncomment this again when we need to fetch and merge
-		// backup details from previous snapshots.
-		// k, _ := kopia.MakeTagKV(kopia.TagBackupID)
-		// bupID := man.Tags[k]
-
-		// bup, err := sw.GetBackup(ctx, model.StableID(bupID))
-		// if err != nil {
-		// 	return nil, nil, err
-		// }
-
 		colls, err := collectMetadata(ctx, kw, man, metadataFiles, tenantID)
 		if err != nil && !errors.Is(err, kopia.ErrNotFound) {
 			// prior metadata isn't guaranteed to exist.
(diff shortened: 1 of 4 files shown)


src/internal/operations/backup.go
```diff
@@ -407,6 +397,119 @@ func consumeBackupDataCollections(
...
+				)
+			}
+
+			deets.Add(
+				newPath.String(),
+				newPath.ShortRef(),
+				newPath.ToBuilder().Dir().ShortRef(),
+				// TODO(ashmrtn): This may need updated if we start using this merge
+				// strategry for items that were cached in kopia.
+				newPath.String() != rr.String(),
```
ashmrtn, line 492: @vkamra I'm not sure if moved items should count as updated. We can iterate on this

src/internal/operations/backup.go
```diff
@@ -407,6 +397,119 @@ func consumeBackupDataCollections(
...
+
+			// Track how many entries we added so that we know if we got them all when
+			// we're done.
+			addedEntries++
+		}
+	}
+
+	if addedEntries != len(toMerge) {
+		return errors.Errorf(
+			"merging details: wanted %v entries but added %v",
```
ryanfkeepers, line 504: For end user clarity
```suggestion
			"incomplete migration of backup details: found %v of %v expected items",
```

src/internal/operations/backup.go
```diff
@@ -407,6 +397,119 @@ func consumeBackupDataCollections(
...
+
+	return false
+}
+
+func mergeDetails(
+	ctx context.Context,
+	ms *store.Wrapper,
+	detailsStore detailsReader,
+	mans []*kopia.ManifestEntry,
+	toMerge map[string]path.Path,
```
ryanfkeepers, line 417: `toMerge` has, in this follow-up PR, revealed itself to be a bit ambiguous.  It describes the goal of the map, but not its contents.  For a dev who doesn't already grok the design at large, the intent is difficult to tie in to a mental model.  Something that more directly describes the value would make it more useful.

I honestly don't have a good replacement name off the top of my head.  `shortRefsFromPrevBackup`?

ryanfkeepers (approved):

src/internal/operations/backup.go
```diff
@@ -407,6 +397,119 @@ func consumeBackupDataCollections(
 	return bu.BackupCollections(ctx, bases, cs, sel.PathService(), oc, tags)
 }
 
+func matchesReason(reasons []kopia.Reason, p path.Path) bool {
```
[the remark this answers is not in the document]
ashmrtn: I was actually thinking we should factor our `Reason` from the kopia package, replace `OwnersCats` with it where needed, and do a little cleanup as to how kopia generates tags for a snapshot. I'm not sure if the `Reason <-> path` check should live in the path package though. That seems like it's adding more functions to an already fairly large interface

PR #1906, opened 2022-12-21
refactors the exchange fetchIDs iters

ryanfkeepers:
## Description
    
Transitions the fetchIDs service iterators to a
set of interfaces to consolidate code across
multiple nearly identical variations of "fetch id
for directory".

This was originally written in another PR (1780), then
separated out to isolate changes.

## Type of change

- [x] :hamster: Trivial/Minor

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/shared.go
+++ b/src/internal/connector/exchange/api/shared.go
@@ -34,13 +34,28 @@ type getIDAndAddtler interface {
 // Generics used here to handle the variation of msoft interfaces
 // that all _almost_ comply with GetValue, but all return a different
 // interface.
-func toValues(a any) ([]getIDAndAddtler, error) {
-	gv, ok := a.(interface{ GetValue() []getIDAndAddtler })
+func toValues[T any](a any) ([]getIDAndAddtler, error) {
+	gv, ok := a.(interface{ GetValue() []T })
 	if !ok {
-		return nil, errors.New("response does not comply with getIDAndAddtler interface")
+		return nil, errors.Errorf("response of type [%T] does not comply with the GetValue() interface", a)
 	}
 
-	return gv.GetValue(), nil
+	items := gv.GetValue()
+	r := make([]getIDAndAddtler, 0, len(items))
+
+	for _, item := range items {
+		var a any
+		a = item
+
+		ri, ok := a.(getIDAndAddtler)
+		if !ok {
+			return nil, errors.Errorf("item of type [%T] does not comply with the getIDAndAddtler interface", item)
+		}
+
+		r = append(r, ri)
+	}
+
+	return r, nil
 }
 
 // generic controller for retrieving all item ids in a container.
@@ -79,13 +94,6 @@ func getItemsAddedAndRemovedFromContainer(
 
 		// iterate through the items in the page
 		for _, item := range items {
-			if item.GetId() == nil {
-				errUpdater(errors.Errorf("item with nil ID"))
-
-				// TODO: Handle fail-fast.
-				continue
-			}
-
 			// if the additional data conains a `@removed` key, the value will either
 			// be 'changed' or 'deleted'.  We don't really care about the cause: both
 			// cases are handled the same way in storage.
--- a/src/internal/connector/exchange/api/shared.go
+++ b/src/internal/connector/exchange/api/shared.go
@@ -3,9 +3,10 @@ package api
 import (
 	"context"
 
+	"github.com/pkg/errors"
+
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/support"
-	"github.com/pkg/errors"
 )
 
 // ---------------------------------------------------------------------------
@@ -14,7 +15,6 @@ import (
 
 type itemPager interface {
 	getPage(context.Context) (pageLinker, error)
-	idsIn(pageLinker) ([]getIDer, error)
 	setNext(string)
 }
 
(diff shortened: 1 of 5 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -85,11 +73,13 @@ func FilterContainersAndFillCollections(
...
 		cID := *c.GetId()
 
-		// this folder exists (probably), do not delete it.
-		delete(deletedPaths, cID)
-
+		currPath, ok := includeContainer(qp, c, scope)
+		if currPath != nil {
+			// this path exists, do not delete it, even if it isn't
+			// included in this backup.
+			delete(tombstones, currPath.String())
```
dadams39, line 80: Comment is confusing. `do not delete` then call delete?

src/internal/connector/exchange/service_iterators.go
```diff
@@ -131,31 +130,30 @@ func FilterContainersAndFillCollections(
...
-	// left in deletedPaths are only the previous paths that did not appear as
-	// children of the root.
-	for fID, dp := range deletedPaths {
+	// A tombstone is a collection path that needs to be marked for deletion.
+	// Tombstones can occur for a number of reasons: the delta token expired,
+	// the container was deleted in flight, or the user deleted the container
+	// between backup runs.  If events combine to both delete and write content
+	// to the same container (ex: container_1 gets deleted, then container_2
+	// gets created with the same name), it is assumed that the backup consumer
+	// processes deletions before creations, making the combined operation safe.
```
dadams39, line 148: this is a little unclear. Are we not the backup consumer?  This probably should be reworded. I would make a suggestion, but it is unclear. I would imagine that  container_1 gets deleted and container_2 is created with the same name, that container_2's `id` is different than the previous and would require that the `id` be updated. 
However, I don't think this is the operation that is being described at all.

src/internal/connector/graph/service_helper.go
```diff
@@ -57,10 +57,13 @@ func CreateHTTPClient() *nethttp.Client {
...
 }
 
+// ---------------------------------------------------------------------------
+// Logging Middleware
+// ---------------------------------------------------------------------------
+
 // LoggingMiddleware can be used to log the http request sent by the graph client
 type LoggingMiddleware struct{}
 
-// Intercept implements the RequestInterceptor interface and decodes the parameters name
```
dadams39, line 63: Comment required for public function

dadams39 (commented):
A few things to be addressed

ryanfkeepers:
@dadams39 I let this branch get out of date from main, and you ended up reviewing a version that had a lot of merge conflicts.  Sorry about that!  In the future, I'd avoid reviewing any branch that has merge conflicts; it'll save you a bit of confusion when trying to scope comments to the actual changes.  Feel free instead to ping me and ask for a rebase so that you can do an accurate review.

The changes should be up to date now.  Your prior comments were all based on changes from other PRs.  I'm happy to walk through those topics outside of this review if you still have questions.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -552,9 +499,109 @@ func FetchMessageIDsFromDirectory(
...
+			if err := graph.IsErrInvalidDelta(err); err != nil {
+				return nil, nil, deltaURL, err
+			}
+
+			return nil, nil, deltaURL, errors.Wrap(err, support.ConnectorStackErrorTrace(err))
+		}
+
+		// each category type responds with a different interface, but all
+		// of them comply with GetValue, which is where we'll get our item data.
+		gv, ok := resp.(interface{ GetValue() []getIDAndAddtler })
```
ashmrtn, line 562: this should be added to the `pageLinker` interface. Yes, it will require a bit more glue code to convert the slice return values, but at least at that point it's clear what the interface actually requires as opposed to having a failure later on on a seemingly unrelated line of code even though the `pageLinker` interface was properly implemented according to the definition

ryanfkeepers: Hah, it's funny to have you mention that, as I went through a process of removing that extra glue for slimness.  Good points, though.

src/internal/connector/exchange/api/shared.go
```diff
@@ -0,0 +1,128 @@
...
+		addedIDs   = []string{}
+		removedIDs = []string{}
+		deltaURL   string
+	)
+
+	for {
+		// get the next page of data, check for standard errors
+		resp, err := pager.getPage(ctx)
+		if err != nil {
+			if err := graph.IsErrDeletedInFlight(err); err != nil {
```
ashmrtn, line 77: can we change the function name on for this and the check for an invalid delta? Just looking at the names, I would assume they don't mutate the original error and can be merged together to simplify things. Looking at the function definitions though that is not the case

src/internal/connector/exchange/api/shared.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return r, nil
+}
+
+// generic controller for retrieving all item ids in a container.
+func getItemsAddedAndRemovedFromContainer(
+	ctx context.Context,
+	pager itemPager,
+	errUpdater func(error),
```
ashmrtn, line 65: although this is passed in I don't think it's ever used in this function. We could probably skip passing it in and just wrap the errors in the caller

ashmrtn (approved):

src/internal/connector/exchange/api/shared.go
ryanfkeepers: We can revisit those in a different PR.

src/internal/connector/exchange/api/shared.go
ryanfkeepers: good eye, thanks

PR #1909, opened 2022-12-22
Stricter checking for how hierarchies are merged

### LINKED ISSUE #1884: don't source items from base snapshot if collection is marked as new
There are some situation around expired/invalid/missing tokens where a folder at a path that was seen previously will be pulled in full and marked as new. In those cases, kopia.Wrapper should ignore the items in the base snapshot instead of merging them with the new collection. This will keep "zombie items" that have actually been deleted from persisting across backups

ashmrtn:
## Description

Add some extra error checking for how the hierarchy can evolve during merging in kopia.Wrapper. Add more tests to solidify this behavior as well.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [x] :robot: Test
- [ ] :hamster: Trivial/Minor

## Issue(s)

* closes #1884 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -576,8 +576,8 @@ func inflateCollectionTree(
 		ownerCats.ServiceCats[serviceCat] = ServiceCat{}
 		ownerCats.ResourceOwners[s.FullPath().ResourceOwner()] = struct{}{}
 
-		// Make sure we haven't moved and made a new instance of the collection at
-		// this path.
+		// Make sure there's only a single collection adding items for any given
+		// path in the new hierarchy.
 		if node.collection != nil {
 			return nil, nil, errors.Errorf("multiple instances of collection at %s", s.FullPath())
 		}
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -455,6 +455,29 @@ func newTreeMap() *treeMap {
 	}
 }
 
+// maybeGetTreeNode walks the tree(s) with roots roots and returns the node
+// specified by pathElements if all nodes on the path exist. If pathElements is
+// nil or empty then returns nil.
+func maybeGetTreeNode(roots map[string]*treeMap, pathElements []string) *treeMap {
+	if len(pathElements) == 0 {
+		return nil
+	}
+
+	dir := roots[pathElements[0]]
+
+	for i := 1; i < len(pathElements); i++ {
+		if dir == nil {
+			return nil
+		}
+
+		p := pathElements[i]
+
+		dir = dir.childDirs[p]
+	}
+
+	return dir
+}
+
 // getTreeNode walks the tree(s) with roots roots and returns the node specified
 // by pathElements. If pathElements is nil or empty then returns nil. Tree nodes
 // are created for any path elements where a node is not already present.
@@ -500,6 +523,9 @@ func inflateCollectionTree(
 	// Allows resolving what the new path should be when walking the base
 	// snapshot(s)'s hierarchy. Nil represents a collection that was deleted.
 	updatedPaths := make(map[string]path.Path)
+	// Temporary variable just to track the things that have been marked as
+	// changed while keeping a reference to their path.
+	changedPaths := []path.Path{}
 	ownerCats := &OwnersCats{
 		ResourceOwners: make(map[string]struct{}),
 		ServiceCats:    make(map[string]ServiceCat),
@@ -508,10 +534,29 @@ func inflateCollectionTree(
 	for _, s := range collections {
 		switch s.State() {
 		case data.DeletedState:
+			changedPaths = append(changedPaths, s.PreviousPath())
+
+			if _, ok := updatedPaths[s.PreviousPath().String()]; ok {
+				return nil, nil, errors.Errorf(
+					"multiple previous state changes to collection %s",
+					s.PreviousPath(),
+				)
+			}
+
 			updatedPaths[s.PreviousPath().String()] = nil
+
 			continue
 
 		case data.MovedState:
+			changedPaths = append(changedPaths, s.PreviousPath())
+
+			if _, ok := updatedPaths[s.PreviousPath().String()]; ok {
+				return nil, nil, errors.Errorf(
+					"multiple previous state changes to collection %s",
+					s.PreviousPath(),
+				)
+			}
+
 			updatedPaths[s.PreviousPath().String()] = s.FullPath()
 		}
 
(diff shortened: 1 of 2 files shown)


src/internal/kopia/upload.go
```diff
@@ -508,10 +534,29 @@ func inflateCollectionTree(
...
 		switch s.State() {
 		case data.DeletedState:
+			changedPaths = append(changedPaths, s.PreviousPath())
+
+			if _, ok := updatedPaths[s.PreviousPath().String()]; ok {
+				return nil, nil, errors.Errorf(
+					"multiple previous state changes to collection %s",
+					s.PreviousPath(),
+				)
+			}
```
ryanfkeepers, line 544: Is this going to error out if we try to both delete and renew a path, as in the case for token expiration?

src/internal/kopia/upload.go
```diff
@@ -531,9 +576,29 @@ func inflateCollectionTree(
 		ownerCats.ServiceCats[serviceCat] = ServiceCat{}
 		ownerCats.ResourceOwners[s.FullPath().ResourceOwner()] = struct{}{}
 
+		// Make sure we haven't moved and made a new instance of the collection at
+		// this path.
```
ryanfkeepers, line 580: This comment is kinda difficult to parse.  Does it mean that a collection in the node implies multiple collections writing data to the same path?

src/internal/kopia/upload.go
ashmrtn: no, it will only error out if we try to do two or more of: delete, move, not move for a given previous path. There's tests included that show the different error cases and I can add one that will succeed for delete and new

src/internal/kopia/upload.go
ashmrtn: if I understand your question correctly yes. Essentially this code is reached for new, not moved, and moved states. Those are the states where we have to worry about new items being added. What we want to ensure is that there's only a single collection adding items for any given path in the _new_ hierarchy

The new hierarchy is based only off the values of `FullPath()` and there are no assumptions/requirements for how collections are processed

src/internal/kopia/upload.go
ryanfkeepers: Hmm, I'm wary of footguns here.  From the producer (gc) side of this, handling a token expiration is getting fuzzy.  The potential outcomes of `notMoved` (ie: just update the directory with the refreshed delta data), `deleted, notMoved` (make a tombstone, but also include the prevPath), and `deleted, new` (make a tombstone, then treat the next collection as new) all feel equally valid when considered in a vacuum.  It takes prior knowledge of kopia's behavior and expectations to choose correctly among them.  That will work in the short term, but I'd feel safer if we could either reduce the combination possibilities, or increase resilience across variations (the former being more preferable for simplicity).

src/internal/kopia/upload.go
ryanfkeepers: > ensure that there's only a single collection adding items for any given path in the new hierarchy

Aha, this ^ is what the comment in code should be.

src/internal/kopia/upload.go
ashmrtn: >reduce the combination possibilities, or increase resilience across variations

I don't think we can really do more of either of these things. Again, we're essentially describing how state has evolved so we need information about how things in the base (old version) have changed and information about the new version of things. Sometimes these changes can be combined into a single collection (i.e. moved/not moved/new states) and sometimes these changes need to be split across two collections (i.e. deleted+new, moved+new, moved+moved)

As a general theme, it may be useful to start by thinking of the outputs of `FullPath()` and `PreviousPath()` and what, taken together, they describe. We already have code for this in the exchange collection stuff. By looking at these, we know 1. where the collection was in in the base snapshot and 2. where the collection should be in the new snapshot. The assumptions that should hold on that set of collections are
1. each collection should have a unique `PreviousPath()` unless `PreviousPath() == nil` (i.e. we have a 1:1 mapping of old collections to their changes unless a collection is new)
2. each collection should have a unique `FullPath()` unless `FullPath() == nil` (i.e. we only source data for a kopia folder from a single collection. We have the nil part to work around the fact that deleted containers use nil for `FullPath()`)

As we've mapped operations on containers in M365 to states in collections in a 1:1 manner, the set of collections we have essentially represent a set of commands that modify the hierarchy. For example, the following hierarchy operations in M365
```text
mv /foo /bar
mkdir /foo
rmdir /baz
mv /inbox /baz
```
would generate the collections (shown in the same order commands ran)
```
{FullPath: /bar, PreviousPath: /foo, State: Moved}
{FullPath: /foo, PreviousPath: <nil>, State: New}
{FullPath: <nil>, PreviousPath: /baz, State: Deleted}
{FullPath: /baz, PreviousPath: /inbox, State: Moved}
```

The major difference between the set of commands (top) and the collections we generate (bottom) is that we don't assume any processing order for the collections. We can get away with this because we have distinct current and previous paths that we track. If we only had `FullPath()` then we'd need to enforce ordering on how collections were consumed in the kopia wrapper

The tricky part is handling token expiry. The complexity for this comes from the fact that we've implicitly mirrored the behavior of a file system when merging the hierarchy in kopia wrapper. This was required because moves and deletes _do_ affect the subtree rooted at the moved/deleted folder in M365. To work around that, we need some way to denote that we still want to affect the subtree of the item (i.e. state specifier) but we don't want to merge items (i.e. some boolean flag). Having the boolean flag will probably make the GC code simpler as well because there's less special handling for states

If we wanted to just bound the number of combinations that could be possibly generated we'd need to control the M365 interface and say things like "a user cannot create a folder in the same spot in the hierarchy with the same name as a deleted folder." Without that, all bets are off as to what changes we'll see

ryanfkeepers: I'm with you on all of that.  Could have been more clear with my language.  A 'reduction of combinations' wasn't meant to apply to the process as a whole, but was meant specifically for token expiration.  A better phrasing would be, "how can we minimize or remove needing to understand kopia in order to handle token expiration correctly?".

Maybe this discussion is outside of this pr, but: the choice to handle expiration as a `delete, new` collection pair seemed reasonable at first blush as a simplification.  Now I'm feeling the opposite: it doesn't feel natural to force a deletion on a folder that isn't being deleted.  It would be more natural to only tombstone folders that are absolutely deleted, and token expirations are not deletions.  They're either `moved-rebuilt` or `not-moved-rebuilt`, which seems to mimic the same indication of the bool being proposed in #1914, potentially allowing us to combine the concepts and avoid the `delete, new` handling of token expiration.

That, I suppose, relies on being able to identify the "rebuilt" portion of the state/flag to only merge subfolders and not items, which seems like the point of 1914.

> would generate the collections (shown in the same order commands ran)

Minor point of contention here (I'm probably pointing out something you're already aware of).  We would get those collections _if_ delta backups were run between each change.  If the backup were run only before and after all changes, the results would reduce to:
```
{FullPath: /bar, PreviousPath: /foo, State: Moved}
{FullPath: /foo, PreviousPath: <nil>, State: New}
{FullPath: /baz, PreviousPath: /inbox, State: Moved} // plus lots of item deletions in /baz.
// also, in both cases, deleted /inbox
```

ashmrtn: >choice to handle expiration as a delete, new collection pair

yes that was a bit of a workaround because I didn't want to have to add more states/flags to the collection. #1914 shows there's a bug in that logic though so we need to abandon that solution. In the future expiry will be handled with `State: [Not]Moved, NoMerge: true` or something similar. It can then be represented as a single collection instead of two

>If the backup were run only before and after all changes, the results would reduce to:

Actually it would still be what I posted earlier. The reason being delta tokens/URLs are tied to container IDs in M365 (hence why we need the path mapping metadata). Just handling the deletions etc based on the IDs and updating the mappings should give the proper set of collections

ryanfkeepers: conversation resolved in sync.  Overview:
* 1914 can simplify token expiration by removing the `deleted, new` tuple handling.
* a bug was uncovered in the gc production of collections, leading to different outcomes than expected.

ryanfkeepers (approved):

PR #1896, opened 2022-12-21
Update folder size and modified time in details

vkamra:
## Description

Caches folder info added during details construction in the details builder and keeps the size/modified time
updated as newer items are added.

## Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [x] :clock1: Yes, but in a later PR
- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :computer: CI/Deployment
- [ ] :hamster: Trivial/Minor

## Issue(s)

* #1850 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #1913, opened 2022-12-22
Address review feedback from #1896

vkamra:
## Description

- Adds accessor methods for size/modified
- Do not initialize folder in test with a modified time

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test
- [x] :hamster: Trivial/Minor

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Folder entries in backup details now include size and modified time for the hierarchy (#1896)
+
 ### Changed
 
 - **Breaking Change**:
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -145,26 +145,9 @@ func (b *Builder) AddFoldersForItem(folders []FolderEntry, itemInfo ItemInfo) {
 		}
 
 		// Update the folder's size and modified time
-		var (
-			itemSize     int64
-			itemModified time.Time
-		)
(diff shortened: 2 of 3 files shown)


rajatkateja (approved):
LGTM. 

One corner case: what happens if there is an empty folder? Do we not back it up at all and hence it never shows in the details?  
And on a related note: is `AddFoldersForItem` always called on an item or is it called on folders too?

vkamra:
> One corner case: what happens if there is an empty folder? Do we not back it up at all and hence it never shows in the details?

Empty folders are not backed up (@ashmrtn - correct me if I'm wrong?)

> And on a related note: is `AddFoldersForItem` always called on an item or is it called on folders too?

Yep - `AddFoldersForItem` is always called for an item.

ashmrtn:
>Empty folders are not backed up

I think we're actually inconsistent on this point. We've never really decided on if there needs to be a fixed behavior

As far as I know, OneDrive will skip empty folders because it can infer which folders are not empty by examining the paths of items. Exchange will create empty folders because we need to find a folder before we know if it has items in it. Incrementals will create new backups that do not have empty folders if the folder was already empty in the base snapshot for the incremental (i.e. folder foo in the base had no items -> it will not appear in the new backup. If folder foo had an item in the base but the item was deleted in the new backup then foo will still appear in the new backup but it will be empty)

rajatkateja:
Yeah, this sounds pretty inconsistent. I think it is fine for now (Ark preview) but can we track making this consistent somewhere just so we don't forget about it? Thanks!

Issue #1914, opened 2022-12-22
add `NoMerge` flag for collections to control merging in kopia


ashmrtn:
Under some conditions, we do not want to merge items in a collection with the base snapshot but we do want to preserve changes (or not change) the subtree rooted at the given collection. One such instance of this is when a delta token expires in M365. When a token expires, Corso will fallback to enumerating all items reachable with that delta endpoint. 

However, because it's doing a full enumeration and GraphConnector has no information about what is in the base snapshot it cannot produce entries in the collection for deleted items

Making a collection marked as deleted (to stop sourcing items from the base) and then making a collection marked as new with the full enumeration of items is also not an option because it will delete the subtree rooted at the folder being enumerated

Adding a new flag instead of making new state(s) for the collection is more flexible because there are times where we'd want to skip merging base snapshot items for a specific container but still propagate changes to the subtree. An example would be moving a container in M365 but having an expired delta token. In that instance, the subfolders should be moved but no items for the root of the subtree should be sourced from the base. Because of this, at least two additional states would be required if we used states only, a `NoMergeMoved` and `NoMergeNotMoved` state

PR #1919, opened 2022-12-22
add doNotMerge func to collections

ryanfkeepers:
## Description

Adds a new func to the data.Collection iface:
DoNotMergeItems() tells kopia to skip the
retention of items from prior snapshots when
generating the new snapshot for this collection.
A primary use case for this flag is when a delta
token expires, preventing an incremental lookup
and forcing gc to re-discover all items in the
container.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1914

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/mockconnector/mock_data_collection.go
+++ b/src/internal/connector/mockconnector/mock_data_collection.go
@@ -15,15 +15,15 @@ import (
 
 // MockExchangeDataCollection represents a mock exchange mailbox
 type MockExchangeDataCollection struct {
-	fullPath        path.Path
-	messageCount    int
-	Data            [][]byte
-	Names           []string
-	ModTimes        []time.Time
-	ColState        data.CollectionState
-	PrevPath        path.Path
-	DeletedItems    []bool
-	doNotMergeItems bool
+	fullPath     path.Path
+	messageCount int
+	Data         [][]byte
+	Names        []string
+	ModTimes     []time.Time
+	ColState     data.CollectionState
+	PrevPath     path.Path
+	DeletedItems []bool
+	DoNotMerge   bool
 }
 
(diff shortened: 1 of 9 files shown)


src/internal/connector/mockconnector/mock_data_collection.go
```diff
@@ -15,14 +15,15 @@ import (
...
-	DeletedItems []bool
+	fullPath        path.Path
+	messageCount    int
+	Data            [][]byte
+	Names           []string
+	ModTimes        []time.Time
+	ColState        data.CollectionState
+	PrevPath        path.Path
+	DeletedItems    []bool
+	doNotMergeItems bool
```
ashmrtn, line 26: nit: can you export this please? I'm going to need it when I wire up the KW side of things
```suggestion
	DoNotMergeItems bool
```

src/internal/data/data_collection.go
```diff
@@ -47,6 +47,15 @@ type Collection interface {
 	// backup along with all items and Collections below them in the hierarchy
 	// unless said items/Collections were moved.
 	State() CollectionState
+	// DoNotMergeItems informs kopia that the collection is rebuilding its contents
+	// from scratch, and that any items currently stored in that collection should
```
ashmrtn, line 51:
```suggestion
	// from scratch, and that any items currently stored at the collections `PreviousPath()` should
```

ashmrtn (approved):

src/internal/connector/mockconnector/mock_data_collection.go
ryanfkeepers: sure can

PR #1921, opened 2022-12-22
set nomerge on delta reset

ryanfkeepers:
## Description

Flag doNotMergeItems for exchange collections
where the delta token was marked as reset (ie,
expired or otherwise unusable).  Do this instead
of making a (delete, new) collection tuple for
that path.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1914

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -87,18 +87,20 @@ func NewCollection(
 	service graph.Servicer,
 	statusUpdater support.StatusUpdater,
 	ctrlOpts control.Options,
+	doNotMergeItems bool,
 ) Collection {
 	collection := Collection{
-		user:           user,
-		data:           make(chan data.Stream, collectionChannelBufferSize),
-		jobs:           make([]string, 0),
-		service:        service,
-		statusUpdater:  statusUpdater,
-		fullPath:       curr,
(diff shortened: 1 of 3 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -87,18 +87,20 @@ func NewCollection(
...
-		data:           make(chan data.Stream, collectionChannelBufferSize),
-		jobs:           make([]string, 0),
-		service:        service,
-		statusUpdater:  statusUpdater,
-		fullPath:       curr,
-		prevPath:       prev,
-		collectionType: collectionType,
-		ctrl:           ctrlOpts,
-		state:          stateOf(prev, curr),
+		collectionType:  collectionType,
```
ryanfkeepers, line 93: decided to alphabetize these.

ashmrtn (approved):
Needs rebasing, but looks good

not something I can comment on directly in this PR but I think [line 77](https://github.com/alcionai/corso/pull/1921/files#diff-67d5d9c3856fb884fe81259eefeb5d7ddb0d695582bf44db4e9ee6e447bc808dL77) in `service_iterators.go` needs updated to use IDs instead of paths. This change should be done in a new PR

Issue #1923, opened 2022-12-22
better handling when GraphConnector encounters errors pulling item IDs for a container


ashmrtn:
Below code links are perma-links so they may not reflect the exact state of the code in main. However, they are useful because they won't drift to something unrelated

With the introduction of incremental backups, GraphConnector needs more complex logic so it can communicate hierarchy state changes to kopia wrapper. Some of these state changes have effects beyond just the container in question, instead applying to subtrees of the hierarchy (ex. move, delete). Operations like these should be carefully considered when handling errors in GraphConnector so we don't mistakenly attempt to delete a subtree instead of producing an empty collection

Situations that stand out as things that could be problematic include
* [container deletions](https://github.com/alcionai/corso/blob/aacb013b60cd4980d401ceb6edba50f6986b3057/src/internal/connector/exchange/service_iterators.go#L110) while fetching item IDs

In cases like these, it may be better to create a new, empty collection and mark it as `DoNotMergeItems`. This will result in an empty collection in the backup but will preserve any subdirectories in kopia rooted at that path

PR #1924, opened 2022-12-22
reset container if deleted in flight, not delete

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* #1923

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -97,13 +97,15 @@ func filterContainersAndFillCollections(
 
 		jobs, newDelta, err := getJobs(ctx, service, qp.ResourceOwner, cID, prevDelta)
 		if err != nil {
-			// race conditions happen, containers might get deleted while
-			// this process is in flight.  If it was deleted, we remake the
-			// tombstone, just to be sure it gets deleted from storage.
 			if graph.IsErrDeletedInFlight(err) == nil {
 				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
 			} else {
-				tombstones[cID] = dp.path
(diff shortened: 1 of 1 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -101,13 +101,15 @@ func filterContainersAndFillCollections(
...
-			// tombstone, just to be sure it gets deleted from storage.
 			if graph.IsErrDeletedInFlight(err) == nil {
 				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
 			} else {
-				tombstones[currPath.String()] = struct{}{}
+				// race conditions happen, containers might get deleted while
+				// this process is in flight.  If that happens, force the collection
+				// to reset which will prevent any old items from being retained in
+				// storage.  If the container (or its children) are sill missing
+				// on the next backup, they'll get tombstoned.
```
ashmrtn, line 111: ☠️

ashmrtn (approved):
Outside of this, we may also need to do the same if we fail parsing the old path for some reason? We won't try to delete the subtree since we have the mapping based on container ID, but we do ignore the delta token in that case. I was having a bit of a time trying to figure that out though cause there were still some PRs in flight for it

ryanfkeepers:
> Outside of this, we may also need to do the same if we fail parsing the old path for some reason?

Would we?  Like you said, we don't lose the subtree.  And new directories (I assume) don't merge existing items, which is what this change controls for.  Otherwise we'd have to start tracking `NewCollection-[DoNot]Merge`, which seems a bit backwards.  So long as we don't create a new folder *and also* run a delta, we should be fine.

Issue #1901, opened 2022-12-21
add flag to force full backup


ashmrtn:
Incremental backups increase the chance that Corso gets into a state where it can't complete a successful backup because new backups depend on the success and correctness of backups before them. Furthermore, the fact that Corso is going to default to trying to make an incremental backup makes it such that the user must delete older backups if they want to force a full backup for any reason.

We should provide a smoother and less destructive experience for making a full backup even when the incrementals code is enabled by providing a (possibly hidden) CLI flag that forces a full backup

ryanfkeepers:
Started with the inverse: a flag to _enable_ incremental backups for exchange.  When we release that behavior and standardize incrementals for exchange, we can use this ticket to remove that feature flag and replace it with a forced full-backup operation.

ashmrtn:
Having a flag to disable a full backup may also require a minor change in how kopia.Wrapper works. Currently `BackupCollections` takes a single list of bases to make the incremental off of. These bases are used for both [building the in-memory hierarchy](https://github.com/alcionai/corso/blob/37f1a3eca5aa2eba21352a1a663eee7263583ab2/src/internal/kopia/wrapper.go#L145) to hand to kopia and [giving to kopia](https://github.com/alcionai/corso/blob/37f1a3eca5aa2eba21352a1a663eee7263583ab2/src/internal/kopia/wrapper.go#L150) to find cached files (aka kopia-assisted incrementals).

To disable delta token incrementals, we should skip handing the bases to the function that builds the in-memory hierarchy. However we should still take advantage of kopia-assisted incrementals

PR #1941, opened 2022-12-23
retain kopia-assist when not on incrementals

ryanfkeepers:
## Description

Adds a flag to the BackupCollections interface
that identifies whether the caller is running an
incremental backup or not.  If they are, kopia
will utilize the previous base snapshots when
building the directory tree.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1901

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/streamstore/streamstore.go
+++ b/src/internal/streamstore/streamstore.go
@@ -45,6 +45,7 @@ const (
 func (ss *streamStore) WriteBackupDetails(
 	ctx context.Context,
 	backupDetails *details.Details,
+	isIncremental bool,
 ) (string, error) {
 	// construct the path of the container for the `details` item
 	p, err := path.Builder{}.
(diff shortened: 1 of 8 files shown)


src/internal/streamstore/streamstore.go
```diff
@@ -80,6 +81,7 @@ func (ss *streamStore) WriteBackupDetails(
 		ss.service,
 		nil,
 		nil,
+		isIncremental,
```
ashmrtn, line 84: this depends on how much of the underlying incremental behavior we want to expose to higher layer. For now since we're already passing `nil` for the set of bases I think it's fine to just hardcode it to `false` and not even expose it as another parameter

ashmrtn (approved):

src/internal/streamstore/streamstore.go
ryanfkeepers: ah, I was looking at the wrong param.  You're right, no point in the flag if the input is nil.

Issue #1777, opened 2022-12-12
Exchange delta token-based incremental backups


ashmrtn:
Reduce the number of Graph API requests for a backup as well as the amount of time a backup takes by leveraging delta tokens in Exchange. This ticket addresses only incremental backups. It does not address restartability at all

This feature requires work on multiple components and includes the following:
- [x] #1612 
- [x] #1685
- [x] #1700 
- [x] #1404 
- [x] #1505
- [x] #1740 
- [x] #1725 
- [x] #1726 
- [x] #1727
- [x] #1779
- [x] #1780
- [x] #1781
- [x] #1800
- [x] #1804 
- [x] #1829
- [x] #1832
- [x] #1866
- [x] #1884
- [x] #1914

-----
Testing
- [x] #1966
- [x] #1967 

-----
Phase 2
- [x] #1901
- [x] #1865
- [x] #1878
- [x] #1812
- [x] #1945 
- [ ] #2068 
- [x] #2022 
- [x] #1954

ashmrtn:
closing as remaining issue is tracked elsewhere

PR #1952, opened 2022-12-23
distinguish added and removed jobs

ryanfkeepers:
## Description

In order to track deleted items, we need to build
separate job sets when aggregating item IDs:
one for added items, and one for removed.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #1777

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -211,10 +211,6 @@ func (col *Collection) streamItems(ctx context.Context) {
 
 	// delete all removed items
 	for _, id := range col.removed {
-		if col.ctrl.FailFast && errs != nil {
-			break
-		}
-
 		semaphoreCh <- struct{}{}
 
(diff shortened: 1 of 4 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -206,15 +209,44 @@ func (col *Collection) populateByOptionIdentifier(ctx context.Context) {
 		errs = support.WrapAndAppend(user, err, errs)
 	}
 
-	for _, identifier := range col.jobs {
+	// delete all removed items
+	for _, id := range col.removed {
+		if col.ctrl.FailFast && errs != nil {
+			break
+		}
+
```
ashmrtn, line 217: nit: could skip this since no errors are generated in this loop
```suggestion
```

ashmrtn (approved):

Issue #1935, opened 2022-12-23
GC: SharePoint: Restore: List Items 


dadams39:
## Issue
Implement restore functionality for SharePoint Lists.

PR #1965, opened 2022-12-27
SharePoint: Mock:  Stream and Colleciton Mock Expansion

dadams39:
## Description
Mocks created for testing `SharePoint.List` restore operations

- [x] :no_entry: No 

## Type of change

- [x] :hamster: Trivial/Minor

## Issue(s)

* related to #1935<issue>

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/mockconnector/mock_data_list.go
+++ b/src/internal/connector/mockconnector/mock_data_list.go
@@ -37,7 +37,7 @@ func (mlc *MockListCollection) DoNotMergeItems() bool {
 }
 
 func (mlc *MockListCollection) PreviousPath() path.Path {
-	return mlc.fullPath
+	return nil
 }
 
 func (mlc *MockListCollection) Items() <-chan data.Stream {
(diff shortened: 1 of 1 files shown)


src/internal/connector/mockconnector/mock_data_list.go
```diff
@@ -1,16 +1,86 @@
...
+func (mlc *MockListCollection) FullPath() path.Path {
+	return mlc.fullPath
+}
+
+func (mlc *MockListCollection) DoNotMergeItems() bool {
+	return false
+}
+
+func (mlc *MockListCollection) PreviousPath() path.Path {
+	return mlc.fullPath
```
ryanfkeepers, line 40: This could have unexpected side effects.  Better for it to be nil until there's an actual previousPath.  This will help ensure that if/when incrementals do get picked up for sharepoint, these mocks always treat their data as being in a "new" state (ie: not an incremental update), as per the expectations of the State field.
```suggestion
	return nil
```

ryanfkeepers (approved):

Issue #1966, opened 2022-12-27
Incrementals e2e/integration tests


ryanfkeepers:
## Issue
Incremental backups require CI integration tests on live data.  These tests should be housed in the `operations` package.  They should cover a non-extensive, happy path screen of incremental behavior.  Phase 1 involves a two-step backup (full, incremental) with no expectations about distinct changes in-between.  Phase 2 involves massaging the data to produce specific changes across one or more incrementals to test the full range of expected behavior.

PR #1973, opened 2022-12-27
add happy-path incremental tests to operations

ryanfkeepers:
## Description

Add a simple happy-path integration test to
operations backups.  This test only attemps to
assert the most basic expectations: that
incrementals are runnable, and that they involve
less data than the initial backup.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #1966

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup_integration_test.go
+++ b/src/internal/operations/backup_integration_test.go
@@ -27,12 +27,17 @@ import (
 // helpers
 // ---------------------------------------------------------------------------
 
+// prepNewBackupOp generates all clients required to run a backup operation,
+// returning both a backup operation created with those clients, as well as
+// the clients themselves.
+//
 //revive:disable:context-as-argument
 func prepNewBackupOp(
 	t *testing.T,
 	ctx context.Context,
 	bus events.Eventer,
 	sel selectors.Selector,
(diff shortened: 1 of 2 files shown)


src/internal/operations/backup_integration_test.go
```diff
@@ -70,22 +75,43 @@ func prepNewBackupOp(
...
+	return bo, acct, kw, ms, closer
+}
+
+// newBackupOp accepts the clients required to compose a backup operation, plus
+// any other metadata, and uses them to generate a new backup operation.  This
+// allows backup chains to utilize the same temp directory and configuration
+// details.
+//
+//revive:disable:context-as-argument
+func newBackupOp(
```
dadams39, line 89: `newBackupOp` --> `testBackupOp`

dadams39 (approved):
nit: The annotated function is very closely named to a function that exists within the package.  Would change that to avoid confusion.

src/internal/operations/backup_integration_test.go
ryanfkeepers: 👍 Like the other PR, I'll save the rename here for the next PR.

PR #1975, opened 2022-12-28
set up integration test framework

ryanfkeepers:
## Description

Adds the base test framework for handling
incremental backups integration testing.  Future
changes will add more tests to the table.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #1966

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup_integration_test.go
+++ b/src/internal/operations/backup_integration_test.go
@@ -38,12 +38,12 @@ const incrementalsDestFolderPrefix = "incrementals_ci_"
 // helpers
 // ---------------------------------------------------------------------------
 
-// prepNewBackupOp generates all clients required to run a backup operation,
+// prepNewTestBackupOp generates all clients required to run a backup operation,
 // returning both a backup operation created with those clients, as well as
 // the clients themselves.
 //
 //revive:disable:context-as-argument
-func prepNewBackupOp(
+func prepNewTestBackupOp(
 	t *testing.T,
 	ctx context.Context,
 	bus events.Eventer,
@@ -86,18 +86,18 @@ func prepNewBackupOp(
 		ms.Close(ctx)
 	}
 
-	bo := newBackupOp(t, ctx, kw, ms, acct, sel, bus, featureFlags, closer)
+	bo := newTestBackupOp(t, ctx, kw, ms, acct, sel, bus, featureFlags, closer)
 
 	return bo, acct, kw, ms, closer
 }
 
-// newBackupOp accepts the clients required to compose a backup operation, plus
+// newTestBackupOp accepts the clients required to compose a backup operation, plus
 // any other metadata, and uses them to generate a new backup operation.  This
 // allows backup chains to utilize the same temp directory and configuration
 // details.
 //
 //revive:disable:context-as-argument
-func newBackupOp(
+func newTestBackupOp(
 	t *testing.T,
 	ctx context.Context,
 	kw *kopia.Wrapper,
@@ -282,7 +282,7 @@ func checkMetadataFilesExist(
 
 // the params here are what generateContainerOfItems passes into the func.
 // the callback provider can use them, or not, as wanted.
-type dataBuilderFunc func(id, now, subject, body string) []byte
+type dataBuilderFunc func(id, timeStamp, subject, body string) []byte
 
 //revive:disable:context-as-argument
 func generateContainerOfItems(
@@ -293,15 +293,15 @@ func generateContainerOfItems(
 	cat path.CategoryType,
 	sel selectors.Selector,
 	tenantID, userID, destFldr string,
-	howMany int,
+	howManyItems int,
 	dbf dataBuilderFunc,
-) (*details.Details, error) {
+) *details.Details {
 	//revive:enable:context-as-argument
 	t.Helper()
 
-	items := make([]incrementalItem, 0, howMany)
+	items := make([]incrementalItem, 0, howManyItems)
 
-	for i := 0; i < howMany; i++ {
+	for i := 0; i < howManyItems; i++ {
 		var (
 			now       = common.Now()
 			nowLegacy = common.FormatLegacyTime(time.Now())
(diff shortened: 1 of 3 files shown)


src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
+//revive:disable:context-as-argument
+func generateContainerOfItems(
+	t *testing.T,
+	ctx context.Context,
+	gc *connector.GraphConnector,
+	service path.ServiceType,
+	cat path.CategoryType,
+	sel selectors.Selector,
+	tenantID, userID, destFldr string,
+	howMany int,
```
dadams39, line 296:
```suggestion
     total int,
```
The variable name would need to be updated within the function.

src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
-	assert.NoError(t, err)
+	collections := []incrementalCollection{{
+		pathElements: []string{destFldr},
+		category:     cat,
+		items:        items,
+	}}
 
-	for _, col := range cols {
-		itemNames := []string{}
+	dest := control.DefaultRestoreDestination(common.SimpleTimeTesting)
```
dadams39, line 325: Weird thought, we call this almost the same through out the code. 
In a different PR should we just make it into the following 2:
`control.DefaultRestoreTestingDestination()`
`control.DefaultRestoreDestination()` 
~jat

src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
+	dest.ContainerName = destFldr
 
-		for item := range col.Items() {
-			assert.Implements(t, (*data.StreamSize)(nil), item)
```
dadams39, line 232: Is the test losing anything by removing this assertion? 

L334 has an error return, is that the error descriptive enough to lose the assertion?

src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
-			)
+	return gc.RestoreDataCollections(ctx, sel, dest, dataColls)
+}
 
-			itemNames = append(itemNames, item.UUID())
-		}
+type incrementalItem struct {
+	name string
+	data []byte
+}
```
dadams39, line 344: I could see the utility of this being in a global package, maybe even mocks or something. It works similarly to the `data.Stream`.  

Not necessary to fix. Just highlighting that within testing we continue to have to make these structs and if they just happened to be in the tool belt, well, they would be available for us within testing.

src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
 
-		assert.ElementsMatchf(
+type incrementalCollection struct {
+	// Elements (in order) for the path representing this collection. Should
+	// only contain elements after the prefix that corso uses for the path. For
+	// example, a collection for the Inbox folder in exchange mail would just be
+	// "Inbox".
```
dadams39, line 350: isn't this just the `path.folders()` call? 

https://github.com/alcionai/corso/blob/20ec708ea357c26afe0076e31f05eacb8597a688/src/pkg/path/resource_path.go#L206

src/internal/operations/backup_integration_test.go
```diff
@@ -191,67 +206,210 @@ func checkMetadataFilesExist(
...
+}
+
+func toDataLayerPath(
+	t *testing.T,
+	service path.ServiceType,
+	tenant, user string,
+	category path.CategoryType,
+	elements []string,
+	isItem bool,
+) (path.Path, error) {
```
dadams39, line 400: Not sure how much we are leveraging the `testing.T` item, but we could lose the error return if `require` was used. Additionally, there are several `toDataLayer`-type names within path, would it be beneficial to refactor this to `fetchTestPath` or something of the like?

src/internal/operations/backup_integration_test.go
```diff
@@ -451,6 +611,165 @@ func (suite *BackupOpIntegrationSuite) TestBackup_Run_exchange() {
...
+	)
+
+	m365, err := acct.M365Config()
+	require.NoError(t, err)
+
+	gc, err := connector.NewGraphConnector(ctx, acct, connector.Users)
+	require.NoError(t, err)
+
+	// generate 2 new folders and add data to them.
+	// need 2 points of data in each folder for testing purposes.
```
dadams39, line 642:
```suggestion
	// require 2 folders with identical data for testing purposes.
```

src/internal/operations/backup_integration_test.go
```diff
@@ -451,6 +611,165 @@ func (suite *BackupOpIntegrationSuite) TestBackup_Run_exchange() {
...
+		},
+	}
+	for _, test := range table {
+		suite.T().Run(test.name, func(t *testing.T) {
+			var (
+				incMB = evmock.NewBus()
+				incBO = newBackupOp(t, ctx, kw, ms, acct, sel.Selector, incMB, ffs, closer)
+			)
+
+			test.updateUserData(t)
```
dadams39, line 739: Is this necessary for this iteration as it is empty?

src/internal/tester/cli.go
```diff
@@ -36,7 +36,6 @@ func StubRootCmd(args ...string) *cobra.Command {
 func NewContext() (context.Context, func()) {
 	level := logger.Info
 
-	fmt.Printf("\n-----\nargs %v\n-----\n", os.Args)
```
dadams39, line 39: From `src/internal/events/mock/mock.go`
+1 on the check annotations:
```go
type MockBus struct {
```
```bash
exported: type name will be used as mock.MockBus by other packages, and that stutters; consider calling this Bus (revive)
```

dadams39 (approved):
These are all nits and can be fixed here or in clean-up PRs. 
Overall, there are a few variable names that could be more descriptive and a few functions/structs that seem very similar to an existing framework buried in the code base. If the code is pushed without the changes, be sure to add a few issues for `tech-debt`.

src/internal/operations/backup_integration_test.go
ryanfkeepers: It's bad form to add testing funcs to a standard package.  With how little we need to produce and modify here, I don't think there's anything wrong with using the standard constructor.

src/internal/operations/backup_integration_test.go
ryanfkeepers: Generally agreed.  Although, and I haven't reviewed all the variations recently, it's possible that each instance is configured to fulfill some specific needs.  That would provide sufficient utility to maintain the apparent duplication.

src/internal/operations/backup_integration_test.go
ryanfkeepers: Per error vs assert/require: sure, that sounds good.

Per the `toDataLayer`, I've talked with ash about a QoL constructor in paths.  The idea has grounds, but isn't of much importance.  Might make a tech debt issue for it later.

src/internal/operations/backup_integration_test.go
ryanfkeepers: It'll be non-empty in all subsequent iterations.

src/internal/tester/cli.go
ryanfkeepers: good old stutters.

src/internal/operations/backup_integration_test.go
ryanfkeepers: We didn't lose this assertion.  The diff is a bit weird, but the check is still there.

Issue #1967, opened 2022-12-27
Incrementals Mocked GC Tests


ryanfkeepers:
## Issues
Incrementals require lower-level testing in the `Connector` package which rely on mocking the graph client to produce controlled response data.  These mocks will allow us to perform quick and versatile checks on the control layer wrapping GC communication.

Currently, we have no mocking package in place for the graph client (nor elsewhere in corso code), and usage of the client is interwoven throughout the connector package.  Considering the high rate of churn that the client experiences, a recommended first step is to separate the two layers so that the connector package acts as a control layer, and another package (perhaps graph) acts as the data layer, with a mockable interface between the two.  This is a larger up-front lift, but will streamline mocked testing for both incrementals and other gc packages in the future.

PR #2003, opened 2022-12-29
first pass migrate graph api to subfolder

ryanfkeepers:
## Description

In order to establish a standard api around our
graph client usage, and thus be able to mock
for testing, we need to migrate graph client
usage into another pacakge.  This is the first
step in that process of refactoring.

All changes are code relocation importing and
exporting may change as needed.  No logic
was altered.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test
- [x] :hamster: Trivial/Minor

## Issue(s)

* #1967

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/api/api.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/microsoft/kiota-abstractions-go/serialization"
+
+	"github.com/alcionai/corso/src/internal/connector/graph"
+)
+
+// ---------------------------------------------------------------------------
+// common types
+// ---------------------------------------------------------------------------
+
+// DeltaUpdate carries details about delta retrieval in aggregators
```
dadams39, line 15: Probably needs a bit more information

src/internal/connector/exchange/api/contacts.go
```diff
@@ -0,0 +1,203 @@
...
+
+			builder = users.NewItemContactFoldersItemContactsDeltaRequestBuilder(*nextLink, gs.Adapter())
+		}
+
+		return nil
+	}
+
+	if len(oldDelta) > 0 {
+		err := getIDs(users.NewItemContactFoldersItemContactsDeltaRequestBuilder(oldDelta, gs.Adapter()))
+		// happy path
```
dadams39, line 178: Could we expand upon this comment?

dadams39 (approved):
a few nits. Linting and other tests will have to pass prior to merging.

ashmrtn:
What's the end goal for the `api` struct? Is it that it's going to implement all functions for all data categories in exchange and those will be the interface?
For example, is the end goal for it to be like
```go
type GraphAPI interface {
  GetMailFolders(...) (folders, error)
  CreateMailFolder(...) error
  CreateMailFolderWithParent(...) error
  GetMessages(...) (messages, error)
  ...

  GetContactFolders(...) (folders, error)
  CreateContactFolder(...) error
  CreateContactFolderWithParent(...) error
  GetContacts(...) (messages, error)
  ...

  // Events functions
}
```

The above seems like it's going to end up as a pretty big interface, which makes mocking annoying because all functions need to be implemented. It may be easier in the long run to instead split things by data category and have multiple implementations of a single interface like
```go
type GraphAPI interface {
  GetFolders(...) (folders, error)
  CreateFolder(...) error
  CreateNestedFolder(...) error
  GetItemsInFolder(...) (items, error)
}

type MailAPI struct {
  ...
}

// Implement interface functions
```

Despite the fact that each implementation works with different data types we're already on the way to wrapping those returned types with our own abstractions that hide the differences. All of the data types also essentially have the same execution flow for backing stuff up or restoring it (may also want to split backup and restore into different interfaces). For backup it's:
1. enumerate folders
2. get item IDs for each folder
3. get item data for each item ID

for restore it's:
1. make folder(s) for restored items
2. restore items in each folder

ryanfkeepers:
> It may be easier in the long run to instead split things by data category

@ashmrtn You're getting a little ahead of the progress here.  Two PRs down the line or so we do exactly this.

PR #2004, opened 2022-12-29
switch trivial for tech-debt

ryanfkeepers:
Swap the `Trivial/Minor` category for `Tech Debt/Cleanup`

- [x] :hamster: Trivial/Minor


meain (approved):
I was thinking about adding one for "Enhancement", mostly so as to not pile it up under "Feature".

ryanfkeepers:
> I was thinking about adding one for "Enhancement", mostly so as to not pile it up under "Feature".

Last company I was at had both of those tags.  The amount of meeting time that got taken up in debate about what work qualifies as one versus the other was astounding.  Imo, just having one is good enough.

PR #2005, opened 2022-12-29
No tester errors 2

ryanfkeepers:
## Type of change

- [x] :hamster: Trivial/Minor


src/internal/tester/integration_runners.go
```diff
@@ -1,11 +1,13 @@
 package tester
```
ryanfkeepers, line 1: This file contains the primary logic of the change.  All other changes in the PR comply with the addition of `t` and removal of the `error` response value, from these functions.

src/internal/tester/integration_runners.go
```diff
@@ -69,3 +69,33 @@ func LogTimeOfTest(t *testing.T) string {
 
 	return now
 }
+
```
ryanfkeepers, line 72: these funcs were moved here from tester/envvars.go.

meain (approved):

PR #2010, opened 2022-12-31
transition api methods to interfaces

ryanfkeepers:
## Description

replaces the new api client methods with interfaces, to prepare for testing funcions with mocks instead of integration.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #1967

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/api.go
+++ b/src/internal/connector/exchange/api/api.go
@@ -87,6 +87,18 @@ func newService(creds account.M365Config) (*graph.Service, error) {
 	return graph.NewService(adapter), nil
 }
 
+func (c Client) Contacts() Contacts {
+	return Contacts{c}
+}
+
+func (c Client) Events() Events {
+	return Events{c}
+}
+
+func (c Client) Mail() Mail {
+	return Mail{c}
+}
+
 // ---------------------------------------------------------------------------
 // helper funcs
 // ---------------------------------------------------------------------------
--- a/src/internal/connector/exchange/event_calendar_cache.go
+++ b/src/internal/connector/exchange/event_calendar_cache.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/pkg/errors"
 
-	"github.com/alcionai/corso/src/internal/connector/exchange/api"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/pkg/path"
 )
@@ -14,7 +13,7 @@ var _ graph.ContainerResolver = &eventCalendarCache{}
 
 type eventCalendarCache struct {
 	*containerResolver
-	ac     api.Client
+	enumer enumerateContainerser
 	userID string
 }
 
@@ -30,7 +29,7 @@ func (ecc *eventCalendarCache) Populate(
 		ecc.containerResolver = newContainerResolver()
 	}
 
-	err := ecc.ac.EnumerateCalendars(ctx, ecc.userID, ecc.addFolder)
+	err := ecc.enumer.EnumerateContainers(ctx, ecc.userID, "", ecc.addFolder)
 	if err != nil {
 		return err
 	}
--- a/src/internal/connector/exchange/event_calendar_cache.go
+++ b/src/internal/connector/exchange/event_calendar_cache.go
@@ -13,7 +13,7 @@ var _ graph.ContainerResolver = &eventCalendarCache{}
 
 type eventCalendarCache struct {
 	*containerResolver
-	enumer enumerateContainerser
+	enumer containersEnumerator
 	userID string
 }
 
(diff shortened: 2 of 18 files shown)


src/internal/connector/exchange/api/api.go
```diff
@@ -87,6 +87,18 @@ func newService(creds account.M365Config) (*graph.Service, error) {
...
+}
+
+func (c Client) Events() Events {
+	return Events{c}
+}
+
+func (c Client) Mail() Mail {
+	return Mail{c}
+}
+
```
dadams39, line 101: This appears a little circular. The creds and service associated are available within the client already. As this is being merged into main. A small comment about about the purpose of these calls would be helpful It's also unclear why there wouldn't be similar calls for onedrive or sharepoint.

src/internal/connector/exchange/event_calendar_cache.go
```diff
@@ -14,7 +13,7 @@ var _ graph.ContainerResolver = &eventCalendarCache{}
 
 type eventCalendarCache struct {
 	*containerResolver
-	ac     api.Client
+	enumer enumerateContainerser
```
dadams39, line 16: Perhaps ContainersEnumerater ?

src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -142,11 +142,11 @@ func (col *Collection) Items() <-chan data.Stream {
...
 	switch category {
 	case path.ContactsCategory:
-		return ac.RetrieveContactDataForUser, serializeAndStreamContact
+		return ac.Contacts().RetrieveContactDataForUser, serializeAndStreamContact
 	case path.EventsCategory:
-		return ac.RetrieveEventDataForUser, serializeAndStreamEvent
+		return ac.Events().RetrieveEventDataForUser, serializeAndStreamEvent
 	case path.EmailCategory:
-		return ac.RetrieveMessageDataForUser, serializeAndStreamMessage
+		return ac.Mail().RetrieveMessageDataForUser, serializeAndStreamMessage
```
dadams39, line 149: I imagine a few PRs away, there is another one where this becomes an interface and it is simply `RetrieveDataForUser`?

src/internal/connector/exchange/folder_resolver_test.go
```diff
@@ -50,14 +50,15 @@ func (suite *CacheResolverSuite) TestPopulate() {
...
+			enumer: ac.Events(),
 		}
 	}
 
 	contactFunc := func(t *testing.T) graph.ContainerResolver {
 		return &contactFolderCache{
 			userID: tester.M365UserID(t),
-			ac:     ac,
+			enumer: ac.Contacts(),
+			getter: ac.Contacts(),
```
dadams39, line 61: Is this a typo? These are both the same.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -16,6 +15,13 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
+type addedAndRemovedItemIDsGetter interface {
+	GetAddedAndRemovedItemIDs(
```
dadams39, line 19: `GetItemIDsAddedAndRemoved`

src/internal/connector/exchange/service_iterators.go
```diff
@@ -89,7 +93,7 @@ func filterContainersAndFillCollections(
 			}
 		}
 
-		added, removed, newDelta, err := getJobs(ctx, qp.ResourceOwner, cID, prevDelta)
+		added, removed, newDelta, err := getter.GetAddedAndRemovedItemIDs(ctx, qp.ResourceOwner, cID, prevDelta)
```
dadams39, line 96: There is a slight stutter here: `getter.Get...`. Refactor could be `getter.ItemIDsAddedAndRemoved()`

dadams39 (approved):
Nits: The changes / suggestions have to do with refactoring for the sake of the API. No observable security concerns.

src/internal/connector/exchange/api/api.go
ryanfkeepers: The comment would be, succinctly, `interface compliance wrappers`.  Not much more to tell than that.

> It's also unclear why there wouldn't be similar calls for onedrive or sharepoint.

Note the package location.  onedrive and sharepoint have no api packages as of yet.

src/internal/connector/exchange/event_calendar_cache.go
ryanfkeepers: Yeah, i like that better.

src/internal/connector/exchange/exchange_data_collection.go
ryanfkeepers: Most certainly!

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Kinda sixes here.

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: If the stutter were a concern, in this case it'd be better to rename the interface than the func.  Due to golang's duck typing, there can be many interfaces that a given struct may fulfill; it's more sane to have each interface avoid stutters than to burden the source func with that worry.

PR #2020, opened 2023-01-02
test service_iterators with mocks

ryanfkeepers:
## Description

Now that we have a mocked api layer in place, we can test filterContainersAndFillCollections without
integration dependencies.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #1967

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -103,7 +103,7 @@ func filterContainersAndFillCollections(
 
 			// race conditions happen, containers might get deleted while
 			// this process is in flight.  If that happens, force the collection
-			// to reset which will prevent any old items from being retained in
+			// to reset. This prevents any old items from being retained in
 			// storage.  If the container (or its children) are sill missing
 			// on the next backup, they'll get tombstoned.
 			newDelta = api.DeltaUpdate{Reset: true}
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -95,18 +95,18 @@ func filterContainersAndFillCollections(
 
 		added, removed, newDelta, err := getter.GetAddedAndRemovedItemIDs(ctx, qp.ResourceOwner, cID, prevDelta)
 		if err != nil {
+			// note == nil check; only catches non-inFlight error cases.
 			if graph.IsErrDeletedInFlight(err) == nil {
 				errs = support.WrapAndAppend(qp.ResourceOwner, err, errs)
-			} else {
-				// race conditions happen, containers might get deleted while
-				// this process is in flight.  If that happens, force the collection
-				// to reset which will prevent any old items from being retained in
-				// storage.  If the container (or its children) are sill missing
-				// on the next backup, they'll get tombstoned.
-				newDelta = api.DeltaUpdate{Reset: true}
+				continue
 			}
 
-			continue
+			// race conditions happen, containers might get deleted while
+			// this process is in flight.  If that happens, force the collection
+			// to reset which will prevent any old items from being retained in
+			// storage.  If the container (or its children) are sill missing
+			// on the next backup, they'll get tombstoned.
+			newDelta = api.DeltaUpdate{Reset: true}
 		}
 
 		if len(newDelta.URL) > 0 {
--- a/src/internal/connector/exchange/service_iterators_test.go
+++ b/src/internal/connector/exchange/service_iterators_test.go
@@ -382,26 +382,26 @@ func (suite *ServiceIteratorsSuite) TestFilterContainersAndFillCollections_incre
 		}
 	)
 
-	prevPath := func(t *testing.T, at string) path.Path {
+	prevPath := func(t *testing.T, at ...string) path.Path {
 		p, err := path.Builder{}.
-			Append(at).
+			Append(at...).
 			ToDataLayerExchangePathForCategory(tenantID, userID, cat, false)
 		require.NoError(t, err)
 
 		return p
 	}
 
+	type endState struct {
+		state      data.CollectionState
+		doNotMerge bool
+	}
+
 	table := []struct {
-		name                  string
-		getter                mockGetter
-		resolver              graph.ContainerResolver
-		dps                   DeltaPaths
-		expectNewColls        int
-		expectNotMovedColls   int
-		expectMovedColls      int
-		expectDeletedColls    int
-		expectMetadataColls   int
-		expectDoNotMergeColls int
+		name     string
+		getter   mockGetter
+		resolver graph.ContainerResolver
+		dps      DeltaPaths
+		expect   map[string]endState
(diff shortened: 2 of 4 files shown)


src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,651 @@
...
+
+		return p
+	}
+
+	table := []struct {
+		name                  string
+		getter                mockGetter
+		resolver              graph.ContainerResolver
+		dps                   DeltaPaths
+		expectNewColls        int
```
ryanfkeepers, line 399: note: planning a small refactor that replaces a baseline count of statuses with ensuring the expected collection contains the expected status.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -95,18 +95,18 @@ func filterContainersAndFillCollections(
 
 		added, removed, newDelta, err := getter.GetAddedAndRemovedItemIDs(ctx, qp.ResourceOwner, cID, prevDelta)
 		if err != nil {
+			// note == nil check; only catches non-inFlight error cases.
```
ashmrtn, line 98: not sure I understand this comment. Given we got an error doesn't that mean the request was inflight? Maybe we're thinking of different spans of time?

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,648 @@
...
+	"github.com/alcionai/corso/src/internal/tester"
+	"github.com/alcionai/corso/src/pkg/account"
+	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/path"
+	"github.com/alcionai/corso/src/pkg/selectors"
+)
+
+// ---------------------------------------------------------------------------
+// mocks
+// ---------------------------------------------------------------------------
```
dadams39, line 26: Get the feeling that these (mocks) would be better in their own file.

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,648 @@
...
+			},
+		},
+		{
+			name: "a little bit of everything",
+			getter: map[string]mockGetterResults{
+				"1": commonResults,  // new
+				"2": commonResults,  // notMoved
+				"3": commonResults,  // moved
+				"4": expiredResults, // moved
+				// "5" gets deleted
```
dadams39, line 547:
```suggestion
				// "5" is deleted
```

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,648 @@
...
+
+	return results.added, results.removed, results.newDelta, results.err
+}
+
+var (
+	_ graph.ContainerResolver = &mockResolver{}
+	_ graph.CachedContainer   = &mockResolvedContainer{}
+)
+
+type (
```
ashmrtn, line 62: can you add a note for this that it could probably be merged with what's in the container resolver tests if we decide we want to do so

dadams39 (approved):
A few nits

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,648 @@
...
+				"5": DeltaPath{
+					delta: "old_delta_url",
+					path:  prevPath(suite.T(), "5", "deleted").String(),
+				},
+			},
+			expect: map[string]endState{
+				"1": {data.NewState, false},
+				"2": {data.NotMovedState, false},
+				"3": {data.MovedState, false},
+				"4": {data.MovedState, true},
```
ashmrtn, line 593: shouldn't this be false unless the token expired?

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -0,0 +1,648 @@
...
+					delta: "old_delta_url",
+					path:  prevPath(suite.T(), "1", "deleted").String(),
+				},
+			},
+			expect: map[string]endState{
+				"1": {data.DeletedState, false},
+			},
+		},
+		{
+			name: "one deleted, one new",
```
ashmrtn, line 476: I know there's already a lot of test cases here, but I think two more (very similar) ones would be good to have:
* delete folder a, make new folder with same path
* move folder a, make new folder with same path as old location of a

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: eh, it's a semantics thing.  `IsErrDeletedInFlight` will return nil if the error doesn't match the sentinel.  This is because the call may wrap the error if it isn't the expected underlying struct, for future errors.As checks.  I might revisit this, as it is confusing (like you're pointing out), but it's sufficient for now.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -95,18 +95,18 @@ func filterContainersAndFillCollections(
...
-				// to reset which will prevent any old items from being retained in
-				// storage.  If the container (or its children) are sill missing
-				// on the next backup, they'll get tombstoned.
-				newDelta = api.DeltaUpdate{Reset: true}
+				continue
 			}
 
-			continue
+			// race conditions happen, containers might get deleted while
+			// this process is in flight.  If that happens, force the collection
```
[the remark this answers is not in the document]
ryanfkeepers: mmm, disagree.  "containers might get deleted while process is in flight" is more stilted.

src/internal/connector/exchange/service_iterators.go
```diff
@@ -95,18 +95,18 @@ func filterContainersAndFillCollections(
...
+			// to reset which will prevent any old items from being retained in
+			// storage.  If the container (or its children) are sill missing
+			// on the next backup, they'll get tombstoned.
```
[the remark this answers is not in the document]
ryanfkeepers: I'm not particularly concerned.  If the reader is this deep in the code they're going to get familiar with it pretty quick.

src/internal/connector/exchange/service_iterators_test.go
ryanfkeepers: This is the only location where they're used.  Migrating them to a separate package is a premature optimization.

src/internal/connector/exchange/service_iterators_test.go
ryanfkeepers: I'll do you one better: they're now merged.

src/internal/connector/exchange/service_iterators_test.go
ryanfkeepers: note line 546 - `"4": expiredResults`

src/internal/connector/exchange/service_iterators_test.go
ryanfkeepers: good ideas!

Issue #2031, opened 2023-01-04
GC: Reduce Imports: Use Adapter for serialization


dadams39:
Reduce the total number of imports and use the serialization adapter within the adapter

PR #2032, opened 2023-01-04
GC:  Framework to use adapter's serialization writer.

dadams39:
## Description
The MS Graph's adapter contains a serialization writer within it. Create a framework to use this writer rather than creating the object within serialization workflows. The writer library was used separately when the package still had many bugs and the most current version was required. This does not change any of the writer instances within the package. Will change out the writer instances in follow-up PRs.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

*related to  #2031 <issue>

## Test Plan
- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -3,7 +3,9 @@ package graph
 import (
 	"context"
 
+	absser "github.com/microsoft/kiota-abstractions-go/serialization"
 	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
+	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/path"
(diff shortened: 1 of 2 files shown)


src/internal/connector/graph/service.go
```diff
@@ -43,6 +45,23 @@ func (s Service) Client() *msgraphsdk.GraphServiceClient {
 	return s.client
 }
 
+// Seraialize writes an M365 parsable object into a byte array using the built-in
+// application/json writer within the adapter.
+func (s Service) Serialize(object absser.Parsable) ([]byte, error) {
+	writer, err := s.adapter.GetSerializationWriterFactory().GetSerializationWriter("application/json")
+	if err != nil || writer == nil {
+		errorMessage := "service unable to create json serialization writer"
+		return nil, errors.Wrap(err, errorMessage)
```
ryanfkeepers, line 54:
```suggestion
		return nil, errors.Wrap(err, "creating json serialization writer")
```

src/internal/connector/graph/service.go
```diff
@@ -43,6 +45,23 @@ func (s Service) Client() *msgraphsdk.GraphServiceClient {
...
+	}
+
+	err = writer.WriteObjectValue("", object)
+	if err != nil {
+		return nil, errors.Wrap(err, "writeObjectValue call failed during Service.Serialize call")
```
ryanfkeepers, line 59:
```suggestion
		return nil, errors.Wrap(err, "writeObjecValue serialization")
```

ryanfkeepers (approved):

PR #2036, opened 2023-01-05
remove DiscreteScopes() from selectors

ryanfkeepers:
## Description

DiscreteScopes is a vestigial func from when scopes contained the list of resource owners to track.  That behavior is no longer in use.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1617

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -208,7 +208,7 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	}
 
 	// should always be 1, since backups are 1:1 with resourceOwners.
-	opStats.resourceCount++
+	opStats.resourceCount = 1
 	opStats.started = true
 	opStats.gc = gc.AwaitStatus()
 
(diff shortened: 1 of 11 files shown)


src/internal/operations/backup.go
```diff
@@ -201,8 +201,8 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		return opStats.writeErr
 	}
 
-	// TODO: should always be 1, since backups are 1:1 with resourceOwners now.
-	opStats.resourceCount = len(data.ResourceOwnerSet(cs))
+	// should always be 1, since backups are 1:1 with resourceOwners.
+	opStats.resourceCount++
```
ashmrtn, line 205: nit: is there ever a time we'll need to call `BackupOp.Run()` multiple times and have resourceCount > 1? Otherwise it seems easier to grok if it's just `opStats.resourceCount = 1` since readers won't need to know the previous value of `opStats.resourceCount`

ashmrtn (approved):

src/internal/operations/backup.go
ryanfkeepers: Either way works for now.  We should probably clean up that tracking to match current behavior in general.

PR #2055, opened 2023-01-07
Don't use OwnersCats struct

ashmrtn:
## Description

Remove OwnersCats so only the Reason struct or tags pass information
between BackupOp and kopia

Instead of having a separate struct (OwnersCats) to fetch previous
snapshots, generate and use reasons. While this results in some repeated
data, it cuts down on the number of distinct structs and simplifies some
of the code for getting previous manifests.

A future PR should create a shared function to create a service/cat tag
given a reason.

Only pass in a set of tags to BackupCollections. This pushes the onus
of generating the tags for later snapshot lookups to BackupOp and
creates a somewhat asymmetric interface as Reason is used for the lookup
but tags is used for the backup. This will be updated later so that both
paths use a common function to convert from Reason->tags.

Despite that, it may result in a cleaner interface with kopia (depending
on how far we want to push it) where tags become the main mean of
communication.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1916 

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -305,8 +305,8 @@ func fetchPrevSnapshotManifests(
 
 		// If we found more recent snapshots then add them.
 		for _, m := range found {
-			found := mans[m.ID]
-			if found == nil {
+			man := mans[m.ID]
+			if man == nil {
 				mans[m.ID] = m
 				continue
 			}
@@ -314,11 +314,11 @@ func fetchPrevSnapshotManifests(
 			// If the manifest already exists and it's incomplete then we should
 			// merge the reasons for consistency. This will become easier to handle
 			// once we update how checkpoint manifests are tagged.
-			if len(found.IncompleteReason) == 0 {
+			if len(man.IncompleteReason) == 0 {
 				continue
 			}
 
-			found.Reasons = append(found.Reasons, m.Reasons...)
+			man.Reasons = append(man.Reasons, m.Reasons...)
 		}
 	}
 
--- a/src/internal/kopia/snapshot_manager.go
+++ b/src/internal/kopia/snapshot_manager.go
@@ -89,6 +89,9 @@ func MakeTagKV(k string) (string, string) {
 // tagsFromStrings returns a map[string]string with tags for all ownersCats
 // passed in. Currently uses placeholder values for each tag because there can
 // be multiple instances of resource owners and categories in a single snapshot.
+// TODO(ashmrtn): Remove in future PR.
+//
+//nolint:unused
 func tagsFromStrings(oc *OwnersCats) map[string]string {
 	if oc == nil {
 		return map[string]string{}
@@ -188,24 +191,17 @@ func fetchPrevManifests(
 	ctx context.Context,
 	sm snapshotManager,
 	foundMans map[manifest.ID]*ManifestEntry,
-	serviceCat ServiceCat,
-	resourceOwner string,
+	reason Reason,
 	tags map[string]string,
 ) ([]*ManifestEntry, error) {
 	tags = normalizeTagKVs(tags)
-	serviceCatKey, _ := MakeServiceCat(serviceCat.Service, serviceCat.Category)
+	serviceCatKey, _ := MakeServiceCat(reason.Service, reason.Category)
 	allTags := normalizeTagKVs(map[string]string{
-		serviceCatKey: "",
-		resourceOwner: "",
+		serviceCatKey:        "",
+		reason.ResourceOwner: "",
 	})
 
 	maps.Copy(allTags, tags)
-
-	reason := Reason{
-		ResourceOwner: resourceOwner,
-		Service:       serviceCat.Service,
-		Category:      serviceCat.Category,
-	}
 
 	metas, err := sm.FindManifests(ctx, allTags)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


src/internal/kopia/snapshot_manager.go
```diff
@@ -275,59 +271,54 @@ func fetchPrevManifests(
...
-
-				// Snapshot can still complete fine, just not as efficient.
+
+			// Snapshot can still complete fine, just not as efficient.
+			continue
+		}
+
+		// If we found more recent snapshots then add them.
+		for _, m := range found {
+			found := mans[m.ID]
```
ryanfkeepers, line 308: I know this is movement of old code, but it just feels weird to re-use the ranged variable here.
```suggestion
		for _, m := range found {
			man := mans[m.ID]
```

src/internal/operations/backup.go
```diff
@@ -335,28 +339,28 @@ func collectMetadata(
...
 
 	for _, sl := range [][]path.CategoryType{pcs.Includes, pcs.Filters} {
 		for _, cat := range sl {
-			k, v := kopia.MakeServiceCat(service, cat)
-			oc.ServiceCats[k] = v
+			reasons = append(reasons, kopia.Reason{
+				ResourceOwner: sel.DiscreteOwner,
+				Service:       service,
+				Category:      cat,
+			})
```
ryanfkeepers, line 359: Thinking out loud: is there ever a time where the list of reasons will have different owners or services?  It feels like a []Reasons is more accurately represented as something like:
```go
type Reason struct {
  ResourceOwner string
  Service path.ServiceType
  Categories []path.CategoryType
}

File: src/internal/operations/backup.go [hunk 3]
```diff
@@ -335,28 +339,28 @@ func collectMetadata(
...
 
 	pcs, err := sel.PathCategories()
 	if err != nil {
-		return &kopia.OwnersCats{}
+		// This is technically safe, it's just that the resulting backup won't be
+		// usable as a base for future incremental backups.
+		return nil
 	}
 
 	for _, sl := range [][]path.CategoryType{pcs.Includes, pcs.Filters} {
```
ryanfkeepers, line 353: Cat could be duplicated across Includes and Filters, yeah?  Does this need to dedupe?

ryanfkeepers (approved):
Some nits/discussions.

Side topic: the `Reason` struct started out with that name for meaning "the reason this manifest is included".  I was pondering over the weekend that spreading the struct around makes it a little opaque relative to the original meaning.  Not pushing hard for a rename, but if it resonates, thoughts on something like `Basis`?

src/internal/operations/backup.go
ashmrtn: >is there ever a time where the list of reasons will have different owners or services?

Not at the moment and we don't seem to have anything like that planned

However, after going through iterations of code with `OwnersCats` (which was similar to what you've proposed) and `Reason`, I kind of like the simplicity of the `Reason` struct. There's no gotchas that need to be known about when iterating through them and it only requires a single for-loop instead of a double for-loop. It also better matches the unit of granularity we need to use for delta token-base incrementals: the (resource owner, service, category) tuple. We're restricted to checking this tuple because that's the granularity that the CLI and selectors allow when making backups

src/internal/operations/backup.go
ryanfkeepers: > Something like backup create --user "*,user1" which would go to selectors for all users and then user1 again. That would eventually cause two backups for user1 to be created

Any case that involves a combined declaration of a wildcard and a static value is already handled.  `*,foo` will get reduced to `*`.

But that assumes a single declaration.  If we get something like:
```go
Include(
  MailFolders(Any()),
  MailFolders("Inbox"),
)
```
then we'll get duplicate collections on the inbox pathing.  IE: the cli is safe from this, but not the sdk.  This could be handled in either selectors or collections, depending on where we want to put the mitigation.

src/internal/operations/backup.go
ryanfkeepers: > only requires a single for-loop instead of a double for-loop

I was imagining this as only passing around a single Reason, not multiples.  The only loop would be over the categories.

Per granularity along the tuple and existence of gotchas, I do think there's a possibility.  Within an operation, we organize data into owner+service silos.  There is no case where we intend to handle multiple users/services at a time.  Multiple reasons with independent declarations provide handling as if that were possible.

ashmrtn: >Multiple reasons with independent declarations provide handling as if that were possible.

I guess I don't see that a drawback really 😅 If we want to ensure it doesn't happen we can just add a validation check to ensure there's only one service/resource owner

> I was imagining this as only passing around a single Reason, not multiples.

that sounds pretty similar to `OwnersCats`. The `snapshot_manager` still needs to break it down into individual (resource owner, service, category) tuples because the CLI still allows that granularity. That means we may have different bases for something like `backup create exchange --data email,contacts` where each selected base will have some (possibly complete) subset of the categories that were sourced from the original

ryanfkeepers: > I guess I don't see that a drawback really
That's fair.  More of a talking point than a concern, really.  All other points are good.  Thanks for the chat.

Issue #2062, opened 2023-01-09
[Bug]: Backup ID is not checked for existence in manifest.


ryanfkeepers:
### What happened?

Manifests can be malformed in two ways: 1/ the manifest exists but has no backupID or the id references non-existing storage.  2/ the manifest and backup exists, but has no detailsID  or the id references non-existing storage.

Malformed manifests can still be utilized for kopia-assisted incrementals (ie: avoids re-downloading existing items), but cannot be used to merge historical backup or details data into the next snapshot. 

As a stop-gap to prevent failures, manifests should be evaluated for correctness.  If malformed, the backup should only be run in kopia-assisted state.

### Corso Version?

Corso v0.0.4

### Where are you running Corso?

mac, linux, s3

PR #2063, opened 2023-01-09
protect against missing backups in manifests

ryanfkeepers:
## Description

If a prior manifest is missing a backup, it should be handled the same way as when a manifest's backup is missing a details ID: the metadata is skipped, and a full backup is performed.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #2062

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -344,7 +344,7 @@ func produceManifestsAndMetadata(
 
 		bID := man.Tags[tk]
 		if len(bID) == 0 {
-			return nil, nil, false, errors.New("manifest missing backup ID")
+			return nil, nil, false, errors.New("snapshot manifest missing backup ID")
 		}
 
 		dID, _, err := sw.GetDetailsIDFromBackupID(ctx, model.StableID(bID))
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -151,7 +151,7 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


src/internal/operations/backup.go
```diff
@@ -151,7 +151,7 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		}
 	}()
 
-	mans, mdColls, err := produceManifestsAndMetadata(ctx, op.kopia, op.store, oc, tenantID, uib)
+	mans, mdColls, canUseMetaData, err := produceManifestsAndMetadata(ctx, op.kopia, op.store, oc, tenantID, uib)
```
ashmrtn, line 154: this works for now, but we're probably going to want something more granular in the future

src/internal/operations/backup.go
```diff
@@ -344,12 +344,22 @@ func produceManifestsAndMetadata(
 
 		bID := man.Tags[tk]
 		if len(bID) == 0 {
-			return nil, nil, errors.New("missing backup id in prior manifest")
+			return nil, nil, false, errors.New("manifest missing backup ID")
```
ashmrtn, line 347:
```suggestion
			return nil, nil, false, errors.New("snapshot manifest missing backup ID")
```

ashmrtn (approved):
can you make tickets for the following please:
* unit tests for `produceManifestsAndMetadata`
* fine-grained control over which bases are used for delta token-based incrementals so that we be as efficient as possible

src/internal/operations/backup.go
ryanfkeepers: agreed

Issue #1877, opened 2022-12-20
Mark backups that have item data but not details as failed


ashmrtn:
Incremental backups need to find a set of one or more base snapshots to source metadata (delta tokens etc) and some subset of ItemInfo entries from. Internally, Corso also breaks up making a backup into two discrete phases of persisting data: first making a snapshot of the item data and then a snapshot of the details entries. This means that there are situations in which Corso could persist the item data snapshot but not the backup details snapshot, leaving things in an unclean state and making the data in the backup inaccessible

Backups that end up in this situation should be marked as failed or have some "repair" command to fix them

- [x] denote these backups as failed or otherwise make them unable to run `backup details` and `restore`
- [x] #1878

ashmrtn:
partial backups should be marked as failed and trying to list the details or restore from them should return an error

PR #2066, opened 2023-01-09
details and restore exit on missing backup

ryanfkeepers:
## Description

Ensures that the details and restore commands
exit without attempting further processing if a
backupID doesn't point to either a valid backup
or backup details, and that the user is appropriately notified.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1877

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/cli/restore/exchange.go
+++ b/src/cli/restore/exchange.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/alcionai/corso/src/cli/print"
 	"github.com/alcionai/corso/src/cli/utils"
 	"github.com/alcionai/corso/src/internal/common"
+	"github.com/alcionai/corso/src/internal/kopia"
 	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/repository"
 )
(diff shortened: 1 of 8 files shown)


src/cli/restore/exchange.go
```diff
@@ -215,23 +216,22 @@ func restoreExchangeCmd(cmd *cobra.Command, args []string) error {
...
-
-	ro, err := r.NewRestore(ctx, backupID, sel.Selector, restoreDest)
+	ro, err := r.NewRestore(ctx, backupID, sel.Selector, dest)
 	if err != nil {
 		return Only(ctx, errors.Wrap(err, "Failed to initialize Exchange restore"))
 	}
 
 	ds, err := ro.Run(ctx)
 	if err != nil {
+		if errors.Is(err, kopia.ErrNotFound) {
```
ashmrtn, line 231: this makes me pause a bit and feels like a code smell. I know that it means we're missing a backup or backup details, but it really feels like we're pulling in info from too low a layer just to say that. It feels like this message should be generated at the layer that actually handles the error and passed back up rather than pulling in the bottom layer just to check (and having the assumption that this layer knows what the bottom layer returns error-wise for this)

ashmrtn (approved):

src/cli/restore/exchange.go
ryanfkeepers: > should be generated at the layer that actually handles the error and passed back up

I'm heavily against this pattern.  It's effectively the same as saying, "the lowest level error should decide the returned http code".  Seems like nice qol, but ultimately ends up coupling layers together by binding concerns across interfaces.  

What I think you mean to point out by noting a smell is that we're slicing the source error details to present a less detailed, but more easily observed, error message, and that this behavior reduces insight and debugging capacity.  That I absolutely agree with.  The ideal solution is to diverge end-user presentation from granular detail. That way we can keep presentation more coarse ("that backup id is unusable"), and not lose debugging insights ("the backup id doesn't have details data").  But ownership of that divergence should be at the cli/api layer, not at the error production.

ashmrtn: >ultimately ends up coupling layers together by binding concerns across interfaces.

I'm a bit confused by this statement. Isn't the current code doing the same though? It's effectively saying that stuff in the kopia package can't change their errors or they will break the CLI

Given that the CLI is mostly a wrapper around the SDK layer, it seems like maybe the SDK should be dealing with this (maybe that will be rolled into error management/logging?)

ryanfkeepers: > It's effectively saying that stuff in the kopia package can't change their errors or they will break the CLI

Not quite the same.  It's saying that if lower level packages change their errors, then the CLI might produce a less accurate message, being unable to specifically correlate that error to a sentinel.  Is it similar?  Yes.  But it's a lot more resilient to actual breaking, and keeps concerns appropriately separated that way.  

> it seems like maybe the SDK should be dealing with this

If by "dealing with" you mean "interpreting errors", I don't think that's the best choice.  Outside of, perhaps, organizing some is/as checks to be more easily consumable, I think we're meeting the needs of the sdk well enough.  Further interpretation/transformation feels like middleware for its own sake.

ashmrtn: mmmm alright. At this point it feels a bit like we're bikeshedding on encapsulation/information hiding at any rate

vkamra: Thoughts on just adding a call to `getBackupAndDetailsFromID` before the operation is Run and if that fails - returning the error below?

ryanfkeepers: Doesn't change much about the error handling, ultimately.  Plus it duplicates validation that we're already expecting `Run()` to cover.  I'd rather the runner maintain its own validation requirements as a centralized precheck than spread those requirements out to callers piecemeal.

Issue #2076, opened 2023-01-10
Log to disk by default


meain:
Currents logs are pushed to stdout by default. This clashes with the progressbar as well. Also, when logging to file we can add more info than just the bare minimum as the user does not see it.

### TBD
- Default log file location (`$PWD/corso-<datetime>.log` or `$HOME/.corso/<datetime>.log` or `/var/log/corso/<datetime>.log`)
  - Have to consider Windows default location
- CLI flag to switch log file location (`--log-file /tmp/corso.log`)
- Flag to switch logging to stdout (`--log-file -`)

PR #2082, opened 2023-01-10
Write logs to disk by default

meain:
## Description

This changes the behavior of logs and writes it to disk instead of `stdout` by default. It also adds a new flag `--log-file` to specify the filename for the log file.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/2076

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/logger/logger.go
```diff
@@ -3,20 +3,26 @@ package logger
...
 	"os"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+var defaultLogLocation = "corso-" + time.Now().UTC().Format("2006-01-02T15-04-05Z") + ".log"
```
meain, line 14: Default log location is currently `$PWD/corso-<timestamp>.log`.

src/pkg/logger/logger.go
```diff
@@ -54,34 +65,48 @@ func AddLogLevelFlag(cmd *cobra.Command) {
...
+	}
+
+	// retrieve the user's preferred log file location
+	// automatically defaults to default log location
+	logfile, err := fs.GetString(logFileFN)
+	if err != nil {
+		return "info", defaultLogLocation
+	}
+
+	if logfile == "-" {
```
meain, line 102: They can still use cobra specific `stdout`, but this is a more common convention.

src/pkg/logger/logger.go
ryanfkeepers, line 14:
```suggestion
var defaultLogLocation = "corso-" + common.FormatNow(common.SimpleDateTime) + ".log"
```

ryanfkeepers: Works for me.  @gmatev you happy with this?  Alternatively, we could co-locate it with the config file location by default.  Or perhaps a ~/.corso folder.

src/pkg/logger/logger.go
```diff
@@ -54,34 +65,48 @@ func AddLogLevelFlag(cmd *cobra.Command) {
...
+		logfile = "stdout"
```
ryanfkeepers, line 103: This should be `stderr`, not stdout.  Stdout will cause logging to interfere with end-of-run application prints, preventing users from piping the results to scripts (ex: `corso ... --json | jq {...}`).

ryanfkeepers (commented):
some concerns about the writer itself, otherwise looks good.

ryanfkeepers (approved):
Approving to unblock this.  Prior comments still need to be addressed.

src/pkg/logger/logger.go
meain: FYI, if we are changing to a different location, we will also have to consider what it will be in Windows.

meain: Another option would be stick to [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) in linux and store it under `$XDG_STATE_HOME` which would be `$HOME/.loca/state/corso/corso-<timestamp>.log`.

As a follow-up, we should probably consider moving the config file to `$XDG_CONFIG_HOME`, which would be `$HOME/.config/corso/config.toml`.

There is a relatively  vocal group of people(myself kinda included) advocating against polluting home directory on linux: https://www.reddit.com/r/linux/comments/971m0z/im_tired_of_folders_littering_my_home_directory/

vkamra: Agree that this should not be home directory or local directory. Suggest we emulate https://kopia.io/docs/advanced/logging/

meain: `:` in `common.SimpleDateTime` is an invalid char for files on Windows. This format is similar to that but `:` replaced with `-`. I can create a new constant if you think that is necessary.

src/pkg/logger/logger.go
meain: This is more of a extra option as `-` is a common convention for `stdin` / `stdout`. If a user wants to pipe to `stderr` they can use "stderr" as the filename.

ntolia (dismissed):
Please also:
- Modify docs/support/bugs-and-features/ to include information on where to find the log file to share
- Modify docs/setup/configuration/ and add a new section (Log files) and specify the default location as well as a reference to the flag.

vkamra (approved):

website/docs/setup/configuration.md
```diff
@@ -126,3 +126,17 @@ directory within the container.
...
+The location of log files varies by operating system:
+
+* On Linux - `~/.cache/corso/logs/<timestamp>.log`
+* On macOS - `~/Library/Logs/corso/logs/<timestamp>.log`
+* On Windows - `%LocalAppData%\corso/logs/<timestamp>.log`
+
+Log file location can be overridden by setting the `--log-file` flag.
+
+:::info
+You can use `stdout` or `stderr` as the `--log-file` location to redirect the logs to "stdout" and "stderr" respectively.
```
vkamra, line 141: Instead of `You can use ...` - change to `Use `stdout` ...`

vkamra: So "Use `stdout` or `stderr` as the `--log-file` location to redirect the logs to "stdout" or "stderr".

PR #2087, opened 2023-01-10
Refactor the Reason struct and how tags for lookups are generated

ashmrtn:
## Description

Solidify the interface between BackupOp and KopiaWrapper by making Reason the de facto way to pass/generate tags for things. The Reason struct now includes a function to generate tags for that instance. KopiaWrapper also now hides the fact that it prefixes tags from other components

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1916 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn:
totally fine with changing the name of the package that holds the Reason struct. I couldn't come up with something good since `communication` makes it sound like networking stuff

src/internal/operations/backup_test.go
```diff
@@ -276,14 +277,20 @@ func makeDetailsEntry(
 	return res
 }
 
+// TODO(ashmrtn): This should belong to some code that lives in the kopia
+// package that is only compiled when running tests.
```
ryanfkeepers, line 281: So long as it isn't referenced in a non-test file, the func is, by default, not compiled into the release binary.  If you want to additionally identify/enforce its purpose so that it doesn't accidentally get referenced, you could take in a testing.T param and call t.Helper().

ashmrtn: it really feels like we should be using build constraints/tags or something for stuff like that rather than requiring unused parameters

ryanfkeepers: I'm not entirely a fan of the unused parameter either, it just makes the purpose more obvious _if_ it's deemed to confusing/available.  Really though, it should be easy to catch ourselves borrowing from test helpers in live code.  For that matter, it shouldn't take much effort of any kind (besides common sense review) to keep the compiled binary clean.

ashmrtn:
I think we're bikeshedding here a bit so I'll go ahead and move this back into the kopia package for now. However, I've left some thoughts on this below

>Likewise, I don't think this package should live on it's own. The purpose of the struct is very targeted to kopia's method of selecting a manifest (again, the "reason" for choosing it)

I'm not really sure I agree here. Tags seem like a perfectly reasonable and generic way to select previous backups (or snapshots) for a tuple without having to already know the backup ID. While the kopia package certainly adds some to this so that tags can be passed when using the kopia CLI for debugging, the concept of tags as a way to select things could be used if we had a different storage system

>it's now vying for identification ownership in a space that's already strongly covered (contented for, even) between Selectors and Paths.

I think to some extent it's alright to have similar groups of information if they make sense in the context they're used. Part of the importance of this is allowing modularity and information hiding. If the info isn't readily available for something, then it makes people think about possible expansions they may want to make and why what they're trying to do may/may not make sense in the broader picture. This also helps out when testing things, especially if some of the groupings are simple structs or interfaces

Selectors and Paths both allow for communicating a lot of additional information that frankly isn't needed when denoting why a snapshot was selected as a base. In fact, Paths can't actually represent this information without adding additional filler values. Though that may make Paths somewhat more difficult to use, it does allow us to ensure that each Path has all the required values so consumers of the package can use it without manually checking everything

ashmrtn:
actually, I guess I'll leave it in the kopia package. I don't want to make the function for getting manifests from kopia public and wrapper.go needs access to that function

ryanfkeepers (approved):

Issue #2086, opened 2023-01-10
CI:  Failure to Checkout with Beta Package Added


dadams39:
Failure during [CI](https://github.com/alcionai/corso/actions/runs/3883931965/jobs/6626321063)
This could be related to memory in the CI or another issue. 

### Blocks

- #2069 
- #2071 
- #2084

dadams39:
Blocker [Upstream-Graph-SDK-129](https://github.com/microsoftgraph/msgraph-sdk-go/issues/129)

dadams39:
Temporarily, we can move in the two files from `Beta`. We will need to solve the problem of compilation with Beta to use the API for backups and restore.

dadams39:
Suggestion from the comments is to increase the swap space on the runner:
```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - name: Set Swap Space
        uses: pierotofy/set-swap-space@master
        with:
          swap-size-gb: 12
```

Need to verify if changes to the configuration will allow for the fix.

dadams39:
Setting the swap space to 15 gb resulted in:
```
Memory and swap:
               total        used        free      shared  buff/cache   available
Mem:           6.8Gi       526Mi       5.2Gi       7.0Mi       1.1Gi       6.0Gi
Swap:          4.0Gi          0B       4.0Gi

NAME          TYPE SIZE USED PRIO
/mnt/swapfile file   4G   0B   -2
```

dadams39:
[Run](https://github.com/alcionai/corso/actions/runs/3960126507/jobs/6783853293) with 12 GB of Swap.
```
Run pierotofy/set-swap-space@master
Run echo "Memory and swap:"
Memory and swap:
               total        used        free      shared  buff/cache   available
Mem:           [6](https://github.com/alcionai/corso/actions/runs/3960126507/jobs/6783853293#step:3:7).8Gi       548Mi       5.1Gi       7.0Mi       1.1Gi       5.9Gi
Swap:          4.0Gi          0B       4.0Gi

NAME          TYPE SIZE USED PRIO
/mnt/swapfile file   4G   0B   -2

Run export SWAP_FILE=$(swapon --show=NAME | tail -n 1)
Setting up swapspace version 1, size = 12 GiB (12884897792 bytes)
no label, UUID=45934725-a0c4-462d-a6a4-668924b10b9e
Run echo "Memory and swap:"
Memory and swap:
[...]

PR #2088, opened 2023-01-10
Patch: Direct Import of Minimal Beta packages for Feature Development

dadams39:
## Description
Imports minimal version of `site_page.go` and `site_pageable.go` to avoid compile time failures for `msgraph-beta-sdk-go`.  

Fields that required additional `Data Types` from the `Beta` packages are commented out in `site_page.go`

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

*related to  #2086<issue>

## Test Plan

- [x] :muscle: Manual
Should fix compile time errors for the package.

### DIFF
--- a/src/internal/connector/sharepoint/site_pageable.go
+++ b/src/internal/connector/sharepoint/site_pageable.go
@@ -1,0 +1,24 @@
+package sharepoint
+
+import (
+	kioser "github.com/microsoft/kiota-abstractions-go/serialization"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+)
+
+// SitePageable adjusted from msgraph-beta-sdk-go for temporary testing
+type SitePageable interface {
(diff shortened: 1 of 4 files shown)


src/internal/connector/sharepoint/site_pageable.go
```diff
@@ -0,0 +1,24 @@
+package sharepoint
+
+import (
+	kioser "github.com/microsoft/kiota-abstractions-go/serialization"
```
ryanfkeepers, line 4: So long as we're aliasing this import (I think I'd prefer not to, but we're already doing it), we should keep the alias standard.

```suggestion
	absser "github.com/microsoft/kiota-abstractions-go/serialization"
```

ryanfkeepers (approved):

PR #2091, opened 2023-01-10
migrate, test manifest+meta producer

ryanfkeepers:
## Description

Adds mocked unit tests for produceManifestsAnd-
Metadata.  For cleanliness, moves that func, and
any funcs called within it, to their own file within operations

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #2062

## Test Plan

- [x] :zap: Unit test


src/internal/operations/manifests.go
```diff
@@ -0,0 +1,210 @@
...
+
+		bID, ok := man.GetTag(kopia.TagBackupID)
+		if !ok {
+			return nil, nil, false, errors.New("snapshot manifest missing backup ID")
+		}
+
+		dID, _, err := gdi.GetDetailsIDFromBackupID(ctx, model.StableID(bID))
+		if err != nil {
+			// if no backup exists for any of the complete manifests, we want
+			// to fall back to a complete backup.
```
ashmrtn, line 94: nit: this comment doesn't quite match the code (yes, I know this was just relocating code). The if-block says that if _any_ previous backup is missing the model/details we fallback to a full backup. If we want to make it such that _all_ previous backups need to be missing the model/details we'll need to check at the end of the loop. That's an optimization we can make in a different PR if we want

src/internal/operations/manifests_test.go
```diff
@@ -0,0 +1,612 @@
...
+		{
+			name: "fetch manifests errors",
+			mr: mockManifestRestorer{
+				mockRestorer: mockRestorer{},
+				mrErr:        assert.AnError,
+			},
+			gdi:       mockGetDetailsIDer{detailsID: did},
+			reasons:   []kopia.Reason{},
+			getMeta:   true,
+			assertErr: assert.Error,
```
ashmrtn, line 465: I'm not sure what the correct behavior here is, but we may actually want to fallback to a full backup and log the error as a warning? Otherwise I think it will stop the backup from running at all

src/internal/operations/manifests_test.go
```diff
@@ -0,0 +1,612 @@
...
+					makeMan(path.EmailCategory, "", "bid"),
+					makeMan(path.EmailCategory, "ir", ""),
+				},
+			},
+			gdi:       mockGetDetailsIDer{detailsID: did},
+			reasons:   []kopia.Reason{},
+			getMeta:   true,
+			assertErr: assert.NoError,
+			assertB:   assert.True,
+			expectDCS: []data.Collection{},
```
ashmrtn, line 542: this should produce metadata no?

src/internal/operations/manifests_test.go
```diff
@@ -0,0 +1,612 @@
...
+			mr: mockManifestRestorer{
+				mockRestorer: mockRestorer{colls: []data.Collection{}},
+				mans:         []*kopia.ManifestEntry{makeMan(path.EmailCategory, "", "bid")},
+			},
+			gdi:       mockGetDetailsIDer{detailsID: did},
+			reasons:   []kopia.Reason{},
+			getMeta:   true,
+			assertErr: assert.NoError,
+			assertB:   assert.True,
+			expectDCS: []data.Collection{},
```
ashmrtn, line 555: here too there should be data in the collection?

src/internal/operations/manifests_test.go
```diff
@@ -0,0 +1,612 @@
...
+			},
+			gdi:       mockGetDetailsIDer{detailsID: did},
+			reasons:   []kopia.Reason{},
+			getMeta:   true,
+			assertErr: assert.NoError,
+			assertB:   assert.True,
+			expectDCS: nil,
+		},
+		{
+			name: "man missing backup id",
```
ashmrtn, line 515: Could you please add a test for missing details ID as well please?

ryanfkeepers: absolutely

src/internal/operations/manifests_test.go
ryanfkeepers: They're not mocked in atm.  I'll make a second pass to handle that.  The table is a little wobbly since it assumes there's a single manifestRestorer, not one per manifest.  When I caught that I figured it'd be better to fix it in a follow up.

src/internal/operations/manifests_test.go
ryanfkeepers: I'm not sure either.  Going to leave it as-is for now, and we can revisit as needed if/when concern arises.

src/internal/operations/manifests.go
ryanfkeepers: I think you're saying the comment is correct and the behavior is wrong.  Both the comment and the code fallback on any missing backup, not if all backups are missing.

Either way, yes, punting.

src/internal/operations/manifests_test.go
ashmrtn: make a ticket for it and bring it up in our upcoming discussion (I know it's not technically tech debt, but does need to be addressed)

ryanfkeepers: Do we need a ticket just to discuss it?

ashmrtn (approved):

src/internal/operations/manifests_test.go
ashmrtn: I suppose not as long as we remember to discuss it

PR #2095, opened 2023-01-10
populate colls in manifest test

ryanfkeepers:
## Description

Adds data.collection checking to the manifests
mocked unit tests.

If this looks good, we can merge it into the parent
branch, then add both to main.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #2062

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/operations/manifests_test.go
+++ b/src/internal/operations/manifests_test.go
@@ -47,6 +47,32 @@ func (mg mockGetDetailsIDer) GetDetailsIDFromBackupID(
 	return mg.detailsID, nil, mg.err
 }
 
+type mockColl struct {
+	id    string // for comparisons
+	p     path.Path
+	prevP path.Path
+}
+
+func (mc mockColl) Items() <-chan data.Stream {
+	return nil
+}
+
+func (mc mockColl) FullPath() path.Path {
+	return mc.p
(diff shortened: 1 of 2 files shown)


src/internal/operations/manifests_test.go
```diff
@@ -607,19 +662,22 @@ func (suite *OperationsManifestsUnitSuite) TestProduceManifestsAndMetadata() {
...
+
+					if rmc.id == mc.id {
 						found = true
 						break
 					}
 				}
 
-				assert.True(t, found, "expected collection is present in results: "+s)
+				assert.True(t, found, "expected collection is present in results: "+mc.id)
 			}
```
ashmrtn, line 681: not sure if this would be any easier, but you could track the expected IDs of collections as `[]string` and then just iterate through the returned collections, add their IDs to another `[]string` and then just call `assert.ElementsMatch(t, expectedIDs, foundIDs)`

ashmrtn (approved):

src/internal/operations/manifests_test.go
ryanfkeepers: good thinking

PR #2114, opened 2023-01-11
GC: Restore: SharePoint add case for Page Category in Workflow. 

### LINKED ISSUE #2113: GC: Restore: SharePoint: Pages: Initiate workflow
- Create stubs for `SharePoint.Pages` for restore workflow.
Allow for the selectors to `path.PagesCategory` to be acknowledged as being called without affecting the overall state of the Corso's restore.

dadams39:
## Description
Fix to allow selectors to include `SharePoint.Pages` during backup without causing backup workflow to fail. 

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

*closes  #2113<issue>

## Test Plan

- [x] :muscle: Manual


src/internal/connector/sharepoint/restore.go
```diff
@@ -76,6 +76,8 @@ func RestoreCollections(
 				deets,
 				errUpdater,
 			)
+		case path.PagesCategory:
+			logger.Ctx(ctx).Infof("backup of pages not supported", dc.FullPath().Category())
```
ashmrtn, line 80: nit: I'd mark this as a warning or error level log message so it's easier to find

src/internal/connector/sharepoint/restore.go
```diff
@@ -76,6 +76,8 @@ func RestoreCollections(
 				deets,
 				errUpdater,
 			)
+		case path.PagesCategory:
```
ashmrtn, line 79: putting this here as it has nicer thread support than just comments in the issue in my opinion

I'm a little out of the loop on the SharePoint stuff, but do we expect to release with only partial SharePoint restore support? I know we'll get everything eventually, but I haven't been following this closely enough to know what the plan is right now

src/internal/connector/sharepoint/restore.go
ryanfkeepers: I vote for error level, and returning an error along with it.  Warning level logs should be avoided.

src/internal/connector/sharepoint/restore.go
dadams39: Yes. In the case of Pages, it is `unusable` until the Beta library can be utilized. Whether this is through creating a separate binary or submodule for beta, it is unclear how it would be incorporated. Lists can be downloaded and restored at this time. As for Libraries, there is a need either create a Corso workflow that would allow for items within restricted directories to be uploaded and migrated to their correct locations.

src/internal/connector/sharepoint/restore.go
dadams39: Resolved within [c25a1ae](https://github.com/alcionai/corso/pull/2114/commits/c25a1ae3c6f816e70058362e0d022369a46af28e)

src/internal/connector/sharepoint/restore.go
ashmrtn: I just a bit confused as this code is in the restore path, not the backup path. I get that we're eventually going to be pulling in the beta version to access more Graph API functionality

src/internal/connector/sharepoint/restore.go
```diff
@@ -76,6 +76,12 @@ func RestoreCollections(
...
+			errorMessage := fmt.Sprintf("backup of pages not supported %s", dc.FullPath().Category())
```
ashmrtn, line 80:
```suggestion
			errorMessage := fmt.Sprintf("restore of %s not supported", dc.FullPath().Category())
```

ashmrtn (approved):

src/internal/connector/sharepoint/restore.go
dadams39: Resolved within [26795fd](https://github.com/alcionai/corso/pull/2114/commits/26795fd6e3d2d3e9169bab64b0eb7047b7a6666b)

src/internal/connector/sharepoint/restore.go
dadams39: Title/comments oriented appropriately.

Issue #1954, opened 2022-12-23
Ensure data collections don't have repeated items


ashmrtn:
Under some conditions Graph API may return the same item multiple times while traversing the items in a container. Currently this will result in the collection pulling the data for the item multiple times and presenting the item to kopia multiple times which will lead to duplicate items in the backup. Collections should be updated to dedupe items prior to pull data and handing them to kopia

Conditions under which items can be seen multiple times usually involve beginning to traverse the container, modifying an item in the container that has already been seen, and then continuing container traversal

ashmrtn:
Received bug reports of this causing incremental backups after the one with duplicates to fail because more items are pulled from the base snapshot than expected when merging backup details. Sample error output:
```
No backups available
Error: 1 error occurred:
        * Failed to run Exchange backup for user user@domain.com: 2 errors occurred:
        * errors prevented the operation from processing
        * merging backup details: incomplete migration of backup details: found 126 of 123 expected items
```

PR #2116, opened 2023-01-11
Dedupe collection item IDs prior to fetching item data

ashmrtn:
## Description

Under some conditions, items can be returned multiple times when enumerating the items in a folder in Exchange. Begin deduping those items by doing the following:
* use a map in ExchangeDataCollection to dedupe on item IDs
* have enumeration function return a single slice of items containing both additions and deletion so a "last-change-wins" policy can be implemented

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1954 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ashmrtn:
saw a bug caused by this in the wild, confirmed through manual testing the following:
* duplicate item IDs are fetched when enumerating IDs when making an incremental
* duplicate items were only saved once in kopia
* another incremental after the one that fetched duplicate IDs succeeded with no errors (no changes were made prior to final incremental)

src/internal/connector/exchange/service_iterators.go
```diff
@@ -125,8 +125,22 @@ func filterContainersAndFillCollections(
 			newDelta.Reset)
 
 		collections[cID] = &edc
-		edc.added = append(edc.added, added...)
-		edc.removed = append(edc.removed, removed...)
+
+		// This results in "last one wins" if there's duplicate entries for an ID
```
vkamra, line 129: Are the delta records ordered? I know the `[]api.DeltaResult` is ordered but asking about the graph delta records we built that slice from?

vkamra (approved):

src/internal/connector/exchange/service_iterators.go
ashmrtn: it's unclear to me if they are or not. The documentation for OneDrive seems to imply they are and the last one should be used. I found some mentions of getting records multiple times and the comments there were to merge the changes. Since we're just dealing with whether or not an item exists it's a bit fuzzier.

I guess we could always keep additions and handle the failure when fetching the item data if we want? I don't think we have code for that at the moment though

vkamra: Going to go ahead and merge this. As we discussed offline - we should probably just switch to "delete wins" because we won't see an item with the same ID reappear after delete

PR #2142, opened 2023-01-13
Switch to "delete wins" merge strategy for Exchange IDs

### LINKED ISSUE #1954: Ensure data collections don't have repeated items

ashmrtn:
## Description

#2116 implemented a "last change wins" strategy for deduping IDs. However, it appears Graph API doesn't really enforce an ordering for returned items so that strategy may not work. Instead, implement a "delete wins" strategy.

Items that are added, deleted, and then restored in M365 will be restored with a different ID

This PR does the following:
* reverts changes to return a unified list of IDs and added/removed status
* switches to a "delete wins" merge strategy for duplicate IDs
* updates tests

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* closes #1954 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators_test.go
+++ b/src/internal/connector/exchange/service_iterators_test.go
@@ -30,7 +30,8 @@ var _ addedAndRemovedItemIDsGetter = &mockGetter{}
 type (
 	mockGetter        map[string]mockGetterResults
 	mockGetterResults struct {
-		items    []api.DeltaResult
+		added    []string
+		removed  []string
 		newDelta api.DeltaUpdate
 		err      error
 	}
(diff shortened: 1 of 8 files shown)


src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -336,143 +333,106 @@ func (suite *ServiceIteratorsSuite) TestFilterContainersAndFillCollections() {
...
-			scope:               allScope,
-			expectErr:           assert.NoError,
-			expectNewColls:      1,
-			expectMetadataColls: 1,
-			expectAdded:         map[string]struct{}{},
-			expectRemoved:       map[string]struct{}{"a1": {}},
 		},
 		{
-			name: "interleaved, final add",
+			name: "remove wins",
```
vkamra, line 394:
```suggestion
			name: "remove for same item wins",
```

vkamra (approved):

Issue #2120, opened 2023-01-12
persist metadata required for incrementals


ashmrtn:
Similar to how [Exchange](https://github.com/alcionai/corso/blob/32b11cdca83f58201dc680e760da62d6ecbab1a1/src/internal/connector/exchange/service_iterators.go#L187) creates metadata files with a `map[M365 folder ID]->folder path` and `map[M365 folder ID]->delta token`, OneDrive needs to create and add metadata files to the returned collections.

OneDrive only has a single endpoint that all items are fetched from so the delta token file will be somewhat different. The folder ID to path map is still required though and should contain all folders discovered during that backup instance

ashmrtn:
closing this, though we need better test coverage for this, especially tests that actually run against data returned by Graph API

PR #2144, opened 2023-01-14
Begin persisting OneDrive/SharePoint library metdata

ashmrtn:
## Description

Start persisting the folder path maps and delta URLs for backed up OneDrive/SharePoint drives. Delta URLs are saved in a map[drive ID]deltaURL while folder IDs are in a map[driveID]map[folder ID]folder path

Needs another patch to properly save the path for folders that match the selector, currently the selector comparison is only on the parent of an item

Later PRs can get the new folder map by taking the map from the previous backup and making changes to it when folder deletions/moves are encountered

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2120 

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/collections.go
```diff
@@ -81,27 +92,76 @@ func (c *Collections) Get(ctx context.Context) ([]data.Collection, error) {
...
+			graph.NewMetadataEntry(graph.DeltaURLsFileName, deltaURLs),
+		},
+		c.statusUpdater,
+	)
+
+	if err != nil {
+		// Technically it's safe to continue here because the logic for starting an
+		// incremental backup should eventually find that the metadata files are
+		// empty/missing and default to a full backup.
+		err = errors.Wrap(err, "making metadata collection")
```
ryanfkeepers, line 149: If we don't intend for the caller to exit on a non-nil error, this should probably just log the error instead of returning it.

ryanfkeepers (approved):

Issue #1774, opened 2022-12-12
Backup and restore OneDrive file permissions


meain:
https://github.com/alcionai/corso-roadmap/issues/13

### TODO
- [x] Prevent name collisions when another file with `.meta` suffix is present
- [x] Handle cases where children have more permissions than parent along with the nested permissins
- [x] Handle cases where children have less permissions than parent
- [ ] Consider file tree a/b, if we first give user1 access to b, then give user1 access to a and remove user1 access to a, they should retain access to b. This is not available after restore.

gmatev:
> Consider file tree a/b, if we first give user1 access to b, then give user1 access to a and remove user1 access to a, they should retain access to b. This is not available after restore.
This I think is tricky since it may have unintended side effects - I don't want the restore of item b to change the permissions of my restore target. 

The appropriate logic should be that permissions are only restored whenever Corso restores a given item or folder specifically. 

Consider the following original hierarchy:
     Folder A
        -> Folder B (shared with user 1 and user 3)
            -> item Y (shared with user 2)
        -> item X (shared with user 3)
        
 (1) Restoring Item X only
      => the resulting item, wherever it is placed, will be shared with User3

 (2) Restoring item Y only 
      => the resulting item, wherever it is placed, will be shared with User 2
      Note: that this should not change permissions on any parent folders where Item Y maybe placed
      Note: Even though all items in Folder B are restored permissions for the parent folder where these items go should not change

 (3) Restoring Folder A (includes everything in it including item Y) and Item X
     => resulting folderA will be shared with user 1 and user 3
     => resulting item Y (within folder A) will be shared with user 2 explicity. Then folder permissions will apply from Folder A as parent
     => resulting item X will be shared with user 3

meain:
@gmatev All of these conditions are satisfied as of now. The issue is only when a file and its parent folder are given the same permissions for the same user, and that too when the file within the folder is given  the permission first.

PR #2148, opened 2023-01-16
Metadata backup for OneDrive

meain:
## Description

This PR adds option to backup and restore additional metadata(currently permissions) for OneDrive.

**This PR also adds the ability to restore empty folders for OneDrive along with their permissions.**

Breaking change: Any old backups will not work as we expect both `.data` and `.meta`/`.dirmeta` files to be available for OneDrive backups.

*Added changelog, docs pending.*

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/issues/1774

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/onedrive/collection.go
```diff
@@ -252,22 +300,49 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+
+				oc.data <- &Item{
+					id:   itemName + DataFileSuffix,
+					data: itemReader,
+					info: itemInfo,
+				}
+			}
+
+			if itemMeta != nil {
+				// TODO(meain): This right now creates two
```
meain, line 318: Should we have a separate progressbar entry for upload metadata? How should that behave?

src/internal/connector/onedrive/collection.go
```diff
@@ -252,22 +300,49 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+				})
+
+				oc.data <- &Item{
+					id:   itemName + metaFileSuffix,
+					data: metaReader,
+					info: itemInfo,
+				}
+			}
 
+			// TODO(meain): Should these counts gets updated now that
```
meain, line 339: How should these counts get updated? Should we consider metadata file as a separate item?

src/internal/connector/onedrive/restore.go
```diff
@@ -118,50 +146,130 @@ func RestoreCollection(
...
+					restoreID, ok := restoredIDs[trimmedName]
+					if !ok {
+						errUpdater(itemData.UUID(), fmt.Errorf("item not available to restore permissions"))
+						continue
+					}
+
+					restorePermissions(ctx, service, drivePath.DriveID, restoreID, parentPerms, meta.Permissions)
+
+					// TODO(meain): Finalize on how we count success
+					metrics.Successes++ // Counted as success when we have also restored metadata
```
meain, line 196: As of now, we add `metrics.Object` when we encounter a new `.data` file and we consider it as success when we have also been able to restore its permissions. LMK if this needs any changes.

src/internal/kopia/wrapper.go
```diff
@@ -366,6 +366,36 @@ func (w Wrapper) RestoreMultipleItems(
...
+	// TODO(meain): details things can avoid storing info about
+	// .dirmeta and we can compute them from the list of files. The
+	// only issue is that we might not be able to restore empty folder
+	// as we will not be aware of them without some mention about
+	// them.
+
+	// This sort is done primarily to order .meta files after .data
+	// files
+	// TODO(meain): This code should be moved into the OneDrive component
+	sort.Slice(paths, func(i, j int) bool {
```
meain, line 395: One thing we still rely on is the fact that we get the stream items with the item before its metadata. It does not matter if we get all the items first, then all the the items, but we cannot handle if we get the metadata first before the actual item data. Hope this is good.

src/internal/connector/onedrive/collection.go
```diff
@@ -252,22 +300,49 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+				if !isFile {
+					metaFileSuffix = DirMetaFileSuffix
+				}
+				metaReader := lazy.NewLazyReadCloser(func() (io.ReadCloser, error) {
+					progReader, closer := observe.ItemProgress(ctx, itemMeta, observe.ItemBackupMsg, itemName+metaFileSuffix, itemMetaSize)
+					go closer()
+					return progReader, nil
+				})
+
+				oc.data <- &Item{
```
meain, line 332: As of now, details file contains the names of metadata files. We can drop that, but one advantage of having them there is so that we are also aware of empty directories. Otherwise we might miss restoring empty directories(which is the current behavior). I am assuming the reason behind us not restoring empty directories as of now was we needed files to create a "Collection".

meain (commented):
@ashmrtn @ryanfkeepers Had a few questions around the changes. Let me know if you have any thoughts around these.

src/internal/connector/onedrive/collection.go
ashmrtn: >might miss restoring empty directories(which is the current behavior)

To be honest, I'm not sure if we've ever actually pinned down the expected behavior of empty directories as far as whether we should back them up and/or restore them. Right now I think we're pretty inconsistent about their handling between OneDrive and Exchange and even within a service between full backup vs. incremental backup

If we keep them in backup details we'll need to add helper code to filter them out when displaying information to the user. I'm not sure I have an answer as to whether they should/shouldn't be in details to start with though

src/internal/kopia/wrapper.go
ashmrtn: yea, I was kind of thinking about this after our chat. Eventually I'd like to remove the ordering requirements from kopia wrapper as it constrains the possible implementations we can choose and how things will work when we may not know what files need selected beforehand (e.x. restoring an email with attachments)

For now I can't think of a different solution though :/ maybe others will have some thoughts

src/internal/connector/onedrive/restore.go
```diff
@@ -42,11 +47,29 @@ func RestoreCollections(
...
+	})
+
+	parentPermissions := map[string][]UserPermission{}
+
 	// Iterate through the data collections and restore the contents of each
 	for _, dc := range dcs {
-		temp, canceled := RestoreCollection(ctx, service, dc, OneDriveSource, dest.ContainerName, deets, errUpdater)
+		parentPerms, ok := parentPermissions[dc.FullPath().String()]
+		if !ok {
+			errUpdater(dc.FullPath().String(), fmt.Errorf("unable to find parent permissions"))
```
ashmrtn, line 62: this may need relaxed if we're only storing permissions when they're not the default

src/internal/connector/onedrive/restore.go
```diff
@@ -118,50 +146,130 @@ func RestoreCollection(
...
 
-			deets.Add(
-				itemPath.String(),
-				itemPath.ShortRef(),
-				"",
-				true,
-				itemInfo)
+			if source == OneDriveSource {
+				name := itemData.UUID()
+				if strings.HasSuffix(name, ".data") {
```
ashmrtn, line 165: nit: this should use the defined constant

src/internal/connector/onedrive/restore.go
```diff
@@ -118,50 +146,130 @@ func RestoreCollection(
...
+					metrics.Objects++
+					metrics.TotalBytes += int64(len(copyBuffer))
+					trimmedName := name[:len(name)-5]
```
ashmrtn, line 168: nit: this should use the defined constant (i.e. `len(dataSuffix)`) or a strings package function if one exists (i.e. `TrimSuffix(dataSuffix)`)

src/internal/connector/onedrive/collection.go
ryanfkeepers: If possible, we should treat it like any other item in the collection.

src/internal/connector/onedrive/restore.go
meain: As of now I am storing the metadata even if it is just the defaults, but it does not really make much sense to store it if there is no change. I'll probably have that in a follow up PR as we might end up changing a few things related to how we update the metrics success.

src/internal/connector/onedrive/collection.go
meain: We are good here then. As of now we do treat them as just another item in the collection. You will get one progressbar when upload the data and one when uploading the metadata.

src/internal/connector/onedrive/restore.go
meain: We might also have to think about if we keep the metadata file in the details if we are going this route. If we are not always storing the metadata, I think we will need to know when to ask for the meta file from kopia.

src/internal/connector/onedrive/collection.go
meain: If we end up omitting the metadata file if the values are just the defaults (as per the discussion [here](https://github.com/alcionai/corso/pull/2148#discussion_r1073074066), we will have to keep the metadata file so as to know when to ask kopia for a metadata file.

src/internal/connector/onedrive/restore.go
```diff
@@ -118,50 +146,130 @@ func RestoreCollection(
...
+
+					trimmedName := name[:len(name)-8]
+					createRestoreFolder(ctx, service, drivePath.DriveID, trimmedName, restoreFolderID, parentPerms, meta.Permissions)
+
+					path := itemPath.String()
+					folderPerms[path[:len(path)-8]] = meta.Permissions
+				} else {
+					if !ok {
+						// TODO(meain): support older backups?
+						errUpdater(itemData.UUID(), fmt.Errorf("invalid backup format, you might be using an old backup"))
```
meain, line 212: Should we consider adding some kind of version information in the backup?

src/internal/connector/onedrive/collection.go
meain: @vkamra any thoughts around what should be going in the details file as well as if we should be restoring empty folders?

src/internal/connector/onedrive/restore.go
meain: Also, do we have to support older backups. There is currently no way of reliably checking if it was backed up using an older version. We could probably change this by changing the format of the details file to include some kind of version information.

src/internal/connector/onedrive/restore.go
ashmrtn: ah, perhaps I misinterpreted some of the other code for fetching permissions or something then. I think what we're kind of running up against here is how we ask kopia for things that we may not know the name of (i.e. "select all files with a `.meta` suffix`). This is something that we may (eventually) need to expand kopia to handle since it seems like email attachments will land in a similar situation (ignoring the ordering issue of things for the moment)

For now, it's fine to always store permissions, even if they're the default. We can come back and optimize later if we find a need for it

src/internal/connector/onedrive/restore.go
ashmrtn: >Should we consider adding some kind of version information in the backup?

which information are you trying to version? Backups do currently include an "on-disk layout version number" (this was that random leading `0001` you were seeing when fetching folder permissions in one of the other WIP PRs for this feature). That is injected, checked, and stripped off by kopia wrapper during backup/restore. As far as versioning the data stored in the files to tie it to changes in some service code (e.x. OneDrive), we don't really have anything like that yet

I think whether we need something like that is dependent on two things:
* if there is reasonable "fallback" behavior we can implement for when we have no metadata files
* if the fallback behavior makes sense to users of corso

src/internal/connector/onedrive/restore.go
```diff
@@ -118,50 +146,141 @@ func RestoreCollection(
...
+
+					trimmedName := strings.TrimSuffix(name, DirMetaFileSuffix)
+					_, err = createRestoreFolder(ctx, service, drivePath.DriveID, trimmedName, restoreFolderID, parentPerms, meta.Permissions)
+					if err != nil {
+						errUpdater(itemData.UUID(), err)
+						continue
+					}
+
+					ipth := itemPath.String()
+					folderPerms[ipth[:len(ipth)-8]] = meta.Permissions
```
ashmrtn, line 219: nit: should this be `timmedName`? (random-looking constants make me nervous 😅)

src/internal/connector/onedrive/restore.go
meain: I might have to look a bit more into what the "on-disk layout version number" is used for. The version number I am suggesting here would be added to the details file. In this current scenario, it would be used to determine if the files were stored with a `.data` and `.meta` suffix or just a 1:1 map of files. Essentially, some way to determine that the backup was from an older version of corso when we hit this if block.

src/internal/connector/onedrive/restore.go
meain: This is the path instead of just the name, let me create a `trimmedPath` variable.

src/internal/connector/onedrive/restore.go
ashmrtn: >might have to look a bit more into what the "on-disk layout version number" is used for.

Right now it's not really used for anything. The idea behind it though is if we start doing things like adding data to the front of items (e.x. store OneDrive filename inline with file data or something) then knowing which version of the data you're reading would become important

src/internal/connector/onedrive/collection.go
ryanfkeepers: Seems reasonable to me.

src/internal/connector/onedrive/restore.go
meain: As of now, I'm skipping adding any version. One thing to keep in mind is that once merged corso will fail when trying to restore with an older backup.

meain:
This PR should now be good to go. It is a relatively big PR, but you can probably track the changes per commit. The core of the logic is within the first 6 commits and the rest is just figuring out how to fix up issues I was running into within the CI. I would suggest looking at the PR commit by commit.

meain:
Breaking change: Any old backups will not work as we expect both `.data` and `.meta`/`.dirmeta` files to be available for OneDrive backups.

vkamra:
> Breaking change: Any old backups will not work as we expect both `.data` and `.meta`/`.dirmeta` files to be available for OneDrive backups.

@meain - can we detect/fallback to not restoring permissions?

meain:
> can we detect/fallback to not restoring permissions?

@vkamra The main reason for this is that file names for backups changed. Previously we saved a file in OneDrive with and entry `filename` in kopia, but now we have `filename.data` and `filename.meta`. We have no way currently to determine if a file named `filename.meta` is the meta file or if it is an actual file with that name from an older backup.

I was considering adding version information into the details file so that we can distinguish between this. I was not certain if we should add this as we are still in Alpha. We cold do this if it is necessary. I can do this in a follow up PR if that is OK as this PR is already pretty big.

One related option is to use [serializationVersion](https://github.com/alcionai/corso/blob/e35c6cd17eaaf0251f21271bba6e4d3089dd61bf/src/internal/kopia/wrapper.go#L30), but that is per file and not for the entire backup.

vkamra:
> > can we detect/fallback to not restoring permissions?
> 
> @vkamra The main reason for this is that file names for backups changed. Previously we saved a file in OneDrive with and entry `filename` in kopia, but now we have `filename.data` and `filename.meta`. We have no way currently to determine if a file named `filename.meta` is the meta file or if it is an actual file with that name from an older backup.
> 
> I was considering adding version information into the details file so that we can distinguish between this. I was not certain if we should add this as we are still in Alpha. We cold do this if it is necessary. I can do this in a follow up PR if that is OK as this PR is already pretty big.

Agree with doing this in a follow up version. I'd recommend we actually introduce a backup format version in the backup model and use that.

src/internal/connector/onedrive/restore.go
```diff
@@ -36,17 +42,48 @@ func RestoreCollections(
...
+	})
+
+	parentPermissions := map[string][]UserPermission{}
+
 	// Iterate through the data collections and restore the contents of each
 	for _, dc := range dcs {
-		temp, canceled := RestoreCollection(ctx, service, dc, OneDriveSource, dest.ContainerName, deets, errUpdater)
+		parentPerms, ok := parentPermissions[dc.FullPath().String()]
+		if !ok {
+			if len(dc.FullPath().Elements()) == 7 {
```
ashmrtn, line 70: nit: this feels kind of brittle since it assumes all the prefix directories for a path have a fixed number of elements. There are some functions for dealing with OneDrive paths that maybe we can leverage. The `path` package below is `src/pkg/path` no the stdlib one
```suggestion
			onedrivePath, err := path.ToOneDrivePath(dc.FullPath())
			if err != nil {
				// some error handling here
			}

			if len(onedrivePath.Folders) == 0 {
```

src/internal/connector/onedrive/collections.go
```diff
@@ -230,8 +230,15 @@ func (c *Collections) UpdateCollections(
 			// already created and partially populated.
 			updatePath(newPaths, *item.GetId(), folderPath.String())
 
+			if c.source != OneDriveSource {
+				continue
+			}
+
+			fallthrough
```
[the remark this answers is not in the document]
ashmrtn: ah, I think I see. I'm guessing the reason permissions for a folder are currently in the parent and not the folder itself is so that we already have them when we need to make the folder during restore?

I guess the question for whether it would be "easy" to keep folder permissions in the folder itself is really whether there is an ordering issue between restoring item permissions and folder permissions. If I remember right, you did say there was a potential ordering issue and that folder permissions should be restored first right?

src/internal/connector/onedrive/restore.go
```diff
@@ -36,17 +42,48 @@ func RestoreCollections(
...
+				// root directory will not have permissions
+				parentPerms = []UserPermission{}
+			} else {
+				errUpdater(dc.FullPath().String(), fmt.Errorf("unable to find parent permissions"))
+			}
+		}
 
-		restoreMetrics.Combine(temp)
+		metrics, folderPerms, permissionIDMappings, canceled = RestoreCollection(ctx, service, dc,
```
ashmrtn, line 78: nit: separate lines for params please

src/internal/connector/onedrive/restore.go
```diff
@@ -262,5 +404,129 @@ func restoreItem(
...
+) (map[string]string, error) {
+	permAdded, permRemoved := getChildPermissions(childPerms, parentPerms)
+
+	for _, p := range permRemoved {
+		err := service.Client().DrivesById(driveID).ItemsById(itemID).
+			PermissionsById(permissionIDMappings[p.ID]).Delete(ctx, nil)
+		if err != nil {
+			return permissionIDMappings, errors.Wrapf(
+				err,
+				"failed to set permissions for item %s. details: %s",
```
ashmrtn, line 492:
```suggestion
				"failed to remove permission for item %s. details: %s",
```

src/internal/connector/graph_connector_test.go
```diff
@@ -368,6 +374,7 @@ func runRestoreBackupTest(
 
 		collections = append(collections, ownerCollections...)
 		totalItems += numItems
+		toatalKopiaItems += kopiaItems
```
ashmrtn, line 377:
```suggestion
		totalKopiaItems += kopiaItems
```

src/internal/connector/graph_connector_test.go
```diff
@@ -907,3 +1026,248 @@ func (suite *GraphConnectorIntegrationSuite) TestMultiFolderBackupDifferentNames
...
+							data:      getTestMetaJSON(suite.T(), suite.secondaryUser, []string{"write"}),
+							lookupKey: "test-file.txt" + onedrive.MetaFileSuffix,
+						},
+					},
+				},
+			},
+		},
+
+		{
+			name:     "FileInsideFolderPermissionsResote",
```
ashmrtn, line 1072:
```suggestion
			name:     "FileInsideFolderPermissionsRestore",
```

src/internal/connector/graph_connector_helper_test.go
```diff
@@ -643,20 +646,42 @@ func compareOneDriveItem(
...
+
+		err = json.Unmarshal(expectedData, &expectedMeta)
+		assert.Nil(t, err)
+		assert.Equal(t, len(expectedMeta.Permissions), len(itemMeta.Permissions), "number of permissions after restore")
+
+		for i, p := range expectedMeta.Permissions {
+			assert.Equal(t, p.Email, itemMeta.Permissions[i].Email)
+			assert.Equal(t, p.Roles, itemMeta.Permissions[i].Roles)
+			assert.Equal(t, p.Expiration, itemMeta.Permissions[i].Expiration)
+		}
```
ashmrtn, line 678: this may need to be more robust in the future. It seems like it currently assumes a fixed order for permissions that is common between both the expected and actual permissions

ashmrtn (commented):
Mostly minor nits. Only thing that's really a question is whether we want folder permissions in the parent folder or the folder they refer to. That can/should be pushed to a different PR if we decide to do it

src/internal/connector/onedrive/collections.go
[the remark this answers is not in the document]
meain: I initially had it within the folder but moved out mainly because of the ordering issue. If we keep the `.dirmeta` file inside the folder we have to make sure we are fetching that file first, but this way we can be certain that by the time a folder contents is to be restored we have created the folder with the right permissions.

src/internal/connector/graph_connector_helper_test.go
meain: Let me add a note in there for now. As of now the tests are only using single permissions as we have only one additional user available `CORSO_SECONDARY_M365_TEST_USER_ID`.

src/internal/connector/sharepoint/restore.go
```diff
@@ -59,14 +59,15 @@ func RestoreCollections(
 
 		switch dc.FullPath().Category() {
 		case path.LibrariesCategory:
-			metrics, canceled = onedrive.RestoreCollection(
+			metrics, _, _, canceled = onedrive.RestoreCollection(
 				ctx,
 				service,
 				dc,
+				[]onedrive.UserPermission{}, // Currently permission data is not stored for sharepoint
 				onedrive.OneDriveSource,
```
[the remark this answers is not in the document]
meain: I had missed this one, but I use the source filed to decide how the backups would be structured at https://github.com/alcionai/corso/blob/266fffa4e0775b1684e0ff835d8757b38eb20567/src/internal/connector/onedrive/restore.go#L193.

ashmrtn (approved):

vkamra (dismissed):
@meain - adding a block on merging so that we can follow up offline on the impact of this on throttling (see https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling and the cost of permission expansion)

Issue #1996, opened 2022-12-29
Refactor the connector package towards interface consumption


ryanfkeepers:
## Issue
The connector package is largely designed around service and category type switches, rather than interfaces.  This is functional, but produces boilerplate with high amounts of noise, couples the core process flow with type fulfillment, and requires a larger mental model than necessary to grok.  The hope is that by swapping out these type-based pivot points with a standard interface, and types that comply with said interface, that he connector code can become both consolidated and more simplified.

## Details
The core graph backup flow is a relatively simple process:
1. for each resource owner, and each requested data category
2. look up all the containers (eg: folders, calendars, etc)
3. fetch all the items in the containers (possibly as an incremental change)
4. compile the results into collection
5. stream items, populating additional item data as needed, to the consumer via each collection

Despite this simplicity, it can be difficult to grok the flow when reading because much of the implementation is coupled with handling per-category differences.  Where it would be possible to call some interface like `typeHandler.GetConainers()`, we instead go through a process of: "given a category enum, switch on the enum to call a factory that produces a getContainersByType func, but error if we don't recognize the category".

Latent issues involve decentralized code: files often provide a single function, and then the many ways it can be called by differing types, causing type fulfillment to get scattered across the package rather than aggregated within that type.  Testing the core flow by way of mocks becomes similarly tangled.  We aren't able to assert expectations about the core process independent of the categorical implementations.

## Value
Much of this debt is the result of frequent upheavals in upstream libraries, rapid iteration on tooling changes and sdk consumer support, and substantial iteration to our own supported features.  To avoid premature design refinement, the team has made conscious effort to avoid design refinements while these changes are ongoing.

We're now entering a cooling period where much of our design and tooling has settled, our long-term patterns are more clearly identified, and we're building abstraction layers for better cohesion with our dependencies.  This puts us in a good position to start looking at design refinement, starting with the proposed refactor.  The end result should be a more readable, testable, and overall maintainable package.

## Work
- [x] code consolidation
- [x] exchange
- [ ] onedrive/sharepoint

ryanfkeepers:
It's Happening!

PR #2150, opened 2023-01-17
add item fetching and serialization to api

ryanfkeepers:
## Description

Adds the per item collection streaming calls to
the api interface.  Primarily migrates a "getItem"
and a "serializeItem" acton into the api pkg, out
from exchange_data_collection.  Building an
ExchangeInfo is now also housed in api.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1996

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/api/events.go
```diff
@@ -190,6 +204,66 @@ func (c Events) GetAddedAndRemovedItemIDs(
...
+	}
+
+	var (
+		err    error
+		writer = kioser.NewJsonSerializationWriter()
+	)
+
+	defer writer.Close()
+
+	if *event.GetHasAttachments() {
```
ashmrtn, line 230: I know this is just moving code around, but this feels more like it should be with fetching the item (or as a separate call somehow) rather than in the serialization code. Maybe it should be moved in a later PR?

src/internal/connector/exchange/api/contacts.go
```diff
@@ -52,12 +56,20 @@ func (c Contacts) DeleteContactFolder(
...
+) (serialization.Parsable, *details.ExchangeInfo, time.Time, error) {
+	cont, err := c.stable.Client().UsersById(user).ContactsById(itemID).Get(ctx, nil)
+	if err != nil {
+		return nil, nil, time.Time{}, err
+	}
+
+	info := ContactInfo(cont)
+	modTime := orNow(cont.GetLastModifiedDateTime())
+
+	return cont, info, modTime, nil
```
ashmrtn, line 72: I'm not really a fan of returning extra values from functions when those values can also be sourced from other returned values (e.x. modTime). It seems like it would be better and less confusing to users of this function to just get the mod time if they need to. This would also remove potential confusion about why the returned mod time doesn't match the mod time in the returned item

If we need to expand an interface to support a `GetLastModifiedDateTime()` call then we can do that

ashmrtn (approved):

src/internal/connector/exchange/api/events.go
ryanfkeepers: Yeah, absolutely agreed.  We can continue to refine this as we go.

src/internal/connector/exchange/api/contacts.go
ryanfkeepers: I'm with you here.  Mostly kept it in because I wasn't sure why we were handling modtime differently in the first place.  Sounds like it's not wanted, so we can consolidate.

ashmrtn: >Mostly kept it in because I wasn't sure why we were handling modtime differently in the first place

If you haven't found the answer already, this is because we've made the items implement the StreamModTime interface or something so that we can do kopia-assisted incrementals. In those cases it's safe to use the current time for modtime (in kopia) so that kopia will just re-upload the data

ryanfkeepers: Fair.  I figure it's all the same if we default the modified time in the exchangeInfo to `now`, like we were doing with the interface value.  Does that concern you at all?

ashmrtn: I think we should keep those separate as well. Defaulting it in the item info will affect what the user sees and can hide issues that may crop up

Issue #2022, opened 2023-01-03
Incrementals - Calendar Event Delta Handling


ryanfkeepers:
## Issue
Delta coverage for calendar events is incomplete.

### Events vs Views
The [Calendar Event](https://learn.microsoft.com/en-us/graph/api/calendar-list-events?view=graph-rest-1.0&tabs=http) object stored by Corso does not support delta queries.  There is an alternative event object, the [Calendar View](https://learn.microsoft.com/en-us/graph/api/event-delta?view=graph-rest-1.0&tabs=http), which is (quoting the docs) "a set of [event](https://learn.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0) resources that have been added, deleted, or updated in a calendarView (a range of events defined by start and end dates)".  It's possible that Corso should store both types.

### Lazy Event Pulling
Since the primary event set does not have delta support, we may have to find alternative ways to minimize data retrieval.  One option is a lazy pull: only perform the full item download if the item's modTime has changed between kopia and GC.

ashmrtn:
this is related to
* #2023

PR #2154, opened 2023-01-17
use delta queries for calendar events

ryanfkeepers:
## Description

Hacks the beta version call into the events api
when iterating through items, allowing us to run
delta-based queries for calendar events.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2022

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/events.go
+++ b/src/internal/connector/exchange/api/events.go
@@ -144,29 +144,25 @@ func (c Events) EnumerateContainers(
 // item pager
 // ---------------------------------------------------------------------------
 
-type eventWrapper struct {
-	models.EventCollectionResponseable
-}
-
-func (ew eventWrapper) GetOdataDeltaLink() *string {
-	return nil
-}
-
 var _ itemPager = &eventPager{}
+
+const (
+	eventBetaDeltaURLTemplate = "https://graph.microsoft.com/beta/users/%s/calendars/%s/events/delta"
+)
 
 type eventPager struct {
 	gs      graph.Servicer
-	builder *users.ItemCalendarsItemEventsRequestBuilder
-	options *users.ItemCalendarsItemEventsRequestBuilderGetRequestConfiguration
+	builder *users.ItemCalendarsItemEventsDeltaRequestBuilder
+	options *users.ItemCalendarsItemEventsDeltaRequestBuilderGetRequestConfiguration
 }
 
 func (p *eventPager) getPage(ctx context.Context) (pageLinker, error) {
 	resp, err := p.builder.Get(ctx, p.options)
-	return eventWrapper{resp}, err
+	return resp, err
 }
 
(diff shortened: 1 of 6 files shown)


src/internal/connector/exchange/api/events.go
```diff
@@ -182,23 +178,54 @@ func (c Events) GetAddedAndRemovedItemIDs(
...
+
+		resetDelta = true
+		errs = nil
+	}
+
+	// Graph SDK only supports delta queries against events on the beta version, so we're
+	// manufacturing use of the beta version url to make the call instead.
+	// See: https://learn.microsoft.com/ko-kr/graph/api/event-delta?view=graph-rest-beta&tabs=http
+	// Note that the delta item body is skeletal compared to the actual event struct.  Lucky
+	// for us, we only need the item ID.  As a result, even though we hacked the version, the
```
ashmrtn, line 214: can you check that items don't get skipped accidentally due to us selecting the bare minimum fields? Tbh, I'm not sure if this is a feature or a "bug" in how Graph works, but using `$select` can sometimes cause a delta endpoint to skip returning results if they don't match what's in the `$select`. For example, not having `isRead` in the select for emails will cause emails where the `isRead` flag has been changed to be skipped because it doesn't update the mod time. This will take some playing around and is probably easiest to do with GraphExplorer/postman

ashmrtn (approved):

src/internal/connector/exchange/api/events.go
ryanfkeepers: Hmm, good call out.  I'll do some manual testing.  This might be especially interesting in this case, because the delta body for an event are only able to contain the id, startTime, and endTime.  Hopefully the delta will catch changes on unrepresented properties, like renames and etc.

ryanfkeepers: Alright, two findings.
1. everything works.  Changes to name, desc, attendees, etc, are all picked up by the delta.
2. we're not allowed to select values in the first place (see the error below).  If anything, we're lucky that the client is tossing the use of options for us.
```json
    "error": {
        "code": "ErrorInvalidUrlQuery",
        "message": "The following parameters are not supported with change tracking over the 'SyncEvents' resource: '$orderby, $filter, $select, $expand, $search'."
    }
```

src/internal/connector/exchange/api/events.go
```diff
@@ -182,23 +178,54 @@ func (c Events) GetAddedAndRemovedItemIDs(
...
+	// response body parses properly into the v1.0 structs and complies with our wanted interfaces.
+	// Likewise, the NextLink and DeltaLink odata tags carry our hack forward, so the rest of the code
+	// works as intended (until, at least, we want to _not_ call the beta anymore).
```
vkamra, line 217: 👍

PR #2155, opened 2023-01-17
extra loggng and error wraps

ryanfkeepers:
## Description

Adds info logging on all throttling responses,
and some extra error handling in container
resolvers.

- [x] :no_entry: No 

## Type of change

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/graph/service_helper.go
+++ b/src/internal/connector/graph/service_helper.go
@@ -1,7 +1,7 @@
 package graph
 
 import (
-	nethttp "net/http"
+	"net/http"
 	"net/http/httputil"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


src/internal/connector/graph/service_helper.go
```diff
@@ -82,6 +82,11 @@ func (handler *LoggingMiddleware) Intercept(
 		return resp, err
 	}
 
+	// special case for supportability: log all throttling cases.
+	if resp.StatusCode == http.StatusTooManyRequests {
+		logger.Ctx(ctx).Infow("graph api throttling", "method", req.Method, "url", req.URL)
```
ashmrtn, line 87: is it safe to log the URL? It seems like it could contain potentially sensitive info

src/internal/connector/support/errors.go
```diff
@@ -89,8 +89,20 @@ func concatenateStringFromPointers(orig string, pointers []*string) string {
...
+func ConnectorStackErrorTraceWrap(e error, prefix string) error {
+	cset := ConnectorStackErrorTrace(e)
+	if len(cset) > 0 {
+		return errors.Wrap(e, prefix+": "+cset)
+	}
+
+	return errors.Wrap(e, prefix)
+}
+
+// ConnectorStackErrorTracew is a helper function that extracts
```
ashmrtn, line 104:
```suggestion
// ConnectorStackErrorTrace is a helper function that extracts
```

ashmrtn (approved):

src/internal/connector/graph/service_helper.go
ryanfkeepers: Something we'll have to tackle when we address log sanatiziation.

Issue #2145, opened 2023-01-14
Improve "licensed user" check for M365 services


vkamra:
We [currently](https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/drive.go#L395) check whether a user has a OneDrive drive by looking at what M365 SKUs the user has a license for.

The check uses a version of this list that we've added to the repo. The check seems brittle and the approach may not scale.

Alternatively - maybe an approach where we query existence of a drive or mailbox might be sufficient.

vkamra:
Adding to current cycle because we do need add a check for Exchange (currently missing)

vkamra:
Exchange fixed with #2156. Need to address OneDrive going forward.

meain:
@vkamra wouldn't https://github.com/alcionai/corso/pull/2161 handle the case for OneDrive? Should we close the issue?

PR #2156, opened 2023-01-18
Check whether the user has an exchange mailbox

vkamra:
## Description

This commit adds logic in discovery and backup to check whether the specified user has
an exchange mailbox that is available/enabled.

If so - the backup is short-circuited to succeed but with "no data"

Going forward - we should be able to move the logic in the OneDrive connector that checks
for a valid drive and license in here.

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2145 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -7,7 +7,9 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/internal/connector/discovery"
 	"github.com/alcionai/corso/src/internal/connector/exchange"
+	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/onedrive"
 	"github.com/alcionai/corso/src/internal/connector/sharepoint"
 	"github.com/alcionai/corso/src/internal/connector/support"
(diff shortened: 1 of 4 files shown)


vkamra:
@meain - take a look at this and please +1 if it looks good. I will follow up with unit tests prior to merge. If it makes sense - we should add the OneDrive check you're working on over here.

src/internal/connector/data_collections.go
```diff
@@ -124,6 +136,29 @@ func verifyBackupInputs(sels selectors.Selector, userPNs, siteIDs []string) erro
...
+	gs graph.Servicer,
+	service path.ServiceType,
+	resource string,
+) (bool, error) {
+	if service == path.SharePointService {
+		// No "enabled" check required for sharepoint
+		return true, nil
+	}
+
+	_, info, err := discovery.User(ctx, gs, resource)
```
meain, line 150: nit: We could probably limit the check to only the service that one is trying to backup instead of both exchange and onedrive.
Other than that, we should be able to follow something similar for OneDrive.

vkamra: Going to leave the check as is for now and make that change in a follow-up (need to get this fix in first)

meain (commented):
As of now trying to do `backup details` and `restore` fails with errors. We should probably give a better message to users when trying to restore something without any data.

vkamra:
> As of now trying to do `backup details` and `restore` fails with errors. We should probably give a better message to users when trying to restore something without any data.

Filed https://github.com/alcionai/corso/issues/2160 to improve

meain (approved):

Issue #2196, opened 2023-01-20
BackupOp doesn't check for errors after GraphConnector fetches items


ashmrtn:
GraphConnector reports the number of errors it hit in the status struct. However, BackupOp never checks this to see if the number of errors was > 0. This can hide some error cases

ashmrtn:
git hammer fix merged so closing for now. See #2207 for tracking of a best-effort approach

PR #2199, opened 2023-01-20
Fail operation if we failed to fetch item data

ashmrtn:
## Description

If GraphConnector reported errors while trying to fetch item data, fail the entire operation. This stops silent failures, but is a big hammer at the moment because there's no checks for items no longer being available.

Error reporting is minimal as well, but some info should be in the log.

**This effectively disables best-effort as any failure will result in the operation being reported as failed**

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2196

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


ashmrtn:
failing with "item not found" errors, likely from a CI purge. Merging #2202 before this can help avoid issues like that

Issue #1969, opened 2022-12-27
Context Aggregation for Logging


ryanfkeepers:
A context logging aggregator package to collect runtime variables throughout the process tree would help debugging and logging quality.

Reading (not necessarily the recommended solution, but good for explanation): https://www.southcla.ws/go-structured-errors/

I've also now created my own library for evaluation here: https://github.com/ryanfkeepers/clues

ryanfkeepers:
Moving forward with a PoC involving Clues.

gmatev:
@ryanfkeepers is this still relevant ?

PR #2214, opened 2023-01-21
adds clues package to begin structured logging

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1969

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/operations/restore.go
+++ b/src/internal/operations/restore.go
@@ -9,6 +9,7 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/clues"
 	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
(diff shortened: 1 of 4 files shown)


src/internal/operations/restore.go
```diff
@@ -129,6 +130,12 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
 
 	detailsStore := streamstore.New(op.kopia, op.account.ID(), op.Selectors.PathService())
 
+	ctx = clues.AddAll(
+		ctx,
+		"tenant_id", op.account.ID(), // TODO: pii
+		"backup_id", op.BackupID,
+		"service", op.Selectors.Service)
+
```
dadams39, line 138: Would it be possible.....to add a ticket for tech debt for this package for commons? There are times when the same `type` of functions in restore and backup.

dadams39 (approved):

src/internal/operations/restore.go
ryanfkeepers: I very specifically want to avoid the use of `common` pkg here.  If we get so far as needing that, then it's a sign that the library itself has an unusable interface.  In which case, it'd be better to improve the clues package than to wrap it.  Ideally, the implementation should always involve a light touch, and not need any syntactical sugar.

PR #2233, opened 2023-01-23
Don't retry if an item was not found

### LINKED ISSUE #2217: don't retry if item is not found
Exchange data collection tries to fetch items that sometimes don't exist anymore. Currently, those cases [result in retries](https://github.com/alcionai/corso/blob/ea2d9cecebe88fa88730f783cf632f22df42254d/src/internal/connector/exchange/exchange_data_collection.go#L260) with increasingly long delays. This can lead to long runtimes if many items are no longer available

Solution:
reduce or remove retries for items that aren't found

ashmrtn:
## Description

When fetching item data, don't backoff and retry if Graph reported the item was not found. In this case, we mark it as succeeded as we can never get the data anyway.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* closes #2217

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -257,6 +257,14 @@ func (col *Collection) streamItems(ctx context.Context) {
 					break
 				}
 
+				// If the data is no longer available just return here and chalk it up
+				// as a success. There's no reason to retry and no way we can backup up
+				// enough information to restore the item anyway.
+				if e := graph.IsErrDeletedInFlight(err); e != nil {
+					atomic.AddInt64(&success, 1)
+					return
+				}
+
 				if i < numberOfRetries {
 					time.Sleep(time.Duration(3*(i+1)) * time.Second)
 				}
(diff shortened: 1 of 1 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -257,6 +257,14 @@ func (col *Collection) streamItems(ctx context.Context) {
 					break
 				}
 
+				// If the data is no longer available just return here and chalk it up
+				// as a success. There's no reason to retry and no way we can backup up
+				// enough information to restore the item anyway.
+				if e := graph.IsErrDeletedInFlight(err); e != nil {
+					atomic.AddInt64(&success, 1)
```
ryanfkeepers, line 264: Let's add a log here, so that we can easily identify it during supportability concerns.  Info level should be fine.

ryanfkeepers (approved):
Looks good.  Is this the only location where we need to add this?

Issue #1970, opened 2022-12-27
Error processing and management


ryanfkeepers:
Error production, management, and tracking throughout corso is largely haphazard, having grown organically within each different package.  A singular, curated error management package would normalize much of the differences, and provide maintenance benefits beyond.

Cases that a singular errors management package would be expected to handle:
1. Coordinating error communication and aggregation.
Throughout graph we have a number of "errUpdater" callback funcs that are passed downstream in order to aggregate lower level errors into a parent multi-err.  We lack a single point of production for parent multi-err, allowing error trees to accrue in place of a singular error list.  Similarly, the functions themselves are often declared anonymously and duplicate similar code, increasing bloat throughout the package.

2. Identifying app behavior after errors.
Corso can be run in one of two error handling modes: fail-fast and best-attempt.  We have various locations which flag the intended state, but application of, and attention to, that flag can vary (or be omitted) depending on the package.  An error manager would provide a central owner for behavior identification.

3. Controlling information flow.
Errors, especially between graph support and pkg/errors, have a variety of options to produce greater or fewer details.  These are things like stack traces, body dumps, etc.  Although those functions exist, we have little means of control for end users to decide how much additional details they want, and when.  An error manager would provide a central owner for toggling greater or lesser amounts of debugging details, as well as the means to coordinate how those details are displayed to the end user through logs and other tools.

4. Normalizing errors for human readable CLI format.
Many of our error states result in a log dump with plenty of chaff info which isn't likely useful to a regular CLI user.  We currently rely on these dumps to identify issues, but as we move towards better logging control and reducing competition between logging and dynamic progress bars, it makes more sense to transform the cli output to a more human readable format.  An error manager would provide a central controller for transforming known error states into well curated outputs in for end users.

vkamra:
#5 is becoming high priority - this may be a good motivation to start this.

gmatev:
@ryanfkeepers is there outstanding work here?

PR #2236, opened 2023-01-23
Add fault pkg and errors struct

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/fault/example_fault_test.go
+++ b/src/pkg/fault/example_fault_test.go
@@ -14,7 +14,6 @@ import (
 
 var (
 	ctrl  any
-	anErr error
 	items = []string{}
 )
 
--- a/src/pkg/fault/example_fault_test.go
+++ b/src/pkg/fault/example_fault_test.go
@@ -1,0 +1,180 @@
+package fault_test
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/alcionai/corso/src/pkg/fault"
+)
+
+// ---------------------------------------------------------------------------
+// mock helpers
+// ---------------------------------------------------------------------------
+
+var (
+	ctrl  any
+	anErr error
+	items = []string{}
+)
+
+type mockController struct {
+	errors any
+}
+
+func connectClient() error      { return nil }
+func dependencyCall() error     { return nil }
+func getIthItem(i string) error { return nil }
+
+// ---------------------------------------------------------------------------
+// examples
+// ---------------------------------------------------------------------------
+
+// ExampleNewErrors highlights assumptions and best practices
+// for generating Errors structs.
+func Example_new() {
+	// Errors should only be generated during the construction of
+	// another controller, such as a new Backup or Restore Operations.
+	// Configurations like failFast are handled during construction.
+	ctrl = mockController{
+		errors: fault.New(false),
+	}
+}
+
+// ExampleErrorsFail describes the assumptions and best practices
+// for setting the Failure error.
+func Example_errors_Fail() {
+	errs := fault.New(false)
+
+	// Fail() should be used to record any error that highlights a
+	// non-recoverable failure in a process.
+	if err := connectClient(); err != nil {
+		errs.Fail(err)
+	}
+
+	// Only the topmost handler of the error should set the Fail() err.
+	// This will normally be the operation controller itself.
+	// IE: Fail() is not Wrap().  In lower levels, errors should get
+	// wrapped and returned like normal, and only handled by errors
+	// at the end.
+	lowLevelCall := func() error {
+		if err := dependencyCall(); err != nil {
+			// wrap here, deeper into the stack
+			return errors.Wrap(err, "dependency")
+		}
+
+		return nil
+	}
+
+	if err := lowLevelCall(); err != nil {
+		// fail here, at the top of the stack
+		errs.Fail(err)
+	}
+}
+
+// ExampleErrorsAdd describes the assumptions and best practices
+// for aggregating iterable or recoverable errors.
+func Example_errors_Add() {
+	errs := fault.New(false)
+
+	// Add() should be used to record any error in a recoverable
+	// part of processing.
+	for _, i := range items {
+		if err := getIthItem(i); err != nil {
+			errs.Add(err)
+		}
+	}
+
+	// In case of failFast behavior, iteration should exit as soon
+	// as an error occurs.  Errors does not expose the failFast flag
+	// directly.  Instead, iterators should check the value of Err().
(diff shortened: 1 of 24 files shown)


ryanfkeepers:
closing and remaking; improper branch base

ryanfkeepers:
Gah.  Nevermind, github doesn't want to let me either make a branch or change the merge.  Oh well.

src/pkg/fault/fault.go
```diff
@@ -0,0 +1,127 @@
...
+	// 1 of the items fails to be retrieved, but the rest of
+	// them succeed, we'd expect to see 1 error added to this
+	// slice.
+	errs []error
+
+	// if failFast is true, the first errs addition will
+	// get promoted to the err value.  This signifies a
+	// non-recoverable processing state, causing any running
+	// processes to exit.
+	failFast bool
```
ashmrtn, line 29: I'm a little hesitant about having this bool because it feels like we're mixing config/control flow info with error aggregation. Who is expected to set this? Are we passing these structs from BackupOp to GC/KW now because it also includes control flow information?

src/pkg/fault/fault.go
```diff
@@ -0,0 +1,127 @@
...
+
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	return e.setErr(err)
+}
+
+// setErr handles setting errors.err.  Sync locking gets
+// handled upstream of this call.
+func (e *errors) setErr(err error) *errors {
```
ashmrtn, line 89: we may eventually want to look into something like [checklocks](https://pkg.go.dev/gvisor.dev/gvisor/tools/checklocks#section-readme) to make sure we follow these comments

src/pkg/fault/fault_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+			e = n.Add(test.add)
+			require.NotNil(t, e)
+
+			test.expect(t, n.Errs())
+		})
+	}
+}
+
+func (suite *FaultErrorsUnitSuite) TestAdd() {
```
ashmrtn, line 159: nit: I would say this is testing adding multiple times since previous tests already check `Add` to some extent. Same with the `Fail` test

src/pkg/fault/example_fault_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if err := lowLevelCall(); err != nil {
+		// fail here, at the top of the stack
+		errs.Fail(err)
+	}
+}
+
+// ExampleErrorsAdd describes the assumptions and best practices
+// for aggregating iterable or recoverable errors.
+func Example_errors_Add() {
```
ashmrtn, line 77: In "practice" the asymmetry between when `Fail` vs. when `Add` should be called seems a little confusing. The rules I'm coming up with looking at the examples and comments are
* `Fail` at the top-level but return errors like normal
* `Add` in the low-level function and check to see if we should exit

While easily summed up, it seems like it could be confusing as to where to put which. We're now also going to have to check both the returned `error` from a function (i.e. something that will turn into a call to `Fail()` at some point in the stack) _and_ the `fault.errors.Err()` value to see if we should exit a function instead of trying to continue

src/pkg/fault/fault.go
ryanfkeepers: > Who is expected to set this?

Whomever calls `fault.New(failFast)`, which should be the operation constructor.

> Are we passing these structs from BackupOp to GC/KW now because it also includes control flow information?

Yes, but not strictly because it includes control flow information.  We want the struct passed around to access its `Adder` interface any place we aggregate errors.  

Configuration dictates control flow, and it makes sense in this instance for control flow that depends on errors to be bound to said errors aggregation.

src/pkg/fault/fault_test.go
ryanfkeepers: True, there's overlap here.  But the difference between Add and Multiple Adds is not, I think, worth calling out.

src/pkg/fault/example_fault_test.go
ryanfkeepers: > The rules I'm coming up with looking at the examples and comments are

That's pretty much correct, though I'm sure I can make that more clear.  I'll go do a pass on the language for improvements.

> We're now also going to have to check both the returned error from a function (i.e. something that will turn into a call to Fail() at some point in the stack) and the fault.errors.Err() value to see if we should exit a function instead of trying to continue

No, we shouldn't be doing that.  Only the users at the interface layer (CLI, SDK) need to investigate both Err() and Errs().

Within corso, error handling should only do one of three things:
1. Return an error, if possible (this is most of our error handling).
2. Add() an error, when within some iteration.
3. Fail() an error; only backup/restore operations should do this.

The majority of our error handling should feel entirely standard.  Ie: just create/wrap/return errors, and check the error that was returned.

ashmrtn: >No, we shouldn't be doing that. Only the users at the interface layer (CLI, SDK) need to investigate both Err() and Errs().

I think we're talking about slightly different things.

Assume you have the following:
```go
func caller() fault.errors {
  errs := fault.NewErrors(true)

  err := callee(&errs)

  // do something with error handling
}
```

How do you check that 1. callee didn't encounter a fatal error and 2. didn't encounter any error while running in fail-fast mode? Wouldn't we by necessity have to check both `fault.errors.Err()` and `err` to determine that? We'd check `Err()` to determine if we failed fast while we'd need to check `err` to see if we encountered a fatal error (e.x. credentials expired)

ryanfkeepers: Ahh, I gotcha.

> 1. callee didn't encounter a fatal error

`if errors.Err() != nil`

 > 2. didn't encounter any error while running in fail-fast mode?
 
 `if errors.Err() != nil`
 
 In fail-fast mode, the first recoverable error automatically promotes itself to the non-recoverable error slot.  IE: "this is the error that caused the operation to end".  
 
 In either case, I'd also say the example is a little different from what I'd expect in usage (I have a more complete example in the examples test now, but to try and shorten it here:)
 
 ```go
 func caller() fault.errors {
  errs := fault.NewErrors(true)

  if err := callee(&errs); err != nil {
    return errs.Fail(err)
  }

  recoverableCallee(&errs)
  
  return errs
}
```

And I realize that we'd need to check both the callee's `if err != nil && errs.Err() != nil`.  So a QoL helper could ease the boilerplate a lot.  I'll think on approaches there.

ashmrtn (approved):

Issue #2028, opened 2023-01-04
Add cli flag to toggle permissions restore


meain:
Follow up to https://github.com/alcionai/corso/issues/1774 . Add a flag `--preserve-permissions` or `--no-permissions` which can be used to decide if we should restore OneDrive permissions when restoring the data. This will be necessary as we might be restoring to a different user or under a different context. The default can probably be to not restore permissions.
The details are still to be finalized. /cc @gmatev

gmatev:
@meain let's make this flag be `--restore-permission`. By default permissions will not be restored. Please also make sure it gets reflected in the docs and the examples for the restore onedrive command.

meain:
Would `Options` or `ToggleFeatures` be a good place to pass this information onto the actual restore job? These currently seem to be just global flags though. Another option would be to add another `struct` specific to restore related operations.
https://github.com/alcionai/corso/blob/e35c6cd17eaaf0251f21271bba6e4d3089dd61bf/src/pkg/control/options.go#L7-L13

One important issue is that all the restore operations go through a single `Run` function irrespective of weather it is onedrive or exchange or any other.
/cc @ashmrtn @ryanfkeepers

ashmrtn:
>all the restore operations go through a single Run function irrespective of weather it is onedrive or exchange or any other.

I'm not as worried about this at the moment. Eventually we may need to figure something out so it doesn't get too sprawled out, but for the moment I think it's alright for service/data categories that don't need the flag to just ignore it

gmatev:
I do expect that we will have options at all levels that are specific to a service, operation, or both, so setting up the right pattern here makes sense to me. 

It may also be required for the proper handling of flags in our automated operations help and docs. If only at the global level, I expect this flag to show up everywhere which will be confusing.

ryanfkeepers:
> Would Options or ToggleFeatures be a good place to pass this information onto the actual restore job?

@meain A flag in `Options` is sufficient here.  Let's not bloat out structure where it's not needed.

ToggleFeatures should only be used for feature toggling control, not for configuration.  It's a blurry line sometimes.  Best gut check is to ask "am I hiding this thing from production usage so that I can release it in a half developed state?".  If so, It's a Toggle.

> let's make this flag be --restore-permissions. By default permissions will not be restored

[...]

PR #2241, opened 2023-01-24
Add `--restore-permissions` flag

meain:
## Description

By default we do not restore permissions for OneDrive. When this flag is provided, we enable restoring permissions to OneDrive.
Will switch from draft once https://github.com/alcionai/corso/pull/2148 is merged.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/2028

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - OneDrive file/folder permissions can now be backed up and restored
-- Add `--restore-permissions` flag to choose weather to restore OneDrive permissions
+- Add `--restore-permissions` flag to toggle restoration of OneDrive permissions
 
 ### Known Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - OneDrive file/folder permissions can now be backed up and restored
+- Add `--restore-permissions` flag to choose weather to restore OneDrive permissions
 
 ### Known Issues
 
--- a/src/cli/restore/onedrive.go
+++ b/src/cli/restore/onedrive.go
@@ -100,6 +100,9 @@ const (
 	oneDriveServiceCommandRestoreExamples = `# Restore file with ID 98765abcdef
 corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd --file 98765abcdef
 
+# Restore file with ID 98765abcdef along with any permissions associated to it
+corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd --file 98765abcdef --restore-permissions
+
 # Restore Alice's file named "FY2021 Planning.xlsx in "Documents/Finance Reports" from a specific backup
 corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd \
       --user alice@example.com --file "FY2021 Planning.xlsx" --folder "Documents/Finance Reports"
--- a/src/cli/restore/onedrive.go
+++ b/src/cli/restore/onedrive.go
@@ -63,6 +63,9 @@ func addOneDriveCommands(cmd *cobra.Command) *cobra.Command {
 			utils.FileFN, nil,
 			"Restore items by file name or ID")
 
+		// permissions restore flag
+		options.AddRestorePermissionsFlag(c)
+
 		// onedrive info flags
 
 		fs.StringVar(
--- a/src/cli/restore/onedrive.go
+++ b/src/cli/restore/onedrive.go
@@ -100,7 +100,7 @@ const (
 	oneDriveServiceCommandRestoreExamples = `# Restore file with ID 98765abcdef
 corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd --file 98765abcdef
 
-# Restore file with ID 98765abcdef along with any permissions associated to it
+# Restore file with ID 98765abcdef along with its associated permissions
 corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd --file 98765abcdef --restore-permissions
 
 # Restore Alice's file named "FY2021 Planning.xlsx in "Documents/Finance Reports" from a specific backup
--- a/src/cmd/factory/impl/common.go
+++ b/src/cmd/factory/impl/common.go
@@ -74,7 +74,7 @@ func generateAndRestoreItems(
 		items:        items,
 	}}
 
-	// TODO: fit the designation to the containers
+	// TODO: fit the destination to the containers
 	dest := control.DefaultRestoreDestination(common.SimpleTimeTesting)
 	dest.ContainerName = destFldr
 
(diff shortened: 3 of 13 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - OneDrive file/folder permissions can now be backed up and restored
+- Add `--restore-permissions` flag to choose weather to restore OneDrive permissions
```
ryanfkeepers, line 20:
```suggestion
- Add `--restore-permissions` flag to toggle restoration of OneDrive permissions
```

src/cli/restore/onedrive.go
```diff
@@ -97,6 +100,9 @@ const (
 	oneDriveServiceCommandRestoreExamples = `# Restore file with ID 98765abcdef
 corso restore onedrive --backup 1234abcd-12ab-cd34-56de-1234abcd --file 98765abcdef
 
+# Restore file with ID 98765abcdef along with any permissions associated to it
```
ryanfkeepers, line 103:
```suggestion
# Restore file with ID 98765abcdef along with its associated permissions
```

src/cmd/factory/impl/common.go
```diff
@@ -73,7 +74,7 @@ func generateAndRestoreItems(
 		items:        items,
 	}}
 
-	// TODO: fit the desination to the containers
+	// TODO: fit the designation to the containers
```
ryanfkeepers, line 77:
```suggestion
	// TODO: fit the destination to the containers
```

src/internal/connector/onedrive/restore.go
```diff
@@ -45,6 +46,8 @@ func RestoreCollections(
 		metrics        support.CollectionMetrics
 		folderPerms    map[string][]UserPermission
 		canceled       bool
+		parentPerms    []UserPermission
+		ok             bool
```
ryanfkeepers, line 50: these variables only exist within the loop.  declaring them outside risks unwanted behavior.  safer to only declare them within the loop.

src/internal/connector/onedrive/restore.go
```diff
@@ -65,18 +68,20 @@ func RestoreCollections(
...
 	// Iterate through the data collections and restore the contents of each
 	for _, dc := range dcs {
-		parentPerms, ok := parentPermissions[dc.FullPath().String()]
-		if !ok {
-			onedrivePath, err := path.ToOneDrivePath(dc.FullPath())
-			if err != nil {
-				errUpdater(dc.FullPath().String(), fmt.Errorf("invalid restore path"))
-			}
+		if opts.RestorePermissions {
+			parentPerms, ok = parentPermissions[dc.FullPath().String()]
```
ryanfkeepers, line 72: parentPerms isn't getting used anywhere that I can see.  
```suggestion
			_, ok = parentPermissions[dc.FullPath().String()]
```

src/internal/connector/onedrive/restore.go
```diff
@@ -65,18 +68,20 @@ func RestoreCollections(
 
 	// Iterate through the data collections and restore the contents of each
 	for _, dc := range dcs {
-		parentPerms, ok := parentPermissions[dc.FullPath().String()]
-		if !ok {
-			onedrivePath, err := path.ToOneDrivePath(dc.FullPath())
-			if err != nil {
-				errUpdater(dc.FullPath().String(), fmt.Errorf("invalid restore path"))
-			}
+		if opts.RestorePermissions {
```
ryanfkeepers, line 71: Nit: we're getting some pretty deeply nested code here.  I think this is a great opportunity to break it out (lines 72-83) into a separate function, both for cleanliness and ease of unit testing.
```go
if opts.RestorePermissions {
  if err := restorePermssion(dc.FullPath()); err != nil {
    errUpdater(err)
  }
}
```

src/internal/connector/onedrive/restore.go
```diff
@@ -65,18 +68,20 @@ func RestoreCollections(
...
-				// root directory will not have permissions
-				parentPerms = []UserPermission{}
-			} else {
-				errUpdater(dc.FullPath().String(), fmt.Errorf("unable to find parent permissions"))
+				if len(onedrivePath.Folders) == 0 {
+					// root directory will not have permissions
+					parentPerms = []UserPermission{}
+				} else {
+					errUpdater(dc.FullPath().String(), fmt.Errorf("unable to find parent permissions"))
+				}
```
ryanfkeepers, line 84: I know I'm picking at old code, but I'm a bit confused here, so I have to ask about it.  So, if we have a non-zero list of folders, then we record an "unable to find" error?  That seems counter-intuitive.

src/internal/connector/onedrive/restore.go
```diff
@@ -207,7 +214,17 @@ func RestoreCollection(
 					restoredIDs[trimmedName] = itemID
 
 					deets.Add(itemPath.String(), itemPath.ShortRef(), "", true, itemInfo)
+
+					// Mark it as success without processing .meta
+					// file if we are not restoring permissions
+					if !enablePermissionsRestore {
+						metrics.Successes++
+					}
 				} else if strings.HasSuffix(name, MetaFileSuffix) {
```
ryanfkeepers, line 223: Nit: this if-else chain is also a great place to break out code into small, testable functions.  The more handling we add to it (like permission restore control here), the more unwieldy it becomes.

src/internal/connector/onedrive/restore.go
```diff
@@ -127,6 +133,7 @@ func RestoreCollection(
 	deets *details.Builder,
 	errUpdater func(string, error),
 	permissionIDMappings map[string]string,
+	enablePermissionsRestore bool,
```
ryanfkeepers, line 136:
```suggestion
	restorePerms bool,
```

ryanfkeepers (approved):
Some nits.  A follow-up PR for cleanup and refactoring could be nice.

src/internal/connector/onedrive/restore.go
meain: It gets used as an arg to RestoreCollections
https://github.com/alcionai/corso/blob/8c91de1ca2be61145cd7c73cd66f62c8fae5642f/src/internal/connector/onedrive/restore.go#L92

src/internal/connector/onedrive/restore.go
meain: We try to get premissions for the parent folder [here](https://github.com/alcionai/corso/blob/8c91de1ca2be61145cd7c73cd66f62c8fae5642f/src/internal/connector/onedrive/restore.go#L72). Failing that we check if the parent is the root, which might not have separate permissions associated to it. If permissions are not available form `parentPermissions` and it is not the root, then we fail saying that we are not able to find permissions of the parent directory.

ryanfkeepers: Hmm, I think a message like `"directory has no permissions"` is more accurate in this case.

meain: Just to clarify, we do have permissions for the folder that we are about to restore, but we are not able to find the permissions for its parent(which should ideally have been restored already) and due to that we are not able to figure out the final permissions for this folder as it has to be computed based on what the parent has. Let me change the error message to "unable to compute item permissions", but let me know if you still think it should be something else.

PR #2246, opened 2023-01-24
standardize deleteContainer func in exch api

ryanfkeepers:
## Description

Minor refactor that came up while hunting bugs.
For the life of me, I cannot find the reason for
read/write counts being off in event incrementals.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #2022

## Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -76,7 +76,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreContact() {
 
 	defer func() {
 		// Remove the folder containing contact prior to exiting test
-		err = suite.ac.Contacts().DeleteContactFolder(ctx, userID, folderID)
+		err = suite.ac.Contacts().DeleteContainer(ctx, userID, folderID)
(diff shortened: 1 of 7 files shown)


src/internal/connector/exchange/restore_test.go
```diff
@@ -266,9 +267,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreExchangeObject() {
 			)
 			assert.NoError(t, err, support.ConnectorStackErrorTrace(err))
 			assert.NotNil(t, info, "item info is populated")
-
-			cleanupError := test.cleanupFunc(ctx, userID, destination)
-			assert.NoError(t, cleanupError)
+			assert.NoError(t, deleters[test.category].DeleteContainer(ctx, userID, destination))
```
ashmrtn, line 270: nit: if this is running as a cleanup function it should probably be run in a defer so we hit it in all cases

ashmrtn (approved):

PR #2251, opened 2023-01-25
Fix queuing in OneDrive populateItems loop

vkamra:
## Description

This addresses an issue in the OneDrive `populateItems` logic where we end up queuing all items in the collection for download. 

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -27,7 +27,7 @@ const (
 	collectionChannelBufferSize = 5
 
 	// TODO: Tune this later along with collectionChannelBufferSize
-	urlPrefetchChannelBufferSize = 5
+	urlPrefetchChannelBufferSizeDefault = 5
 
 	// Max number of retries to get doc from M365
 	// Seems to timeout at times because of multiple requests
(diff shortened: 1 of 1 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -253,6 +255,9 @@ func (oc *Collection) populateItems(ctx context.Context) {
 			}
 
 			itemReader := lazy.NewLazyReadCloser(func() (io.ReadCloser, error) {
```
ryanfkeepers, line 257: Thoughts on moving this up to line 211 instead?  That would treat the item download as part of the lazy read.  The only performance gain we achieve by separating the lazy reader and the item download is the pre-download for the first 5 items, and then we're bound to kopia multi-threading.

dadams39: +1

vkamra: We discussed a bit offline - we also get some error handling for the URL i.e. this helps validate the URL is actually valid. Follow-up from the discussion was that we will need to handle the case where the URL might expire for long-running operations - so it might make sense to do what you're suggesting (defer the reader creation and GET to when we start downloading the file and also add some revalidation logic in the reader)

PR #2266, opened 2023-01-26
add itemClient for re-usable od item downloads

### LINKED ISSUE #2262: OneDrive connectors overflow os file descriptors
## Issue
For a user with multi-thousands of oneDrive files, a corso backup can consume in excess of 1700 file descriptors (http sockets, in this case).  This is untenable for environments with small ulimit restrictions on file descriptors.

## Acceptance Criteria
* Investigation: discover what is causing/allowing the socket count to get out of hand.
* Fix: minimize socket consumption.

ryanfkeepers:
## Description

onedrive currently constructs a new http client
for every file it downloads.  This causes the OS
to generate extra sockets, and hang onto them
after the download is complete.  Replacing these
one-off clients with a singular, re-used client-
which is the behavior and standard suggested
for golang http clients- minimizes system
resource consumption.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* closes #2262

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -4,11 +4,13 @@ package onedrive
 import (
 	"context"
 	"io"
+	"net/http"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"github.com/pkg/errors"
 	"github.com/spatialcurrent/go-lazy/pkg/lazy"
 
 	"github.com/alcionai/corso/src/internal/connector/graph"
@@ -43,6 +45,9 @@ var (
 
 // Collection represents a set of OneDrive objects retrieved from M365
 type Collection struct {
+	// configured to handle large item downloads
+	itemClient *http.Client
+
 	// data is used to share data streams with the collection consumer
 	data chan data.Stream
 	// folderPath indicates what level in the hierarchy this collection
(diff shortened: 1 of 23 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -64,12 +69,13 @@ type Collection struct {
 
 // itemReadFunc returns a reader for the specified item
 type itemReaderFunc func(
-	ctx context.Context,
+	hc *http.Client,
```
vkamra, line 72: Why remove `ctx`?

src/internal/connector/onedrive/collection.go
```diff
@@ -198,11 +205,16 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
-	for _, item := range oc.driveItems {
+	for id, item := range oc.driveItems {
 		if oc.ctrl.FailFast && errs != nil {
 			break
 		}
 
+		if item == nil {
+			errUpdater(id, errors.New("nil item"))
+			continue
+		}
```
vkamra, line 216: Is this expected?

src/internal/connector/onedrive/collection.go
vkamra: Nevermind - I guess it's no longer used.

vkamra (approved):

src/internal/connector/onedrive/collection.go
ryanfkeepers: yup

src/internal/connector/onedrive/collection.go
ryanfkeepers: Not particularly, but I had a test that paniced on it.  Figured we're better off protecting ourselves against a panic than risking it.

PR #2268, opened 2023-01-26
wrap item get in lazy reader

### LINKED ISSUE #2262: OneDrive connectors overflow os file descriptors

ryanfkeepers:
## Description

Wraps the onedrive item download control into
the lazy reader creation, so that items are not
fetched from graph until kopia decides it wants
the bytes.  This should only occurr after other
checks, like mod-time comparison, have passed,
thus giving us kopia-assists for bakup.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* closes #2262

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -251,8 +251,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 				for i := 1; i <= maxRetries; i++ {
 					_, itemData, err = oc.itemReader(oc.itemClient, item)
-					if err == nil || graph.IsErrTimeout(err) == nil {
-						// retry on Timeout type errors, break otherwise.
+					if err == nil {
 						break
 					}
 
@@ -271,13 +270,22 @@ func (oc *Collection) populateItems(ctx context.Context) {
 				progReader, closer := observe.ItemProgress(ctx, itemData, observe.ItemBackupMsg, itemName, itemSize)
 				go closer()
 
-				// Item read successfully, add to collection
-				atomic.AddInt64(&itemsRead, 1)
-				// byteCount iteration
-				atomic.AddInt64(&byteCount, itemSize)
-
 				return progReader, nil
 			})
+
+			// This can cause inaccurate counts.  Right now it counts all the items
+			// we intend to read.  Errors within the lazy readCloser will create a
+			// conflict: an item is both successful and erroneous.  But the async
+			// control to fix that is more error-prone than helpful.
+			//
+			// TODO: transform this into a stats bus so that async control of stats
+			// aggregation is handled at the backup level, not at the item iteration
+			// level.
+			//
+			// Item read successfully, add to collection
+			atomic.AddInt64(&itemsRead, 1)
+			// byteCount iteration
+			atomic.AddInt64(&byteCount, itemSize)
 
 			oc.data <- &Item{
 				id:   itemName,
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -207,7 +207,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 	for id, item := range oc.driveItems {
 		if oc.ctrl.FailFast && errs != nil {
-			continue
+			break
 		}
 
 		if item == nil {
@@ -231,10 +231,10 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 			switch oc.source {
 			case SharePointSource:
-				itemInfo.SharePoint = sharePointItemInfo(item, *item.GetSize())
+				itemInfo.SharePoint = sharePointItemInfo(item, itemSize)
 				itemInfo.SharePoint.ParentPath = parentPathString
 			default:
-				itemInfo.OneDrive = oneDriveItemInfo(item, *item.GetSize())
+				itemInfo.OneDrive = oneDriveItemInfo(item, itemSize)
 				itemInfo.OneDrive.ParentPath = parentPathString
 			}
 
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -223,57 +223,62 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+				}
+
+				// display/log the item download
 				progReader, closer := observe.ItemProgress(ctx, itemData, observe.ItemBackupMsg, itemName, itemSize)
 				go closer()
+
+				// Item read successfully, add to collection
+				atomic.AddInt64(&itemsRead, 1)
+				// byteCount iteration
+				atomic.AddInt64(&byteCount, itemSize)
```
ryanfkeepers, line 277: Currently stuck on this, and line 293 (can't include in comment span, it's the `reportAsCompleted()`) with an async control issue.

If we run the code as-is, the loop completes and `reportAsCompleted` gets called before all items count their successes and size.  However, if we move these lines back to their original spot (272), then we get a success and byte count before we successfully read the item.  IE: any kopia assisted skip or error would cause the count/size to be off.

The waitGroup and eager downloads previously protected us from this condition.  Ideal refactor would be for stats to accrue asynchronously in a bus, rather than be bound up in the collection instance.  In the interim, a smaller hack might be to have the caller block until all items are completed (like operations does with GC and collections), and calling `reportAsCompleted` in a goroutine that's blocked on the wg.

ryanfkeepers: @vkamra Tried some async control, ended up overwhelmed with it.  Going to punt to the easy route for now and allow successes and sizes to accrue outside the lazy reader.  Opens up the possibility of mismatched success/error counts.  But we're getting into some larger design rework to handle it.  So, at least as a temporary change, this state keeps things moving.

An upcoming PR will address expired URL retries.

src/internal/connector/onedrive/collection.go
```diff
@@ -223,52 +223,65 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+				// display/log the item download
 				progReader, closer := observe.ItemProgress(ctx, itemData, observe.ItemBackupMsg, itemName, itemSize)
 				go closer()
+
 				return progReader, nil
 			})
 
+			// This can cause inaccurate counts.  Right now it counts all the items
+			// we intend to read.  Errors within the lazy readCloser will create a
+			// conflict: an item is both successful and erroneous.  But the async
```
ashmrtn, line 278: are the errors propagated to the status updater eventually? Do we need to worry about triggering log messages/failures for the number of items expected not being equal to the sum of the successes and failures in the status updater code?

src/internal/connector/onedrive/collection_test.go
```diff
@@ -235,18 +255,28 @@ func (suite *CollectionUnitTestSuite) TestCollectionReadError() {
 
 			mockItem := models.NewDriveItem()
 			mockItem.SetId(&testItemID)
+			mockItem.SetId(&testItemID)
```
ashmrtn, line 258: nit: duplicate line
```suggestion
```

ashmrtn (approved):

src/internal/connector/onedrive/collection.go
```diff
@@ -223,52 +223,65 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
 			itemReader := lazy.NewLazyReadCloser(func() (io.ReadCloser, error) {
+				// Read the item
+				var (
+					itemData io.ReadCloser
+					err      error
+				)
+
+				for i := 1; i <= maxRetries; i++ {
+					_, itemData, err = oc.itemReader(oc.itemClient, item)
+					if err == nil || graph.IsErrTimeout(err) == nil {
```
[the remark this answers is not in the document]
ashmrtn: >We should check how kopia handles dir entries for items that fail.

closing the loop on this. Checked the kopia code, it does not return a dir entry if an item has failed so that item will not be available in future backups nor will it be taken into account for kopia-assisted incrementals. Expanding a unit test slightly to ensure files that had errors during read do not result in a file that can be read from kopia

src/internal/connector/onedrive/collection.go
```diff
@@ -207,7 +207,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 	for id, item := range oc.driveItems {
 		if oc.ctrl.FailFast && errs != nil {
-			break
```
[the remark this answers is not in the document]
ryanfkeepers: Ah, yep, that was leftover from a different poc.

src/internal/connector/onedrive/collection.go
```diff
@@ -223,52 +223,65 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+						break
+					}
+
+					if i < maxRetries {
+						time.Sleep(1 * time.Second)
+					}
+				}
+
+				// check for errors following retries
+				if err != nil {
```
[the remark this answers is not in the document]
ryanfkeepers: I'll leave that to a different PR.

src/internal/connector/onedrive/collection.go
ryanfkeepers: depends on the error; if the lazy reader returns the error, kopia wrapper will catch it.  Yes, if errors occur, counts will not be equal in the status updater.  That's not a blocker atm.

Issue #2271, opened 2023-01-26
[Bug]: Unable to backup OneDrive


hmkias:
### What happened?

Testing One Drive, encountering "No Data" Status. Checked the SKU and One drive data, everything looks Ok. 

### Corso Version?

version: v0.1.0

### Where are you running Corso?

Centos 7 VM 

### Relevant log output

```shell
corso backup create onedrive --user <xxxxxx@xxxxxxx.onmicrosoft.com> --log-level debug
2023-01-26T06:54:58.469Z        DEBUG   config/config.go:174    found config file       {"configFile": "/root/.corso.toml"}
Connecting to repository:     13s done 
Connecting to M365:      0s done 
Connecting to M365:      0s done 
Connecting to M365:      0s done 
Connecting to M365:      0s done 
Discovering items to backup:      0s done 
Discovered 0 items to backup   
Backing up data:         done 
2023-01-26T06:55:14.720Z        DEBUG   kopia/upload.go:289     reading item    {"path": "71433e16-3db6-476f-946f-b5f01fd93cdf/onedriveMetadata/details/details/details"}
  Started At            ID                                    Status              Resource Owner                       
  2023-01-26T06:55:13Z  8436a1ef-387f-49f0-8406-0e8bcca8d75b  No Data (0 errors)  <xxxxxx@xxxxxxx.onmicrosoft.com>
2023-01-26T06:55:16.852Z        DEBUG   config/config.go:174    found config file       {"configFile": "/root/.corso.toml"}
```

ntolia:
Logs from Discord showed that Corso is not recognizing the M365 SKU in use (ENTERPRISEPREMIUM). We are moving away from this method of identifying OneDrive licensing in the near future.

```
2023-01-26T07:22:54.738Z        DEBUG   config/config.go:174    found config file       {"configFile": "/root/.corso.toml"}
2023-01-26T07:23:13.421Z        DEBUG   connector/data_collections.go:163       Creating OneDrive collections   {"user": "ayyanar@M365x47292038.onmicrosoft.com"}
2023-01-26T07:23:13.746Z        DEBUG   onedrive/drive.go:128   User ayyanar@M365x47292038.onmicrosoft.com does not have a license for OneDrive
2023-01-26T07:23:13.747Z        DEBUG   operations/backup.go:193        Backed up 0 directories and 0 files
[...]

PR #2275, opened 2023-01-26
Remove OneDrive license GUID check

vkamra:
## Description

This check is brittle and shouldn't be needed. It should be sufficient to check whether a user has a 
drive or not (similar to what we now do for exchange mailboxes).

Validated with the existing integration test.

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2271

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/drive.go
+++ b/src/internal/connector/onedrive/drive.go
@@ -79,7 +79,7 @@ func userDrives(ctx context.Context, service graph.Servicer, user string) ([]mod
 			detailedError := support.ConnectorStackErrorTrace(err)
 			if strings.Contains(detailedError, userMysiteURLNotFound) ||
 				strings.Contains(detailedError, userMysiteNotFound) {
-				logger.Ctx(ctx).Debugf("User %s does not have a drive", user)
+				logger.Ctx(ctx).Infof("User %s does not have a drive", user)
 				return make([]models.Driveable, 0), nil // no license
 			}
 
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/drive.go
```diff
@@ -116,21 +67,11 @@ func siteDrives(ctx context.Context, service graph.Servicer, site string) ([]mod
...
 
-	hasDrive, err := hasDriveLicense(ctx, service, user)
-	if err != nil {
-		return nil, errors.Wrap(err, user)
-	}
-
-	if !hasDrive {
-		logger.Ctx(ctx).Debugf("User %s does not have a license for OneDrive", user)
-		return make([]models.Driveable, 0), nil // no license
-	}
```
vkamra, line 132: The check below is sufficient.

gmatev (approved):

src/internal/connector/onedrive/drive.go
gmatev: @vkamra How does this work if the user drive is not provisioned yet?

vkamra: The check on line 80 below catches that case.

Issue #2267, opened 2023-01-26
OneDrive does not use kopia assisted incrementals


ryanfkeepers:
## issue
The onedrive lazy reader only wraps the item transmission, not its download.  This means that even in the case where kopia-assisted incrementals could be used, they still incur the majority of the effort in the transaction.

## Acceptance Criteria
Migrate the http download of the item into the lazy reader so that kopia assistance is maximized on older file backups.

vkamra:
Fixed

PR #2283, opened 2023-01-26
re-fetch file download url after expiration

ryanfkeepers:
## Description

If a drive item goes over its 1 hour jwt expiration to download the backing file, re-fetch the item
and use the new download url to get the file.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2267

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -26,7 +26,13 @@ const (
 	errCodeMailboxNotEnabledForRESTAPI = "MailboxNotEnabledForRESTAPI"
 )
 
-var err401Unauthorized = errors.New("401 unauthorized intercepted")
+var (
+	Err401Unauthorized = errors.New("401 unauthorized")
+	// normally the graph client will catch this for us, but in case we
+	// run our own client Do(), we need to translate it to a timeout type
+	// failure locally.
+	Err429TooManyRequests = errors.New("429 too many requests")
+)
 
 // The folder or item was deleted between the time we identified
 // it and when we tried to fetch data for it.
@@ -121,6 +127,27 @@ func isTimeoutErr(err error) bool {
 	}
 }
 
+type ErrThrottled struct {
+	common.Err
+}
+
+func IsErrThrottled(err error) error {
+	if errors.Is(err, Err429TooManyRequests) {
+		return err
+	}
+
+	if asThrottled(err) {
+		return err
+	}
+
+	return nil
+}
+
+func asThrottled(err error) bool {
+	e := ErrThrottled{}
+	return errors.As(err, &e)
+}
+
 type ErrUnauthorized struct {
 	common.Err
 }
@@ -129,16 +156,12 @@ func IsErrUnauthorized(err error) error {
 	// TODO: refine this investigation.  We don't currently know if
 	// a specific item download url expired, or if the full connection
 	// auth expired.
-	if errors.Is(err, err401Unauthorized) {
+	if errors.Is(err, Err401Unauthorized) {
 		return err
 	}
 
 	if asUnauthorized(err) {
 		return err
-	}
-
-	if hasErrorCode(err, "foo") {
-		return ErrInvalidDelta{*common.EncapsulateError(err)}
 	}
 
 	return nil
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -26,9 +26,7 @@ const (
 	errCodeMailboxNotEnabledForRESTAPI = "MailboxNotEnabledForRESTAPI"
 )
 
-var (
-	err401Unauthorized = errors.New("401 unauthorized intercepted")
-)
+var err401Unauthorized = errors.New("401 unauthorized intercepted")
 
 // The folder or item was deleted between the time we identified
 // it and when we tried to fetch data for it.
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -31,7 +31,8 @@ var (
 	// normally the graph client will catch this for us, but in case we
 	// run our own client Do(), we need to translate it to a timeout type
 	// failure locally.
-	Err429TooManyRequests = errors.New("429 too many requests")
+	Err429TooManyRequests    = errors.New("429 too many requests")
+	Err503ServiceUnavailable = errors.New("503 Service Unavailable")
 )
 
 // The folder or item was deleted between the time we identified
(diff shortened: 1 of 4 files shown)


src/internal/connector/graph/errors.go
```diff
@@ -102,6 +104,51 @@ func asTimeout(err error) bool {
...
+		return err
+	}
+
+	if asUnauthorized(err) {
+		return err
+	}
+
+	if hasErrorCode(err, "foo") {
+		return ErrInvalidDelta{*common.EncapsulateError(err)}
+	}
```
ashmrtn, line 142: I'm guessing this should be removed, refined, and readded later?

src/internal/connector/graph/service_helper.go
```diff
@@ -94,31 +94,51 @@ func (handler *LoggingMiddleware) Intercept(
...
+		}
+
+		if resp.StatusCode != http.StatusTooManyRequests && (resp.StatusCode/100) != 2 {
+			logger.Ctx(ctx).Infow("graph api error", "status", resp.Status, "method", req.Method, "url", req.URL)
+		}
+	}
+
+	// this is a hack to address 401 responses from graph api not coming
+	// through as errors.
+	if err == nil && resp.StatusCode == http.StatusUnauthorized {
```
ashmrtn, line 140: 🥲

src/internal/connector/graph/service_helper.go
```diff
@@ -94,31 +94,51 @@ func (handler *LoggingMiddleware) Intercept(
...
 		logger.Ctx(ctx).Errorw("non-2xx graph api response", metadata...)
+	} else {
+		// special case for supportability: log all throttling cases.
+		if resp.StatusCode == http.StatusTooManyRequests {
+			logger.Ctx(ctx).Infow("graph api throttling", "method", req.Method, "url", req.URL)
+		}
+
+		if resp.StatusCode != http.StatusTooManyRequests && (resp.StatusCode/100) != 2 {
+			logger.Ctx(ctx).Infow("graph api error", "status", resp.Status, "method", req.Method, "url", req.URL)
+		}
```
ashmrtn, line 135: having this in the else block will only log the starting message when we're not running in debug mode. This can make tracking down throttling more difficult. Maybe we should just always log this even if it means we'll see repeated messages in the log? At least then the "graph api throttling" part will be stable

src/internal/connector/onedrive/collection.go
```diff
@@ -251,7 +252,27 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+						break
+					}
+
+					if graph.IsErrUnauthorized(err) != nil {
+						// assume unauthorized requests are a sign of an expired
+						// jwt token, and that we've overrun the available window
+						// to download the actual file.  Re-downloading the item
+						// will refresh that download url.
+						di, diErr := getDriveItem(ctx, oc.service, oc.driveID, itemID)
+						if diErr != nil {
```
ashmrtn, line 265: nit: should probably make sure we don't error if the item was deleted

src/internal/connector/graph/service_helper.go
ashmrtn: Can you add a comment to the code explaining how 401 responses would appear to higher layers of the stack if we didn't have this. That way people can really see why we need this

ashmrtn (approved):

src/internal/connector/graph/service_helper.go
```diff
@@ -94,31 +94,51 @@ func (handler *LoggingMiddleware) Intercept(
...
 
 	if (resp.StatusCode / 100) == 2 {
-		return resp, err
-	}
-
-	// special case for supportability: log all throttling cases.
-	if resp.StatusCode == http.StatusTooManyRequests {
-		logger.Ctx(ctx).Infow("graph api throttling", "method", req.Method, "url", req.URL)
-	}
+		if logger.DebugAPI || os.Getenv(logGraphRequestsEnvKey) != "" {
```
vkamra, line 97: Let's dump response here (like line 114) but without the body i.e. httputil.DumpResponse(resp, false)

src/internal/connector/graph/service_helper.go
vkamra: Why is this happening? We saw this with 429 errors also

src/internal/connector/graph/service_helper.go
ryanfkeepers: I'm not too concerned.  The debugging mode isn't representative of any of our standard use cases, and will automatically capture throttling responses as part of the non-2xx log.

src/internal/connector/onedrive/collection.go
ryanfkeepers: This can be followed up in another PR.

src/internal/connector/graph/service_helper.go
vkamra: We shouldn't catch this here. Let's move the error handling into `downloadItem` where we should check resp.StatusCode

src/internal/connector/onedrive/item.go
```diff
@@ -25,21 +25,25 @@ const (
 	downloadURLKey = "@microsoft.graph.downloadUrl"
 )
 
+// generic drive item getter
```
vkamra, line 28: @ryanfkeepers - we need to add a unit test for the reader with a mock client that simulates the errors we hit. The refactor you did actually enables that pretty easily (🤞 )

ryanfkeepers: Sounds good.  I'm unaware of the specific errors, if there's a link it'd be good to record those in a follow-up issue.

src/internal/connector/onedrive/collection.go
```diff
@@ -251,7 +252,32 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+
+						item = di
+
+						continue
+
+					} else if graph.IsErrTimeout(err) == nil &&
+						graph.IsErrThrottled(err) == nil &&
+						graph.IsSericeUnavailable(err) == nil {
+						// TODO: graphAPI will provides headers that state the duration to wait
+						// in order to succeed again.  The one second sleep won't cut it here.
```
meain, line 278: I believe the `Retry-After` header is only available for throttled requests, but they are handled internally within the [default retry handler](https://github.com/microsoft/kiota-http-go/blob/af8d9d1d0fbf3841ff2dc1d87b754ef0d47b2368/retry_handler.go#L127) within kiota.

ryanfkeepers: That gets handled in _some_ cases.  Fun fact about the default retry handler: it doesn't work on GET requests.

```
func (middleware RetryHandler) isRetriableRequest(req *nethttp.Request) bool {
	isBodiedMethod := req.Method == "POST" || req.Method == "PUT" || req.Method == "PATCH"
```

vkamra: > Nevermind me! I should read the whole func before I respond. Thanks for the reminder.

Same :). I made the same assumption when I was reading the code with @meain last time

Issue #2252, opened 2023-01-25
OneDrive permissions backup: Toggle to disable permissions backup


meain:
Add a hidden toggle so that ark can disable permissions backup for now, primarily as we have a much higher chance of hitting throttling with the permissions being stored.

Ref: https://github.com/alcionai/corso/pull/2148#pullrequestreview-1268580524

Issue #2230, opened 2023-01-23
OneDrive Metadata Backup/Restore: Add ability to restore older backup without metadata


meain:
This can be done by adding versioning to backups so that we can detect if the data was stored with a single file or multiple files. This will also prove useful as we start adding more features for other backups as well.

### TBD
- How would this affect restartability / incrementals?
- Versioning bump as of now is only applicable to OneDrive, should we have something which lets us bump version based on if we are backing up OneDrive or something else?

Ref: https://github.com/alcionai/corso/pull/2148#issuecomment-1399900891

meain:
@vkamra @gmatev Should we show a warning or error out if there are no permissions available as it is an older backup but the `--restore-permissions` flag was passed?

PR #2296, opened 2023-01-27
Add hidden flag to disable permissions backup

meain:
## Description

This will let users completely disable backing up permissions altogether.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/2252

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - OneDrive file/folder permissions can now be backed up and restored
 - Add `--restore-permissions` flag to toggle restoration of OneDrive permissions
+- Add a hidden `--disable-permissions-backup` to completely disable backing up permissions
```
ryanfkeepers, line 21: Do we want to announce hidden flags in the changelog? 🤔

src/cli/options/options.go
```diff
@@ -78,3 +82,16 @@ func DisableIncrementals() func(*pflag.FlagSet) {
...
+
+// Adds the hidden '--disable-permissions-backup' cli flag which, when
+// set, disables backing up permissions.
+func DisablePermissionsBackup() func(*pflag.FlagSet) {
+	return func(fs *pflag.FlagSet) {
+		fs.BoolVar(
+			&disablePermissionsBackup,
+			"disable-permissions-backup",
+			false,
+			"Disable backing up permissions for OneDrive")
```
ryanfkeepers, line 94:
```suggestion
			"Disable backing up item permissions for OneDrive")
```

src/pkg/control/options.go
```diff
@@ -75,4 +75,10 @@ type Toggles struct {
 	// DisableIncrementals prevents backups from using incremental lookups,
 	// forcing a new, complete backup of all data regardless of prior state.
 	DisableIncrementals bool `json:"exchangeIncrementals,omitempty"`
+
+	// DisablePermissionsBackup is used to disable backups for
+	// permissions. This can be used by other tools to completely
+	// disable the backup and restore functionality for backups as
+	// they consume more API credits and can cause increased throttling.
```
ryanfkeepers, line 82:
```suggestion
	// DisablePermissionsBackup is used to disable backups of item
	// permissions. Permission metadata increases graph api call count,
	// so disabling their retrieval when not needed is advised.
```

ryanfkeepers (approved):
Some nits, but looks good overall

src/internal/connector/onedrive/collection.go
```diff
@@ -258,9 +267,38 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+				if !oc.ctrl.ToggleFeatures.DisablePermissionsBackup {
+					itemMeta, itemMetaSize, err = oc.itemMetaReader(ctx, oc.service, oc.driveID, item)
+				} else {
+					// We are still writing the metadata file but with
+					// empty permissions as we are not sure how the
+					// restore will be called.
+					itemMeta = io.NopCloser(strings.NewReader("{}"))
+					itemMetaSize = 2
+					err = nil
+				}
```
ashmrtn, line 301: it's a little hard to follow when we set err = nil so we hit the break in the else-block. It would be more straightforward to read if we just make an if-block that breaks when needed and let everything else have it's own control flow
```suggestion
				if oc.ctrl.ToggleFeatures.DisablePermissionsBackup {
					// We are still writing the metadata file but with
					// empty permissions as we are not sure how the
					// restore will be called.
					itemMeta = io.NopCloser(strings.NewReader("{}"))
					itemMetaSize = 2

					break
				}

				itemMeta, itemMetaSize, err = oc.itemMetaReader(ctx, oc.service, oc.driveID, item)
```

src/internal/connector/onedrive/collection.go
```diff
@@ -258,9 +267,38 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
+			// TODO(meain): refactor these multiple retry logic
 			for i := 1; i <= maxRetries; i++ {
-				itemInfo, itemData, itemMeta, err = oc.itemReader(ctx, oc.service, oc.driveID, item)
+				if !oc.ctrl.ToggleFeatures.DisablePermissionsBackup {
+					itemMeta, itemMetaSize, err = oc.itemMetaReader(ctx, oc.service, oc.driveID, item)
+				} else {
+					// We are still writing the metadata file but with
+					// empty permissions as we are not sure how the
+					// restore will be called.
+					itemMeta = io.NopCloser(strings.NewReader("{}"))
```
ashmrtn, line 298: given that we're deciding to store "empty" permissions and we now have flags for restoring and backing up permissions it feels like we should be printing a warning or some other message if a user tries to do `restore permissions(backup no permissions)`

This should be done in a separate PR if we decide to add it, but wanted to surface it here for discussion

meain: I guess that would be useful. I had a similar question in https://github.com/alcionai/corso/issues/2230#issuecomment-1402331725 . The problem with warning when we backup with `--disable-permissions-backup` is that we are not adding any data that as of now mentions that we avoided backing up permissions of it just didn't have any permissions.

ashmrtn: made #2357 to track this

src/internal/connector/onedrive/collection.go
meain: Woops, this is such an obvious refactor :D

ashmrtn: no worries, it can be easy to miss stuff like this if there's a lot of other things going on too 😄

Issue #2299, opened 2023-01-27
Use LargeFile http client for attachment downloads.


ryanfkeepers:
## Issue
Email and event attachment downloads are prone to ctx timeouts because they use the standard http client, which has a 90 second timeout.  We recently introduced a LargeItem client specifically for downloading items with loose timeout concerns.  We should use that for attachment downloads instead.

## Acceptance Criteria
* When constructing an `/exchange/api Client{}`, generate and store a LargeItem http client in the api.Client struct.
* Move attachment downloads out of the Serialize() func.  It should probably live in `GetItem()` instead.
* Use the LargeItem http client to download attachments.

PR #2301, opened 2023-01-27
quick graph client refactor

ryanfkeepers:
## Description

A couple minor refactors.  First, combines the
graph/service_helper.go and graph/service.go
files into a single file (just lift'n'shift, no logic
changes).  Second, renames CreateHTTPClient
to just HTTPClient.  Third, replaces the new
LargeItemClient constructor with extensible
options for the http client, starting with a
NoTimeout option.  This will provide longer
term maintenance over multiple constructors.

This change sets up QoL for building multiple services
and clients within api layer handlers, which will appear
in an immediately following PR.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #2299

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/graph/service.go
```diff
@@ -47,7 +64,7 @@ func (s Service) Client() *msgraphsdk.GraphServiceClient {
 
 // Seraialize writes an M365 parsable object into a byte array using the built-in
 // application/json writer within the adapter.
-func (s Service) Serialize(object absser.Parsable) ([]byte, error) {
+func (s Service) Serialize(object serialization.Parsable) ([]byte, error) {
```
dadams39, line 67: Please do not change the library alias. The abstraction serialization library deals with the parsables and other key indicators. Additionally, the change adds more lines to inspect which is not necessary at this time.

src/internal/connector/graph/service.go
```diff
@@ -61,6 +78,90 @@ func (s Service) Serialize(object absser.Parsable) ([]byte, error) {
...
+// ---------------------------------------------------------------------------
+
+type clientConfig struct {
+	noTimeout bool
+}
+
+type option func(*clientConfig)
+
+// populate constructs a clientConfig according to the provided options.
+func (c *clientConfig) populate(opts ...option) *clientConfig {
```
dadams39, line 92: Request a name change. We have a lot of populate functions. It would be nice to distinguish from other methods.

src/internal/connector/graph/service.go
```diff
@@ -120,3 +221,72 @@ type ContainerResolver interface {
...
+	var (
+		ctx       = req.Context()
+		resp, err = pipeline.Next(req, middlewareIndex)
+	)
+
+	if resp == nil {
+		return resp, err
+	}
+
+	if (resp.StatusCode / 100) == 2 {
```
dadams39, line 246: Add comment to identify to describe which errors are being caught in the 200 range

src/internal/connector/graph/service.go
```diff
@@ -61,6 +78,90 @@ func (s Service) Serialize(object absser.Parsable) ([]byte, error) {
...
+func NoTimeout() option {
+	return func(c *clientConfig) {
+		c.noTimeout = true
+	}
+}
+
+// CreateAdapter uses provided credentials to log into M365 using Kiota Azure Library
+// with Azure identity package. An adapter object is a necessary to component
+// to create  *msgraphsdk.GraphServiceClient
+func CreateAdapter(tenant, client, secret string, opts ...option) (*msgraphsdk.GraphRequestAdapter, error) {
```
ashmrtn, line 120: I'm curious as to the choice of passing a vararg of function pointers rather than just a `clientConfig` struct that's then consumed by `HTTPClient()`. It seems like the logic would be easier to follow on the latter while still allowing the user to modify the full set of parameters. It may also be more flexible as we wouldn't be bound to functions that only take `*clientConfig` as a parameter (e.x. being able to set the timeout to any integer vs. having different functions for different integer values)

src/internal/connector/graph/service.go
ryanfkeepers: I'm not confident about there being dangers here.  The alias removal doesn't prevent the library from identifying the Parsable interface, and there's no collision with other library names.  Arbitrary aliasing's primary effect is to interfere with automatic import adjustments form formatters and other linters, since the alias cannot always be removed.  The alias seems to me like abstraction for its own sake, which isn't particularly helpful.

src/internal/connector/graph/service.go
ryanfkeepers: The populate funcs that I see are all methods on discrete structs, which gives distinct space for each one to clearly describe its behavior without conflict of other methods.  Changing the name here is akin to asking to change the pattern in all cases throughout the code.  Is that a change you believe will resolve confusion in long term maintenance of the app?

src/internal/connector/graph/service.go
ryanfkeepers: I don't follow your statement.  The 200 range should not include errors.

src/internal/connector/graph/service.go
```diff
@@ -120,3 +221,72 @@ type ContainerResolver interface {
 	// Items returns the containers in the cache.
 	Items() []CachedContainer
 }
+
+// ---------------------------------------------------------------------------
+// Client Middleware
+// ---------------------------------------------------------------------------
+
+// LoggingMiddleware can be used to log the http request sent by the graph client
+type LoggingMiddleware struct{}
```
vkamra, line 230: Code movement only here - correct?

ryanfkeepers: yes

src/internal/connector/graph/service.go
```diff
@@ -61,6 +78,90 @@ func (s Service) Serialize(object absser.Parsable) ([]byte, error) {
...
+func (c *clientConfig) populate(opts ...option) *clientConfig {
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	return c
+}
+
+// apply updates the http.Client with the expected options.
+func (c *clientConfig) apply(hc *http.Client) {
```
vkamra, line 101: Add a small unit test for this please. Let's validate that apply does the right thing for the http client.

vkamra:
No blockers from my side other than the unit test request. Will let others approve when they're ready.

src/internal/connector/graph/service.go
ryanfkeepers: > I'm curious as to the choice of passing a vararg of function pointers rather than just a clientConfig struct that's then consumed by HTTPClient().

I like this pattern because it makes the base (ie, no options) constructor look like it produces a viable default value without further input.  This keeps the majority of our cases the most simple.  Requiring a configuration struct adds chaff complexity to the majority of default construction.  Empty and nil configurations can be confusing, as the caller doesn't know at first glance if they're _expected to_ be configuring further.  Options read more naturally like like deviations from the default.

> (e.x. being able to set the timeout to any integer vs. having different functions for different integer values)

Oh, this is a very easy change.  `NoTimeout()` -> `Timeout(0)`.  These funcs don't need to be bools.

src/internal/connector/graph/service.go
ryanfkeepers: Can do.

src/internal/connector/graph/service.go
ashmrtn: >Oh, this is a very easy change. NoTimeout() -> Timeout(0). These funcs don't need to be bools.

Ah, I see. I still find that rather annoying to read because now there's multiple levels of indirection to parse through just to figure out what parameters are being passed. For example, we have function call for param -> functor -> populate function -> apply function and then the reader has to realize that it's actually the union of all populate  calls which is kind of difficult to hold in your head at once. Also things like `Timeout(0), Timeout(5)` become valid and then ordering is thrown in on top of everything else

It still seems like a simple `DefaultConfig()` would give known good values even if it adds a little bit of chaff

ryanfkeepers: > Also things like Timeout(0), Timeout(5) become valid and then ordering is thrown in on top of everything else

Sure, but that's also a bit contrived, don't you think?  Under a microscope, yes, I absolutely agree there's weirdness.  But that weirdness is escaped in real usage by duplicate configurations being obviously in contention.  Now, some of this is on us as curators, sure.  For example, `NoTimeout(), Timeout(N)`, is both very weird and less obvious.  But that can be avoided by ensuring that we have either Timeout() or NoTimeout(), and not both.

> I still find that rather annoying to read because now there's multiple levels of indirection to parse through just to figure out what parameters are being passed.

In the perspective of calling the constructor, or handling it?  Personally, I think it's a tradeoff where usage of the constructor gets more readable, with only a minor increase in complexity when processing construction.

We're both soapboxing a bit about design philosophy and preferred patterns here.  This is good, we want to pin down the patterns that we prefer (or prefer to avoid) within the code.  But I also feel like this is the wrong place to resolve the discussion.  Are you okay with treating this as a two-way door that we can come back to after we resolve code standards elsewhere?

ashmrtn: >Sure, but that's also a bit contrived, don't you think?

It depends. If we end up with lots of different settings then it can become easy to miss stuff like this because it's easy to skim over long lists of parameters or just think "I'll just add this at the end because it's not set already"

>In the perspective of calling the constructor, or handling it?

From the perspective of calling the constructor. Unless someone really reads the code well and is familiar with this pattern then it's something that's easy to miss or misunderstand (e.x. me thinking we couldn't have function parameters) which harms long-term maintainability of the codebase

>Are you okay with treating this as a two-way door that we can come back to after we resolve code standards elsewhere?

Only so long as we have a fixed date where we resolve this question and make time to actually go through the codebase and do cleanup. I don't want to get into the habit of saying we'll resolve things elsewhere and then never revisiting them

ryanfkeepers: > I don't want to get into the habit of saying we'll resolve things elsewhere and then never revisiting them

P2 tech debt, yeah? 😁

ashmrtn: 🥲 RIP P2 tech debt. Approving for now as there's an out-of-band discussion for determining and laying down standards

ryanfkeepers:
@dadams39 @ashmrtn Understanding that we're have some discussions to resolve about code standards at large, is there anything in this PR outside of that which needs to be triaged before moving forward?  The PR that follows on this is a blocker for the next Corso release.

ashmrtn (approved):

src/internal/connector/graph/service.go
dadams39: Apologies, not an error. Asking for comments on the different clauses there are few
`# Successful Response`

dadams39:
> @dadams39 @ashmrtn Understanding that we're have some discussions to resolve about code standards at large, is there anything in this PR outside of that which needs to be triaged before moving forward? The PR that follows on this is a blocker for the next Corso release.

I did not reject the PR. Comments were points to consider, not stop progress.

ryanfkeepers:
> I did not reject the PR. Comments were points to consider, not stop progress.

Of course not.  I didn't mean to imply that anything was rejected; my apologies if my statement accidentally suggested otherwise.  But approval is still required to move forward, and I hadn't gotten that yet.

> RIP P2 tech debt

Haha, this is not necessarily P2.  I will say, however, that this is the fourth or fifth location where I've introduced this pattern.  It will require a larger discussion to address no matter the feelings on its usage in this PR.

src/internal/connector/graph/service.go
ryanfkeepers: Ah, gotcha. That's a good idea, thanks, I'll add it now.

PR #2302, opened 2023-01-27
refactor exchange attachment downloads

ryanfkeepers:
## Description

Three changes: adds a LargeItem servicer to
the exchange api client.  Migrates attachment
downloads in mail and events out of serialize
and into the GetItem func.  Finally, utilizes the
largeItem servicer to download attachments,
instead of the standard servicer.

A follow-up PR will add mocked test cases for
these changes.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2299

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/api.go
+++ b/src/internal/connector/exchange/api/api.go
@@ -61,7 +61,7 @@ type Client struct {
 	stable graph.Servicer
 
 	// The largeItem graph servicer is configured specifically for
-	// downloading large items.  Specificlaly for use when handling
+	// downloading large items.  Specifically for use when handling
 	// attachments, and for no other use.
 	largeItem graph.Servicer
 }
@@ -69,7 +69,12 @@ type Client struct {
 // NewClient produces a new exchange api client.  Must be used in
 // place of creating an ad-hoc client struct.
 func NewClient(creds account.M365Config) (Client, error) {
-	s, li, err := newServices(creds)
+	s, err := newService(creds)
+	if err != nil {
+		return Client{}, err
+	}
+
+	li, err := newLargeItemService(creds)
 	if err != nil {
 		return Client{}, err
 	}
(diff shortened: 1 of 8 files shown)


src/internal/connector/exchange/api/api.go
```diff
@@ -57,37 +59,51 @@ type Client struct {
 	// The stable service is re-usable for any non-paged request.
 	// This allows us to maintain performance across async requests.
 	stable graph.Servicer
+
+	// The largeItem graph servicer is configured specifically for
+	// downloading large items.  Specificlaly for use when handling
```
ashmrtn, line 64:
```suggestion
	// downloading large items.  Specifically for use when handling
```

src/internal/connector/exchange/api/api.go
```diff
@@ -57,37 +59,51 @@ type Client struct {
...
-	return Client{creds, s}, nil
+	return Client{creds, s, li}, nil
 }
 
 // service generates a new service.  Used for paged and other long-running
 // requests instead of the client's stable service, so that in-flight state
 // within the adapter doesn't get clobbered
 func (c Client) service() (*graph.Service, error) {
-	return newService(c.Credentials)
+	s, _, err := newServices(c.Credentials)
```
ashmrtn, line 84: I think this should have the first part of `newServices` and `newServices` should call this function instead of the other way around. It's usually better to build larger things from smaller things rather than getting smaller things by throwing out part of larger things. Plus, then there won't be any questions as to whether or not the HTTP clients that we create and then discard in this function are consuming resources

ntolia:
@ryanfkeepers Even though it is indicated here, I don't see a Changelog update.

src/internal/connector/exchange/api/api.go
ryanfkeepers: Sure, I'm good with that.

ryanfkeepers:
@ntolia Ack, thanks for catching that.  They've been added.

src/internal/connector/exchange/api/mail.go
```diff
@@ -107,6 +108,26 @@ func (c Mail) GetItem(
...
+				UsersById(user).
+				MessagesById(itemID).
+				Attachments().
+				Get(ctx, nil)
+			if err == nil {
+				mail.SetAttachments(attached.GetValue())
+				break
+			}
+		}
+
```
vkamra, line 124: Not a blocker - but for supportability - we should accumulate the errors and return those if we run out of retries.

ryanfkeepers: That feels a little bit like overpopulation of errors to me.  For example, if we had automatic retry/backoff in the client, instead of here in the code, we wouldn't add an error per retry, would we?

A log per retry, perhaps, sounds a bit more appropriate.

vkamra: A debug log per retry - would be useful - yes. This will help to debug when the issue is repro'able and we also have the option of asking the bug reporter to reproduce.

For the error message - what I was suggesting was that *if* we do end up failing the request eventually (after retries) - it might be interesting to see the entire chain of errors (from each retry) and not just the last one. This helps debug issues where it's not the same error each time but a chain of different errors. 

Not a blocker here because for the most part the last error will be the most relevant.

ryanfkeepers: Sounds good, thanks for the insight.

vkamra (approved):

PR #2305, opened 2023-01-27
add getItem unit with retries

### LINKED ISSUE #2299: Use LargeFile http client for attachment downloads.

ryanfkeepers:
## Description

separates item get/retry when iterating over
items in exchange collection streams, and adds
a mocked unit test on that func.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* closes #2299

## Test Plan

- [x] :zap: Unit test


src/internal/connector/exchange/exchange_data_collection_test.go
```diff
@@ -160,3 +166,47 @@ func (suite *ExchangeDataCollectionSuite) TestNewCollection_state() {
...
+	}{
+		{
+			name:  "happy",
+			items: mockItemer{},
+			expectErr: func(t *testing.T, err error) {
+				assert.NoError(t, err)
+			},
+		},
+		{
+			name:  "an error",
```
ashmrtn, line 184: this test is good to have, but it would also be good to have a test where we actually count the number of times the retry is called. That will help ensure the retry is actually working as expected. It adds a little more test complexity but it can help track down issues as well

One of the recent OneDrive PRs I have does a test to check the number of retries and it programmatically generates the mock results by reusing the variable containing the number of retries https://github.com/alcionai/corso/pull/2306/files#diff-a63d5466edbe22450c15d02fa6f9dd0cf04785314552c509e3c8c4d334e85978R120

ashmrtn (approved):
looks good, though adding a test to ensure retries happen as expected would be nice

src/internal/connector/exchange/exchange_data_collection_test.go
ryanfkeepers: oh yeah, not a bad idea.  Added that.

Issue #2264, opened 2023-01-25
create OneDrive api package similar to exchange/api package


ashmrtn:
Testing OneDrive incrementals support with mocks requires the ability to mock calls to external services. OneDrive should be updated to have an interface/package structure similar to that of exchange/api to support this testing

ashmrtn:
closing for now as there's been enough refactoring that backup calls can be mocked except for getting item data. Currently there's enough to mock out things for testing incrementals

PR #2306, opened 2023-01-27
Use mocks to test getting drive info from Graph API

ashmrtn:
## Description

Make a separate interface for fetching drive information from Graph API. Interface allows for better testing via mocks

Merges SharePoint and OneDrive code for getting drives.

Also fixes potential bug where not all drives would be fetched. This could have occurred because the previous implementation for both SharePoint and OneDrive were not checking for paginated results

Viewing by commit is recommended

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #2264

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/drive.go
```diff
@@ -7,101 +7,117 @@ import (
...
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to retrieve site drives. site: %s, details: %s",
-			site, support.ConnectorStackErrorTrace(err))
+	if !retry {
+		numberOfRetries = 0
 	}
 
-	return r.GetValue(), nil
-}
+	for done := false; !done; {
```
ryanfkeepers, line 73: I'm unsure about the need for this outer loop.  I think you could achieve the same effect by setting `numberOfRetries = 0` on line 93, and continuing instead of breaking.  It might feel a little strange to do that; but imo that's less mental juggling than keeping track of both the done state and the loop iteration while reading through the block, and easily knowing what's coming next (admittedly, I had a difficult time doing that).

src/internal/connector/onedrive/drive.go
```diff
@@ -7,101 +7,117 @@ import (
...
+		// Retry Loop for Drive retrieval. Request can timeout
+		for i := 0; i <= numberOfRetries; i++ {
+			page, err := pager.GetPage(ctx)
+			if err == nil {
```
ryanfkeepers, line 77: I think this would flow better if `err != nil` was handled here instead of the success path.  I'm not seeing anything that requires the success handling to go first, so the inversion does weird things for my golang-standards expectations.  Likewise, it feels a bit strange that the happy path "exit point" is a break at the very center of the code block, as opposed to the end of it.

ryanfkeepers (approved):
Some nits, but generally looks good.

Question on refactoring process: in the exchange api we've combined both the page handling and the individual calls in that layer.  Are we planning to do the same here in onedrive (not immediately, even; just, as a future goal in general)?

src/internal/connector/onedrive/drive.go
ashmrtn: The outer loop is required because we may get multiple pages of results. Basically the outer loop is for pages, the inner loop is for retries. I'll add some comments to the code which will hopefully make things clearer

ryanfkeepers: Right, that's understood.  But a retry loop that resets its ticker to 0 does the same thing as the outer loop.

ashmrtn:
>in the exchange api we've combined both the page handling and the client communication in that layer. Are we planning to do the same here in onedrive (not immediately, even; just, as a future goal in general)?

I'm not completely sure at the moment. There's a couple of different factors that could affect the final design we end up with. These factors don't necessarily apply to the fetching the drives, but they will apply to fetching items (which I was planning on having follow a similar pattern as this)

First is that restartability needs to be able to persist (most of) the results page for every iteration. Essentially it needs, at the very minimum, to persist the next link (if any), delta link (if any), and item IDs. This is so we can resume fetching items if we crash part-way through while still having an "atomic" write unit (i.e. a page of items and next/delta links persisted as a single file in kopia)

Second, OneDrive is different from Exchange in that we get all items for the drive instead of items for a folder. This means that unless we start passing callback functions into the pager to process items we may require a lot of memory to store the set of items to return to the caller for processing after pagination. Storing results would also undermine the above point about restartability so it's kind of off the table I suppose

PR #2320, opened 2023-01-30
add clues, fault to op/restore.go

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/operations/restore.go
+++ b/src/internal/operations/restore.go
@@ -342,7 +342,7 @@ func formatDetailsForRestoration(
 	ctx context.Context,
 	sel selectors.Selector,
 	deets *details.Details,
-	errs fault.Adder,
+	errs *fault.Errors,
 ) ([]path.Path, error) {
 	fds, err := sel.Reduce(ctx, deets)
 	if err != nil {
@@ -355,6 +355,10 @@ func formatDetailsForRestoration(
 	)
 
 	for i := range fdsPaths {
+		if errs.Errs() != nil {
+			return nil, errs.Err()
+		}
+
 		p, err := path.FromDataLayerPath(fdsPaths[i], true)
 		if err != nil {
 			errs.Add(clues.
@@ -379,9 +383,5 @@ func formatDetailsForRestoration(
 		return paths[i].String() < paths[j].String()
 	})
 
-	if errs != nil {
-		return nil, err
-	}
-
-	return paths, nil
+	return paths, errs.Err()
 }
--- a/src/internal/operations/restore.go
+++ b/src/internal/operations/restore.go
@@ -25,6 +25,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/fault"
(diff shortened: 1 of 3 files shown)


meain (approved):

src/internal/operations/restore.go
```diff
@@ -142,8 +143,10 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
...
 	if err != nil {
-		opStats.readErr = errors.Wrap(err, "restore")
-		return nil, opStats.readErr
+		op.Errors.Fail(errors.Wrap(err, "retrieving backup"))
+		opStats.readErr = op.Errors.Err()
+
+		return nil, op.Errors.Err()
 	}
 
 	ctx = clues.Add(ctx, "resource_owner", bup.Selector.DiscreteOwner)
```
dadams39, line 152: Would the clues need to be added to the clause at L145

src/internal/operations/restore.go
```diff
@@ -159,10 +162,12 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
 		},
 	)
 
-	paths, err := formatDetailsForRestoration(ctx, op.Selectors, deets)
+	paths, err := formatDetailsForRestoration(ctx, op.Selectors, deets, op.Errors)
 	if err != nil {
-		opStats.readErr = err
-		return nil, err
+		op.Errors.Fail(err)
+		opStats.readErr = op.Errors.Err()
```
dadams39, line 168: Is there a change that `op.Errors` is nil?

src/internal/operations/restore.go
```diff
@@ -175,8 +180,10 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
 
 	dcs, err := op.kopia.RestoreMultipleItems(ctx, bup.SnapshotID, paths, opStats.bytesRead)
 	if err != nil {
-		opStats.readErr = errors.Wrap(err, "retrieving service data")
-		return nil, opStats.readErr
+		op.Errors.Fail(errors.Wrap(err, "retrieving service data"))
+		opStats.readErr = op.Errors.Err()
```
dadams39, line 184: Same question as L168 --> Do we need a guard check in case op.Errors is nil

src/internal/operations/restore.go
```diff
@@ -274,34 +286,34 @@ func formatDetailsForRestoration(
...
 
 			continue
 		}
 
 		paths[i] = p
 	}
 
-	if errs != nil {
-		return nil, errs
+	if err != nil {
```
dadams39, line 315: Is this not a typo? `errs` is updated with clues and then it is not returned or checked if it is nil at L315

dadams39 (changes requested):

src/internal/operations/restore.go
ryanfkeepers: None

src/internal/operations/restore.go
ryanfkeepers: no

src/internal/operations/restore.go
ryanfkeepers: it is, good catch.

dadams39 (approved):

PR #2322, opened 2023-01-30
OneDrive Items API for mocking

ashmrtn:
## Description

Create a pager for drive items that allows for better testing via mocking. Increased testing will come in later PRs

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #2264

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/api/drive.go
+++ b/src/internal/connector/onedrive/api/drive.go
@@ -50,9 +50,6 @@ func NewItemPager(
 	res := &driveItemPager{
 		gs:      gs,
 		options: requestConfig,
-		// TODO: Specify a timestamp in the delta query
-		// https://docs.microsoft.com/en-us/graph/api/driveitem-delta?
-		// view=graph-rest-1.0&tabs=http#example-4-retrieving-delta-results-using-a-timestamp
 		builder: gs.Client().DrivesById(driveID).Root().Delta(),
 	}
 
--- a/src/internal/connector/onedrive/api/drive.go
+++ b/src/internal/connector/onedrive/api/drive.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 
+	msdrives "github.com/microsoftgraph/msgraph-sdk-go/drives"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	mssites "github.com/microsoftgraph/msgraph-sdk-go/sites"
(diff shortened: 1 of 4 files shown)


src/internal/connector/onedrive/api/drive.go
```diff
@@ -12,6 +13,68 @@ import (
...
+		QueryParameters: &msdrives.ItemRootDeltaRequestBuilderGetQueryParameters{
+			Top:    &pageCount,
+			Select: fields,
+		},
+	}
+
+	res := &driveItemPager{
+		gs:      gs,
+		options: requestConfig,
+		// TODO: Specify a timestamp in the delta query
```
meain, line 53: Just wondering if this comment is still relevant?

src/internal/connector/onedrive/drive.go
```diff
@@ -190,23 +194,29 @@ func collectItems(
...
+		err = collector(ctx, driveID, driveName, vals, oldPaths, newPaths, excluded)
 		if err != nil {
 			return "", nil, nil, err
 		}
 
-		if r.GetOdataDeltaLink() != nil && len(*r.GetOdataDeltaLink()) > 0 {
-			newDeltaURL = *r.GetOdataDeltaLink()
+		nextLink, deltaLink := gapi.NextAndDeltaLink(page)
+
+		if len(deltaLink) > 0 {
```
meain, line 209: I know the logic is from before, but graph api either gives a `nextLink` or a `deltaLink` . Don't think both will be available.

meain (approved):

src/internal/connector/onedrive/api/drive.go
ashmrtn: Good catch! I think we can remove it. I don't think we want to start with a timestamp for queries based on how the rest of the system is built

src/internal/connector/onedrive/drive.go
ashmrtn: yea, I think that's what I've read as well. At the moment I believe this pattern is used in a few different places in the codebase so I'd kind of like to keep with the pattern. If we change it, I'd suggest a separate PR that changes all instances instead of a more ad-hoc approach. After all, having consistency in handling of things like this can help with thinking through how parts of the system work

Issue #2287, opened 2023-01-27
Standardize HTTP retry handling


vkamra:
The current retry handling for graph and OneDrive requests is spread out across the code.

The retry itself cannot always be in middleware - often needs to be at a higher layer - but the retry implementation itself can be consolidated/standardized, avoid code duplication.

We should leverage something like https://github.com/hashicorp/go-retryablehttp to implement the retry/backoff mechanism.

meain:
[Here](https://github.com/microsoft/kiota-http-go/blob/main/retry_handler.go) is the default retry handler from kiota and [here](https://github.com/alcionai/corso/blob/0f39b23a00c4375f74ef9b6f5d7dfb820a570cbf/src/internal/connector/graph/service.go#L149) is where we add it to corso.

meain:
Adding another error where I have seen failures when trying locally:
```
Error: 1 error occurred:
        * Failed to run Exchange backup for user user@mail.com: 2 errors occurred:
        * errors prevented the operation from processing
        * retrieving data to backup: getting user: getting user by id: Get "https://graph.microsoft.com/v1.0/users/Jo.....rj.onmicrosoft.com": net/http: TLS handshake timeout
```

ashmrtn:
closing as the PR linked to this was merged

PR #2328, opened 2023-01-31
Retry handling for delta queries in Exchange

meain:
## Description

Added retry handling for delta queries in OneDrive. Also, bumping time timeout for graph api calls from 90s to 3m as we were seeing client timeouts for graph api calls. ~Haven't added retry for every request in exchange as I'm hoping https://github.com/alcionai/corso/issues/2287 will be a better way to handle this.~

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


vkamra:
Should we add this for message fetch as well in collection population logic (for messages at least)? This would be a temporary fix till we got #2287 in.

ntolia:
- Please do add a bug fix changelog entry 
- If we don't have retried oN other EX calls, isn't it possible that large backups could fail there instead?

meain:
@ntolia I've added a CHANGELOG entry. As for you second comment, the original desc is not correct anymore and I have since added retries for every request made by exchange. Missed updating the original comment.

src/internal/connector/exchange/api/contacts.go
```diff
@@ -61,7 +61,16 @@ func (c Contacts) GetItem(
...
 ) (serialization.Parsable, *details.ExchangeInfo, error) {
-	cont, err := c.stable.Client().UsersById(user).ContactsById(itemID).Get(ctx, nil)
+	var (
+		cont models.Contactable
+		err  error
+	)
+
+	runWithRetry(func() error {
+		cont, err = c.stable.Client().UsersById(user).ContactsById(itemID).Get(ctx, nil)
+		return err
```
dadams39, line 71: Could need to run `support.ConnectorStackErrorTraceWrap()` to get the full error for non-network related failure details

src/internal/connector/exchange/api/contacts.go
```diff
@@ -81,7 +90,14 @@ func (c Contacts) GetAllContactFolderNamesForUser(
 		return nil, err
 	}
 
-	return c.stable.Client().UsersById(user).ContactFolders().Get(ctx, options)
+	var resp models.ContactFolderCollectionResponseable
+
+	runWithRetry(func() error {
```
dadams39, line 95: Or we could put the support error wrap within runWithRetry

src/internal/connector/exchange/api/api.go
```diff
@@ -153,3 +153,21 @@ func HasAttachments(body models.ItemBodyable) bool {
...
+
+		if !graph.IsErrTimeout(err) && !graph.IsInternalServerError(err) {
+			break
+		}
+
+		if i < numberOfRetries {
+			time.Sleep(time.Duration(3*(i+1)) * time.Second)
+		}
+	}
+}
```
ryanfkeepers, line 173: I'd prefer this return the final error.  That way we can centralize formatting, like in Danny's suggestions.
```suggestion
func runWithRetry(run func() error) err {
        var err errror

	for i := 1; i <= numberOfRetries; i++ {
		err = run()
		if err == nil {
			return nil
		}

		if !graph.IsErrTimeout(err) && !graph.IsInternalServerError(err) {
			break
		}

		if i < numberOfRetries {
			time.Sleep(time.Duration(3*(i+1)) * time.Second)
		}
	}
	
	return support.ConnectorStackErrorTraceWrap(err, "")
}
```

src/internal/connector/graph/errors.go
```diff
@@ -135,16 +137,16 @@ func IsErrUnauthorized(err error) bool {
...
 
-func IsSericeUnavailable(err error) bool {
-	if errors.Is(err, Err503ServiceUnavailable) {
+func IsInternalServerError(err error) bool {
+	if errors.Is(err, Err500InternalServerError) {
 		return true
 	}
 
-	e := ErrUnauthorized{}
+	e := ErrInternalServerError{}
```
ryanfkeepers, line 149: woops.  Good catches.

ryanfkeepers (approved):

src/internal/connector/graph/errors.go
```diff
@@ -135,21 +142,18 @@ func IsErrUnauthorized(err error) bool {
...
 		return true
 	}
 
-	e := ErrUnauthorized{}
-	if errors.As(err, &e) {
-		return true
-	}
+	e := ErrInternalServerError{}
 
-	return true
```
ryanfkeepers, line 152: Critical point of failure: this caused retries handlers watching for this error to retry.

src/internal/connector/graph/errors.go
```diff
@@ -17,6 +17,7 @@ import (
 // ---------------------------------------------------------------------------
 
 const (
+	errCodeActivityLimitReached        = "activityLimitReached"
```
dadams39, line 20: Is this a throttling error?

src/internal/connector/graph/errors.go
vkamra: 🤦 - nice catches all around

vkamra: Add a unit test please?

src/internal/connector/graph/errors.go
ryanfkeepers: yup

src/internal/connector/graph/errors.go
meain: Woops 🤦🏼

Issue #2329, opened 2023-01-31
Corso logging improvements


gmatev:
The following are a number of improvements that we should make to Corso logging to improve supportability and troubleshooring for real world users:
- [ ] ~Log file names currently use UTC - switch to local time for easier identification of the right log. Time stamps inside logs are already in local time~
- [ ] Time stamps inside logs should be in UTC
- [x]  Print out log file name being used with the CLI output. Allows a user who is invoking using CLI to correlate which log file has more details
- [x]  Log Corso version at the beginning of logs
- [x]  Include the command being executed (with appropriate masking of PII like emails addresses and folder names) at the beginning of the log
- [x]  Include final success or failure of command to make the result obvious in the log
- [x]  ~Number of items discovered / backed up is not PII and should not be masked~
- [x]  Long running operations should include INFO logs as they make progress, for example every 1000 items processed
- [x]  Logs should also capture relevant stdout/stderr information that Corso will otherwise output as errors

ryanfkeepers:
> Log file names currently use UTC - switch to local time for easier identification of the right log. Time stamps inside logs are already in local time

Clarifying, the request here is to use local time for the file name, right?

ntolia:
@gmatev @ryanfkeepers Best practices recommend always recording all times in UTC. It becomes a big mess otherwise when we start combining logs from different places in the future (e.g., logs from Corso and AWS). Based on previous experiences, I highly recommend we standardize everything to UTC instead of local time zone.

dadams39:
Suggestion is to clarify retry failure(s):
```bash
2023-01-30T08:52:08.367Z        INFO    graph/service.go:293    graph api error {"status": "500 Internal Server Error", "method": "GET", "url": https://graph.microsoft.com/v1.0/users/user%40errors.net/messages/<messageID>/attachments}
```
In this particular case, this is a failure which looks very similar to an API Timeout and which is retried. Users should be easily able to distinguish between the system resending information and the system exiting after x retries.

ryanfkeepers:
> I highly recommend we standardize everything to UTC instead of local time zone

Yep, I was going to push similarly.  There's also some [side documentation in the repo](https://github.com/alcionai/corso/blob/main/design/time_standards.md) affirming that expectation.  Glad we're agreed about it.

ntolia:
> Users should be easily able to distinguish between the system resending information and the system exiting after x retries.

[...]

PR #2331, opened 2023-01-31
add version to clues, log cli command and flags

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #2329 

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/alcionai/clues"
 	"github.com/spf13/cobra"
 
 	"github.com/alcionai/corso/src/cli/backup"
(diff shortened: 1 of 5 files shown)


src/cli/cli.go
```diff
@@ -94,6 +117,8 @@ func Handle() {
 	loglevel, logfile := logger.PreloadLoggingFlags()
 	ctx, log := logger.Seed(ctx, loglevel, logfile)
 
+	ctx = clues.AddAll(ctx, "corso_version", version.CurrentVersion())
```
vkamra, line 120: Does this get logged in every log/error message?

ryanfkeepers: Yes

vkamra: Yeah - I think it'll be overkill. Having it in the log file should be good enough (maybe we put in CLI output as well as log file?) 

We should also print on the command line the log file full-path we are using (this will help the user identify which log file to look at)

vkamra (changes requested):
Remove version from clues. Otherwise LGTM.

vkamra (approved):

vkamra:
Removed the code that added version to clues. Going to queue @ryanfkeepers

vkamra:
Added the version to information getting logged.

PR #2335, opened 2023-01-31
add clues, fault to selectors

ryanfkeepers:
## Description

Adds clues and fault handling to selectors pkg.
Some bleed upward into the CLI occured from
where the cli directly calls selectors.Reduce.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/selectors.go
+++ b/src/pkg/selectors/selectors.go
@@ -5,9 +5,11 @@ import (
 	"encoding/json"
 	"strings"
 
+	"github.com/alcionai/clues"
 	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/filters"
 	"github.com/alcionai/corso/src/pkg/path"
 )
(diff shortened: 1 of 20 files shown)


src/pkg/selectors/selectors.go
```diff
@@ -272,7 +279,7 @@ func selectorAsIface[T any](s Selector) (T, error) {
 		a, err = func() (any, error) { return s.ToSharePointRestore() }()
 		t = a.(T)
 	default:
-		err = errors.New("service not supported: " + s.Service.String())
+		err = clues.Stack(ErrorUnrecognizedService, errors.New(s.Service.String()))
```
ashmrtn, line 282: What is the stack trace for this going to look like? Since we're not wrapping the original error is seems like we could end up with stack traces for both? Stack traces can generally be printed with `%+v` when using `github.com/pkg/errors`

ashmrtn (approved):

src/pkg/selectors/selectors.go
ryanfkeepers: Honestly, I haven't quite investigated stack trace outputs.  That's something I can chase down at a later date.  

Sufficed to say: I'll make it useful.

ashmrtn: for stack traces I'd prefer to have them working sooner rather than later. They've been invaluable for finding issues so losing them would set back debugging efforts quite a bit

PR #2336, opened 2023-01-31
add clues, fault to small /internal pkgs

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/common/time.go
+++ b/src/internal/common/time.go
@@ -4,6 +4,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/alcionai/clues"
 	"github.com/pkg/errors"
 )
 
@@ -85,7 +86,10 @@ var (
 	}
 )
 
-var ErrNoTimeString = errors.New("no substring contains a known time format")
+var (
+	ErrNoTimeString        = errors.New("no substring contains a known time format")
+	errParsingStringToTime = errors.New("parsing string as time.Time")
+)
 
 // Now produces the current time as a string in the standard format.
 func Now() string {
@@ -132,7 +136,7 @@ func FormatLegacyTime(t time.Time) string {
 // the provided string.  Always returns a UTC timezone value.
 func ParseTime(s string) (time.Time, error) {
 	if len(s) == 0 {
-		return time.Time{}, errors.New("cannot interpret an empty string as time.Time")
+		return time.Time{}, clues.Stack(errParsingStringToTime, errors.New("empty string"))
 	}
 
 	for _, form := range formats {
@@ -142,14 +146,14 @@ func ParseTime(s string) (time.Time, error) {
 		}
 	}
 
-	return time.Time{}, errors.New("unable to parse time string: " + s)
+	return time.Time{}, clues.Stack(errParsingStringToTime, errors.New(s))
 }
 
 // ExtractTime greedily retrieves a timestamp substring from the provided string.
 // returns ErrNoTimeString if no match is found.
 func ExtractTime(s string) (time.Time, error) {
 	if len(s) == 0 {
-		return time.Time{}, errors.New("cannot extract time.Time from an empty string")
+		return time.Time{}, clues.Stack(errParsingStringToTime, errors.New("empty string"))
 	}
 
 	for _, re := range regexes {
@@ -159,5 +163,5 @@ func ExtractTime(s string) (time.Time, error) {
 		}
 	}
 
-	return time.Time{}, errors.Wrap(ErrNoTimeString, s)
+	return time.Time{}, clues.Stack(ErrNoTimeString, errors.New(s))
 }
--- a/src/internal/streamstore/streamstore.go
+++ b/src/internal/streamstore/streamstore.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"io"
 
+	"github.com/alcionai/clues"
 	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/internal/data"
@@ -52,17 +53,16 @@ func (ss *streamStore) WriteBackupDetails(
 			ss.tenant,
 			collectionPurposeDetails,
 			ss.service,
-			false,
-		)
+			false)
 	if err != nil {
-		return "", err
+		return "", clues.Stack(err).WithClues(ctx)
 	}
 
 	// TODO: We could use an io.Pipe here to avoid a double copy but that
 	// makes error handling a bit complicated
 	dbytes, err := json.Marshal(backupDetails)
 	if err != nil {
-		return "", errors.Wrap(err, "marshalling backup details")
+		return "", clues.Wrap(err, "marshalling backup details").WithClues(ctx)
 	}
 
 	dc := &streamCollection{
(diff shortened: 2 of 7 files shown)


src/internal/common/time.go
```diff
@@ -142,14 +146,14 @@ func ParseTime(s string) (time.Time, error) {
 		}
 	}
 
-	return time.Time{}, errors.New("unable to parse time string: " + s)
+	return time.Time{}, clues.Stack(errParsingStringToTime, errors.New(s))
```
ashmrtn, line 149: what does the printed output of this look like? If the string `s` isn't towards the end it could make it harder to quickly parse the logs (as a human that is). I often find it harder to find stuff in the center of strings than the beginning or end

src/internal/streamstore/streamstore.go
```diff
@@ -115,7 +114,9 @@ func (ss *streamStore) ReadBackupDetails(
 
 	// Expect only 1 data collection
 	if len(dcs) != 1 {
-		return nil, errors.Errorf("expected 1 details data collection: %d", len(dcs))
+		return nil, clues.New("greater than 1 details data collection found").
```
ashmrtn, line 117:
```suggestion
		return nil, clues.New("more than 1 details data collection found").
```

src/internal/streamstore/streamstore.go
```diff
@@ -128,20 +129,20 @@ func (ss *streamStore) ReadBackupDetails(
 	for {
 		select {
 		case <-ctx.Done():
-			return nil, errors.New("context cancelled waiting for backup details data")
+			return nil, clues.New("context cancelled waiting for backup details data").WithClues(ctx)
```
ashmrtn, line 132: these are reported as errors (return-value wise) so we should be associating stack traces with them. I'm not sure off the top of my head if `clues.New` has that functionality yet

src/internal/streamstore/streamstore.go
```diff
@@ -156,7 +157,7 @@ func (ss *streamStore) DeleteBackupDetails(
 ) error {
 	err := ss.kw.DeleteSnapshot(ctx, detailsID)
 	if err != nil {
-		return errors.Wrap(err, "deleting backup details failed")
+		return errors.Wrap(err, "deleting backup details")
```
ashmrtn, line 160: should this be using clues?

ryanfkeepers: Nope. `errors.Wrap` is perfectly fine.  And since the called function accepts `ctx`, we assume that the error creator in that func already appended structured data to the error.

src/internal/streamstore/streamstore.go
ryanfkeepers: We'll get there.

src/internal/streamstore/streamstore.go
```diff
@@ -78,10 +78,9 @@ func (ss *streamStore) WriteBackupDetails(
 		nil,
 		[]data.Collection{dc},
 		nil,
-		false,
-	)
+		false)
 	if err != nil {
-		return "", nil
+		return "", errors.Wrap(err, "storing details in repository")
```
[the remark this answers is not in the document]
ryanfkeepers: It's all good, see answer below.

src/internal/common/time.go
ryanfkeepers: Yup!  That was a big driver for the Stack method, actually: not having to push data out of order just to wrap/attach an error.  There's a good display of expectations in the Clues tests, but here it will look like this:
```
parent wrapper message: error parsing string to time: <value of s> 
```

src/internal/streamstore/streamstore.go
ashmrtn: tbh I'd prefer to continue having stack traces associated with log messages even if it requires us to do something like `clues.New(errors.New(...))` in the interim. Stack traces have been very helpful in tracking down exact locations of things and I don't want to lose them while we build out functionality

src/internal/streamstore/streamstore.go
ryanfkeepers: A large thrust of the design is for the two packages to be interchangeable.  I understand the want for consistency, but the goal of the package is not to replace errors.  The lack of `clues.Is()` and `clues.As()` is intentional.  It's a common expectation for packages of this kind to interface cohesively with current uses of errors and pkg/errors so that contributors aren't required to touch every point of error handling in a repo when they trade out libraries.  

Having been in this position before: my expectation is that contributors will feel less overwhelmed by code that often uses errors, and sometimes uses clues, than they would if the repo used only clues.  The latter changes the total topography of error handling (and cognitive load to understand it, even if only as an assumption, unknowns can be impenetrable).  By showcasing interoperability with errors, we can prove that the behavior remains compliant with handling that is well known and understood, at a minimum.  That provides a baseline level of comfort and eases cognitive load.

ashmrtn: To be clear, the following trying to put down the clues package nor is that my intention. What I'm trying to show is that there's a lot of edge cases and open questions that can pop up. Some of these I'm trying to surface here but others we may not realize until we need to work around them in implementations or worse, we see bugs appearing due to unexpected behavior

>By showcasing interoperability with errors, we can prove that the behavior remains compliant with handling that is well known and understood, at a minimum.

I think this is kind of a red herring honestly. I think what we can say is that clues is interoperable with `error` because `error` only requires `Error() string`. What we can't easily say is that the clue package is interoperable with whatever third-party error library we may be using (in our case `pkg/errors` in most places). Since we've already mostly abandoned `error` in favor of `pkg/errors`, having interoperability with just `error` is not as useful as interoperability with what we commonly use in the codebase.

If we want to be able to seamlessly use either `pkg/errors` or `clue`, what we want is for clue to be a wrapper around `pkg/error`, not to just be able to match the same `error` interface

>expectation is that contributors will feel less overwhelmed by code that often uses errors, and sometimes uses clues, than they would if the repo used only clues.

Looking at the code that's here, I'm confused by when I should use clues and when I should use errors. If I don't use clues, am I going to hamstring debugging because we don't get enough context information? If clues is not meant to be a full replacement for whatever errors package we're using, do I need to be careful about what I return? There doesn't seem to be a way to get back the wrapped/stacked errors, so what if I need context from clues but also access to some interface on the wrapped/stacked error? It seems like that functionality would become inaccessible, which makes interoperability difficult and requires more thinking about the entire call chain rather than how to handle things in an individual function. Adding handling for specific functions like `func (c *clue) Is(target error) bool` to clues is possible, but that makes clues more of a wrapper around a specific error library rather than a separate interoperable package. At best we'll get compile errors when we don't implement things. At worst we'll get silent failures because a lot of things like `As(target error)` have type assertions and just exit with false if the assertion fails

These sorts of questions start to come up as we begin integrating things, but they are really easy to overlook when just starting to work on a codebase that uses multiple packages for error handling, especially if one isn't intimately familiar with the libraries being used. If using clues doesn't give all the expected interfaces, it can easily lead to bugs because engineers may assume that it does provide all the stuff they need, especially since they're often treated as the same `error` type as other things. Yes, I get the builtin `error` type only really has `Error()` in it's interface, but once we start using other packages like `pkg/errors` the assumption kind of becomes "all errors we return have some common interface I can assume and it's the interface of the third-party package being used"

ashmrtn (approved):
approving as we had an out of band chat. Also there has been an update to the clues package to support stack traces

Issue #2122, opened 2023-01-12
deserialize metadata from previous backups in OneDrive


ashmrtn:
description in #2144 states what metadata structs are being persisted and will need deserialized

PR #2341, opened 2023-02-01
Patch and test some edge cases for deserializing OneDrive metadata

ashmrtn:
## Description

Currently no folder path entry is stored for the root folder. This leads to not having a folders map but having a valid delta token if all items are in the root of the drive.

Update the code to add at least an empty folders map entry so we don't think we have missing metadata. Add tests to check for these edge cases.

Long-term we may want to add a folder entry for the root folder. Whether we do or not may depend on how we decide to set the state field for that collection.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2122 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -180,7 +180,15 @@ func deserializeMetadata(
 
 		// Go through and remove partial results (i.e. path mapping but no delta URL
 		// or vice-versa).
-		for k := range prevDeltas {
+		for k, v := range prevDeltas {
+			// Remove entries with an empty delta token as it's not useful.
+			if len(v) == 0 {
+				delete(prevDeltas, k)
+				delete(prevFolders, k)
+			}
+
+			// Remove entries without a folders map as we can't tell kopia the
+			// hierarchy changes.
 			if _, ok := prevFolders[k]; !ok {
 				delete(prevDeltas, k)
 			}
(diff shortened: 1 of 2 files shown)


ashmrtn:
@meain tagging you here as there's some stuff about folder path metadata you may need to change when setting state fields

(from PR description)
>Long-term we may want to add a folder entry for the root folder. Whether we do or not may depend on how we decide to set the state field for that collection.

src/internal/connector/onedrive/collections.go
```diff
@@ -180,7 +180,15 @@ func deserializeMetadata(
 
 		// Go through and remove partial results (i.e. path mapping but no delta URL
 		// or vice-versa).
-		for k := range prevDeltas {
+		for k, v := range prevDeltas {
+			// Remove entries with an empty delta token as it's not useful.
+			if len(v) == 0 {
```
meain, line 185: Just wondering what is the scenario where this would happen.

meain (approved):

src/internal/connector/onedrive/collections.go
ashmrtn: I don't expect this to happen at the moment, but I do want to make sure we handle it if it does. Right now the code that serializes the delta tokens has it's own check to ensure no 0 length tokens are added to the metadata. However, having a check here makes the code a little more flexible because even if the other check is accidentally removed things should (hopefully) continue working as expected

PR #2344, opened 2023-02-01
Issue 2329 logfile 1

ryanfkeepers:
## Description

Prevents showing the log file location when calling commands that defer to help output or env details.
Certain cases aren't caught, such as when calling a command with no flags (ex: `corso backup create exchange`).
In those cases we catch the lack of flags and manually display the usage within the command handler.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #2329

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 
 	"github.com/alcionai/corso/src/cli/backup"
 	"github.com/alcionai/corso/src/cli/config"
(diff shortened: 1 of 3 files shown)


src/cli/cli.go
```diff
@@ -50,6 +51,13 @@ func preRun(cc *cobra.Command, args []string) error {
 		flagSl = append(flagSl, f)
 	}
 
+	avoidTheseCommands := []string{
+		"corso", "env", "help", "backup", "details", "list", "restore", "delete", "repo", "init", "connect",
+	}
+	if len(logger.LogFile) > 0 && !slices.Contains(avoidTheseCommands, cc.Use) {
```
meain, line 57: Maybe we should have the check `len(logger.LogFile) == 0` check to `PreloadLoggingFlags` and have it use the default.

meain (approved):

src/cli/cli.go
ryanfkeepers: The default should already be established by this point, iirc.

PR #2364, opened 2023-02-01
add fault to repository repository funcs

ryanfkeepers:
## Description

adds fault.Errors to backupDetails() and backups().

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange.go
+++ b/src/cli/backup/exchange.go
@@ -16,7 +16,6 @@ import (
 	"github.com/alcionai/corso/src/internal/model"
 	"github.com/alcionai/corso/src/pkg/backup"
 	"github.com/alcionai/corso/src/pkg/backup/details"
-	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
 	"github.com/alcionai/corso/src/pkg/selectors"
@@ -472,10 +471,9 @@ func detailsExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	defer utils.CloseRepo(ctx, r)
 
-	ds, errs := runDetailsExchangeCmd(ctx, r, backupID, opts)
-	if errs.Err() != nil {
-		// TODO: log/display iterated errors
-		return Only(ctx, errs.Err())
+	ds, err := runDetailsExchangeCmd(ctx, r, backupID, opts)
+	if err != nil {
+		return Only(ctx, err)
 	}
 
 	if len(ds.Entries) == 0 {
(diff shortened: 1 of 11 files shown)


neha-Gupta1 (approved):

src/cli/backup/exchange.go
```diff
@@ -502,13 +503,14 @@ func runDetailsExchangeCmd(
 		return nil, errs.Fail(err)
 	}
 
-	d, _, err := r.BackupDetails(ctx, backupID)
-	if err != nil {
-		if errors.Is(err, kopia.ErrNotFound) {
+	d, _, errs := r.BackupDetails(ctx, backupID)
+	// TODO: log/track recoverable errors
+	if errs.Err() != nil {
+		if errors.Is(errs.Err(), kopia.ErrNotFound) {
```
dadams39, line 509: Do we need to add Clues here?

src/cli/backup/exchange_integration_test.go
```diff
@@ -396,8 +397,9 @@ func (suite *PreparedBackupExchangeIntegrationSuite) TestExchangeDetailsCmd() {
 			bID := suite.backupOps[set]
 
 			// fetch the details from the repo first
-			deets, _, err := suite.repo.BackupDetails(ctx, string(bID))
-			require.NoError(t, err)
+			deets, _, errs := suite.repo.BackupDetails(ctx, string(bID))
+			require.NoError(t, errs.Err())
+			require.Empty(t, errs.Errs())
```
dadams39, line 402: Should this be `deets` also?

src/cli/backup/onedrive.go
```diff
@@ -394,13 +395,14 @@ func runDetailsOneDriveCmd(
...
-		if errors.Is(err, kopia.ErrNotFound) {
+	d, _, errs := r.BackupDetails(ctx, backupID)
+	// TODO: log/track recoverable errors
+	if errs.Err() != nil {
+		if errors.Is(errs.Err(), kopia.ErrNotFound) {
 			return nil, errs.Fail(errors.Errorf("no backup exists with the id %s", backupID))
 		}
 
-		return nil, errs.Fail(errors.Wrap(err, "Failed to get backup details in the repository"))
+		return nil, errs.Fail(errors.Wrap(errs.Err(), "Failed to get backup details in the repository"))
```
dadams39, line 405: Should the original error be included in L401?

src/cli/utils/testdata/opts.go
```diff
@@ -497,8 +498,11 @@ func (MockBackupGetter) Backup(
...
 }
 
-func (MockBackupGetter) Backups(context.Context, []model.StableID) ([]*backup.Backup, error) {
-	return nil, errors.New("unexpected call to mock")
+func (MockBackupGetter) Backups(
+	context.Context,
+	[]model.StableID,
+) ([]*backup.Backup, *fault.Errors) {
+	return nil, fault.New(false).Fail(errors.New("unexpected call to mock"))
 }
```
dadams39, line 506: The naming is interesting because it always returns an error.

dadams39 (approved):
There are a few nits, but nothing to stop the process.

src/cli/utils/testdata/opts.go
ryanfkeepers: It's a bit of a standard pattern for mocks.  The mock has to comply with an interface; but if certain calls on the interface aren't expected, it returns an error to showcase that the mock isn't set up to handle that function.

src/cli/backup/exchange.go
ryanfkeepers: Good question, and no.  The general handling is that if were passing a `ctx` into an internal func, then we should assume _that_ func handles packing clues into the error.  There's no harm if we over-pack the data.  But this helps minimize boilerplate.

src/cli/backup/exchange_integration_test.go
ryanfkeepers: I think deets gets used below, and is therefore checked in that way.

src/cli/backup/onedrive.go
ryanfkeepers: It is the original error

PR #2368, opened 2023-02-02
Run gci on beta package

meain:
## Description

Followup to https://github.com/alcionai/corso/pull/2340 . Runs `gci write --skip-generated -s 'standard,default,prefix(github.com/alcionai/corso)'` to properly order headers.

This is generated files as mentioned in the previous PR, but I thought we might as well order the headers as well if we are formatting it. Let me know if leaving this out was intentional though.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test


ashmrtn (approved):
Thanks for the follow up. I think it's worth merging as it also snags some other import ordering things outside the betasdk package. It's not strictly necessary we completely format the betasdk package as it's added to an [exclude rule in golangci-lint](https://github.com/alcionai/corso/blob/6f44a507b57d9f616465884a1f830043fa031d3c/src/.golangci.yml#L108). Mostly I wanted to run gofmt on it so that when I ran `gofmt -w -s .` in `/src` in the repo I wouldn't get extra changes tracked in git 😅

PR #2370, opened 2023-02-02
Fix some edge cases around OneDrive permissions backup

meain:
## Description

This fixes tying to parse link shares as permissions and some error/retry handling.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -191,7 +191,7 @@ func oneDriveItemMetaInfo(
 
 	perm, err := service.Client().DrivesById(driveID).ItemsById(*itemID).Permissions().Get(ctx, nil)
 	if err != nil {
-		return Metadata{}, errors.Wrapf(err, "failed to get item permissions %s", *itemID)
+		return Metadata{}, err
 	}
 
 	uperms := filterUserPermissions(perm.GetValue())
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -194,9 +194,21 @@ func oneDriveItemMetaInfo(
 		return Metadata{}, errors.Wrapf(err, "failed to get item permissions %s", *itemID)
 	}
 
+	uperms := filterUserPermissions(perm.GetValue())
+
+	return Metadata{Permissions: uperms}, nil
+}
+
+func filterUserPermissions(perms []models.Permissionable) []UserPermission {
 	up := []UserPermission{}
 
-	for _, p := range perm.GetValue() {
+	for _, p := range perms {
+		if p.GetGrantedToV2() == nil {
+			// For link shares, we get permissions without a user
+			// specified
+			continue
+		}
+
 		roles := []string{}
 
 		for _, r := range p.GetRoles() {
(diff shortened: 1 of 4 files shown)


src/internal/connector/onedrive/item.go
```diff
@@ -191,12 +191,24 @@ func oneDriveItemMetaInfo(
 
 	perm, err := service.Client().DrivesById(driveID).ItemsById(*itemID).Permissions().Get(ctx, nil)
 	if err != nil {
-		return Metadata{}, errors.Wrapf(err, "failed to get item permissions %s", *itemID)
```
ashmrtn, line 194: since this error isn't wrapped elsewhere please keep at least the part that says we failed fetching the permissions. I'm unsure if we need/should keep the item ID

src/internal/connector/onedrive/item.go
```diff
@@ -191,12 +191,24 @@ func oneDriveItemMetaInfo(
...
+
+func filterUserPermissions(perms []models.Permissionable) []UserPermission {
 	up := []UserPermission{}
 
-	for _, p := range perm.GetValue() {
+	for _, p := range perms {
+		if p.GetGrantedToV2() == nil {
+			// For link shares, we get permissions without a user
+			// specified
+			continue
```
ashmrtn, line 209: since we're skipping these please add a note about this as a `known issue` in the changelog. We'll probably also have to do a website doc update for it at some point

src/internal/connector/onedrive/item_test.go
```diff
@@ -257,3 +258,72 @@ func (suite *ItemIntegrationSuite) TestDriveGetFolder() {
...
+	readPerm, readUperm := getPermsUperms(permID, userID, []string{"read"})
+	readWritePerm, readWriteUperm := getPermsUperms(permID, userID2, []string{"read", "write"})
+
+	noPerm, _ := getPermsUperms(permID, userID, []string{"read"})
+	noPerm.SetGrantedToV2(nil) // eg: link shares
+
+	cases := []struct {
+		name   string
+		perms  []models.Permissionable
+		uperms []UserPermission
```
ashmrtn, line 297: nit: it would be nice if this were `expectedPerms` or something else that denoted it's an expected result. Right now it's hard to tell if it's what the result should be or an input for the test

ashmrtn (approved):

src/internal/connector/onedrive/item.go
meain: This was done so that the errors returned from the client are passed as is to the checks [here](https://github.com/alcionai/corso/blob/18a8c350f7e9e5e12a3296c1b87e1d032841654d/src/internal/connector/onedrive/collection.go#L296). But let me wrap it before adding it to `errUpdater`.

src/internal/connector/onedrive/item.go
meain: For links shares, IIUC while we can recreate link shares, I don't think we can keep the same URLs. So I'm not sure if there is a point in restoring link shares, at least as of now.

src/internal/connector/onedrive/item_test.go
meain: `uperms` is essentially the `expectedPerms` unless I am missing something. Let me change the name.

Issue #2123, opened 2023-01-12
set `State` and `DoNotMerge` fields in Collections when fetching items


ashmrtn:
The `PreviousPath`, `State`, and `DoNotMerge` fields in the updated `data.Collection` interface inform kopia of how to handle `Collection`s when merging hierarchies during a delta token-based incremental backup. These fields must be populated by the OneDrive component when it is creating the collections. Populating these fields with non-default values will require getting changes from Graph API using a delta token (not just the full item enumeration at the delta endpoint)

The Exchange [code](https://github.com/alcionai/corso/blob/32b11cdca83f58201dc680e760da62d6ecbab1a1/src/internal/connector/exchange/service_iterators.go#L30) contains an example of these being populated, though the logic is spread across several functions when also considering handling token expiration and missing metadata

This [comment](https://github.com/alcionai/corso/pull/1909#discussion_r1055729484) givens an example of what collection states should be produced for a series of operations and gives a mental model for how to think about the states.

The `DoNotMerge` field is used to denote times when all data in the collection has been enumerated again. This is required if the delta token expired, the metadata could not be deserialized/verified, or any other event that would lead to a full enumeration of the items in the collection with no diffs from the previous backup

PR #2371, opened 2023-02-02
Pass in prev delta to collectItems

meain:
## Description

Pass the previous delta url fetched from metadata to collectItems and make sure we are using that when fetching the pages.

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/issues/2123

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/drive.go
```diff
@@ -172,22 +183,41 @@ func collectItems(
...
+		pager.SetNext(prevDelta)
+	}
+
 	for {
 		page, err := pager.GetPage(ctx)
+
+		if !triedPrevDelta && graph.IsErrInvalidDelta(err) {
+			logger.Ctx(ctx).Infow("Invalid previous delta link", "link", prevDelta)
+
+			triedPrevDelta = true // TODO(meain): Do we need this check?
```
ryanfkeepers, line 211: > Do we need this check?

probably not

ryanfkeepers (approved):

src/internal/connector/onedrive/drive.go
```diff
@@ -172,22 +183,41 @@ func collectItems(
...
+			invalidPrevDelta = true
+
+			pager.SetNext("")
```
ashmrtn, line 214: hmmm I think we may actually have to adjust the pager interface for this. I think we'll need something like `Reset()` or such. I have not tested how graph SDK works, but I think if this is set to just `""` then it won't actually have a URL to send a request to. Having a `Reset` function will require either passing in the drive ID again or storing the drive ID when the pager is created. Storing the ID is probably better since it will allow the pager to be used more easily without worrying if the drive ID is also passed to the function that is trying to use it

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1431,3 +1432,94 @@ func delItem(
...
+				{deltaLink: &delta},
+			},
+			prevDeltaSuccess: true,
+		},
+		{
+			name:     "invalid prev delta",
+			deltaURL: delta,
+			items: []deltaPagerResult{
+				{nextLink: &next, err: deltaError},
+				{deltaLink: &delta}, // works on retry
```
ashmrtn, line 1477: nit: could you use a different value here besides `next` and `delta` please just so we can be certain we're returning what we think we are? It's being a bit paranoid, but it also helps reduce the number of assumptions about how the code could be broken 😅

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1431,3 +1432,94 @@ func delItem(
...
+				itemPager,
+				"",
+				"General",
+				collectorFunc,
+				"",
+			)
+
+			require.ErrorIs(suite.T(), err, test.err)
+			require.Equal(suite.T(), test.deltaURL, delta.URL)
+			require.Equal(suite.T(), !test.prevDeltaSuccess, delta.Reset)
```
ashmrtn, line 1522: nit: just for readability if a test fails it's useful at times to add some text to comparisons of types like bools
```suggestion
			require.Equal(suite.T(), !test.prevDeltaSuccess, delta.Reset, "delta reset")
```

src/internal/connector/onedrive/drive.go
meain: Yeah, I guess this specific error is only if the delta is invalid. I am being paranoid.

src/internal/connector/onedrive/drive.go
meain: Good catch. I'll add a `Reset` method with `driveID` added to `driveItemPager`.

src/internal/connector/onedrive/collections_test.go
meain: I'm not sure I understand you.
*Related: The next is useless there and I have removed it*

ashmrtn: oh, I actually totally misread this bit of code. You can ignore this comment 😅

ashmrtn (approved):

PR #2375, opened 2023-02-02
add fault/clues to the rest of internal/kopia

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/conn.go
+++ b/src/internal/kopia/conn.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/alcionai/clues"
 	"github.com/kopia/kopia/fs"
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/blob"
@@ -17,7 +18,6 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 	"github.com/pkg/errors"
 
-	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/pkg/storage"
 )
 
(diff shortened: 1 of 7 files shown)


src/internal/kopia/conn.go
```diff
@@ -389,14 +386,19 @@ func checkCompressor(compressor compression.Name) error {
...
 }
 
 func (w *conn) LoadSnapshots(
 	ctx context.Context,
 	ids []manifest.ID,
 ) ([]*snapshot.Manifest, error) {
-	return snapshot.LoadSnapshots(ctx, w.Repository, ids)
+	mans, err := snapshot.LoadSnapshots(ctx, w.Repository, ids)
+	if err != nil {
+		return nil, clues.Stack(err).WithClues(ctx)
```
meain, line 398: Just wondering if we should have  the `if err != nil` check withing the `Stack` call? This might simplify the logic in a few places. Just as a reference [pkg/errors:Wrap](https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L184-L187) was doing it.

meain (approved):

src/internal/kopia/conn.go
ryanfkeepers: Not a bad idea, but we'd have to keep this pattern here anyway.  The trouble isn't `clues.Stack()`, it's `.WithClues(ctx)`.  If we call `clues.Stack(nil).WithClues(ctx)` we'd segfault.

PR #2377, opened 2023-02-02
replace multierror with fault in setSites

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange.go
+++ b/src/cli/backup/exchange.go
@@ -16,6 +16,7 @@ import (
 	"github.com/alcionai/corso/src/internal/model"
 	"github.com/alcionai/corso/src/pkg/backup"
 	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
 	"github.com/alcionai/corso/src/pkg/selectors"
@@ -272,20 +273,23 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	sel := exchangeBackupCreateSelectors(user, exchangeData)
 
-	users, err := m365.UserPNs(ctx, acct)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
+
+	users, err := m365.UserPNs(ctx, acct, errs)
 	if err != nil {
 		return Only(ctx, errors.Wrap(err, "Failed to retrieve M365 user(s)"))
 	}
 
 	var (
-		errs *multierror.Error
-		bIDs []model.StableID
+		merrs *multierror.Error
+		bIDs  []model.StableID
 	)
 
 	for _, discSel := range sel.SplitByResourceOwner(users) {
 		bo, err := r.NewBackup(ctx, discSel.Selector)
 		if err != nil {
-			errs = multierror.Append(errs, errors.Wrapf(
+			merrs = multierror.Append(merrs, errors.Wrapf(
 				err,
 				"Failed to initialize Exchange backup for user %s",
 				discSel.DiscreteOwner,
@@ -296,7 +300,7 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 		err = bo.Run(ctx)
 		if err != nil {
-			errs = multierror.Append(errs, errors.Wrapf(
+			merrs = multierror.Append(merrs, errors.Wrapf(
 				err,
 				"Failed to run Exchange backup for user %s",
 				discSel.DiscreteOwner,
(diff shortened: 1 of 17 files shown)


src/internal/connector/graph_connector.go
```diff
@@ -338,5 +351,5 @@ func getResources(
 		return nil, clues.Stack(err).WithClues(ctx).WithAll(graph.ErrData(err)...)
 	}
 
-	return resources, errs.ErrorOrNil()
+	return resources, errs.Err()
```
ashmrtn, line 354: this is going to hide non-catastrophic errors which is different than the current behavior. That's what we want in the long run, but for right now it's not clear if this will end up hiding errors from us especially since `errs.Errs()` is never checked in the cli code (or other code in this PR it seems)

src/internal/operations/restore.go
```diff
@@ -316,5 +320,5 @@ func formatDetailsForRestoration(
 		paths[i] = p
 	}
 
-	return paths, nil
+	return paths, errs.Err()
```
ashmrtn, line 323: same here, this can hide errors and the current code may not be setup to deal with that

src/cmd/factory/impl/common.go
```diff
@@ -110,8 +111,10 @@ func getGCAndVerifyUser(ctx context.Context, userID string) (*connector.GraphCon
 		return nil, account.Account{}, errors.Wrap(err, "finding m365 account details")
 	}
 
-	// build a graph connector
-	gc, err := connector.NewGraphConnector(ctx, acct, connector.Users)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
+
+	gc, err := connector.NewGraphConnector(ctx, acct, connector.Users, errs)
```
ashmrtn, line 117: in this and other non-cli commands/helpers we should probably check the value of `errs.Errs()` so we can surface any failures. We really want this to keep the current behavior of the code until we're sure we have all parts of corso ready for best effort. Just checking the returned `err` won't show non-catastrophic failures so it may change behavior

src/cli/backup/exchange.go
```diff
@@ -273,20 +273,23 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	sel := exchangeBackupCreateSelectors(user, exchangeData)
 
-	users, err := m365.UserPNs(ctx, acct)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
+
+	users, err := m365.UserPNs(ctx, acct, errs)
```
ashmrtn, line 279: along the same lines as other comments but flagging here too. The CLI should probably check `errs.Errs()` because if we don't we may change the behavior of corso when we don't mean to

src/cmd/factory/impl/common.go
ryanfkeepers: Honestly what's probably easiest right now is to swap default behavior from bestEffort to failFast.  That way we catch every possible failure.  Then, when we're satisfied with the integration, we can back off to bestEffort again.

src/cli/backup/sharepoint.go
```diff
@@ -315,7 +319,10 @@ func sharePointBackupCreateSelectors(
 		}
 	}
 
-	union, err := gc.UnionSiteIDsAndWebURLs(ctx, sites, weburls)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
```
ashmrtn, line 323: should we just pass `errs` into this function and check it in the caller?

src/cmd/factory/impl/common.go
```diff
@@ -114,7 +115,10 @@ func getGCAndVerifyUser(ctx context.Context, userID string) (*connector.GraphCon
 	}
 
 	// build a graph connector
-	gc, err := connector.NewGraphConnector(ctx, graph.HTTPClient(graph.NoTimeout()), acct, connector.Users)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
```
ashmrtn, line 119: in this and the other stuff in `cmd` package in this PR shouldn't we check if `errs` contains anything? We check the returned `err` from the function, but seems like we're still missing checking `errs`

src/pkg/fault/fault.go
```diff
@@ -40,9 +40,12 @@ type ErrorsData struct {
...
 	return &Errors{
-		mu:       &sync.Mutex{},
-		errs:     []error{},
-		failFast: failFast,
+		mu:   &sync.Mutex{},
+		errs: []error{},
+		// TODO: respect the failfast input.  But while we're
+		// in transition towards fault support, and not completely
+		// checking errs, it's better for us to catch everything.
+		failFast: true,
```
ashmrtn, line 48: we need to make sure this is well documented and folks know about it until we change it. Otherwise it becomes a bit confusing if people just look at the function signature

ashmrtn (approved):

src/pkg/fault/fault.go
ryanfkeepers: Yup.  Per comments in sync, I will probably revert this but find a way to run in failFast for standard behavior.

PR #2380, opened 2023-02-02
adding fault/clues to non-service connector pkgs

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector.go
+++ b/src/internal/connector/graph_connector.go
@@ -98,7 +98,7 @@ func NewGraphConnector(
 	// For now this keeps things functioning if callers do pass in a selector like
 	// "*" instead of.
 	if r == AllResources || r == Users {
-		if err = gc.setTenantUsers(ctx); err != nil {
+		if err = gc.setTenantUsers(ctx, errs); err != nil {
 			return nil, errors.Wrap(err, "retrieving tenant user list")
 		}
 	}
@@ -129,11 +129,11 @@ func (gc *GraphConnector) createService() (*graph.Service, error) {
 // setTenantUsers queries the M365 to identify the users in the
 // workspace. The users field is updated during this method
 // iff the returned error is nil
-func (gc *GraphConnector) setTenantUsers(ctx context.Context) error {
+func (gc *GraphConnector) setTenantUsers(ctx context.Context, errs *fault.Errors) error {
 	ctx, end := D.Span(ctx, "gc:setTenantUsers")
 	defer end()
 
-	users, err := discovery.Users(ctx, gc.Owners.Users())
+	users, err := discovery.Users(ctx, gc.Owners.Users(), errs)
 	if err != nil {
 		return err
 	}
--- a/src/internal/connector/uploadsession/uploadsession.go
+++ b/src/internal/connector/uploadsession/uploadsession.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/pkg/errors"
+	"github.com/alcionai/clues"
 	"gopkg.in/resty.v1"
 
 	"github.com/alcionai/corso/src/pkg/logger"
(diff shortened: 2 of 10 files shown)


src/cli/backup/exchange.go
```diff
@@ -272,20 +273,23 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	sel := exchangeBackupCreateSelectors(user, exchangeData)
 
-	users, err := m365.UserPNs(ctx, acct)
+	// TODO: log/print recoverable errors
+	errs := fault.New(false)
```
dadams39, line 277: Should this error be involved in the logic in line 279?

dadams39: It's not entirely what its purpose is when there is `merrs` and `ferrs`.

src/internal/connector/graph_connector.go
```diff
@@ -79,26 +81,26 @@ func NewGraphConnector(ctx context.Context, acct account.Account, r resource) (*
...
+		return nil, clues.Wrap(err, "creating api client").WithClues(ctx)
 	}
 
 	// TODO(ashmrtn): When selectors only encapsulate a single resource owner that
 	// is not a wildcard don't populate users or sites when making the connector.
 	// For now this keeps things functioning if callers do pass in a selector like
 	// "*" instead of.
 	if r == AllResources || r == Users {
-		if err = gc.setTenantUsers(ctx); err != nil {
+		if err = gc.setTenantUsers(ctx, errs); err != nil {
```
dadams39, line 97: Is there a particular reason why these errors aren't wrapped with clues?

src/internal/connector/uploadsession/uploadsession.go
```diff
@@ -57,15 +57,15 @@ func (iw *writer) Write(p []byte) (n int, err error) {
...
+			"upload_id", iw.id,
+			"upload_chunk_size", rangeLength,
+			"upload_offset", iw.lastWrittenOffset,
+			"upload_size", iw.contentLength)
 	}
 
 	// Update last offset
 	iw.lastWrittenOffset = endOffset
 
-	logger.Ctx(context.Background()).Debugf("Response: %s", resp.String())
```
dadams39, line 68: Are we sure that we don't want to log this any longer?

src/internal/kopia/conn.go
```diff
@@ -87,22 +72,22 @@ func NewConn(s storage.Storage) *conn {
 func (w *conn) Initialize(ctx context.Context) error {
 	bst, err := blobStoreByProvider(ctx, w.storage)
 	if err != nil {
-		return errors.Wrap(err, errInit.Error())
+		return errors.Wrap(err, "initializing storage")
```
dadams39, line 75: Should we use clues here as well?

src/internal/kopia/model_store.go
```diff
@@ -472,26 +496,17 @@ func (ms *ModelStore) Delete(ctx context.Context, s model.Schema, id model.Stabl
...
-		ctx,
-		ms.c,
-		repo.WriteSessionOptions{Purpose: "ModelStoreDelete"},
-		func(innerCtx context.Context, w repo.RepositoryWriter) error {
-			return w.DeleteManifest(innerCtx, id)
-		},
-	)
-
-	return errors.Wrap(err, "deleting model")
-}
```
dadams39, line 488: Is this correct?

src/internal/connector/graph_connector.go
ryanfkeepers: As in, why do we use `errors.Wrap()` instead of `clues.Wrap()`?  Clues is interchangeable with errors, so it won't matter if you wrap using one or the other (unless you need the other clues interface to add additional structured data, too).

src/internal/connector/uploadsession/uploadsession.go
ryanfkeepers: We can always put it back in if we need it.

src/internal/kopia/model_store.go
ryanfkeepers: yup.

src/cli/backup/exchange.go
ryanfkeepers: `merrs` will eventually go away.  The cli refactor is going to be the last piece of this migration, so if you see code in a halfway state like this, it's pending total completion, which is why we have both.

dadams39 (approved):

PR #2381, opened 2023-02-02
add fault/clues to exchange data_collections

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/data_collections_test.go
+++ b/src/internal/connector/data_collections_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/alcionai/corso/src/internal/connector/sharepoint"
 	"github.com/alcionai/corso/src/internal/tester"
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
@@ -105,7 +106,8 @@ func (suite *ConnectorDataCollectionIntegrationSuite) TestExchangeDataCollection
 				nil,
 				connector.credentials,
 				connector.UpdateStatus,
-				control.Options{})
+				control.Options{},
+				fault.New(true))
 			require.NoError(t, err)
 
 			assert.Empty(t, excludes)
@@ -201,7 +203,12 @@ func (suite *ConnectorDataCollectionIntegrationSuite) TestDataCollections_invali
 
 	for _, test := range tests {
 		suite.T().Run(test.name, func(t *testing.T) {
-			collections, excludes, err := connector.DataCollections(ctx, test.getSelector(t), nil, control.Options{})
+			collections, excludes, err := connector.DataCollections(
+				ctx,
+				test.getSelector(t),
+				nil,
+				control.Options{},
+				fault.New(true))
 			assert.Error(t, err)
 			assert.Empty(t, collections)
 			assert.Empty(t, excludes)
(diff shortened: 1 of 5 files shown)


src/internal/connector/data_collections_test.go
```diff
@@ -98,13 +99,16 @@ func (suite *ConnectorDataCollectionIntegrationSuite) TestExchangeDataCollection
 
 	for _, test := range tests {
 		suite.T().Run(test.name, func(t *testing.T) {
+			errs := fault.New(true)
```
ashmrtn, line 102: this is unrelated to this PR but I did want to surface it. Maybe at some point we should make a const in the fault package that is `FailFast` so that this reads `fault.New(fault.FailFast)` instead of a random bool value. It's a small thing, but helps with readability as people don't have to go open up another package or something to figure out what the mysterious bool is

src/internal/connector/exchange/data_collections.go
```diff
@@ -105,13 +105,13 @@ func parseMetadataCollections(
...
 				if err != nil {
-					return nil, errors.New("decoding metadata json")
+					return nil, clues.New("decoding metadata json").WithClues(ctx)
 				}
 
 				switch item.UUID() {
 				case graph.PreviousPathFileName:
 					if _, ok := found[category]["path"]; ok {
-						return nil, errors.Errorf("multiple versions of %s path metadata", category)
+						return nil, clues.Wrap(clues.New(category.String()), "multiple versions of path metadata").WithClues(ctx)
```
ashmrtn, line 114: couple of questions for this and the call below:
* does context already contain the category?
* should we be adding category as a clue and not part of the error message? It seems more like related metadata to help with debugging rather than part of the root cause of the error

src/internal/connector/exchange/data_collections.go
```diff
@@ -185,24 +185,27 @@ func DataCollections(
...
-			return nil, support.WrapAndAppend(user, err, errs)
+			errs.Add(err)
+			continue
 		}
 
 		collections = append(collections, dcs...)
 	}
 
-	return collections, errs
+	return collections, errs.Err()
```
ashmrtn, line 208: I think we need to be careful here, it seems like it has the possibility to hide errors if we're running in not fail fast mode

Some questions about this:
* since we now pass in `fault.Errors` is it required to return this?
* is it possible that we need to return the union of `errs.Err()` and `errs.Errs()`? Or different ones depending on if we're in fail fast mode?

src/internal/connector/exchange/data_collections.go
```diff
@@ -214,7 +217,7 @@ func getterByType(ac api.Client, category path.CategoryType) (addedAndRemovedIte
 	case path.ContactsCategory:
 		return ac.Contacts(), nil
 	default:
-		return nil, fmt.Errorf("category %s not supported by getFetchIDFunc", category)
+		return nil, clues.Wrap(clues.New(category.String()), "category not supported")
```
ashmrtn, line 220: maybe this should also be a metadata tag instead of part of the error?

ashmrtn (approved):

src/internal/connector/exchange/data_collections.go
ryanfkeepers: > does context already contain the category?

Probably

> should we be adding category as a clue and not part of the error message? It seems more like related metadata to help with debugging rather than part of the root cause of the error

Depends on who you imagine consumes the error.  If this is only reviewed in logging, then it can stay as metadata.  If we imagine an end user (even if that's us) receiving it as, say, a cli output, the extra data can be really helpful.

To be honest, I don't think there's a clear rule for how to choose one or the other.  I'm a little arbitrary about the implementation myself.  But I also see this as a benign situation either way, so I don't put much emphasis a "correct" version.

src/internal/connector/exchange/data_collections.go
ryanfkeepers: > since we now pass in fault.Errors is it required to return this?

Yes.  Consider the prior implementation with multi-errors:

```go
var errs error

for := range stuffToDo() {
  if failFast && errs != nil {
    break
  }
  doStuff()
}

return errs.ErrorOrNil()
```

The primary difference between the two implementations is that multierrors would always return an error if one occurred, while fault will only return an error if it is considered a non-recoverable case, which includes checking failFast.  IE: we're now opening  up the ability to continue processing in bestEffort cases.

Returning errs.Err() is necessary because the adding errors via fault provides no guarantees about non-recoverable (or failFast) error cases.  This ensures we don't hide or lose exit cases, while still optionally supporting best effort.

PR #2382, opened 2023-02-03
add fault/clues throughout exchange backup

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/contact_folder_cache.go
+++ b/src/internal/connector/exchange/contact_folder_cache.go
@@ -3,10 +3,10 @@ package exchange
 import (
 	"context"
 
+	"github.com/alcionai/clues"
 	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/internal/connector/graph"
-	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
@@ -26,7 +26,7 @@ func (cfc *contactFolderCache) populateContactRoot(
 ) error {
 	f, err := cfc.getter.GetContainerByID(ctx, cfc.userID, directoryID)
 	if err != nil {
-		return support.ConnectorStackErrorTraceWrap(err, "fetching root folder")
+		return clues.Wrap(err, "fetching root folder")
 	}
 
 	temp := graph.NewCacheFolder(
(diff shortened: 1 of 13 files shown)


src/internal/connector/exchange/contact_folder_cache.go
```diff
@@ -26,13 +26,13 @@ func (cfc *contactFolderCache) populateContactRoot(
 ) error {
 	f, err := cfc.getter.GetContainerByID(ctx, cfc.userID, directoryID)
 	if err != nil {
-		return support.ConnectorStackErrorTraceWrap(err, "fetching root folder")
+		return clues.Wrap(err, "fetching root folder")
```
ashmrtn, line 29: I think this still needs `ConnectorStackErrorTrace` to get details
```suggestion
		return clues.Wrap(support.ConnectorStackErrorTrace(err), "fetching root folder")
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -63,8 +64,8 @@ func filterContainersAndFillCollections(
 	}
 
 	for _, c := range resolver.Items() {
-		if ctrlOpts.FailFast && errs != nil {
-			return errs
+		if errs.Failed() {
+			return errs.Err()
```
ashmrtn, line 68: a bit concerned here too about only returning one of possibly several errors. Seems like either this function should no longer return an error and rely on the in/out param `errs` or should return all errors in `errs`

ashmrtn (approved):

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: I feel like there's some misunderstanding of the fault package behavior we need to clear up.  We can't lose or hide errors, except by ignoring their population when we handle errors upstream.  What we're doing here isn't deciding _what_ errors to return.  It's only a consideration of _when_ to stop processing and return.

src/internal/connector/exchange/contact_folder_cache.go
ryanfkeepers: I haven't gotten around to updating the api package yet (being downstream, it required cascading changes to all upstream callers to implement, which is why this PR went first), but the expectation is that the API will pack that data into the error for this caller by use of [gaph.ErrData()](https://github.com/alcionai/corso/blob/main/src/internal/connector/graph/errors.go#L185).

PR #2383, opened 2023-02-03
extra panic protection in operations

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Test Plan

- [x] :green_heart: E2E


src/internal/operations/backup.go
```diff
@@ -589,6 +596,11 @@ func (op *BackupOperation) persistResults(
 			opStats.writeErr)
 	}
 
+	if opStats.gc == nil {
+		op.Status = Failed
+		return errors.New("data population never completed")
```
zackrossman, line 601: do we want to fail here? other option is to set `op.Status = NoData` and return nil. Not saying I know the right answer, just raising the option

src/internal/operations/backup.go
```diff
@@ -589,6 +596,11 @@ func (op *BackupOperation) persistResults(
 			opStats.writeErr)
 	}
 
+	if opStats.gc == nil {
```
dadams39, line 599: We check if either are nil on line ~582
```go
if opStats.readErr != nil || opStats.writeErr != nil {
		op.Status = Failed
```
Should we only be checking if gc.successful at this point... after the guard check of course?

src/internal/operations/restore.go
```diff
@@ -250,6 +257,11 @@ func (op *RestoreOperation) persistResults(
 			opStats.writeErr)
 	}
 
+	if opStats.gc == nil {
+		op.Status = Failed
+		return errors.New("data restoration never completed")
+	}
+
```
dadams39, line 264: Same question as in backup

src/internal/operations/backup.go
ryanfkeepers: No need at this time.

src/internal/operations/backup.go
ryanfkeepers: we do want to fail.  But I"m glad you pointed it out, because the other stats needed to get moved above this check.

src/internal/operations/backup.go
```diff
@@ -589,15 +596,21 @@ func (op *BackupOperation) persistResults(
...
+	op.Results.BytesUploaded = opStats.k.TotalUploadedBytes
+	op.Results.ItemsWritten = opStats.k.TotalFileCount
+	op.Results.ResourceOwners = opStats.resourceCount
+
+	if opStats.gc == nil {
+		op.Status = Failed
+		return errors.New("data population never completed")
+	}
+
 	if opStats.readErr == nil && opStats.writeErr == nil && opStats.gc.Successful == 0 {
```
zackrossman, line 609: I'm just noticing that there is already a check `if opStats.readErr != nil || opStats.writeErr != nil {` on line 589. Don't need to do it in this PR but we can probably remove the extraneous check from this line

ryanfkeepers: Same thing that danny pointed out.

zackrossman (approved):
LGTM, pending the tests

PR #2386, opened 2023-02-03
handle error from bu.backupCollections

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -445,24 +445,22 @@ func consumeBackupDataCollections(
 		cs,
 		nil,
 		tags,
-		isIncremental,
-	)
+		isIncremental)
+	if err != nil {
+		if kopiaStats == nil {
+			return nil, nil, nil, err
+		}
+
+		return nil, nil, nil, errors.Wrapf(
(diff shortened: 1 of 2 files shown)


src/internal/operations/backup.go
```diff
@@ -445,24 +445,22 @@ func consumeBackupDataCollections(
...
-			)
-		} else {
-			err = errors.Errorf(
-				"kopia snapshot failed with %v catastrophic errors and %v ignored errors",
-				kopiaStats.ErrorCount,
-				kopiaStats.IgnoredErrorCount,
-			)
-		}
+		err = errors.Errorf(
+			"kopia snapshot failed with %v catastrophic errors and %v ignored errors",
```
dadams39, line 468: > "kopia snapshot failed with %v catastrophic errors and %v ignored errors",

Does this need to be %d for integer or is are we good with %v?

zackrossman (approved):
Error handling looks good. If it is possible, unit tests would be nice for this change.

PR #2387, opened 2023-02-03
Adds more retries to OneDrive API calls

meain:
## Description

Adds more reties to handle timeout issues.

## Does this PR need a docs update or release note?

- [x] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #2389, opened 2023-02-03
Standardize HTTP retry handling

neha-Gupta1:
## Description

Handle Exponential retrial of HTTP call in case of failure from endpoint using https://pkg.go.dev/github.com/cenkalti/backoff/v4

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #https://github.com/alcionai/corso/issues/2287

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ryanfkeepers (commented):
You have the correct idea here, but the scope of its application is in a spot that makes it difficult to re-use elsewhere.  What we ultimately want is a single declaration of `backoff`, that all other callers can utilize to ensure they properly retry calls.  Centralized handling minimizes the amount of imports, configuration, and maintenance we need add for any given retry/backoff handling.

A good place to centralize that handling would be the `/connector/graph` package.  See Meain's recent PR that's doing something similar, but without an external library: https://github.com/alcionai/corso/pull/2387.  Injecting `backoff` into that location will give us maximum re-use with minimum fuss.

Issue #2388, opened 2023-02-03
Exchange backup fails if user has multiple calendars with the same name


meain:
### What happened?

When we have multiple calendars with the same name, [this](https://github.com/alcionai/corso/blob/8d01d1b397f19cb965015ccb820be7a6a6899fb1/src/internal/kopia/upload.go#L675-L679) check fails.

Error: 1 error occurred:
        * Failed to run Exchange backup for user Jo...L@8...onmicrosoft.com: 2 errors occurred:
        * errors prevented the operation from processing
        * backing up service data: building kopia directories: inflating collection tree: multiple instances of collection at 4d603060-18d6-4764-b9be-4cb794d32b69/exchange/J..L@8..rj.onmicrosoft.com/events/Calendar

### Corso Version?

Corso version: dev-v0.2.0-48-g8d01d1b39

### Relevant log output

```shell
github.com/alcionai/corso/src/internal/operations.(*BackupOperation).do
        /home/meain/dev/alcion/corso/src/internal/operations/backup.go:245
github.com/alcionai/corso/src/internal/operations.(*BackupOperation).Run
        /home/meain/dev/alcion/corso/src/internal/operations/backup.go:132
github.com/alcionai/corso/src/cli/backup.createExchangeCmd
        /home/meain/dev/alcion/corso/src/cli/backup/exchange.go:297
github.com/spf13/cobra.(*Command).execute
        /tmp/g/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:916
github.com/spf13/cobra.(*Command).ExecuteC
        /tmp/g/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044
github.com/spf13/cobra.(*Command).Execute
        /tmp/g/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/spf13/cobra.(*Command).ExecuteContext
        /tmp/g/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:961
github.com/alcionai/corso/src/cli.Handle
        /home/meain/dev/alcion/corso/src/cli/cli.go:132
main.main
        /home/meain/dev/alcion/corso/src/corso.go:8
runtime.main
        /nix/store/v6i0a6bfx3707airawpc2589pbbl465r-go-1.19.5/share/go/src/runtime/proc.go:250
[...]

PR #2395, opened 2023-02-03
add pathEquals to filter set

ryanfkeepers:
## Description

Adds a pathEquals filter, and a corresponding
option for setting pathEquals in the scope options.
This will enable tests to restrict calendar lookups
to only the primary calendar folder, instead of
including all children underneath.

- [x] :no_entry: No 

## Type of change

- [x] :robot: Test

## Issue(s)

* #2388

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add versions to backups so that we can understand/handle older backup formats
 
 ### Fixed
-- Backing up a calendar that has the same name as the default calendar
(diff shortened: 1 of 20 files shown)


ashmrtn (approved):

ashmrtn:
hrrmmmm it almost seems like now that we've redone the calendar paths stuff we should actually try to use a contains filter for CI testing to ensure we can actually handle the situation we've patched. Otherwise we need to give selectors with the precise paths we need which kind of seems like it defeats the purpose of things

Equals matcher could still be useful though so welcome to merge if you want

PR #2397, opened 2023-02-03
Alternative way to handle 2-level calendars hierarchy

ashmrtn:
## Description

All calendars except the default are nested under a "Other Calendars" folder. Having a non-default calendar named the same as the default calendar does not cause problems when fetching the default calendar by name. Only the default calendar will be returned in that situation.

This fixes the bug where we had multiple collections for the same path but representing different folders.

Also updates the restore execution path to handle the new nested folder structure.

Backup, incremental backup, and restore flows tested manually

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2388

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/data_collections_test.go
+++ b/src/internal/connector/exchange/data_collections_test.go
@@ -537,7 +537,7 @@ func (suite *DataCollectionsIntegrationSuite) TestEventsSerializationRegression(
 		},
 		{
 			name:     "Birthday Calendar",
-			expected: "Birthdays",
+			expected: calendarOthersFolder + "/Birthdays",
 			scope: selectors.NewExchangeBackup(users).EventCalendars(
 				[]string{calendarOthersFolder + "/Birthdays"},
 				selectors.PrefixMatch(),
--- a/src/internal/connector/exchange/data_collections_test.go
+++ b/src/internal/connector/exchange/data_collections_test.go
@@ -539,7 +539,7 @@ func (suite *DataCollectionsIntegrationSuite) TestEventsSerializationRegression(
 			name:     "Birthday Calendar",
 			expected: "Birthdays",
 			scope: selectors.NewExchangeBackup(users).EventCalendars(
-				[]string{"Birthdays"},
+				[]string{calendarOthersFolder + "/Birthdays"},
 				selectors.PrefixMatch(),
 			)[0],
 		},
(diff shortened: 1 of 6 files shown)


ryanfkeepers (commented):
There's a fundamental problem with the approach that doesn't exist in the other variation: if/when we add calendar groups, assuming we have a path like `/ten/srv/usr/events/Other Calendars/...`, I assume we'd store the group by the group name as a level above the calendar folders (eg: `/events/<calGroup>/<calendars>...`).  In that case we'll have a naming conflict anytime someone has a group named `Other Calendars`.

ashmrtn:
> There's a fundamental problem with the approach that doesn't exist in the other variation: if/when we add calendar groups, assuming we have a path like `/ten/srv/usr/events/Other Calendars/...`, I assume we'd store the group by the group name as a level above the calendar folders (eg: `/events/<calGroup>/<calendars>...`). In that case we'll have a naming conflict anytime someone has a group named `Other Calendars`.

That's an excellent point. I did not think that far ahead when writing this

If we happen to switch the naming scheme in kopia to use something other than display names then we'll be fine. Otherwise we'll need to plan around that if we want to store that as part of the hierarchy

I'm not sure if we'll need calendar groups in the hierarchy though. Some limited manual testing showed calendars in other groups showed up with a query to `calendars` endpoint, making them appear a bit more like labels?

ryanfkeepers:
> Some limited manual testing showed calendars in other groups showed up with a query to calendars endpoint, making them appear a bit more like labels?

Huh, you're right.  Wonder why those labels are hidden outside of direct lookup.  Well, good enough for me.  If it happens, I'm happy to approach it by storing IDs.  I will admit I'm a little iffy about the naming convention here.  But I don't really want to block on that.

ryanfkeepers (approved):

src/internal/connector/exchange/data_collections_test.go
```diff
@@ -537,7 +537,7 @@ func (suite *DataCollectionsIntegrationSuite) TestEventsSerializationRegression(
 		},
 		{
 			name:     "Birthday Calendar",
-			expected: "Birthdays",
+			expected: calendarOthersFolder + "/Birthdays",
```
vkamra, line 540: 👏

PR #2399, opened 2023-02-04
fix user query filtering on null userType

ryanfkeepers:
## Description

UserType can be null for users created before
2014.  In order to not filter them out with the
other guest users, we have to amend our user
query to only exclude guests, and retain all
other persons.

## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -59,6 +59,11 @@ const (
 // require more fine-tuned controls in the future.
 // https://stackoverflow.com/questions/64044266/error-message-unsupported-or-invalid-query-filter-clause-specified-for-property
 //
+// ne 'Guest' ensures we don't filter out users where userType = null, which can happen
+// for user accounts created prior to 2014.  In order to use the `ne` comparator, we
+// MUST include $count=true and the ConsistencyLevel: eventual header.
+// https://stackoverflow.com/questions/49340485/how-to-filter-users-by-usertype-null
+//
 //nolint:lll
 var userFilterNoGuests = "onPremisesSyncEnabled eq true OR userType ne 'Guest'"
 
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 
+	absser "github.com/microsoft/kiota-abstractions-go"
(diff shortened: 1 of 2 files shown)


src/internal/connector/discovery/api/users.go
```diff
@@ -59,13 +60,21 @@ const (
 // https://stackoverflow.com/questions/64044266/error-message-unsupported-or-invalid-query-filter-clause-specified-for-property
 //
 //nolint:lll
-var userFilterNoGuests = "onPremisesSyncEnabled eq true OR userType eq 'Member'"
+var userFilterNoGuests = "onPremisesSyncEnabled eq true OR userType ne 'Guest'"
```
gmatev, line 63: `onPremisesSyncEnabled` this was `null` for their users as well I think. Are we sure this is something we want?

src/internal/connector/discovery/api/users.go
```diff
@@ -59,13 +60,21 @@ const (
...
+
 	return &users.UsersRequestBuilderGetRequestConfiguration{
+		Headers: headers,
 		QueryParameters: &users.UsersRequestBuilderGetQueryParameters{
 			Select: []string{userSelectID, userSelectPrincipalName, userSelectDisplayName},
 			Filter: fs,
+			Count:  &t,
 		},
 	}
 }
```
dadams39, line 80: This makes sense. We add the documentation as the source for these changes so they aren't removed by accident in the future.

src/internal/connector/discovery/api/users.go
vkamra: Unclear - I can't find any clear documentation on whether we can have "guest on-prem users" - that would be the concern with this check.

If we remove it - run the risk of not discovering all users
If we keep it - and the check is incorrect - will discover users but backup will probably fail with a "no drive" or "no mailbox found".

Leaning towards the latter for now i.e. keep it.

gmatev: Fine by me

vkamra (approved):

src/internal/connector/discovery/api/users.go
ryanfkeepers: Good call, I added additional explanation/references above the filter const.

PR #2405, opened 2023-02-06
Add a flag to toggle OneDrive delta incrementals

meain:
.## Description

This will help us while this is still in development. We should be
able to add in features without affecting anything else.

Once we have it in a more or less stable state this can be removed
completely.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


meain:
I guess this can be deferred for later. As of now, we ca make use of the fact that we don't get a deleted item until we use a prev delta as well as for new items, as long as we don't really have the oldPaths the behavior should remain the same.

PR #2406, opened 2023-02-06
add retry middleware

neha-Gupta1:
## Description

Added retry middleware which handles retrial of all HTTP request if,
- tooManyRequests
- serviceUnavailable
- gatewayTimeout
- internalServerError 
 
with default no of retrials being 3.

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/issues/2287

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
 
 	return support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
 }
+
+// RetryHandler handles transient HTTP responses and retries the request given the retry options
+type RetryHandler struct {
```
meain, line 370: Since this is mostly pulled from [kiota](https://github.com/microsoft/kiota-http-go/blob/main/retry_handler.go), lets keep it in separate file.

src/internal/connector/graph/service.go
```diff
@@ -160,6 +172,40 @@ func HTTPClient(opts ...option) *http.Client {
...
+			) bool {
+				return true
+			}},
+		},
+	}
+}
+
+// GetDefaultMiddlewaresWithOptions creates a default slice of middleware for the Graph Client.
+func GetMiddlewaresWithOptions(options *msgraphgocore.GraphClientOptions) []khttp.Middleware {
+	kiotaMiddlewares := GetMiddlewares()
```
meain, line 197: nit: This should probably be `GetKiotaMiddlewares`, else it might be confused as a non "WithOptions" version of `GetMiddlewaresWithOptions`, or we could completely remove this indirection.

meain (commented):
LGTM, just a few things
1. There are few other places where we have retries. Look for [maxRetries](https://github.com/alcionai/corso/blob/ac8fe1e9c142538856e18e1ba03bfab97717bff6/src/internal/connector/onedrive/collection.go#L37) and [numberOfRetries](https://github.com/alcionai/corso/blob/ac8fe1e9c142538856e18e1ba03bfab97717bff6/src/internal/connector/exchange/api/api.go#L20).
2. We can delete `RunWithRetry` once we get this working well.
3. Handling of context deadline timeout

neha-Gupta1:
I am yet to remove the `RunWithRetry` function call from many places. 
And still testing the handling of context deadline timeout.

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+}
+
+type RetryHandlerOptions struct {
+	// Callback to determine if the request should be retried
+	ShouldRetry func(delay time.Duration, executionCount int, request *http.Request, response *http.Response) bool
+	// The maximum number of times a request can be retried
+	MaxRetries int
+	// The delay in seconds between retries
+	DelaySeconds int
+}
```
ryanfkeepers, line 382: The indirection here (handler holding an options struct holding actual options) doesn't seem particularly helpful, except for the injection of all the extra interfaces associated with it. Personally, I think the code here could be greatly improved by cutting out all the abstracting and instead using values as plainly and straightforwardly as possible.

Start with a simple RetryHandler configuration, like so, and only add onto that as required.
```go
type RetryHandler struct {
  MaxRetries int
  MinimumBackoff time.Duration
}
```

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+	if !ok {
+		reqOption = &middleware.options
+	}
+
+	return middleware.retryRequest(ctx, pipeline,
+		middlewareIndex, reqOption, req, response, 0, 0, observabilityName)
+}
+
+func (middleware RetryHandler) retryRequest(ctx context.Context,
+	pipeline khttp.Pipeline,
```
ryanfkeepers, line 464: Nit: code standards, all params should be inline, or each on their own line (again, similar types can be stacked).
```suggestion
func (middleware RetryHandler) retryRequest(
	ctx context.Context,
	pipeline khttp.Pipeline,
```

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+		return response, err
+	}
+
+	reqOption, ok := req.Context().Value(retryKeyValue).(retryHandlerOptionsInt)
+	if !ok {
+		reqOption = &middleware.options
+	}
+
+	return middleware.retryRequest(ctx, pipeline,
+		middlewareIndex, reqOption, req, response, 0, 0, observabilityName)
```
ryanfkeepers, line 460: Nit: code standards, all params should be inline, or each on their own line (similar types can be stacked).
```suggestion
	return middleware.retryRequest(
	  ctx,
	  pipeline,
	  middlewareIndex,
	  reqOption,
	  req,
	  response,
	  0, 0,
	  observabilityName)
```

src/internal/connector/graph/errors.go
```diff
@@ -25,6 +25,10 @@ const (
 	errCodeSyncStateNotFound           = "SyncStateNotFound"
 	errCodeResourceNotFound            = "ResourceNotFound"
 	errCodeMailboxNotEnabledForRESTAPI = "MailboxNotEnabledForRESTAPI"
+	tooManyRequests                    = 429
+	serviceUnavailable                 = 503
+	gatewayTimeout                     = 504
+	internalServerError                = 500
```
ryanfkeepers, line 31: Golang's http library has const values for all of these.  We should use those, not declare our own.

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+) (*http.Response, error) {
+	obsOptions := khttp.GetObservabilityOptionsFromRequest(req)
+	ctx := req.Context()
+
+	var (
+		span              trace.Span
+		observabilityName string
+	)
+
+	if obsOptions != nil {
```
ryanfkeepers, line 439: is telemetry required for the baseline implementation of the retry handler?  If not, we should save its addition for a follow-up PR.  Keep this one as small as possible.  Besides, we already have a telemetry package, and may want to integrate with that.

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+
+		req = req.WithContext(ctx)
+	}
+
+	response, err := pipeline.Next(req, middlewareIndex)
+	if err != nil {
+		return response, err
+	}
+
+	reqOption, ok := req.Context().Value(retryKeyValue).(retryHandlerOptionsInt)
```
ryanfkeepers, line 454: I see the retry options being pulled out of the context here, but not being added to it.  Not sure if this is intended as a means to overwrite the current configuration or if it's for some other purpose.  For the moment I'd suggest leaving it out.  We can address ad-hoc customization at a future time.

src/internal/connector/graph/service.go
```diff
@@ -160,6 +172,40 @@ func HTTPClient(opts ...option) *http.Client {
...
+// GetDefaultMiddlewares creates a new default set of middlewares for the Kiota request adapter
+func GetMiddlewares() []khttp.Middleware {
+	return []khttp.Middleware{
+		// khttp.NewRetryHandler(),
+		khttp.NewRedirectHandler(),
+		khttp.NewCompressionHandler(),
+		khttp.NewParametersNameDecodingHandler(),
+		khttp.NewUserAgentHandler(),
+		&LoggingMiddleware{},
+		&RetryHandler{
```
ryanfkeepers, line 184: As per the meeting, our retry handler should be last to occur in the middleware stack (ie, idx 0), not first.

src/internal/connector/graph/service.go
```diff
@@ -160,6 +172,40 @@ func HTTPClient(opts ...option) *http.Client {
 	return httpClient
 }
 
+// GetDefaultMiddlewares creates a new default set of middlewares for the Kiota request adapter
+func GetMiddlewares() []khttp.Middleware {
+	return []khttp.Middleware{
+		// khttp.NewRetryHandler(),
```
ryanfkeepers, line 178: We should probably keep the khttp retry handler in place.  Continue letting it handle retries in the way that it prefers (ex, handling throttling, etc), and only concern ourselves with the handling that's left over.

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+		executionCount < options.GetMaxRetries() &&
+		cumulativeDelay < time.Duration(absoluteMaxDelaySeconds)*time.Second &&
+		options.GetShouldRetry()(cumulativeDelay, executionCount, req, resp) {
+		executionCount++
+		delay := middleware.getRetryDelay(req, resp, options, executionCount)
+		cumulativeDelay += delay
+
+		req.Header.Set(retryAttemptHeader, strconv.Itoa(executionCount))
+
+		if req.Body != nil {
```
ryanfkeepers, line 484: What is this block doing?

src/internal/connector/graph/service.go
```diff
@@ -319,3 +365,181 @@ func RunWithRetry(run func() error) error {
...
+			s, ok := req.Body.(io.Seeker)
+			if ok {
+				_, err := s.Seek(0, io.SeekStart)
+				if err != nil {
+					return &http.Response{}, err
+				}
+			}
+		}
+
+		if observabilityName != "" {
```
ryanfkeepers, line 494: same question again on telemetry.

src/internal/connector/graph/service.go
neha-Gupta1: We can do that but the issue is - we won't be able to control the `no. of retries` or `delays` in that case. And have to go with default khttp values.

src/internal/connector/graph/service.go
```diff
@@ -297,25 +357,23 @@ func (handler *LoggingMiddleware) Intercept(
...
+	if err != nil && !IsErrTimeout(err) {
+		return response, err
+	}
 
-		if i < numberOfRetries {
-			time.Sleep(time.Duration(3*(i+2)) * time.Second)
-		}
+	reqOption, ok := req.Context().Value(retryKeyValue).(RetryHandlerOptions)
+	if !ok {
+		reqOption = middleware.options
```
ryanfkeepers, line 375: I'm still not sure this is a helpful pattern for us, when these configurations are already accessible through the RetryHandler itself.  It implies that configuration is expected to change on a per-call basis, and I'm not sure we have reason to support that at this time.  Is there a situation where you expect us to need this in the immediate future?

ryanfkeepers (commented):
Lots of little things, but overall the changes are much improved already.  The reduced complexity from removing interfaces and extra code will really help the middleware stay readable and maintainable.  I think you have room to take one more step forward in that direction, which is what I've focused most of the comments on.  It's very close to being in the right state.

src/internal/connector/graph/retry_middleware.go
```diff
@@ -0,0 +1,131 @@
...
+type RetryHandler struct {
+	// default options to use when evaluating the response
+	options RetryHandlerOptions
+}
+
+type RetryHandlerOptions struct {
+	// request never retried if flag set to true
+	NoRetry bool
+	// The maximum number of times a request can be retried
+	MaxRetries int
```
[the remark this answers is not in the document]
neha-Gupta1: scenario A-
- user wants No Retries-
```
input provided-
MaxRetries- 0
```

- user wants default Retries-

```
no input provided. But since `MaxRetries` is a `int` we get the default value 0-
MaxRetries -0
```
In both the scenarios we will get 0 as `MaxRetries` and it would be difficult to decide retries.
Please correct me if I am missing something in this scenario.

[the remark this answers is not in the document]
neha-Gupta1: The current code is written in a way that - if `noRetries` is set to true no other configuration will be taken in consideration.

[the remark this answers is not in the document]
neha-Gupta1: Since default value of `MaxRetries(type int)` is 0. It was difficult to know if 0 was send or input or if it is a default value. 
To handle this, 
- we ask user for graph.MaxRetries(). If provided as graph.MaxRetries(0), we internally set `noRetry` flag to true. 
Please suggest if there is a better way to handle it.

src/internal/connector/graph/retry_middleware.go
```diff
@@ -0,0 +1,131 @@
...
+	resp *http.Response,
+	executionCount int,
+	cumulativeDelay time.Duration,
+	respErr error,
+) (*http.Response, error) {
+	if options.MaxRetries < 1 {
+		options.MaxRetries = defaultMaxRetries
+	}
+
+	if (respErr != nil || middleware.isRetriableErrorCode(req, resp.StatusCode)) &&
```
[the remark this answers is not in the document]
neha-Gupta1: there are few scenarios where `respErr == nil` but status code is a non successful status code. Retrial should still happen on basis of the other or condition i.e. (middleware.isRetriableErrorCode(req, resp.StatusCode)).

src/internal/connector/graph/retry_middleware.go
```diff
@@ -0,0 +1,131 @@
...
+				if err != nil {
+					return &http.Response{}, err
+				}
+			}
+		}
+
+		time.Sleep(delay)
+
+		response, err := pipeline.Next(req, middlewareIndex)
+		if err != nil && !IsErrTimeout(err) {
```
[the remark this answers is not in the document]
neha-Gupta1: Correct. The client hasn't transformed a non-2xx response into an error till this point.

src/internal/connector/graph/retry_middleware.go
```diff
@@ -0,0 +1,131 @@
...
+// Client Middleware
+// ---------------------------------------------------------------------------
+
+// RetryHandler handles transient HTTP responses and retries the request given the retry options
+type RetryHandler struct {
+	// default options to use when evaluating the response
+	options RetryHandlerOptions
+}
+
+type RetryHandlerOptions struct {
```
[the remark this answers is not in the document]
neha-Gupta1: For doing that 3 more variable will be passed through multiple functions. Worried if the code will become less readable because of that.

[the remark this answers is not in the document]
ryanfkeepers: Multiple parameters (especially only 3) isn't a readability concern.

src/internal/connector/graph/retry_middleware.go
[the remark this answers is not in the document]
ryanfkeepers: I'd recommend adding the flag to the constructor options, but not the retryHandler struct.  Ex:
```go
type RetryHandler struct {
    MaxRetries int
    MinDelay time.Duration
}

type clientConfig struct {
    noTimeout bool
    maxRetires int
    overrideRetryCount bool
}

func MaxRetries(max int) option {
    return func(c *clientConfig) {
        c.overrideRetryCount = true
        c.maxRetries = max
    }
}

func (c *clientConfig) applyMiddlewareConfig(handler *RetryHandler) {
    if c.overrideRetryCount {
         handler.maxRetries = c.maxRetries
    }
}
```

src/internal/connector/graph/service.go
```diff
@@ -99,6 +106,24 @@ func (c *clientConfig) populate(opts ...option) *clientConfig {
...
+			retryoptions.MaxRetries = c.retry.MaxRetries
+		} else {
+			retryoptions.MaxRetries = defaultMaxRetries
+		}
+		if c.retry.Delay > 0 {
+			retryoptions.Delay = c.retry.Delay
+		} else {
+			retryoptions.Delay = defaultDelay
+		}
+	}
```
ryanfkeepers, line 124: nit: for code cleanliness, prefer returns over nested else blocks.
```suggestion
	if c.retry.NoRetry {
		retryoptions = &RetryHandlerOptions{NoRetry: true}
		return
	}

	retryoptions.MaxRetries = defaultMaxRetries
	if c.retry.MaxRetries > 0 {
		retryoptions.MaxRetries = c.retry.MaxRetries
	}
	
	retryoptions.Delay = defaultDelay
	if c.retry.Delay > 0 {
		retryoptions.Delay = c.retry.Delay
	}
```

ryanfkeepers (approved):
Approving to unblock.  Wouldn't mind seeing the final changes as suggested, but the general behavior is sufficiently solid.  Thanks for hearing out all my concerns, it's looking great!

Issue #2242, opened 2023-01-24
add items to delete list in GraphConnector


ashmrtn:
Deleted and moved items will need to be moved to a "global exclude list" in kopia wrapper that tell it not to source those items from the base snapshot. This will avoid duplicates of the items appearing when things are moved or deleted. This is only required for items as we maintain a separate map[folder ID]->folder path for folders

Items must be added to this list because we do not know the previous location the item existed at. This precludes us from marking the item as deleted in the previous location and new in the new location

ashmrtn:
some of this can only be done after #2148 has been merged because that will add some extra files that need managed

meain:
While we have the initial changes, I believe we need https://github.com/alcionai/corso/issues/1535 to finalize and close this out and so keeping this open. As we have to decide on how the naming convention(with ids) works out to decide how to expand(if necessary) the exclude list as well as make use of the exclude list.

ashmrtn:
I think we'll need to expand the ID list at least a little bit. We should probably mark `.meta` files as deleted when an item is deleted

ashmrtn:
resolved in #2557

PR #2407, opened 2023-02-06
Set proper PreviousPath and State for OneDrive collections

meain:
## Description

This wires up OneDrive incrementals to properly set PreviousPath and State values in the Collections.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/issues/2123
* https://github.com/alcionai/corso/issues/2242

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -136,7 +136,9 @@ func NewCollection(
 }
 
 // Figures out the state of a collection using prev and current path
-// FIXME(meain): Same as exchange_data_collection.go:stateOf
+// This is similar to the one in exchange, but instead of checking the
+// equivalence of folders, it checks it for the actual item.
+// TODO(meain): Why is exchange checking folders?
 func stateOf(prev, curr path.Path) data.CollectionState {
 	if curr == nil || len(curr.String()) == 0 {
 		return data.DeletedState
@@ -146,7 +148,7 @@ func stateOf(prev, curr path.Path) data.CollectionState {
 		return data.NewState
 	}
 
-	if curr.Folder() != prev.Folder() {
+	if curr.String() != prev.String() {
 		return data.MovedState
 	}
 
@@ -169,10 +171,13 @@ func (oc *Collection) FullPath() path.Path {
 	return oc.folderPath
 }
 
-// TODO(ashmrtn): Fill in with previous path once GraphConnector compares old
-// and new folder hierarchies.
 func (oc Collection) PreviousPath() path.Path {
-	return nil
+	return oc.prevPath
+}
+
+func (oc *Collection) SetPreviousPath(prevPath path.Path) {
+	oc.prevPath = prevPath
+	oc.state = stateOf(prevPath, oc.folderPath)
 }
 
 // TODO(ashmrtn): Fill in once GraphConnector compares old and new folder
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -69,6 +69,14 @@ type Collection struct {
 	itemMetaReader itemMetaReaderFunc
 	ctrl           control.Options
 
+	// PrevPath is the previous hierarchical path used by this collection.
+	// It may be the same as fullPath, if the folder was not renamed or
+	// moved.  It will be empty on its first retrieval.
+	prevPath path.Path
+
+	// Specifies if it new, moved/rename or deleted
+	state data.CollectionState
+
 	// should only be true if the old delta token expired
 	doNotMergeItems bool
 }
@@ -92,6 +100,7 @@ type itemMetaReaderFunc func(
 func NewCollection(
 	itemClient *http.Client,
 	folderPath path.Path,
+	prevPath path.Path,
 	driveID string,
 	service graph.Servicer,
 	statusUpdater support.StatusUpdater,
@@ -102,6 +111,7 @@ func NewCollection(
 	c := &Collection{
 		itemClient:      itemClient,
 		folderPath:      folderPath,
+		prevPath:        prevPath,
 		driveItems:      map[string]models.DriveItemable{},
 		driveID:         driveID,
 		source:          source,
@@ -109,6 +119,7 @@ func NewCollection(
 		data:            make(chan data.Stream, collectionChannelBufferSize),
 		statusUpdater:   statusUpdater,
 		ctrl:            ctrlOpts,
+		state:           stateOf(prevPath, folderPath),
 		doNotMergeItems: doNotMergeItems,
 	}
 
@@ -122,6 +133,24 @@ func NewCollection(
 	}
 
 	return c
+}
+
+// Figures out the state of a collection using prev and current path
+// FIXME(meain): Same as exchange_data_collection.go:stateOf
+func stateOf(prev, curr path.Path) data.CollectionState {
+	if curr == nil || len(curr.String()) == 0 {
+		return data.DeletedState
+	}
+
+	if prev == nil || len(prev.String()) == 0 {
+		return data.NewState
+	}
+
+	if curr.Folder() != prev.Folder() {
+		return data.MovedState
+	}
+
+	return data.NotMovedState
 }
 
 // Adds an itemID to the collection
@@ -149,7 +178,7 @@ func (oc Collection) PreviousPath() path.Path {
 // TODO(ashmrtn): Fill in once GraphConnector compares old and new folder
 // hierarchies.
 func (oc Collection) State() data.CollectionState {
-	return data.NewState
+	return oc.state
 }
 
 func (oc Collection) DoNotMergeItems() bool {
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -119,7 +119,7 @@ func NewCollection(
 		data:            make(chan data.Stream, collectionChannelBufferSize),
 		statusUpdater:   statusUpdater,
 		ctrl:            ctrlOpts,
-		state:           stateOf(prevPath, folderPath),
+		state:           data.StateOf(prevPath, folderPath),
 		doNotMergeItems: doNotMergeItems,
 	}
 
@@ -133,26 +133,6 @@ func NewCollection(
 	}
 
 	return c
-}
-
-// Figures out the state of a collection using prev and current path
-// This is similar to the one in exchange, but instead of checking the
-// equivalence of folders, it checks it for the actual item.
-// TODO(meain): Why is exchange checking folders?
-func stateOf(prev, curr path.Path) data.CollectionState {
-	if curr == nil || len(curr.String()) == 0 {
-		return data.DeletedState
-	}
-
-	if prev == nil || len(prev.String()) == 0 {
-		return data.NewState
-	}
-
-	if curr.String() != prev.String() {
-		return data.MovedState
-	}
-
-	return data.NotMovedState
 }
 
 // Adds an itemID to the collection
@@ -177,11 +157,9 @@ func (oc Collection) PreviousPath() path.Path {
 
 func (oc *Collection) SetPreviousPath(prevPath path.Path) {
 	oc.prevPath = prevPath
-	oc.state = stateOf(prevPath, oc.folderPath)
+	oc.state = data.StateOf(prevPath, oc.folderPath)
 }
 
-// TODO(ashmrtn): Fill in once GraphConnector compares old and new folder
-// hierarchies.
 func (oc Collection) State() data.CollectionState {
 	return oc.state
 }
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -373,6 +373,7 @@ func (c *Collections) UpdateCollections(
 	items []models.DriveItemable,
 	oldPaths map[string]string,
 	newPaths map[string]string,
+	visitedPaths map[string]string,
 	excluded map[string]struct{},
 	invalidPrevDelta bool,
 ) error {
@@ -405,7 +406,11 @@ func (c *Collections) UpdateCollections(
 
 		switch {
 		case item.GetFolder() != nil, item.GetPackage() != nil:
-			var prevPath path.Path
+			var (
+				prevPath path.Path
+				col      data.BackupCollection
+				found    bool
+			)
 
 			prevColPath, ok := oldPaths[*item.GetId()]
 			if ok {
@@ -466,7 +471,11 @@ func (c *Collections) UpdateCollections(
 			// already created and partially populated.
 			updatePath(newPaths, *item.GetId(), folderPath.String())
 
-			col, found := c.CollectionMap[folderPath.String()]
+			pth, pok := visitedPaths[*item.GetId()]
+			if pok {
+				col, found = c.CollectionMap[pth]
+			}
+
 			if !found {
 				// Skipping new folders here as for non empty folders
 				// we will end up creating a folder in the next case
@@ -491,15 +500,18 @@ func (c *Collections) UpdateCollections(
 				// we encountered a file, this gives us a chance
 				// to mark the container as moved or not moved
 				// instead of new.
-				ccol, ok := col.(*Collection)
+				ocol, ok := col.(*Collection)
 				if !ok {
 					logger.Ctx(ctx).Errorw("unable to cast onedrive collection", "error", err)
 					return err
 				}
 
-				ccol.SetPreviousPath(prevPath)
-				c.CollectionMap[folderPath.String()] = ccol
+				delete(c.CollectionMap, pth)
+				ocol.SetPreviousPath(prevPath)
+				c.CollectionMap[folderPath.String()] = ocol
 			}
+
+			visitedPaths[*item.GetId()] = folderPath.String()
 
 			if c.source != OneDriveSource {
 				continue
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -406,6 +406,7 @@ func (c *Collections) UpdateCollections(
 		switch {
 		case item.GetFolder() != nil, item.GetPackage() != nil:
 			var prevPath path.Path
+
 			prevColPath, ok := oldPaths[*item.GetId()]
 			if ok {
 				prevPath, err = path.FromDataLayerPath(prevColPath, false)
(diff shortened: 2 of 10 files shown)


src/internal/connector/onedrive/collections.go
```diff
@@ -405,13 +405,45 @@ func (c *Collections) UpdateCollections(
 
 		switch {
 		case item.GetFolder() != nil, item.GetPackage() != nil:
+			prevColPath, ok := oldPaths[*item.GetId()]
```
ashmrtn, line 408:
```suggestion
			var prePath path.Path

			prevColPath, ok := oldPaths[*item.GetId()]
			if ok {
				prevPath, err := path.FromDataLayerPath(prevColPath, false)
				if err != nil {
					logger.Ctx(ctx).Errorw("invalid path for deleted item", "error", err)
					return err
				}
			}
```

src/internal/connector/onedrive/collections.go
```diff
@@ -434,15 +466,62 @@ func (c *Collections) UpdateCollections(
 			// already created and partially populated.
 			updatePath(newPaths, *item.GetId(), folderPath.String())
 
+			if ok {
```
ashmrtn, line 469: We don't need this if check. We can just run the code below in all cases though maybe only make a new collection if it was moved so we don't create new empty collections accidentally

If we keep the if check I think there's an edge case where the user made a new folder (at some point, maybe between backups), we've picked up the new folder in the results but haven't gotten all results, the user moves the folder, and then we get a new result with the folder move. I'm not sure if Graph will ever return a sequence like that, but it may be best if we plan on handling it since it would be difficult to track down

src/internal/connector/onedrive/collections.go
```diff
@@ -434,15 +466,62 @@ func (c *Collections) UpdateCollections(
...
+					// to mark the container as moved or not moved
+					// instead of new.
+					ccol, ok := col.(*Collection)
+					if !ok {
+						logger.Ctx(ctx).Errorw("unable to cast onedrive collection", "error", err)
+						return err
+					}
+
+					ccol.SetPreviousPath(prevPath)
+					c.CollectionMap[folderPath.String()] = ccol
```
ashmrtn, line 505: this could result in the collection ending up in the map multiple times maybe (see earlier comment about maybe picking up both "halves" of a folder move). We should probably do the following in these cases:
1. get the current path of the collection
2. set the collection's FullPath path to `folderPath`
3. `delete(c.CollectionMap, previousFullPath.String())`
4. add the folder to the map at the current path

src/internal/connector/onedrive/collections.go
```diff
@@ -434,15 +466,62 @@ func (c *Collections) UpdateCollections(
...
 
 			fallthrough
 
 		case item.GetFile() != nil:
-			if item.GetDeleted() != nil {
+			if !invalidPrevDelta && item.GetFile() != nil {
+				// Always add a file to the excluded list. If it was
+				// deleted, we want to avoid it. If it was
+				// renamed/moved/modified, we still have to drop the
+				// original one and repull the new one.
```
ashmrtn, line 520: I think this works for now, though we may want to refine it in the future to only add it to the exclude list if the delta was non-empty and valid. Empty deltas mean we're pulling everything and not merging with a base so we don't need to exclude stuff

src/internal/connector/onedrive/collections.go
```diff
@@ -451,17 +530,19 @@ func (c *Collections) UpdateCollections(
...
-
 			if !found {
-				// TODO(ashmrtn): Compare old and new path and set collection state
-				// accordingly.
+				var prevPath path.Path
+				// TODO(meain): better check for root
+				if strings.HasSuffix(collectionPath.Folder(), "root:") && !invalidPrevDelta {
+					// Root collection is always added with not moved flag
+					prevPath = collectionPath
+				}
```
ashmrtn, line 540: may be better to handle this by adding the root folder to the set of metadata we save. Then we don't need extra logic for every item

src/internal/connector/onedrive/collections.go
```diff
@@ -434,15 +466,62 @@ func (c *Collections) UpdateCollections(
...
+						prevPath,
+						driveID,
+						c.service,
+						c.statusUpdater,
+						c.source,
+						c.ctrl,
+						invalidPrevDelta,
+					)
+					c.CollectionMap[folderPath.String()] = col
+					c.NumContainers++
```
ashmrtn, line 492: we only want to make new collections if the folder was moved. Otherwise we risk backup up empty folders which we've avoided thus far in OneDrive (not sure if we should rethink that at some point)
```suggestion
					if prevPath != nil {
						col := NewCollection(
							c.itemClient,
							folderPath,
							prevPath,
							driveID,
							c.service,
							c.statusUpdater,
							c.source,
							c.ctrl,
							invalidPrevDelta,
						)
						c.CollectionMap[folderPath.String()] = col
						c.NumContainers++
					}
```

src/internal/connector/onedrive/collections.go
meain: Not sure what you mean by empty delta, but we do have a check to see if we are pulling from a saved delta query or a new one with the `invalidPrevDelta` flag.

src/internal/connector/onedrive/collections.go
meain: Looking at the graph API response, if we create a folder, create a file in it, then rename the folder to a separate name, we just get two delta entries one for the folder with the new name and one for the file with the path as the new location of the folder.

src/internal/connector/onedrive/collections.go
meain: I think the same [comment about delta only returning latest state](https://github.com/alcionai/corso/pull/2407/files#r1098458571) applies here as well.

src/internal/connector/onedrive/collections.go
meain: I've added in a TODO there for now, let me pick it up in a follow up PR. It requires a bit more changes in the code.

src/internal/connector/onedrive/collection.go
```diff
@@ -122,6 +135,26 @@ func NewCollection(
 	return c
 }
 
+// Figures out the state of a collection using prev and current path
+// This is similar to the one in exchange, but instead of checking the
+// equivalence of folders, it checks it for the actual item.
+// TODO(meain): Why is exchange checking folders?
+func stateOf(prev, curr path.Path) data.CollectionState {
```
ashmrtn, line 142: This can be done in a follow up PR, but we should centralize this function since it's probably going to be used a lot. I guess maybe we can put it in the `data` package since it's independent from the service making the collections?

The Exchange version of this function also has tests associated with it which will need to be moved

src/internal/connector/onedrive/collection.go
```diff
@@ -138,16 +171,19 @@ func (oc *Collection) FullPath() path.Path {
...
+	return oc.prevPath
+}
+
+func (oc *Collection) SetPreviousPath(prevPath path.Path) {
+	oc.prevPath = prevPath
+	oc.state = stateOf(prevPath, oc.folderPath)
 }
 
 // TODO(ashmrtn): Fill in once GraphConnector compares old and new folder
 // hierarchies.
```
ashmrtn, line 184: This TODO can be removed
```suggestion
```

src/internal/connector/onedrive/collections.go
```diff
@@ -404,13 +405,45 @@ func (c *Collections) UpdateCollections(
...
+			if ok {
+				prevPath, err = path.FromDataLayerPath(prevColPath, false)
+				if err != nil {
+					logger.Ctx(ctx).Errorw("invalid path for deleted item", "error", err)
+					return err
+				}
+			}
+
 			if item.GetDeleted() != nil {
+				if !ok {
```
ashmrtn, line 420: nit: can you have conditional branches based on `prevPath` instead? It feels a little more descriptive than `ok` when reading the code 😅 
```suggestion
				if prevPath == nil {
```

src/internal/connector/onedrive/collections.go
```diff
@@ -433,15 +466,57 @@ func (c *Collections) UpdateCollections(
...
+						c.source,
+						c.ctrl,
+						invalidPrevDelta,
+					)
+					c.CollectionMap[folderPath.String()] = col
+					c.NumContainers++
+				}
+			} else {
+				// If we have previously created a collection when
+				// we encountered a file, this gives us a change
```
ashmrtn, line 491:
```suggestion
				// we encountered a file, this gives us a chance
```

src/internal/connector/onedrive/collections.go
```diff
@@ -433,15 +466,57 @@ func (c *Collections) UpdateCollections(
...
+				// to mark the container as moved or not moved
+				// instead of new.
+				ccol, ok := col.(*Collection)
+				if !ok {
+					logger.Ctx(ctx).Errorw("unable to cast onedrive collection", "error", err)
+					return err
+				}
+
+				ccol.SetPreviousPath(prevPath)
+				c.CollectionMap[folderPath.String()] = ccol
```
ashmrtn, line 501: hmmmm, there's still a potential issue here I think. I think it's still possible to end up with the collection in the map 2x (or more) if we have an order of operations like

1. start fetching delta results
2. see folder at path1
3. folder moves to path2
4. see folder at path2

Looking up the folder by the current path in our map won't suffice because the path has changed and the lookup will return nil. We may need to expand Collections to have a second map of folder ID -> *Collection? then we could do something like

```go
col, found := c.collectionsByID[*item.GetId()]
if !found {
  // Do other things like the code already has
} else {
  ccol, ok := col.(*Collection)
  if !ok {
    // log and/or return error
  }

  delete(c.CollectionMap, col.FullPath().String())
  ccol.SetPreviousPath(prevPath)
  c.CollectionMap[folderPath.String()] = ccol
}
```

If you add an additional map to `Collections` (I haven't thought of alternative approaches so you may come up with a different solution) please make a separate PR for it. We can then review and merge that PR before this one goes in

src/internal/connector/onedrive/collections.go
ashmrtn: mkay, that will work as well. Maybe add a comment somewhere that `invalidPrevDelta` also includes an empty delta so that folks know. Or if there's some part of the code that makes that clear then it's fine (I may have forgotten how `invalidPrevDelta`'s value is assigned 😅)

src/internal/connector/onedrive/collection.go
```diff
@@ -122,6 +135,26 @@ func NewCollection(
 	return c
 }
 
+// Figures out the state of a collection using prev and current path
+// This is similar to the one in exchange, but instead of checking the
+// equivalence of folders, it checks it for the actual item.
+// TODO(meain): Why is exchange checking folders?
```
ashmrtn, line 141: I think maybe this comment needs updated? State should be associated with folder paths (or whatever the equivalent for folders is for a given external service). Individual item deletion/creation is controlled by adding `data.Stream` instances to a collection

src/internal/connector/onedrive/collections.go
meain: It was not properly handling empty case until now, I've added it [here](https://github.com/alcionai/corso/pull/2407/commits/d061fc983ae100f9493626629f52ff62cb7f7415).

src/internal/connector/onedrive/collections.go
meain: From my testing, it does not seem like a rename in between fetching delta queries result in multiple delta objects. That said, there is "The same item may appear more than once in a delta feed, for various reasons. You should use the last occurrence you see." in the [docs](https://learn.microsoft.com/en-us/graph/api/driveitem-delta?view=graph-rest-1.0&tabs=http#remarks) which might point to a possibility of this.

My best guess is that they are sort of paging on ids on the graph backend and he skip token has that information.

src/internal/connector/onedrive/collection.go
meain: Fixed it, https://github.com/alcionai/corso/pull/2407#discussion_r1097737786 was the culprit.

src/internal/connector/onedrive/collections.go
meain: I think I have a relatively simpler solution. You can find it in [this commit](https://github.com/alcionai/corso/pull/2407/commits/0220124eddfcda8cdc8e2289a66a71988e65c763). Did not bother making a separate PR as we are not changing anything in the original Collections interface. I believe this should work.

meain: However any thoughts on how we would handle "The same item may appear more than once in a delta feed, for various reasons. You should use the last occurrence you see." for items?

src/internal/data/data_collection_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+func TestStateOf(t *testing.T) {
```
ashmrtn, line 12: we've been trying to have all tests use testify. can you please make a testify test suite for this and run it with that?
```suggestion
type DataCollectionSuite struct {
  suite.Suite
}

func TestDataCollectionSuite(t *testing.T) {
  suite.Run(t, new(DataCollectionSuite)
}

func (suite *DataCollectionSuite) TestStateOf(t *testing.T) {
```

src/internal/connector/onedrive/collections.go
```diff
@@ -405,13 +406,49 @@ func (c *Collections) UpdateCollections(
...
+					prevPath,
+					driveID,
+					c.service,
+					c.statusUpdater,
+					c.source,
+					c.ctrl,
+					invalidPrevDelta,
+				)
+
+				c.CollectionMap[prevColPath] = col
```
ashmrtn, line 450: I think indexing `CollectionMap` with the previous path could cause issues in some cases. I think we may hit a problem if we encounter the following:
1. delete known folder at path `/foo`
2. create new folder at path `/foo`

In that case, the newly created folder will clobber the entry for the deleted one. The folder itself will end up with the correct items in the snapshot (new folders shouldn't be merged) but it will cause any subfolders of the old folder to be added to the snapshot instead of being deleted (e.x. if a folder `/foo/bar` existed it will be a subfolder of the new `/foo` folder instead of being part of the deletion of `/foo`)

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1222,7 +1328,41 @@ func (suite *OneDriveCollectionsSuite) TestGet() {
...
+							driveItem("file", "file", testBaseDrivePath+"/folder", true, false, false),
+							driveItem("folder", "folder2", testBaseDrivePath, false, true, false),
+						},
+						deltaLink: &delta,
+					},
+				},
+			},
+			errCheck: assert.NoError,
+			expectedCollections: map[string][]string{
+				folderPath:     {"file"},
```
ashmrtn, line 1352: with the current code indexing on folder paths I think this is actually incorrect. I think the key we'd need would be testBaseDrivePath+"/folder2" since this test renames the folder

See my other comment about the PR in general though as I'm not sure right now if indexing by collection name is a good idea

ashmrtn (commented):
First, sorry, I forgot how involved this can all end up because there's always edge cases

Second, this PR is snowballing on us and it's going to become unmanageable soon so we should break it into chunks. Moving the `StateOf` call to the `data` package should be done in a separate PR. That code is already isolated and can be easily approved/merged. This PR can then make use of it

Next, after reviewing the code again and seeing a possible issue with deletions and newly created collections, we should probably actually transition `CollectionMap` to use folder IDs as keys instead of paths. The folder ID is available in the information that Graph API returns, so it should be just swapping out the key information where needed. That should also be a separate PR to main that this PR will then be based on. Tests will need some updating, but there's already an idea of IDs so hopefully it's just swapping out values. The `TestGet` test changes will be slightly more involved as there will be the chance of multiple collections with the same path being returned. Those collections will just have different states. Since we don't really want to look inside the `Collections` struct for our results, I guess we can make the `expectedCollections` be `map[string]map[data.CollectionState][]string` and add that as part of the indexing [here](https://github.com/alcionai/corso/blob/adbe85b47fb3e9dc9536556517907a9a6315a31f/src/internal/connector/onedrive/collections_test.go#L1556)?

Once the other two PRs are made, this one can make use of them and focus specifically on getting all the states set and such. This can either wait for the others to merge or base off of them as part of a chain

src/internal/data/data_collection_test.go
meain: Resolved in https://github.com/alcionai/corso/pull/2455

src/internal/connector/onedrive/collections.go
meain: Hmm, indexing by name is starting to look more and more flaky. Let me switch to indexing by id. The reason why I was reluctant initially was because we were creating collection when we first encounter an item(this is done so as to avoid empty collections), but at that point the id of a collection is not available. We could store somewhere else and use it when we first find an item, but my worry was weather graph would serve us with a item before the corresponding directory (I have not seen this happen yet).

meain: Thinking about we, we might have to start ordering the collections in the order of how they appear. I was initially thinking we could add last modified timestamp to collections, but deletes do not have timestamp associated with them. I think out best bet would be to keep a list of ids in the order of when we encounter them and use that to decide the order in which we return the final set of collections.

meain: Opened https://github.com/alcionai/corso/pull/2460 for ordering.

ashmrtn: as we discussed out of band, kopia is order insensitive as it only examines the previous and current paths of a collection. This means it can't do "tree operations" on collections because it assumes collections know their current location. Instead, we'll need to fixup the path for collections that we already have if one of the parent collections moves. For example, if we have `/foo/bar` because we saw an item for it and then we see the equivalent of `mv /foo /baz` we'll need to update the `FullPath()` of `/foo/bar` to be `/baz/bar`

Since we now have IDs for each collection each map lookup should be easier. The new path of the collection after the parent move will probably be in `newPaths` as that's updated below

I don't think a variable to track the paths we've seen or the order collections arrive in will be required

Issue #2373, opened 2023-02-02
Separate integration from smoke testing


ryanfkeepers:
## Issue
Corso integration tests are flooded and tend to overlap behavior across layers.  For instance, at least three layers of testing depend on running a complete backup, and that can occur multiple times across tests on similar datasets in each layer.  This bloats our CI runtime without significant gain.

## Proposal
Divide integration tests into two buckets: integration, and smoke.

The integration bucket would contain lower-level tests that still rely on dependencies (ex: testing against kopia, or graphAPI) while minimizing multiple-package coverage (ex: testing both kopia and graphapi).  These tests will be run on every CI build.

The smoke bucket contains all higher-level e2e testing.  Anything that includes multiple packages or runs at an api layer (ex: e2e cli testing).  Smoke tests do not run during the CI testing step.  Instead, they are run when building main and on deployments.

## Other Details
* Tests categorization engagement should be identified using env variables (ie: continuing the current paradigm for separating CI suites from unit suites).
* Tests should remain associated with the package that the test is starting from.  An additional testing package for "complete e2e" testing may be a useful addition as we curate that set of tests.
* Producing alerts for smoke test failure should be included in this change.

## Not Included
* Separating out code layers to facilitate and implement mocked testing.

## Effort, Value, Impact
Effort: major - large-scale refactoring of our test layout  
Value: large - faster CI means faster time to deployment, faster turnaround on identifying/resolving errors, and an easier time testing locally which allows devs and contributors to rely less on GH CI to identify failures.  
Impact: medium - two stage testing (ci/smoke) could delay identification of e2e bugs

PR #2410, opened 2023-02-06
Proof of concept wrapper for suite.Suite

ashmrtn:
## Description

Proof of concept for how to wrap suite.Suite and insert our own logic for determining if a test should be skipped or not.

Adds a new env var for running only integration tests

May require future work to add unit tests with the same (non-default) labels as their integration tests

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #2373

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector_disconnected_test.go
+++ b/src/internal/connector/graph_connector_disconnected_test.go
@@ -22,12 +22,16 @@ import (
 // Disconnected Test Section
 // ---------------------------------------------------------------
 type DisconnectedGraphConnectorSuite struct {
-	suite.Suite
+	*tester.UnitSuite
 }
 
 func TestDisconnectedGraphSuite(t *testing.T) {
+	s := &DisconnectedGraphConnectorSuite{
+		UnitSuite: tester.NewUnitSuite(t),
+	}
+
 	tester.LogTimeOfTest(t)
-	suite.Run(t, new(DisconnectedGraphConnectorSuite))
+	suite.Run(t, s)
 }
 
 func (suite *DisconnectedGraphConnectorSuite) TestBadConnection() {
--- a/src/internal/connector/graph_connector_disconnected_test.go
+++ b/src/internal/connector/graph_connector_disconnected_test.go
@@ -70,8 +70,15 @@ func (suite *DisconnectedGraphConnectorSuite) TestBadConnection() {
 	}
 
 	for _, test := range table {
-		suite.T().Run(test.name, func(t *testing.T) {
-			gc, err := NewGraphConnector(ctx, graph.HTTPClient(graph.NoTimeout()), test.acct(t), Users, fault.New(true))
+		suite.Run(test.name, func() {
+			t := suite.T()
+
+			gc, err := NewGraphConnector(
+				ctx,
+				graph.HTTPClient(graph.NoTimeout()),
+				test.acct(t),
+				Users,
+				fault.New(true))
 			assert.Nil(t, gc, test.name+" failed")
 			assert.NotNil(t, err, test.name+" failed")
 		})
--- a/src/internal/connector/graph_connector_disconnected_test.go
+++ b/src/internal/connector/graph_connector_disconnected_test.go
@@ -22,12 +22,12 @@ import (
 // Disconnected Test Section
 // ---------------------------------------------------------------
 type DisconnectedGraphConnectorSuite struct {
-	*tester.UnitSuite
+	tester.Suite
 }
 
 func TestDisconnectedGraphSuite(t *testing.T) {
 	s := &DisconnectedGraphConnectorSuite{
-		UnitSuite: tester.NewUnitSuite(t, nil),
+		Suite: tester.NewUnitSuite(t, nil),
 	}
 
 	tester.LogTimeOfTest(t)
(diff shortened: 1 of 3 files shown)


ashmrtn:
opening this to start a discussion on the topic and see if this pattern is a good idea

src/internal/tester/integration_runners.go
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 const (
-	CorsoLoadTests                                = "CORSO_LOAD_TESTS"
+	CorsoLoadTests = "CORSO_LOAD_TESTS"
+	// Run only integration tests.
+	CorsoIntegrationTests = "CORSO_INTEGRATION_TESTS"
+	// Run all tests, unit and integration.
 	CorsoCITests                                  = "CORSO_CI_TESTS"
```
ryanfkeepers, line 18: IMO: `Integration` should be 1:1 with `CI`.  We should have one flag or the other, but not both.

src/internal/tester/suite.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func NewUnitSuite(t *testing.T, excludeGroups ...string) *UnitSuite {
+	dontRunOn(
+		t,
+		append(
+			[]string{CorsoIntegrationTests},
+			excludeGroups...,
+		)...,
+	)
```
ryanfkeepers, line 16: This seems extraneous.  Unit testing should only take a  couple seconds.  Weeding it out of integration tests kinda feels like throwing a single cup of water out of a pool.

If we hit a point in CI design where we want to separately run one pass on unit tests and another pass on integration tests, that's the time to add something like this.  But at the moment I don't see much use in it.

src/internal/tester/suite.go
```diff
@@ -0,0 +1,39 @@
...
+func NewIntegrationSuite(t *testing.T, includeGroups ...string) *IntegrationSuite {
+	RunOnAny(
+		t,
+		append(
+			[]string{CorsoCITests, CorsoIntegrationTests},
+			includeGroups...,
+		)...,
+	)
+
+	return new(IntegrationSuite)
```
ryanfkeepers, line 34: Something I've been noodling about a PoC like this is what the best way would be to assert _both_ the env flags and the required env vars in a single constructor.  It'd be really nice to be able to do something like:
```go
func TestDisconnectedGraphSuite(t *testing.T) {
	suite.Run(t, &DisconnectedGraphConnectorSuite{
		Suite: tester.NewIntegrationSuite(
			t,
			[]string{CorsoCITests},
			[][]string{m365Creds, awsCreds},
		),
	})
}
```
And expect everything to be checked from there.

ashmrtn: It should be possible to add those checks to the constructor functions that I've made. I'd really like to make `UnitSuite` and `IntegrationSuite` unexported, but unfortunately testify doesn't have a predefined interface for `suite.Suite`. I could make one, but that feels a bit error prone when we start considering testify updates etc

The alternative to the constructor function I've made is to use a `SetupSuite()` call, but then the user will need to ensure they make a call to `suite.UnitSuite.SetupSuite()` if they create their own `SetupSuite()` in a test

src/internal/tester/suite.go
ashmrtn: sure, that's fine. I wasn't sure if we wanted them completely separate or not. Easy enough to make them all run and remove the extra env var I added

src/internal/tester/suite.go
ryanfkeepers: > but unfortunately testify doesn't have a predefined interface for suite.Suite

That's not entirely true.  `suite.Run()` doesn't take a suite.Suite, it takes a [TestingSuite](https://pkg.go.dev/github.com/stretchr/testify/suite#Run) interface.  This would allows us to provide just about any wrapper that provides passthroughs to an underlying suite.

ryanfkeepers: Right.  I think this is something we can/should just play around with for a while (dabble a little here and a little there) until we find a version that we like.  We have the time to be picky.

src/internal/tester/suite.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/stretchr/testify/suite"
+)
+
+type Suite interface {
+	suite.TestingSuite
+	Run(name string, subtest func()) bool
+}
+
+func NewUnitSuite(t *testing.T, envSets [][]string) *unitSuite {
```
ryanfkeepers, line 14: should unit tests ever depend on env vars?  Seems like an anti-pattern to me.

src/internal/connector/graph_connector_disconnected_test.go
```diff
@@ -21,12 +21,16 @@ import (
...
-	suite.Suite
+	tester.Suite
 }
 
 func TestDisconnectedGraphSuite(t *testing.T) {
+	s := &DisconnectedGraphConnectorSuite{
+		Suite: tester.NewUnitSuite(t, nil),
+	}
+
 	tester.LogTimeOfTest(t)
```
ryanfkeepers, line 32: Oh hey, we should call this in NewUnitSuite and NewIntegrationSuite.

ryanfkeepers (approved):
I like it.  Don't think we'll get any cleaner of a wrap than this.  Maybe add a quick set of examples in the tester package as documentation?

src/internal/tester/suite.go
ashmrtn: excellent point. It did feel a bit odd to have there. Easy enough to remove though

src/internal/connector/graph_connector_disconnected_test.go
ashmrtn: hmmm actually I'm not certain this will be useful if we call it when we make a new suite instance. It will only tell us when the entire suite started, not when each test started

ryanfkeepers: I'm not sure it's even useful as it stands 😅

Issue #2423, opened 2023-02-07
Swap calendar names for container IDs in kopia


ryanfkeepers:
## Issue
In order to protect against name collisions from shared or like-named calendars, we need to stop persisting calendars by their display names and use their underlying m365 ID instead.

## Acceptance Criteria
[] Swap out calendar display names for IDs.
[] Add a path string with display names into the item details. https://github.com/alcionai/corso/issues/2412
[] When running selectors.Reduce, replace the path with the display name path string (when available).
[] Update the details display output to reference display names instead of IDs.
[] Check selector usage in GC to ensure path comparisons follow the new format.

ryanfkeepers:
Remaining changes to close: manual release testing and merger to main branch.

PR #2427, opened 2023-02-07
add locationRef to details

ryanfkeepers:
## Description

Adds a new reference to the details ent: location-
ref.  The location holds the human-readable
version of the item's location in whatever m365
service sourced the item.  Hookup is incomplete,
following PRs will fill out functionality.

Also adds a LocationPather interface to data_
collections to pass this data back and forth
between producers and consumers.

Should be safe to merge into main.

## Type of change

- [x] :sunflower: Feature
- [x] :bug: Bugfix

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/container_resolver_test.go
+++ b/src/internal/connector/exchange/container_resolver_test.go
@@ -595,7 +595,7 @@ func (suite *FolderCacheIntegrationSuite) TestCreateContainerDestination() {
 
 	for _, test := range tests {
 		suite.T().Run(test.name, func(t *testing.T) {
-			folderID, err := CreateContainerDestinaion(
+			folderID, err := CreateContainerDestination(
 				ctx,
 				m365,
 				test.pathFunc1(t),
@@ -608,7 +608,7 @@ func (suite *FolderCacheIntegrationSuite) TestCreateContainerDestination() {
 			_, err = resolver.IDToPath(ctx, folderID)
 			assert.NoError(t, err)
 
-			secondID, err := CreateContainerDestinaion(
+			secondID, err := CreateContainerDestination(
 				ctx,
 				m365,
 				test.pathFunc2(t),
--- a/src/internal/connector/exchange/container_resolver_test.go
+++ b/src/internal/connector/exchange/container_resolver_test.go
@@ -26,16 +26,19 @@ type mockContainer struct {
 	displayName *string
 	parentID    *string
 	p           *path.Builder
+	l           *path.Builder
 }
 
 //nolint:revive
 func (m mockContainer) GetId() *string { return m.id }
 
 //nolint:revive
-func (m mockContainer) GetParentFolderId() *string { return m.parentID }
-func (m mockContainer) GetDisplayName() *string    { return m.displayName }
-func (m mockContainer) Path() *path.Builder        { return m.p }
-func (m mockContainer) SetPath(p *path.Builder)    {}
+func (m mockContainer) GetParentFolderId() *string  { return m.parentID }
+func (m mockContainer) GetDisplayName() *string     { return m.displayName }
+func (m mockContainer) Location() *path.Builder     { return m.l }
+func (m mockContainer) SetLocation(p *path.Builder) {}
+func (m mockContainer) Path() *path.Builder         { return m.p }
+func (m mockContainer) SetPath(p *path.Builder)     {}
 
 func strPtr(s string) *string {
 	return &s
@@ -168,7 +171,7 @@ func (suite *FolderCacheUnitSuite) TestAddFolder() {
 					parentID:    nil,
 				},
 				nil,
-			),
+				nil),
 			check: assert.Error,
 		},
 		{
@@ -180,7 +183,7 @@ func (suite *FolderCacheUnitSuite) TestAddFolder() {
 					parentID:    nil,
 				},
 				path.Builder{}.Append("foo"),
-			),
+				path.Builder{}.Append("loc")),
 			check: assert.NoError,
 		},
 		{
@@ -192,7 +195,7 @@ func (suite *FolderCacheUnitSuite) TestAddFolder() {
 					parentID:    &testParentID,
 				},
 				path.Builder{}.Append("foo"),
-			),
+				path.Builder{}.Append("loc")),
 			check: assert.Error,
 		},
 		{
@@ -204,7 +207,7 @@ func (suite *FolderCacheUnitSuite) TestAddFolder() {
 					parentID:    &testParentID,
 				},
 				path.Builder{}.Append("foo"),
-			),
+				path.Builder{}.Append("loc")),
 			check: assert.Error,
 		},
 		{
@@ -216,7 +219,7 @@ func (suite *FolderCacheUnitSuite) TestAddFolder() {
 					parentID:    &testParentID,
 				},
 				nil,
-			),
+				nil),
 			check: assert.NoError,
 		},
 	}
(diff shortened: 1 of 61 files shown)


src/internal/connector/exchange/container_resolver_test.go
```diff
@@ -595,7 +595,7 @@ func (suite *FolderCacheIntegrationSuite) TestCreateContainerDestination() {
 
 	for _, test := range tests {
 		suite.T().Run(test.name, func(t *testing.T) {
-			folderID, err := CreateContainerDestinaion(
+			folderID, err := CreateContainerDestination(
```
ashmrtn, line 598: the phantom T strikes again!

src/internal/data/data_collection.go
```diff
@@ -70,6 +70,12 @@ type Stream interface {
 	Deleted() bool
 }
 
+// Locationer provides a LocationPath describing the path with Display Names
```
ashmrtn, line 73:
```suggestion
// LocationPather provides a LocationPath describing the path with Display Names
```

src/internal/data/data_collection.go
```diff
@@ -70,6 +70,12 @@ type Stream interface {
...
+// instead of canonical IDs
+type LocationPather interface {
+	LocationPath() path.Path
```
ashmrtn, line 76: This is something we haven't fully discussed, but are we comfortable with always returning a Path that has all the leading elements like tenant, service, user, and category?

I guess if we intend to use this for selectors it's kind of required right? After all, Path is meant to help with selector comparisons by providing index-independent access to fields like `Category()` etc

src/internal/kopia/upload.go
```diff
@@ -186,20 +187,29 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 
 	parent := d.repoPath.ToBuilder().Dir()
 
+	var locationFolders string
+	if d.locationPath != nil {
+		locationFolders = d.locationPath.Folder()
```
ashmrtn, line 192: `Folder()` does not provide something that can be turned into a `path.Path` later on. It's really only the part after the prefix and it will continue to include things like the drive ID and `root:` for OneDrive paths if we don't make further changes

Not everything needs to be solved in this PR, but trying to give a sense in how it will play out in the long run

src/pkg/backup/details/details.go
```diff
@@ -150,11 +162,30 @@ func FolderEntriesForPath(parent *path.Builder) []folderEntry {
...
+		}
 	}
 
 	return folders
 }
 
+// assumes the pb contains a path like:
+// <tenant>/<service>/<owner>/<category>/<logical_containers>...
+// and returns a string with only <logical_containers>/...
+func locationRefOf(pb *path.Builder) *path.Builder {
```
ashmrtn, line 177: we should probably plan to move this to some service/category specific code down the road. Depending on how we want the location refs to look it could vary between them

We should probably also refactor the code that adds folders to be more efficient, but that's also another PR

src/internal/data/data_collection.go
ashmrtn: I see we're actually storing a subset of this information. Future PRs can revisit if we need to return an actual Path from this or if something less strict like a string would be alright

src/pkg/backup/details/details.go
```diff
@@ -233,9 +268,21 @@ type DetailsEntry struct {
...
 	ShortRef  string `json:"shortRef"`
 	ParentRef string `json:"parentRef,omitempty"`
+
+	// LocationRef contains the logical path structure by its human-readable
+	// display names.  IE:  If an item is located at "/Inbox/Important", we
+	// hold that string in the LocationRef, while the actual IDs of each
+	// container are used for the RepoRef.
+	// LocationRef only holds the container values, and does not include
+	// the metadata prefixes (tenant, service, owner, etc) found in the
+	// repoRef.
```
ashmrtn, line 278: If we intend to store only a subset of the data in the reporef, how do we intend to deal with selector comparisons on that metadata for `Reduce()`? Part of this could be me being unfamiliar with the selectors package, but if we currently have selectors doing something like `entry.RepoRef.Category() == sel.category` then won't we end up splitting things in a bit of an awkward fashion? Seems like some comparisons would end up against the LocationRef while others would use the RepoRef

We don't need a full implementation right now, and I realize that may be difficult given the main thing to address at the moment is Calendars, but it's good to have an idea of where we plan to go

src/pkg/backup/details/details_test.go
```diff
@@ -207,7 +234,7 @@ func (suite *DetailsUnitSuite) TestDetailsModel_Path() {
 					Entries: test.ents,
 				},
 			}
-			assert.Equal(t, test.expectRefs, d.Paths())
+			assert.Equal(t, test.expectRepoRefs, d.Paths())
```
ashmrtn, line 237: nit: we probably want `ElementsMatch()`?
```suggestion
			assert.ElementsMatch(t, test.expectRepoRefs, d.Paths())
```

src/internal/connector/exchange/container_resolver_test.go
ryanfkeepers: Hehe, I think that one has been hiding around for half a year, actually.

src/internal/data/data_collection.go
ryanfkeepers: whoops, thanks

src/internal/data/data_collection.go
ryanfkeepers: I think it's most holistic for us to maintain the full path here.  Passing around paths that arbitrarily do or do not comply with the standard format is bound produce later gotchas and headaches, not to mention increasing cognitive load to track which is which.

src/pkg/backup/details/details_test.go
ryanfkeepers: always gets me.

src/pkg/backup/details/details.go
ryanfkeepers: If/when you look at the next PR, I think this will get cleared up.

src/internal/kopia/upload.go
ryanfkeepers: Right, I think that's correct for what I'm aiming to implement here.  It should, if I'm not mistaken, match the value we get out of `locationRefOf(path)` in `details.go`.

ashmrtn: Just putting here for completeness, one of my comments in another PR mentioned that this actually needs to be the escaped version of the string (`Folder()` is not escaped, I don't remember why at this point :( ). Original [comment](https://github.com/alcionai/corso/pull/2429#discussion_r1099229056)

src/internal/operations/backup.go
```diff
@@ -573,10 +573,11 @@ func mergeDetails(
 				newPath.String(),
 				newPath.ShortRef(),
 				newPath.ToBuilder().Dir().ShortRef(),
+				"", // TODO Location Ref,
```
ashmrtn, line 576: we need to fix this before #2429 merges. Otherwise we'll break incremental backups for calendars

src/internal/operations/backup.go
```diff
@@ -573,10 +573,11 @@ func mergeDetails(
...
 				itemUpdated,
 				item)
 
-			folders := details.FolderEntriesForPath(newPath.ToBuilder().Dir())
+			folders := details.FolderEntriesForPath(newPath.ToBuilder().Dir(), nil)
```
ashmrtn, line 580: also needs fixed prior to #2429 merging

ashmrtn (approved):

PR #2429, opened 2023-02-07
use locationRef in reduce when available

ryanfkeepers:
## Description

When reducing selectors, if the details entry has
a LocationRef available, replace the repoRef
folders section with the locationRef values.
This allows users to continue using folder display
names as scope parameters while using folder
IDs in the repoRef path.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/selectors/exchange_test.go
+++ b/src/pkg/selectors/exchange_test.go
@@ -833,10 +833,6 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 		return deets
 	}
 
-	arr := func(s ...string) []string {
-		return s
-	}
-
 	table := []struct {
 		name         string
 		deets        *details.Details
@@ -861,7 +857,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.AllData())
 				return er
 			},
-			arr(contact),
+			[]string{contact},
 		},
 		{
 			"event only",
@@ -871,7 +867,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.AllData())
 				return er
 			},
-			arr(event),
+			[]string{event},
 		},
 		{
 			"mail only",
@@ -881,7 +877,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.AllData())
 				return er
 			},
-			arr(mail),
+			[]string{mail},
 		},
 		{
 			"all",
@@ -891,7 +887,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.AllData())
 				return er
 			},
-			arr(contact, event, mail),
+			[]string{contact, event, mail},
 		},
 		{
 			"only match contact",
@@ -901,7 +897,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.Contacts([]string{"cfld"}, []string{"cid"}))
 				return er
 			},
-			arr(contact),
+			[]string{contact},
 		},
 		{
 			"only match contactInSubFolder",
@@ -911,7 +907,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.ContactFolders([]string{"cfld1/cfld2"}))
 				return er
 			},
-			arr(contactInSubFolder),
+			[]string{contactInSubFolder},
 		},
 		{
 			"only match contactInSubFolder by prefix",
@@ -921,7 +917,7 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 				er.Include(er.ContactFolders([]string{"cfld1/cfld2"}, PrefixMatch()))
 				return er
 			},
-			arr(contactInSubFolder),
+			[]string{contactInSubFolder},
 		},
 		{
 			"only match contactInSubFolder by leaf folder",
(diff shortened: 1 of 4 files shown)


src/pkg/selectors/exchange_test.go
```diff
@@ -1041,6 +1041,233 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
...
+		suite.T().Run(test.name, func(t *testing.T) {
+			ctx, flush := tester.NewContext()
+			defer flush()
+
+			errs := mock.NewAdder()
+
+			sel := test.makeSelector()
+			results := sel.Reduce(ctx, test.deets, errs)
+			paths := results.Paths()
+			assert.Equal(t, test.expect, paths)
```
ashmrtn, line 1265:
```suggestion
			assert.ElementsMatch(t, test.expect, paths)
```

src/pkg/selectors/exchange_test.go
```diff
@@ -1041,6 +1041,233 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
 	}
 }
 
+func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce_locationRef() {
```
ashmrtn, line 1044: I understand that this test showcases how things will work when we can get everything switched over to storing the LocationRef. However, we're not currently in that situation so I think it would be best to have this test ensure that email and contacts selectors continue to properly compare the path elements in the RepoRef while the events selector compares the path elements in the LocationRef

Since I think we currently lack the sort of e2e tests for the above I think this is the best we can do to make sure things work as expected in the broader system

src/pkg/selectors/exchange_test.go
```diff
@@ -1041,6 +1041,233 @@ func (suite *ExchangeSelectorSuite) TestExchangeRestore_Reduce() {
...
+						ItemType: itype,
+					},
+				},
+			})
+		}
+
+		return deets
+	}
+
+	arr := func(s ...string) []string {
```
ashmrtn, line 1093: 🏴‍☠️

src/pkg/selectors/scopes.go
```diff
@@ -317,6 +317,27 @@ func reduce[T scopeT, C categoryT](
 			continue
 		}
 
+		// if the details entry has a locationRef specified, use those folders in place
+		// of the repoRef folders, so that scopes can match against the display names
+		// instead of container IDs.
+		if len(ent.LocationRef) > 0 {
+			pb, err := path.Builder{}.
```
ashmrtn, line 324: we may need to add a helper function to the path package for this. I think you want `path.UnescapeAndAppend(path.Split(path.TrimTrailingSlash(ent.LocationRef))...)` and we need to make sure that the folders were output with escaping, which I [don't think is happening right now](https://github.com/alcionai/corso/blob/27c1e5c51140b238c6c3cbc3df79b290e37704fe/src/pkg/path/resource_path.go#L210) (it should be making a Builder and calling `String()` for escaping) :/

May be best to temporarily make a new function for `EscapedFolder()` until we can reconcile the current calls to `Folder()` that assumes no escaping

src/pkg/selectors/exchange_test.go
ryanfkeepers: aw c'mon.  I'm saving us, like, five characters a pop here.  Isn't that worth it?

(seriously though, this is old code, and I also think this is hilarious;  what was I thinking?)

ashmrtn: oh, I was going for "pirate" and that's what came up lol. Stick with `arr` I think it's great too

src/pkg/selectors/exchange_test.go
ryanfkeepers: There's a catch here: this code doesn't really recognize expectations like that from the broader system.  I could comment out the mail and event tests here, which is probably sufficient for the time being.  But the reality of the selectors code is that it doesn't (and shouldn't) evaluate whether or not the thing is logically supported as a use case.  It only cares that it can run the provided data correctly.  IE: this test isn't meant to show that we _are_ handling locations for mail and contacts, just that we _can_.

I could add a comment explaining this, if that helps?

ashmrtn: mmmm logically I think that explanation makes sense. I think it still leaves a hole in our testing where we can't actually say that these changes don't break the current functionality for email and contacts because we have no tests ensuring it works as expected

src/pkg/selectors/scopes.go
ryanfkeepers: Is the necessity of escaping just for ensuring parity between builder.String() and path.EscapedFolder()?

ashmrtn: we need escaping so `Split()` works properly. Otherwise we can't differentiate between the joined versions of `["this", "is/", "a", "path"]` and `["this", "is", "", "a", "path"]`

I realize there's a debate we could have about the handling of empty elements, but right now the path package is setup such that `Split` [takes an escaped path](https://github.com/alcionai/corso/blob/27c1e5c51140b238c6c3cbc3df79b290e37704fe/src/pkg/path/path.go#L582) and constructors for `Builder` [drop empty elements](https://github.com/alcionai/corso/blob/27c1e5c51140b238c6c3cbc3df79b290e37704fe/src/pkg/path/path.go#L158)

src/pkg/selectors/exchange_test.go
ryanfkeepers: I think we'll have to go elsewhere for that.  Probably at the collections, to enforce that the location isn't populated for email and contacts.

Also, it should just work, even if things slip through?  Because that value _should_ be a copy of our the current folders for those types.

ashmrtn: mmm I suppose that's true. Just not having automated tests for system-wide invariants makes me a bit nervous because it's easy to forget to check them 😅

ashmrtn (commented):
this is ready for approval but holding off until another part of the code is updated. This should only be merged after we fixup the LocationRef when merging backup details. Otherwise we may end up not matching events if they were sourced from a base snapshot. [Comment](https://github.com/alcionai/corso/pull/2427/files#r1099269567) for the code that needs updated

ashmrtn (approved):

PR #2430, opened 2023-02-07
populate the location path values

ryanfkeepers:
## Description

Adds location path values to all exchange types.
Only propagates these values if the collection is
an event category.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/events.go
+++ b/src/internal/connector/exchange/api/events.go
@@ -209,10 +209,11 @@ func (c Events) EnumerateContainers(
 				continue
 			}
 
-			temp := graph.NewCacheFolder(cd, path.Builder{}.Append(*cd.GetDisplayName()))
-
-			err = fn(temp)
-			if err != nil {
+			temp := graph.NewCacheFolder(
+				cd,
+				path.Builder{}.Append(*cd.GetDisplayName()),
+				path.Builder{}.Append(*cd.GetDisplayName()))
+			if err := fn(temp); err != nil {
 				errs = multierror.Append(err, errs)
 				continue
 			}
--- a/src/internal/connector/exchange/api/mail.go
+++ b/src/internal/connector/exchange/api/mail.go
@@ -198,8 +198,7 @@ func (c Mail) EnumerateContainers(
 		}
 
 		for _, v := range resp.GetValue() {
-			temp := graph.NewCacheFolder(v, nil)
-
+			temp := graph.NewCacheFolder(v, nil, nil)
 			if err := fn(temp); err != nil {
 				errs = multierror.Append(errs, errors.Wrap(err, "iterating mail folders delta"))
 				continue
--- a/src/internal/connector/exchange/contact_folder_cache.go
+++ b/src/internal/connector/exchange/contact_folder_cache.go
@@ -29,7 +29,8 @@ func (cfc *contactFolderCache) populateContactRoot(
 		return support.ConnectorStackErrorTraceWrap(err, "fetching root folder")
 	}
 
-	temp := graph.NewCacheFolder(f,
+	temp := graph.NewCacheFolder(
+		f,
 		path.Builder{}.Append(baseContainerPath...), // storage path
 		path.Builder{}.Append(baseContainerPath...)) // display location
 	if err := cfc.addFolder(temp); err != nil {
--- a/src/internal/connector/exchange/contact_folder_cache.go
+++ b/src/internal/connector/exchange/contact_folder_cache.go
@@ -29,8 +29,9 @@ func (cfc *contactFolderCache) populateContactRoot(
 		return support.ConnectorStackErrorTraceWrap(err, "fetching root folder")
 	}
 
-	temp := graph.NewCacheFolder(f, path.Builder{}.Append(baseContainerPath...))
-
+	temp := graph.NewCacheFolder(f,
+		path.Builder{}.Append(baseContainerPath...), // storage path
+		path.Builder{}.Append(baseContainerPath...)) // display location
 	if err := cfc.addFolder(temp); err != nil {
 		return errors.Wrap(err, "adding resolver dir")
 	}
--- a/src/internal/connector/exchange/event_calendar_cache.go
+++ b/src/internal/connector/exchange/event_calendar_cache.go
@@ -44,7 +44,9 @@ func (ecc *eventCalendarCache) populateEventRoot(ctx context.Context) error {
 		return errors.Wrap(err, "fetching calendar "+support.ConnectorStackErrorTrace(err))
 	}
 
-	temp := graph.NewCacheFolder(f, path.Builder{}.Append(container))
+	temp := graph.NewCacheFolder(f,
+		path.Builder{}.Append(container), // storage path
+		path.Builder{}.Append(container)) // display location
 	if err := ecc.addFolder(temp); err != nil {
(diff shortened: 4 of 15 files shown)


src/internal/connector/exchange/api/events.go
```diff
@@ -209,10 +209,11 @@ func (c Events) EnumerateContainers(
 				continue
 			}
 
-			temp := graph.NewCacheFolder(cd, path.Builder{}.Append(*cd.GetDisplayName()))
-
-			err = fn(temp)
-			if err != nil {
+			temp := graph.NewCacheFolder(
+				cd,
+				path.Builder{}.Append(*cd.GetDisplayName()),
```
ashmrtn, line 214: I think this should be the ID?
```suggestion
				path.Builder{}.Append(*cd.GetId()),
```

src/internal/connector/exchange/event_calendar_cache.go
```diff
@@ -44,7 +44,9 @@ func (ecc *eventCalendarCache) populateEventRoot(ctx context.Context) error {
 		return errors.Wrap(err, "fetching calendar "+support.ConnectorStackErrorTrace(err))
 	}
 
-	temp := graph.NewCacheFolder(f, path.Builder{}.Append(container))
+	temp := graph.NewCacheFolder(f,
+		path.Builder{}.Append(container), // storage path
```
ashmrtn, line 48: this should be the ID
```suggestion
		path.Builder{}.Append(*f.GetId()), // storage path
```

src/internal/connector/exchange/api/events.go
ryanfkeepers: In a future PR, yes.

src/internal/connector/exchange/contact_folder_cache.go
```diff
@@ -29,8 +29,9 @@ func (cfc *contactFolderCache) populateContactRoot(
 		return support.ConnectorStackErrorTraceWrap(err, "fetching root folder")
 	}
 
-	temp := graph.NewCacheFolder(f, path.Builder{}.Append(baseContainerPath...))
-
+	temp := graph.NewCacheFolder(f,
```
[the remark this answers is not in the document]
ryanfkeepers: damn, no, that's on me.  Good catch.

src/internal/connector/exchange/event_calendar_cache.go
ryanfkeepers: next PR!

src/internal/connector/exchange/event_calendar_cache.go
```diff
@@ -91,7 +93,9 @@ func (ecc *eventCalendarCache) AddToCache(ctx context.Context, f graph.Container
 		return errors.Wrap(err, "validating container")
 	}
 
-	temp := graph.NewCacheFolder(f, path.Builder{}.Append(calendarOthersFolder, *f.GetDisplayName()))
+	temp := graph.NewCacheFolder(f,
+		path.Builder{}.Append(calendarOthersFolder, *f.GetDisplayName()), // storage path
```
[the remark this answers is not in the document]
ryanfkeepers: soon

src/internal/connector/exchange/service_functions.go
```diff
@@ -86,44 +86,62 @@ func PopulateExchangeContainerResolver(
...
+		qp.ResourceOwner,
+		category,
+		false)
 	// Containers without a path (e.g. Root mail folder) always err here.
 	if err != nil {
-		return nil, false
+		return nil, nil, false
 	}
 
 	directory = pb.String()
```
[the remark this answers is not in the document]
ryanfkeepers: Oh, yes, thank you.

src/internal/connector/exchange/service_functions.go
```diff
@@ -86,44 +86,62 @@ func PopulateExchangeContainerResolver(
 }
 
 // Returns true if the container passes the scope comparison and should be included.
-// Also returns the path representing the directory.
+// Returns:
+// - the path representing the directory as it should be stored in the repository.
+// - the human-readable path using display names.
+// - true if the path passes the scope comparison.
 func includeContainer(
```
ashmrtn, line 93: actually, I think this function needs a good once-over. It's confusing because we have both `loc` and `locPath` but `locPath` is based off the path pulled from the container resolver. `loc` also does not seem to be returned at all so even if the container resolver has it we're not making use of it

ashmrtn (changes requested):
needs revisions for `includeContainer` but otherwise looks pretty good

src/internal/connector/exchange/service_functions.go
```diff
@@ -118,14 +121,18 @@ func includeContainer(
...
-		qp.Credentials.AzureTenantID,
-		qp.ResourceOwner,
-		category,
-		false)
-	// Containers without a path (e.g. Root mail folder) always err here.
-	if err != nil {
-		return nil, nil, false
+	var locPath path.Path
+
+	if loc != nil {
```
ashmrtn, line 126: we're still
1. comparing only against locPath
2. returning locPath

If we're only using locPath, then what's the point of having loc at all?

src/internal/connector/exchange/service_functions.go
ryanfkeepers: Not entirely sure what you're looking for with a refactor here.  Things are functional (outside of having had a bug on 121, and setting the locPath to the pb path).  And we need all the component parts.

src/internal/connector/exchange/api/mail.go
```diff
@@ -198,8 +198,7 @@ func (c Mail) EnumerateContainers(
 		}
 
 		for _, v := range resp.GetValue() {
-			temp := graph.NewCacheFolder(v, nil)
-
+			temp := graph.NewCacheFolder(v, nil, nil)
```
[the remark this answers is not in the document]
ryanfkeepers: I think we're now covered here, across various PRs.

src/internal/connector/exchange/service_functions.go
ryanfkeepers: mm, I think I'd rather keep locPath nil throughout the whole passdown at this time.  Theoretically, in the future, if/when we expand this design to all kopia storage, loc will _never_ be nil.  In the meantime, if its nil, then it should be considered unsupported for that use case.  This will also help avoid "I didn't realize I wasn't populating it" gotchas.

That said, for purposes of directory comparison, just in this func, we already are falling back to the pb directory if loc is nil.

ashmrtn: > for purposes of directory comparison, just in this func, we already are falling back to the pb directory if loc is nil.

Are we though? all comparisons are done against `directory` which is set to `locPath.Folder()` on 138. `locPath` is only set if `loc != nil`

ryanfkeepers: Ugh.  Apparently I have more than one branch touching this code.  It's fixed _elsewhere_ ☹️ .  I'll make a change here, too.

ashmrtn: ah that would explain the confusion. Honestly you're doing great managing all the branches that are out right now

ashmrtn (approved):

PR #2432, opened 2023-02-07
allow escaping the path folder

ryanfkeepers:
## Description

to be added to 2423-selectors before it merges.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -189,7 +189,7 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 
 	var locationFolders string
 	if d.locationPath != nil {
-		locationFolders = d.locationPath.Folder(false)
+		locationFolders = d.locationPath.Folder(true)
 	}
 
 	cp.deets.Add(
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -189,7 +189,7 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 
 	var locationFolders string
 	if d.locationPath != nil {
-		locationFolders = d.locationPath.Folder()
+		locationFolders = d.locationPath.Folder(false)
 	}
 
 	cp.deets.Add(
--- a/src/pkg/path/resource_path.go
+++ b/src/pkg/path/resource_path.go
@@ -213,12 +213,8 @@ func (rp dataLayerResourcePath) Folder(escape bool) string {
 		return join(fs)
 	}
 
-	for i := range fs {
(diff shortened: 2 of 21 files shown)


src/internal/kopia/upload.go
```diff
@@ -189,7 +189,7 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 
 	var locationFolders string
 	if d.locationPath != nil {
-		locationFolders = d.locationPath.Folder()
+		locationFolders = d.locationPath.Folder(false)
```
ashmrtn, line 192: this should be true 😝 
```suggestion
		locationFolders = d.locationPath.Folder(true)
```

src/pkg/path/resource_path.go
```diff
@@ -201,13 +201,24 @@ func (rp dataLayerResourcePath) lastFolderIdx() int {
...
+	if !escape {
+		return join(fs)
+	}
+
+	for i := range fs {
+		f := fs[i]
+		fs[i] = escapeElement(f)
+	}
+
+	return join(fs)
```
ashmrtn, line 221: since we know we need it escaped you could just make a builder and call `String()` if you want
```suggestion
	return Builder{}.Append(folders...).String()
```

src/pkg/selectors/exchange.go
```diff
@@ -598,7 +598,7 @@ func (ec exchangeCategory) pathValues(p path.Path) map[categorizer]string {
 	}
 
 	return map[categorizer]string{
-		folderCat: p.Folder(),
+		folderCat: p.Folder(false),
```
ashmrtn, line 601: what is this used for? I'm unsure from just this snippet if it should be escaped or not

src/internal/kopia/upload.go
ryanfkeepers: false

src/pkg/path/resource_path.go
ryanfkeepers: Huh, I guess I could.

src/pkg/selectors/exchange.go
ryanfkeepers: This is the folder input value that the scopes compare against.  Off the top of my head it should be unescaped.  But it's late, and I might be forgetting things.

ashmrtn: mmmm I think I get what you're saying and I think you're right that it should be unescaped

ashmrtn (approved):

PR #2433, opened 2023-02-08
use ID for cal path instead of display name

ryanfkeepers:

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/event_calendar_cache.go
```diff
@@ -94,7 +94,7 @@ func (ecc *eventCalendarCache) AddToCache(ctx context.Context, f graph.Container
 	}
 
 	temp := graph.NewCacheFolder(f,
-		path.Builder{}.Append(calendarOthersFolder, *f.GetDisplayName()), // storage path
+		path.Builder{}.Append(calendarOthersFolder, *f.GetId()),          // storage path
```
ashmrtn, line 97: this feels so anti-climatic after all the other stuff 😛 

Anyway, there's a few places this needs updated actually. See comments from the previous PR:
* https://github.com/alcionai/corso/pull/2430#discussion_r1099501415
* https://github.com/alcionai/corso/pull/2430#discussion_r1099502588

ryanfkeepers: It's the best way to have it happen!

ashmrtn (approved):

Issue #1535, opened 2022-11-16
Use item IDs for OneDrive file names in kopia


ashmrtn:
OneDrive files currently use the file name of the file in OneDrive as the file name in kopia. However, this is not compatible with using delta token information to update files during incremental backups. Delta tokens only give the ID of the file and if the file was deleted there's no way to retrieve the name of the file. Unless some other way of resolving IDs -> file names is created, it may be best to store files by ID in kopia

ashmrtn:
#2496 needs updated to use (reporef, filename) when generating shortref in backup details

PR #2434, opened 2023-02-08
Add Fetch() to RestoreCollection

ashmrtn:
## Description

Add a function to fetch a file from the collection synchronously. This will help avoid data dependencies on the restore path created by splitting item information across multiple kopia files

Fetch function is currently unoptimized, though deeper analysis of memory footprint should be done before changing

Viewing by commit will help reduce chaff from updating tests to comply with the new interface

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1535

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/data_collection.go
+++ b/src/internal/kopia/data_collection.go
@@ -1,7 +1,11 @@
 package kopia
 
 import (
+	"context"
 	"io"
+
+	"github.com/alcionai/clues"
+	"github.com/kopia/kopia/fs"
 
 	"github.com/alcionai/corso/src/internal/data"
 	"github.com/alcionai/corso/src/pkg/path"
@@ -13,8 +17,10 @@ var (
 )
 
 type kopiaDataCollection struct {
-	path    path.Path
-	streams []data.Stream
+	path         path.Path
+	streams      []data.Stream
+	snapshotRoot fs.Entry
+	counter      ByteCounter
 }
 
 func (kdc *kopiaDataCollection) Items() <-chan data.Stream {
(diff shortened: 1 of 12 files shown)


src/internal/kopia/data_collection.go
```diff
@@ -35,6 +41,25 @@ func (kdc kopiaDataCollection) FullPath() path.Path {
...
+	if kdc.snapshotRoot == nil {
+		return nil, clues.New("no snapshot root")
+	}
+
+	p, err := kdc.FullPath().Append(name, true)
+	if err != nil {
+		return nil, clues.Wrap(err, "creating item path")
+	}
+
+	// TODO(ashmrtn): We could possibly hold a reference to the folder this
```
meain, line 57: Wondering what the use would be to storing the reference to the folder.

meain (approved):

src/internal/kopia/data_collection.go
ashmrtn: it really depends on how we want to play things. Right now we store only the root and require kopia to traverse all directories on the path of an item if we call `Fetch`. If we stored the directory reference for the collection then kopia would only need to access the data in the held reference instead of traversing a tree

The tradeoff is that we'd end up holding references to as many directories as we're restoring instead of holding a single reference to the root and temporarily holding references to directories as we traverse them

PR #2436, opened 2023-02-08
Add `make fmt` for src

meain:
## Description

Adds make command for code formatting.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -15,6 +15,10 @@ lint: check-lint-version
 	golangci-lint run
 	staticcheck ./...
 
+fmt:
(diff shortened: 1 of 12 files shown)


ryanfkeepers (approved):

meain:
> I'm curious as to why these things (especially import ordering) aren't causing lint failures in CI

Yeah, I am a bit concerned about that too. Except for the betasdk(which we skip), I can' think of a reason why it would not have already failed in CI.

PR #2442, opened 2023-02-08
set fullPath using IDs in cache resolver

ryanfkeepers:

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -52,7 +52,7 @@ func (cr *containerResolver) IDToPath(
 	ctx context.Context,
 	folderID string,
 	useIDInPath bool,
-) (*path.Builder, error) {
+) (*path.Builder, *path.Builder, error) {
 	return cr.idToPath(ctx, folderID, 0, useIDInPath)
 }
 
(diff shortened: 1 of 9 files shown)


src/internal/connector/exchange/container_resolver.go
```diff
@@ -74,13 +76,21 @@ func (cr *containerResolver) idToPath(
...
 	}
 
-	fullPath := parentPath.Append(*c.GetDisplayName())
+	toAppend := *c.GetDisplayName()
+	if useIDInPath {
+		toAppend = *c.GetId()
+	}
+
+	fullPath := parentPath.Append(toAppend)
+	locPath := parentPath.Append(*c.GetDisplayName())
```
ashmrtn, line 90: we're going to need to return the locPath as well and then use that in place of `parentPath` here. Otherwise we'll end up with locPaths that have mixed IDs/display names

ryanfkeepers: ooh, yup, good catch

src/internal/connector/exchange/container_resolver.go
```diff
@@ -51,38 +51,48 @@ type containerResolver struct {
...
 	}
 
-	fullPath := parentPath.Append(*c.GetDisplayName())
+	toAppend := *c.GetDisplayName()
+	if useIDInPath {
+		toAppend = *c.GetId()
+	}
+
+	fullPath := parentPath.Append(toAppend)
+	locPath := parentLoc.Append(*c.GetDisplayName())
```
ashmrtn, line 90: do we need a nil check around this call? I honestly don't remember if this is guaranteed to be non-nil at this point

src/internal/connector/exchange/mail_folder_cache_test.go
```diff
@@ -18,10 +18,11 @@ const (
 	// top-level folders right now.
 	//nolint:lll
 	testFolderID = "AAMkAGZmNjNlYjI3LWJlZWYtNGI4Mi04YjMyLTIxYThkNGQ4NmY1MwAuAAAAAADCNgjhM9QmQYWNcI7hCpPrAQDSEBNbUIB9RL6ePDeF3FIYAABl7AqpAAA="
-
 	//nolint:lll
 	topFolderID = "AAMkAGZmNjNlYjI3LWJlZWYtNGI4Mi04YjMyLTIxYThkNGQ4NmY1MwAuAAAAAADCNgjhM9QmQYWNcI7hCpPrAQDSEBNbUIB9RL6ePDeF3FIYAAAAAAEIAAA="
+	//nolint:lll
 	// Full folder path for the folder above.
+	expectedIDPath     = "AAMkAGZmNjNlYjI3LWJlZWYtNGI4Mi04YjMyLTIxYThkNGQ4NmY1MwAuAAAAAADCNgjhM9QmQYWNcI7hCpPrAQDSEBNbUIB9RL6ePDeF3FIYAAAAAAEIAAA=/subFolder/subsubfolder"
```
ashmrtn, line 25: this appears to be unused
```suggestion
```

ashmrtn (approved):
seems like maybe we should have a test for setting the ID and not having the display names match the IDs

src/internal/connector/exchange/container_resolver.go
ryanfkeepers: oof, yeah.

PR #2457, opened 2023-02-09
Use ids for CollectionsMap for OneDrive

meain:
## Description

This will help us with delta backups where we might run into a single
folder changing multiple times.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/pull/2407

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -376,8 +376,26 @@ func (c *Collections) UpdateCollections(
 	excluded map[string]struct{},
 	invalidPrevDelta bool,
 ) error {
+	// This can only be made local to a page as the folder rename
+	// might happen between pages. Every page in the graph response
+	// will have all the parents for that item.
+	visitedPaths := map[string]string{}
+
 	for _, item := range items {
 		if item.GetRoot() != nil {
+			collectionPath, err := GetCanonicalPath(
+				// TODO(meain): Is there a better way to get root path?
+				fmt.Sprintf("/drives/%s/root:", driveID),
+				c.tenant,
+				c.resourceOwner,
+				c.source,
+			)
+			if err != nil {
+				return err
+			}
+
+			visitedPaths[collectionPath.String()] = *item.GetId()
+
 			// Skip the root item
 			continue
 		}
@@ -425,6 +443,8 @@ func (c *Collections) UpdateCollections(
 				return err
 			}
 
+			visitedPaths[folderPath.String()] = *item.GetId()
+
 			// Moved folders don't cause delta results for any subfolders nested in
 			// them. We need to go through and update paths to handle that. We only
 			// update newPaths so we don't accidentally clobber previous deletes.
@@ -454,8 +474,12 @@ func (c *Collections) UpdateCollections(
 			// TODO(ashmrtn): Figure what when an item was moved (maybe) and add it to
 			// the exclude list.
 
-			col, found := c.CollectionMap[collectionPath.String()]
+			collectionID, ok := visitedPaths[collectionPath.String()]
+			if !ok {
+				return fmt.Errorf("unable to find collection id for %s", collectionPath.String())
+			}
 
+			col, found := c.CollectionMap[collectionID]
 			if !found {
 				// TODO(ashmrtn): Compare old and new path and set collection state
 				// accordingly.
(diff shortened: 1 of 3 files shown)


src/internal/connector/onedrive/collections.go
```diff
@@ -376,8 +376,26 @@ func (c *Collections) UpdateCollections(
...
+	// This can only be made local to a page as the folder rename
+	// might happen between pages. Every page in the graph response
+	// will have all the parents for that item.
+	visitedPaths := map[string]string{}
+
 	for _, item := range items {
 		if item.GetRoot() != nil {
+			collectionPath, err := GetCanonicalPath(
+				// TODO(meain): Is there a better way to get root path?
+				fmt.Sprintf("/drives/%s/root:", driveID),
```
ashmrtn, line 388: unfortunately I don't think so. Though maybe we should pull the pattern itself into a const and just reference it here?
```suggestion
				// TODO(meain): Is there a better way to get root path?
				fmt.Sprintf(rootDrivePattern, driveID),
```

src/internal/connector/onedrive/collections.go
```diff
@@ -454,8 +480,12 @@ func (c *Collections) UpdateCollections(
 			// TODO(ashmrtn): Figure what when an item was moved (maybe) and add it to
 			// the exclude list.
 
-			col, found := c.CollectionMap[collectionPath.String()]
+			collectionID, ok := visitedPaths[collectionPath.String()]
```
ashmrtn, line 483: I think we actually do have the ID of the parent folder (along with the path) when looking at an item. I think it's at `item.GetParentReference().GetId()`. This is part of the `ParentReference` field that is returned which the API doc states is of type [itemReference](https://learn.microsoft.com/en-us/graph/api/resources/itemreference?view=graph-rest-1.0). That way the rest of the code can continue to stay the way it was and we'll just swap the map update to `c.CollectionMap[*item.GetParentReference().GetId()] = col`

I think we probably don't have the parent ID set in our tests right now because we've never needed it before :/ Graph Explorer shows it's populated for everything but the root folder item though (i.e. items in root have it set to root's ID, but root the folder item does not have it set -- see below)

Root folder parent reference:
```
"parentReference": {
    "driveType": "business",
    "driveId": "some-drive-ID"
},
```

src/internal/connector/onedrive/collections_test.go
```diff
@@ -638,9 +662,10 @@ func (suite *OneDriveCollectionsSuite) TestUpdateCollections() {
 				nil,
 				control.Options{ToggleFeatures: control.Toggles{EnablePermissionsBackup: true}})
 
+			// items := append([]models.DriveItemable{driveRootItem("root")}, tt.items...)
```
ashmrtn, line 665: nit: stray line?
```suggestion
```

src/internal/connector/onedrive/collections.go
ashmrtn: hmmm also the comments in the code may be off on what this path actually is. Looking at returned results in Graph Explorer I see stuff like the below which seems to imply we should use "drive" instead of "drives"
```
"parentReference": {
    "driveType": "business",
    "driveId": "some-drive-ID",
    "id": "an-ID",
    "path": "/drive/root:"
},
```

meain: Which query is this for? I am getting `drives` in the response. In fact I had changed some references of `drive` to  `drives` in the test assuming they were misspelt. This is from a delta response:
``` json
"parentReference": {
    "driveType": "business",
    "driveId": "b!54S....8PlFxA",
    "id": "01GI....LRRZ",
    "path": "/drives/b!54S....8PlFxA/root:"
},
```

ashmrtn: oh dear that's confusing :/ The query I pulled my data from was `https://graph.microsoft.com/v1.0/me/drive/root/delta` in Graph Explorer and signed into the dev account assigned to me

Looking at the code and using the query in there `https://graph.microsoft.com/v1.0/drives/<drive ID>/root/delta` yields paths with `drives` instead of `drive` like you've been seeing. I guess I should be more careful with the queries that I use 😅

ashmrtn (approved):
This looks great. Sorry for the churn on tests

PR #2458, opened 2023-02-09
Update to PR template

meain:
## Description

This is a small QOL improvement to the PR template to fix some annoyances.

When creating small PR(usually a few commits), the first commit is usually has the title and description we need. GitHub does take this information and dump it into the editor, but we end up having to cut paste `## Description` from the end of what GitHub added to the top. This change avoids that. Also, now that description is no more in bold with underlines, I've reduced the heading level of the other ones.

*Just a personal opinion, let me know if folks have a difference in opinion*.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test


.github/pull_request_template.md
```diff
@@ -1,14 +1,14 @@
...
+
+#### Does this PR need a docs update or release note?
 
 - [ ] :white_check_mark: Yes, it's included
 - [ ] :clock1: Yes, but in a later PR
-- [ ] :no_entry: No 
+- [ ] :no_entry: No
 
-## Type of change
+#### Type of change
```
ryanfkeepers, line 11: There is no `####` in markdown.  3 hashes is the maximum.  Also, (not a strongly held opinion) I'd recommend keeping with two.

meain: [GFM](https://github.github.com/gfm/#atx-headings) - the flavor of markdown GitHub uses, does have up to 6. The reason why I moved away from 2 to 4 was primarily because of the heading underlines, but we can drop back to 2. The main reason for bringing this up was the `## Description` block.

|Before|After|
|---|---|
|||

dadams39 (approved):

PR #2460, opened 2023-02-09
Make sure OneDrive collections are returned in order 

meain:
## Description

We were using a map so as to not create duplicate collections but this
causes issues with ordering of the collections. This is not
necessarily important as of now, but will be needed when we work on
delta incrementals and have to handle deletes, updates etc.

ref: https://github.com/alcionai/corso/pull/2407#discussion_r1101377901

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -376,8 +376,26 @@ func (c *Collections) UpdateCollections(
 	excluded map[string]struct{},
 	invalidPrevDelta bool,
 ) error {
+	// This can only be made local to a page as the folder rename
+	// might happen between pages. Every page in the graph response
+	// will have all the parents for that item.
+	visitedPaths := map[string]string{}
+
 	for _, item := range items {
 		if item.GetRoot() != nil {
+			collectionPath, err := GetCanonicalPath(
+				// TODO(meain): Is there a better way to get root path?
+				fmt.Sprintf("/drives/%s/root:", driveID),
+				c.tenant,
+				c.resourceOwner,
+				c.source,
+			)
+			if err != nil {
+				return err
+			}
+
+			visitedPaths[collectionPath.String()] = *item.GetId()
+
 			// Skip the root item
 			continue
 		}
@@ -425,6 +443,8 @@ func (c *Collections) UpdateCollections(
 				return err
 			}
 
+			visitedPaths[folderPath.String()] = *item.GetId()
+
 			// Moved folders don't cause delta results for any subfolders nested in
 			// them. We need to go through and update paths to handle that. We only
 			// update newPaths so we don't accidentally clobber previous deletes.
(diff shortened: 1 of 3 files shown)


src/internal/connector/onedrive/collections.go
```diff
@@ -61,9 +61,10 @@ type Collections struct {
 
 	ctrl control.Options
 
-	// collectionMap allows lookup of the data.BackupCollection
+	// CollectionMap allows lookup of the data.BackupCollection
 	// for a OneDrive folder
-	CollectionMap map[string]data.BackupCollection
+	CollectionMap      map[string]data.BackupCollection
+	CollectionOrdering []string
```
meain, line 67: Keeping this public because of sharepoint test `TestUpdateCollections`.

meain:
> I'm not sure I understand where ordering comes into play. Is this to get the most recent update to something or a different thing?

Kinda. Here are some example scenarios:

Previous metadata: {id1: folder1}
Delta:
{id1: moved to folder2}  // or {id1: delete folder1}
{id2: create folder1}

If we have subfolders for folder1, how would that be handled? Also, now that you bring this up, I wanted to bring up a discussion on  [this](https://github.com/alcionai/corso/blob/f9ae56cc01d59932af0434d7e619955d2b4fda6b/src/internal/connector/onedrive/collections.go#L428). If we move a folder, do we have to go to every folder under that and move create collections to mark a move?

ashmrtn:
The below depend on what we've seen for collections so far. I haven't played around to see all the different responses you can get cause it's a bit difficult in graph explorer, but I've been thinking of the following example for some of these:

Assume starting directory layout as
```text
- root
  - foo
    - bar
      - file1
  - baz
```

1. start fetching items via delta query
  * first page of items has new file `/root/foo/bar/file2` -> make new collection for folder `/root/foo/bar`
2. user moves directory `/root/foo` to be a child of directory `/root/baz`
3. get next page of delta results
  * see move for folder `/root/foo` to `/root/baz/foo` but no change for `/root/foo/bar`

Now, on to the question you have with the example above that we can refer to

>If we have subfolders for folder1, how would that be handled?

if there's no collection(s) for the subfolder(s) then everything is fine (i.e. they weren't updated and neither were their items). Kopia does "tree operations" for moves/deletes, so it will pickup on the fact that subfolders also need to be moved (there's a [test](https://github.com/alcionai/corso/blob/b174c632b744d1bd7de353e74db676406e2bca8f/src/internal/kopia/upload_test.go#L1437) for moves and another somewhere for deletes and deletes+subfolder moves)

if we have the situation like the above example then we would need to go through the set of collections and find the ones that are subfolders and adjust their paths/mark them as moved. This is required because kopia takes the path a collection has as the location of the collection. It can't handle "tree operations" on collection paths. Again, this would only be for the collections we have at that time, not all possible subfolders of a folder (i.e. not all subfolders in the previous paths map). We may be able to leverage the updated paths map for this step, but I haven't really thought it through

The paths map is updates all subfolders when a folder is moved because we need correct paths for any subsequent backups we do. Without that information we'll see mismatches in the paths even if folders haven't actually moved (i.e. will compare old path to current (unchanged-since-last-backup) path and find they don't match)

meain:
Ahh, makes sense. We will have to update the paths of all the collections that in the new delta when a parent moves/gets deleted. My initial idea was along the lines of send this to kopia in the order that it happened so that the "tree operations" in kopia would take care of it.

PR #2462, opened 2023-02-09
find calendars by id when restoring

ryanfkeepers:
## Description

Now that calenders are stored by ID, we have a
problem with restore destinations: the dest is
a name, not an ID.  Even though we create the
dest during restoration, we don't have a way to
reference the ID of the created dest, which
causes additional restores to fail on a "folder
already exists" post error.  This change adds in
the missing dest -> id lookup.

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_restore.go
+++ b/src/internal/connector/exchange/service_restore.go
@@ -554,8 +554,9 @@ func CreateContainerDestination(
 			newCache = true
 			directoryCache = ecc
 		} else if did := directoryCache.DestinationNameToID(dest); len(did) > 0 {
-			// calendars are cached by ID, not name, so we need to look up
-			// the destination id before falling back to destination name
+			// calendars are cached by ID in the resolver, not name, so once we have
+			// created the destination calendar, we need to look up its id and use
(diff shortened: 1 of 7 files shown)


src/internal/connector/exchange/service_restore.go
```diff
@@ -548,16 +553,22 @@ func CreateContainerDestination(
 			caches[category] = ecc
 			newCache = true
 			directoryCache = ecc
+		} else if did := directoryCache.DestinationNameToID(dest); len(did) > 0 {
+			// calendars are cached by ID, not name, so we need to look up
+			// the destination id before falling back to destination name
+			dest = did
 		}
 
+		folders := append([]string{dest}, directory.Folders()...)
```
ashmrtn, line 562: can you add a quick comment here to say that this update to the path is so that it's resolved properly in the container resolver please. Otherwise it looks a little weird by itself since we're technically mixing IDs and display names (in theory)

ashmrtn (approved):

PR #2463, opened 2023-02-09
add --on-error flag to replace --fail-fast

ryanfkeepers:
## Description

Adds and exposes the flag on-error that allows
users to choose between 'fail' and 'recover' (ie:
fail-fast and best-effort) error handling modes.

Also sets some fault constructors to fail-fast
so that we don't let errors slip through.

## Type of change

- [x] :sunflower: Feature

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 ### Added
 
+- New Flag `--on-error` which controls recoverable error handling behavior.  Accepts `--on-error fail` (default) to fail out immediately, and `--on-error recover` to proceed with the operation.
+
 ### Fixed
 - Support for item.Attachment:Mail restore
 - Errors from duplicate names in Exchange Calendars
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 ### Added
 
-- New Flag `--on-error` which controls recoverable error handling behavior.  Accepts `--on-error fail` (default) to fail out immediately, and `--on-error recover` to proceed with the operation.
+- New Flag `--on-error` which controls recoverable error handling behavior.  Accepts `--on-error fail` to fail out immediately, and `--on-error continue` (default) to proceed with the operation.
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


meain (commented):
Code changes look good, I'll let others to approve on the behaviour/flag change.

ryanfkeepers:
@gmatev @vkamra could I get a review on the general change here, please?  Abin has already given the granular code changes a thumbs up.

vkamra:
@ryanfkeepers - this also changes the default behavior for backup - correct? The flag looks good - I do want to understand the motivation behind the behavior change.

gmatev:
If I understand the behavior change correctly, we now default to stopping immediately as opposed to best effort. This is a change form before.

Is this correct? If yes, why the change? Seems to me that we want to do as much as we can so the default should be the opposite.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (alpha)
 ### Added
 
+- New Flag `--on-error` which controls recoverable error handling behavior.  Accepts `--on-error fail` (default) to fail out immediately, and `--on-error recover` to proceed with the operation.
```
gmatev, line 11: The option name of `recover` seems weird. 

My understanding is that this is a more general parameter. A more appropriate value is probably `continue`

ryanfkeepers:
@vkamra @gmatev Good questions.  The concern came from reviews across the error handling migration.  Since the codebase currently has scattered error handling, we always failFast in some places, even when we could continue.  Worries were expressed that a switch to best effort in those locations might cause overly lax handling in error cases with unexpected consequences.

This change to default to failFast behavior is to ensure that we will continue to hard stop at those locations while the handling migration is underway, and that we can return the default to best effort once things have settled.  The new flag format will help support this transition to ensure that if or when we swap the default again, existing flag declarations will maintain the same behavior (and be easier to script altogether).

CHANGELOG.md
ryanfkeepers: Good call, will change.

vkamra:
@ryanfkeepers - couple of follow ups:
1) `recover` needs to be changed to `continue`
2) Can we hold of on changing the default value for --on-error? My concern is that we will start failing earlier in locations where we do not want to. For places where we absolutely need to fail hard - we shouldn't be using `opts.FailFast` anyway.

vkamra (approved):

PR #2465, opened 2023-02-09
GC: Item attachment contact support

### LINKED ISSUE #2426: Item Attachment: Contact Support
Item.Attachments of OdataType `Contacts` require special transformations prior to being uploaded.

dadams39:
## Description
The logic for sanitizing Contactable data for restoration of `ItemAttachable.Contact` types. 

Contact `Item.Attachment`s required the removal of:
- `odata.Context` 
- `ETag` 
- `ParentFolder`

Otherwise, the following error occurs on POST. 
```bash
UnableToDeserializePostBody were unable to deserialize
```

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [x] :bug: Bugfix

## Issue(s)

* closes #2426<issue>

## Test Plan
- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -319,3 +319,16 @@ func RunWithRetry(run func() error) error {
 
 	return support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
 }
+
+// UnwrapStringPointer helper method for unwrapping strings
+// Microsoft Graph saves many variables as string pointers.
+// Function will safely check if the point is nil prior to
+// dereferencing the pointer. If the pointer is nil,
+// an empty string is returned.
+func UnWrapStringPointer(ptr *string) string {
+	if ptr == nil {
+		return ""
+	}
+
+	return *ptr
+}
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -320,7 +320,7 @@ func RunWithRetry(run func() error) error {
 	return support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
 }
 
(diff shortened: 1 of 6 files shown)


src/internal/connector/graph/service.go
```diff
@@ -319,3 +319,16 @@ func RunWithRetry(run func() error) error {
 
 	return support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
 }
+
+// UnWrapStringPointer helper method for unwrapping strings
+// Microsoft Graph saves many variables as string pointers.
+// Function will safely check if the point is nil prior to
+// dereferencing the pointer. If the pointer is nil,
+// an empty string is returned.
+func UnWrapStringPointer(ptr *string) string {
```
ryanfkeepers, line 328: This could see a lot of usage.  I'm not against that, but three things.

1/ since it's such a frequent concern, I think we can get away with a really compact func name.  Ex: moving it to a package like `ptr` and naming the func `Val`.  Having `ptr.Val(x)` everywhere is much easier on the eyes than `graph.UnWrapStringPointer(x)`.

2/ We have a handful of validations funcs that want to know if the value was populated or not.  Just dealing with a string can make that confusing, because an empty string is not always unpopulated.  So a return value of string, bool, ex: `Val(p *string) (string, bool)` and handling it as `s, ok := Val(x)` is helpful in those cases.  (or maybe a second func like ValOK(x), where needed).

3/ This is a prime use case for generics.  `func Val[T any](ptr *T) (T, bool)` solves every type, not just strings.

ryanfkeepers (approved):

src/internal/connector/graph/service.go
vkamra: 👍 for (1) here

PR #2467, opened 2023-02-10
ensure display folder matches location, if possible

ryanfkeepers:

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #2423

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -141,13 +141,20 @@ func FolderEntriesForPath(parent, location *path.Builder) []folderEntry {
 	lfs := locationRefOf(location)
 
 	for len(parent.Elements()) > 0 {
-		nextParent := parent.Dir()
+		var (
+			nextParent = parent.Dir()
+			lr         string
+			dn         = parent.LastElem()
+		)
 
 		// TODO: We may have future cases where the storage hierarchy
 		// doesn't match the location hierarchy.
-		var lr string
 		if lfs != nil {
 			lr = lfs.String()
+
+			if len(lfs.Elements()) > 0 {
+				dn = lfs.LastElem()
+			}
 		}
 
 		folders = append(folders, folderEntry{
@@ -158,7 +165,7 @@ func FolderEntriesForPath(parent, location *path.Builder) []folderEntry {
 			Info: ItemInfo{
 				Folder: &FolderInfo{
 					ItemType:    FolderItem,
-					DisplayName: parent.Elements()[len(parent.Elements())-1],
+					DisplayName: dn,
 				},
 			},
 		})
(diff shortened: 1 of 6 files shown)


src/pkg/backup/details/details_test.go
```diff
@@ -724,3 +724,162 @@ func (suite *DetailsUnitSuite) TestUpdateItem() {
...
+		},
+		{
+			name:   "single depth parent only",
+			parent: basePath.Append(fnords...),
+			expect: folderEntriesFor(fnords, nil),
+		},
+		{
+			name:     "single depth with location",
+			parent:   basePath.Append(fnords...),
+			location: basePath.Append(beau...),
```
ashmrtn, line 849: nice

src/pkg/path/path.go
```diff
@@ -213,6 +213,14 @@ func (pb Builder) Dir() *Builder {
 	}
 }
 
+func (pb Builder) LastElem() string {
+	if len(pb.elements) == 0 {
+		return ""
+	}
+
+	return pb.elements[len(pb.elements)-1]
```
ashmrtn, line 221: really makes me wish for negative indexing sometimes

src/pkg/backup/details/details.go
```diff
@@ -141,13 +141,20 @@ func FolderEntriesForPath(parent, location *path.Builder) []folderEntry {
...
 
 		// TODO: We may have future cases where the storage hierarchy
 		// doesn't match the location hierarchy.
-		var lr string
 		if lfs != nil {
 			lr = lfs.String()
+
+			if len(lfs.Elements()) > 0 {
+				dn = lfs.LastElem()
+			}
```
ashmrtn, line 157: Not a blocker since we currently control inputs to this but... it seems like the behavior when we "bottom out" is that the stuff earlier in the path won't have display names correct? Since we're starting at the end (complete path) and removing elements as we go

Just, I think we need to make time to refactor this code in general 😅

ashmrtn (approved):

src/pkg/path/path.go
ryanfkeepers: I'm a little surprised the `/x/slices` package doesn't have this (and head, tail, etc) in it.

src/pkg/backup/details/details.go
ryanfkeepers: Agreed.  We def need a refactor, because we should be constructing folders by walking up the tree using only the location path, and not the full repo ref.  Ex: for every folder we're getting folder elems for the prefix (tenant/service/owner/type), which is wasteful, and not representative of anything outside of corso.  Also, it's supposed to be location by preference.

But before we do that, we need to ensure location is populated for all the data we produce.

ashmrtn (approved):

Issue #2447, opened 2023-02-08
Remove data dependency in OneDrive restore


ashmrtn:
OneDrive restore currently has an ordering dependency between creating items and restoring item permissions. This should be removed so we don't begin relying on the behavior of individual components for restore to function properly

#1944 and #2434 added an inline call to the RestoreCollection that allows fetching files in the same collection by name. This will remove the ordering dependency

PR #2468, opened 2023-02-10
centralize folder creation and permission setting for OneDrive restore

ashmrtn:
## Description

Rearrange how folder permissions are restored. Instead of restoring them when reading the meta file in the parent folder restore them when iterating through the collection itself.

This solution leans more heavily on the idea of having a map[folder path]->folder permissions and uses that to lookup what to restore

Folder permissions are still read and added to the map as they are encountered

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2447 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -332,8 +332,6 @@ func RestoreCollection(
 					// RestoreOp, so we still need to handle them in some way.
 					continue
 				} else if strings.HasSuffix(name, DirMetaFileSuffix) {
-					// Only the first version needed to deserialize the permission for
-					// child folders here. Later versions can request permissions inline.
 					if !restorePerms {
 						continue
 					}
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/restore.go
```diff
@@ -358,29 +380,46 @@ func RestoreCollection(
...
+		id,
+		parentPermissions,
+		folderPermissions,
+		permissionIDMappings)
+	if err != nil {
+		return id, permissionIDMappings, err
+	}
 
-	return *folderItem.GetId(), nil
+	return id, permissionIDMappings, nil
```
meain, line 413: We can probably drop the `if err != nil` check here.
``` go
return id, permissionIDMappings, err
```

src/internal/connector/onedrive/restore.go
```diff
@@ -288,40 +332,18 @@ func RestoreCollection(
...
-						drivePath.DriveID,
-						trimmedName,
-						restoreFolderID,
-					)
-					if err != nil {
-						errUpdater(itemData.UUID(), err)
-						continue
-					}
-
+					// Only the first version needed to deserialize the permission for
```
meain, line 335: Not sure what you mean by "first version" here.

meain (approved):

Issue #2472, opened 2023-02-10
OneDrive permissions are randomly not backed up


meain:
I have not been able to pin point what is causing this, but OneDrive permissions are randomly not getting backed up. The permissions are being fetched (ie available in [`itemMeta`](https://github.com/alcionai/corso/blob/78e8e1a4a83ce1f6cbd5a3a05549296ec2a6ed91/src/internal/connector/onedrive/collection.go#L292)), but what ends up in kopia is just empty permissions(ie `{}`). It creates the `.meta` file but with empty content. The `.meta` file is also present in the  `details` file.

These runs somehow seems to not call [`LazyReadCloser` for `metaReader`](https://github.com/alcionai/corso/blob/78e8e1a4a83ce1f6cbd5a3a05549296ec2a6ed91/src/internal/connector/onedrive/collection.go#L390) (for any of the .meta files). Not sure how it ends up with empty permissions(`{}`) if this call is not happening.

This is never reproducible in the tests, at least I have not seen it in CI nor in my local runs for tests.

PR #2473, opened 2023-02-10
Fix OneDrive permissions not getting backed up

meain:
## Description

When we do a new backup, we look up the old permissions assuming they are accurate, but the permissions from the previous backup might be incorrect as we may have run it with permissions backup disabled.

It can also happen the other way round where even if we ask it to disable permissions backup, it can still end up having permissions which are pulled from the previous backup.

This fixes this behavior by setting the mod time for the backup info to the current time.

- [x] Add tests
- [x] Address TODOs in the PR

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* fixes https://github.com/alcionai/corso/issues/2472

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -402,10 +402,27 @@ func (oc *Collection) populateItems(ctx context.Context) {
 					return progReader, nil
 				})
 
+				// TODO(meain): Remove this once we change to always
+				// backing up permissions. Until then we cannot rely
+				// on weather the previous data is what we need as the
+				// user might have not backup up permissions in the
+				// previous run.
+				metaItemInfo := details.ItemInfo{}
+				metaItemInfo.OneDrive = &details.OneDriveInfo{
+					Created:    itemInfo.OneDrive.Created,
+					ItemName:   itemInfo.OneDrive.ItemName,
+					DriveName:  itemInfo.OneDrive.DriveName,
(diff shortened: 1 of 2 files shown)


ashmrtn:
hmmmm I think we need to be careful with this else it could cause unexpected behavior for SDK consumers. What is it that we need to do for this issue? Do we need to ensure kopia always saves the data and doesn't use cached data at all or something else?

meain:
> Do we need to ensure kopia always saves the data and doesn't use cached data at all or something else?

That is pretty much it. Since we don't know if the permissions are saved without or without the `--enable-permissions-backup` flag, we can never be certain as to if that information is still valid.

ashmrtn:
ah, I see now, thanks for the info. Just to recap some stuff that was discussed offline...

Setting the mod time to the current time should ensure the file is always backed up (by disabling kopia-assisted incrementals). We may want to make that more optimized in the future somehow, but I don't think we have the infrastructure to do so right now.

Beyond that, we should remove metadata items from backup details. Backup details is becoming more of a store for the "user visible" items in a backup and not so much a record of all the extra things that corso stores. Removing metadata items from backup details can be done in a separate PR though

ashmrtn (approved):

PR #2487, opened 2023-02-13
Consolidate/move OneDrive GraphConnector tests

ashmrtn:
## Description

Now that we're also testing permissions behavior the OneDrive tests have gotten kind of large. This PR consolidates some of the code for tests and moves the OneDrive tests to their own file

It also sets things up so it's easier to add tests for different backup layouts in the future

Viewing by commit may make reviews easier

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #2447

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph_connector_helper_test.go
```diff
@@ -1030,7 +1031,7 @@ func collectionsForInfo(
 			baseExpected[info.items[i].lookupKey] = info.items[i].data
 
 			// We do not count metadata files against item count
-			if service != path.OneDriveService ||
+			if countMeta || service != path.OneDriveService ||
```
meain, line 1047: nit: The name `countMeta` seems a bit misleading. How about we just pass in the version number of the backup and use that here?

meain (approved):
LGTM, just a small nit.

src/internal/connector/graph_connector_helper_test.go
ashmrtn: yea, lets try that for now. We may need to do something else in the future if we start getting a lot of different versions we need to switch behavior on, but I guess we'll figure out when we get there ;)

PR #2491, opened 2023-02-14
add clues/fault to exchange restore

ryanfkeepers:

- [ ] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/attachment.go
```diff
@@ -46,54 +47,52 @@ func uploadAttachment(
...
-			// TODO: (rkeepers) Update to support PII protection
-			logger.Ctx(ctx).Infow(msg,
-				"err", err,
-				"attachment_name", name,
-				"attachment_type", attachmentType,
-				"internal_item_type", getItemAttachmentItemType(prev),
-				"attachment_id", ptr.Val(prev.GetId()),
-			)
+			logger.Ctx(ctx).
+				With("err", err).
```
ashmrtn, line 74: nit: I don't think we need to worry about number of characters output to the log so perhaps "error" would be better? ;) We should also probably make a standard for this field cause it will be rather common in our log output so it should be made easy to search for

```suggestion
				With("error", err).
```

src/internal/connector/exchange/service_restore.go
```diff
@@ -362,14 +345,15 @@ func RestoreExchangeDataCollections(
...
 	}
 
-	status := support.CreateStatus(ctx,
+	status := support.CreateStatus(
+		ctx,
 		support.Restore,
 		len(dcs),
 		metrics,
-		errs,
+		errs.Err(),
```
ashmrtn, line 353: since this only returns the first error and restore operation currently gets error information through the `AwaitStatus()` call isn't this going to hide other errors that were encountered for this particular collection?

ashmrtn (approved):
Generally looks good, question about error(s) sent to the status updater and how that will affect visibility into the system until we have the fault package through all of corso

src/internal/connector/exchange/attachment.go
ryanfkeepers: I've seen both cases be standard.  Let's put a pin on this and I can ensure everything uses the same format as a cleanup at the end.

src/internal/connector/exchange/service_restore.go
ryanfkeepers: Exposing the other errors relies on the controller reviewing errs.Errs().  Nothing here is lost, ultimately.  First priority is to ensure we're tracking whether or not an error happened, and for that we only need to find one.  The long term goal is to remove error tracking from support.Status altogether.

PR #2492, opened 2023-02-14
add fault & clues to exchange api

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/common/ptr/pointer.go
+++ b/src/internal/common/ptr/pointer.go
@@ -1,5 +1,7 @@
 package ptr
 
+import "time"
+
 // ptr package is a common package used for pointer
 // access and deserialization.
 
@@ -18,4 +20,26 @@ func Val[T any](ptr *T) T {
 	}
 
 	return *ptr
+}
+
+// ValOK behaves the same as Val, except it also gives
+// a boolean response for whether the pointer was nil
+// (false) or non-nil (true).
+func ValOK[T any](ptr *T) (T, bool) {
+	if ptr == nil {
+		return *new(T), false
+	}
+
+	return *ptr, true
+}
+
+// OrNow returns the value of the provided time, if the
+// parameter is non-nil.  Otherwise it returns the current
+// time in UTC.
+func OrNow(t *time.Time) time.Time {
+	if t == nil {
+		return time.Now().UTC()
+	}
+
+	return *t
 }
--- a/src/internal/connector/exchange/api/api.go
+++ b/src/internal/connector/exchange/api/api.go
@@ -3,12 +3,13 @@ package api
 import (
 	"context"
 	"strings"
-	"time"
 
+	"github.com/alcionai/clues"
 	"github.com/microsoft/kiota-abstractions-go/serialization"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/pkg/account"
 )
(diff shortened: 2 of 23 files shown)


src/internal/common/ptr/pointer.go
```diff
@@ -19,3 +21,25 @@ func Val[T any](ptr *T) T {
...
+// a boolean response for whether the pointer was nil
+// (false) or non-nil (true).
+func ValOK[T any](ptr *T) (T, bool) {
+	if ptr == nil {
+		return *new(T), false
+	}
+
+	return *ptr, true
+}
+
```
dadams39, line 35: The wording is a little misleading in comparison with the logic. If it was nil, the first clause should return true and the second one is false. Otherwise, the comments require updating.

src/internal/connector/exchange/api/api.go
```diff
@@ -122,34 +123,24 @@ func newLargeItemService(creds account.M365Config) (*graph.Service, error) {
 // checkIDAndName is a helper function to ensure that
 // the ID and name pointers are set prior to being called.
 func checkIDAndName(c graph.Container) error {
-	idPtr := c.GetId()
-	if idPtr == nil || len(*idPtr) == 0 {
-		return errors.New("folder without ID")
+	id := ptr.Val(c.GetId())
```
dadams39, line 126: L23 `numberOfRetries` is complaining about being unused.

src/internal/connector/exchange/api/contacts.go
```diff
@@ -125,32 +130,42 @@ func (c Contacts) EnumerateContainers(
...
+		if !ok {
 			break
 		}
 
-		builder = users.NewItemContactFoldersItemChildFoldersRequestBuilder(*resp.GetOdataNextLink(), service.Adapter())
+		builder = users.NewItemContactFoldersItemChildFoldersRequestBuilder(link, service.Adapter())
 	}
 
-	return errs.ErrorOrNil()
+	return errs.Err()
```
dadams39, line 168: Do we no longer want to use the `OrNil` package here?

src/internal/connector/exchange/api/events.go
```diff
@@ -244,33 +234,30 @@ func (c Events) GetAddedAndRemovedItemIDs(
...
+		var (
+			builder = users.NewItemCalendarsItemEventsDeltaRequestBuilder(oldDelta, service.Adapter())
+			pgr     = &eventPager{service, builder, nil}
+		)
 
 		added, removed, deltaURL, err := getItemsAddedAndRemovedFromContainer(ctx, pgr)
 		// note: happy path, not the error condition
 		if err == nil {
-			return added, removed, DeltaUpdate{deltaURL, false}, errs.ErrorOrNil()
+			return added, removed, DeltaUpdate{deltaURL, false}, err
```
dadams39, line 252: If err == nil, why would we return the error rather than nil?

dadams39 (approved):
A few nits

src/internal/common/ptr/pointer.go
ryanfkeepers: mm, no, I think it's correct.  Consider this in comparison to indexing map values:
```go
m := map[int]any{}
v, ok := m[key] // ok is true if the key holds an entry in the map.  ie: "its present"

p *int = 0
v, ok := ptr.ValOK(p) // ok is true if p is non-nil.  ie: "its present"
```

src/internal/connector/exchange/api/contacts.go
ryanfkeepers: Nope.  `ErrorOrNil` is a silly hack and shouldn't be needed.

src/internal/connector/exchange/api/events.go
ryanfkeepers: hah, I was on autopilot.  Good catch.

PR #2493, opened 2023-02-14
update items to accept ctx, fault.Errors

ryanfkeepers:
## Description

In order for corso to track recoverable errors,
we need to pass a fault.Errors struct into the
items stream.  As long as we're doing that, we
might as well pass along the available ctx as well.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -11,6 +11,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/alcionai/clues"
 	"github.com/microsoft/kiota-abstractions-go/serialization"
 
 	"github.com/alcionai/corso/src/internal/connector/graph"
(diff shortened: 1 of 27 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -307,40 +293,24 @@ func getItemWithRetries(
...
-		if graph.IsErrDeletedInFlight(err) {
-			return nil, nil, err
-		}
-
-		if i < numberOfRetries {
-			time.Sleep(time.Duration(3*(i+1)) * time.Second)
-		}
-	}
-
+	item, info, err := items.GetItem(ctx, userID, itemID, errs)
```
meain, line 296: Good catch on the retry logic. :D

meain (approved):

PR #2494, opened 2023-02-14
Begin storing OneDrive file names in metadata

ashmrtn:
## Description

This starts to store the name of the OneDrive file/directory in the corso metadata file for the OneDrive file/directory. This data is not yet used anywhere

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1535

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/item.go
```diff
@@ -186,17 +224,17 @@ func oneDriveItemInfo(di models.DriveItemable, itemSize int64) *details.OneDrive
 func oneDriveItemMetaInfo(
```
meain, line 224: nit: Should we change the name to suggest that we are only returning permissions now instead of entire metadata.

src/internal/connector/onedrive/item.go
```diff
@@ -61,18 +62,55 @@ func oneDriveItemMetaReader(
 	service graph.Servicer,
 	driveID string,
 	item models.DriveItemable,
-) (io.ReadCloser, int, error) {
-	meta, err := oneDriveItemMetaInfo(ctx, service, driveID, item)
-	if err != nil {
-		return nil, 0, err
+	fetchPermissions bool,
+) (r io.ReadCloser, size int, err error) {
```
meain, line 66: Just wondering if we should have a conversation about using naked returns in the codebase.

From a quick grep, it looks like we only use naked returns when we deal with `recover()` as of now:
```
$ rg '[^A-z]\([^\(]*,?\w+ \w+\) \{'
# recover
internal/operations/backup.go:func (op *BackupOperation) Run(ctx context.Context) (err error) {
internal/operations/restore.go:func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.Details, err error) {

# tests (not naked returns though)
internal/connector/uploadsession/uploadsession_test.go:func (mr *mockReader) Read(b []byte) (n int, err error) {
internal/kopia/model_store_test.go:             func(innerCtx context.Context, w repo.RepositoryWriter) (innerErr error) {

# not naked returns even though named
internal/kopia/upload.go:func (rw *backupStreamReader) Read(p []byte) (n int, err error) {
internal/kopia/upload.go:func (rw *restoreStreamReader) Read(p []byte) (n int, err error) {
```

ryanfkeepers: My vote goes to always using naked returns, unless required by a recover.  Post processing in defer can be similarly achieved by separating code into two funcs, with less confusion to boot.

src/internal/connector/onedrive/collection_test.go
```diff
@@ -343,92 +345,6 @@ func (suite *CollectionUnitTestSuite) TestCollectionReadError() {
 	}
 }
 
-func (suite *CollectionUnitTestSuite) TestCollectionDisablePermissionsBackup() {
```
ashmrtn, line 346: this is now an e2e test in graph_connector_onedrive_test.go. The code it was exercising was changed such that it was no longer easy to test in a unit test

ashmrtn:
this should be mergeable without ill effects. It just persists data we don't use yet

src/internal/connector/onedrive/item.go
```diff
@@ -61,18 +62,40 @@ func oneDriveItemMetaReader(
...
 	if err != nil {
+		// Keep this in an if-block because if it's not then we have a weird issue
+		// of having no value in error but golang thinking it's non nil because of
+		// the way interfaces work.
+		err = clues.Wrap(err, "fetching item permissions")
+	} else {
+		meta.Permissions = perms
+	}
+
+	metaJSON, serializeErr := json.Marshal(meta)
```
meain, line 81: Should we bother trying to marshall if we hit an error in 76?

meain (approved):
Small nit, approving as nothing major.

src/internal/connector/onedrive/item.go
ashmrtn: that's a good question. I think it would be best if we do because then we have the highest change of recovering some data down the line even if we had issues. Since the file name is now in the metadata file it's harder for us to ensure we have all the things we need to restore an item properly

If we skip marshalling data then there's the chance we'll end up with the item data but not the item name (depending on how fail-fast semantics evolve as well). If we do marshall it then maybe we won't have permissions, but we'll at least have the file name

PR #2495, opened 2023-02-14
New restore path for items with name in metadata file

ashmrtn:
## Description

Add code to restore items that have their file name stored in the corso metadata file instead of as the name of the kopia file.

This code is not executed as the conditional for it will not trigger (comparison to max int)

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1535 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"runtime/trace"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


src/internal/connector/onedrive/restore.go
```diff
@@ -390,6 +381,155 @@ func RestoreCollection(
...
+	}
+
+	metaReader := permsFile.ToReader()
+	meta, err := getMetadata(metaReader)
+	metaReader.Close()
+
+	if err != nil {
+		err = clues.Wrap(err, "deserializing item metadata")
+		return details.ItemInfo{}, err
+	}
```
meain, line 438: nit: Should we move this block to a util function.

src/internal/connector/onedrive/restore.go
```diff
@@ -390,6 +381,155 @@ func RestoreCollection(
...
+	if !restorePerms {
+		return itemInfo, nil
+	}
+
+	// Fetch item permissions from the collection and restore them.
+	metaName := trimmedName + MetaFileSuffix
+
+	permsFile, err := fetcher.Fetch(ctx, metaName)
+	if err != nil {
+		err = clues.Wrap(err, "getting item metadata")
```
meain, line 427: I know this is code that I had written, but now that I look at it, having the name available in the error might end up being useful.

meain (approved):

Issue #1762, opened 2022-12-09
Emit `Repo Connect` event


gmatev:
Emit a `Repo Connect` event (same properties as Repo Init) but only when `corso repo connect XXXX` CLI command succeeds (this should not be done during the repository connections established for other CLI operations)

This will allow us to distinguish repo connect CLI command from other operations

vkamra:
@neha-Gupta1 - take a look at this one. This would involve firing the [Repo Connect](https://github.com/alcionai/corso/blob/main/src/internal/events/events.go#L28) event when repo.Connect() is called from `connectS3Cmd`.

We could do this inside `repo.Connect()` but need to make sure this only happens when repo.Connect is invoked as part of the CLI command

neha-Gupta1:
Merged in https://github.com/alcionai/corso/pull/2502

PR #2502, opened 2023-02-14
Handle repo connect event on - repo connect command

neha-Gupta1:
## Description

Create a `repo connect` event on CLI's repo connect command 

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature

## Issue(s)
https://github.com/alcionai/corso/issues/1762

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/control/options.go
```diff
@@ -11,6 +11,7 @@ type Options struct {
 	FailFast           bool            `json:"failFast"`
 	RestorePermissions bool            `json:"restorePermissions"`
 	ToggleFeatures     Toggles         `json:"ToggleFeatures"`
+	RepoConnect        bool            `json:"-"`
```
ryanfkeepers, line 14: I think `options` is the wrong owner for this flag.  The struct is used to aggregate runtime configurations: settings that could cascade to many different places throughout processing.  What you're solving in this PR is a one-off flag that only involves the repository.Connect call.  Placing the flag into options will expose it to a lot more topology, and that might confuse contributors who are working with options downstream.

My recommendation is to add the flag to the Connect() parameters themselves.  That way it's clear that this change only affects that one function call, and isn't going to cause confusion elsewhere.

src/cli/repo/s3.go
```diff
@@ -183,7 +183,7 @@ func connectS3Cmd(cmd *cobra.Command, args []string) error {
 		return Only(ctx, errors.Wrap(err, "Failed to parse m365 account config"))
 	}
 
-	r, err := repository.Connect(ctx, a, s, options.Control())
+	r, err := repository.Connect(ctx, a, s, true, options.Control())
```
vkamra, line 186: @ryanfkeepers @neha-Gupta1 - thoughts on just adding a `func ConnectWrapper(...)` helper in the `repo` package that connects to the repo and fires the event? Seems cleaner.

neha-Gupta1: Seems to be a cleaner way. If @ryanfkeepers agrees with it.

vkamra: @neha-Gupta1 - to save you the wait till tomorrow - how about we move forward with that approach. I think @ryanfkeepers will be on board and if not I will owe you both some ☕

ryanfkeepers (approved):

src/cli/repo/s3.go
ryanfkeepers: I am in such deep coffee debt.

PR #2503, opened 2023-02-14
Mark collections as Deleted if not available and invalid delta

meain:

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* Final entry : closes https://github.com/alcionai/corso/issues/2242

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -344,6 +344,13 @@ func (c *Collections) Get(
 		folderPaths[driveID] = map[string]string{}
 		maps.Copy(folderPaths[driveID], paths)
 
+		logger.Ctx(ctx).Infow(
+			"persisted metadata for drive",
+			"num_paths_entries",
+			len(paths),
+			"num_deltas_entries",
+			numDeltas)
+
 		if !delta.Reset {
 			maps.Copy(excludedItems, excluded)
 			continue
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -371,8 +371,8 @@ func (c *Collections) Get(
 
 			_, found = modifiedPaths[p]
 			if found {
-				// Path has been updated with something else already
-				// instead of being modified
+				// Original folder was deleted and new folder with the
+				// same name/path was created in its place
 				continue
 			}
 
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -344,14 +344,52 @@ func (c *Collections) Get(
 		folderPaths[driveID] = map[string]string{}
 		maps.Copy(folderPaths[driveID], paths)
 
-		maps.Copy(excludedItems, excluded)
+		if !delta.Reset {
+			maps.Copy(excludedItems, excluded)
+			continue
+		}
 
-		logger.Ctx(ctx).Infow(
-			"persisted metadata for drive",
-			"num_paths_entries",
-			len(paths),
-			"num_deltas_entries",
-			numDeltas)
+		// Set all folders in previous backup but not in the current
+		// one with state deleted
+		modifiedPaths := map[string]struct{}{}
+		for _, p := range c.CollectionMap {
+			modifiedPaths[p.FullPath().String()] = struct{}{}
+		}
+
+		for i, p := range oldPaths {
+			_, found := paths[i]
+			if found {
+				continue
+			}
+
+			_, found = modifiedPaths[p]
+			if found {
+				// Path has been updated with something else already
+				// instead of being modified
+				continue
+			}
+
+			delete(paths, i)
+
+			prevPath, err := path.FromDataLayerPath(p, false)
+			if err != nil {
+				return nil, map[string]struct{}{},
+					clues.Wrap(err, "invalid previous path").WithClues(ctx).With("deleted_path", p)
+			}
+
+			col := NewCollection(
+				c.itemClient,
+				nil,
+				prevPath,
+				driveID,
+				c.service,
+				c.statusUpdater,
+				c.source,
+				c.ctrl,
+				true,
+			)
+			c.CollectionMap[i] = col
+		}
 	}
 
 	observe.Message(ctx, observe.Safe(fmt.Sprintf("Discovered %d items to backup", c.NumItems)))
--- a/src/internal/connector/onedrive/collections_test.go
+++ b/src/internal/connector/onedrive/collections_test.go
@@ -1531,6 +1531,89 @@ func (suite *OneDriveCollectionsSuite) TestGet() {
 			expectedDelList: map[string]struct{}{"file": {}},
 			doNotMergeItems: false,
 		},
+		{
+			name:   "OneDrive_OneItemPage_NoPrevDelta_DeleteNonExistantFolder",
+			drives: []models.Driveable{drive1},
+			items: map[string][]deltaPagerResult{
+				driveID1: {
+					{
+						err: getDeltaError(),
+					},
+					{
+						items: []models.DriveItemable{
+							driveRootItem("root"),
+							driveItem("folder2", "folder2", driveBasePath1, "root", false, true, false),
+							driveItem("file", "file", driveBasePath1+"/folder2", "folder2", true, false, false),
+						},
+						deltaLink: &delta,
+					},
+				},
+			},
+			errCheck: assert.NoError,
+			prevFolderPaths: map[string]map[string]string{
+				driveID1: {
+					"root":   rootFolderPath1,
+					"folder": folderPath1,
+				},
+			},
+			expectedCollections: map[string]map[data.CollectionState][]string{
+				expectedPath1(""):         {data.NotMovedState: {"folder2"}},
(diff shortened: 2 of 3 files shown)


src/internal/connector/onedrive/collections.go
```diff
@@ -325,6 +328,33 @@ func (c *Collections) Get(
...
 
+		// Set all folders in previous backup but not in the current
+		// one with DoNotMerge
+		if c.ctrl.ToggleFeatures.EnableOneDriveDeltaIncrementals && delta.Reset {
+			for i, p := range oldPaths {
+				_, found := paths[i]
+				if !found {
+					prevPath, err := path.FromDataLayerPath(p, false)
+					if err != nil {
+						return nil, map[string]struct{}{}, clues.Wrap(err, fmt.Sprintf("invalid previous path for deleted item %s", p))
```
ryanfkeepers, line 339: logging standards: keep messages short and simple, don't fmt data.  Also, since the erroring function didn't take in a `ctx`, you'll want to plug in the clues from the context (WithClues(ctx)).
```suggestion
						return nil, map[string]struct{}{}, clues.Wrap(err, "invalid previous path").WithClues(ctx).With("deleted_path", p)
```

src/internal/connector/onedrive/collections.go
```diff
@@ -325,6 +328,33 @@ func (c *Collections) Get(
...
 		maps.Copy(folderPaths[driveID], paths)
 
+		// Set all folders in previous backup but not in the current
+		// one with DoNotMerge
+		if c.ctrl.ToggleFeatures.EnableOneDriveDeltaIncrementals && delta.Reset {
+			for i, p := range oldPaths {
+				_, found := paths[i]
+				if !found {
+					prevPath, err := path.FromDataLayerPath(p, false)
+					if err != nil {
```
ryanfkeepers, line 338: Nit: deep nesting always feels like a smell to me.  Some suggestions:
```suggestion
		if !c.ctrl.ToggleFeatures.EnableOneDriveDeltaIncrementals || !delta.Reset {
		    maps.Copy(excludedItems, excluded) // pulled up from 358
		    continue
		}
		for i, p := range oldPaths {
			_, found := paths[i]
			if found {
			    continue
			}
			
			prevPath, err := path.FromDataLayerPath(p, false)
			if err != nil {
```

src/internal/connector/onedrive/collections.go
```diff
@@ -255,7 +256,7 @@ func (c *Collections) Get(
 	ctx context.Context,
 	prevMetadata []data.RestoreCollection,
 ) ([]data.BackupCollection, map[string]struct{}, error) {
-	prevDeltas, _, err := deserializeMetadata(ctx, prevMetadata)
+	prevDeltas, allOldPaths, err := deserializeMetadata(ctx, prevMetadata)
```
ryanfkeepers, line 259: Nit: for clarity
```suggestion
	prevDeltas, oldPathsByDriveID, err := deserializeMetadata(ctx, prevMetadata)
```

src/internal/connector/onedrive/collections.go
```diff
@@ -325,6 +328,33 @@ func (c *Collections) Get(
 		folderPaths[driveID] = map[string]string{}
 		maps.Copy(folderPaths[driveID], paths)
 
+		// Set all folders in previous backup but not in the current
+		// one with DoNotMerge
+		if c.ctrl.ToggleFeatures.EnableOneDriveDeltaIncrementals && delta.Reset {
+			for i, p := range oldPaths {
```
ryanfkeepers, line 334: `for i, p` is confusing when `oldPaths` is a map; `i` suggests an index, as if oldPaths is an array.
```suggestion
			for k, p := range oldPaths {
			// alt: for folderID, p := range oldPaths {
```

src/internal/connector/onedrive/collections.go
```diff
@@ -325,6 +328,33 @@ func (c *Collections) Get(
 		folderPaths[driveID] = map[string]string{}
 		maps.Copy(folderPaths[driveID], paths)
 
+		// Set all folders in previous backup but not in the current
+		// one with DoNotMerge
```
ashmrtn, line 332: We actually want to tell kopia to delete them. Deleting them is safe because we must not have seen them during our item enumeration because they were deleted in OneDrive (we lookup folders by ID in the map so renames/moves don't affect this)

This is to catch situation like
```text
1. create backup and store Corso metadata
2. user deletes folder (e.x. /root/foo)
3. delta token expires
4. backup launched
5. all existing files and folders discovered (/root/foo not found since it was deleted)
```

steps 2 and 3 above can be swapped, the only important thing is that they both happen before the next backup

if we didn't make a collection and mark `/root/foo` as deleted then kopia wrapper would add it to the backup because kopia wrappers's default right now is "if it's not explicitly mentioned, assume it still exists at the same location with the same items"

```suggestion
		// one with state deleted.
```

src/internal/connector/onedrive/collections.go
ashmrtn: +1 for less nesting

Small typo in the suggestion above: it should only copy excluded items if incrementals were enabled and the delta wasn't reset (above has extra negation in there)

it could probably be turned into something like
```go
if !c.ctrl.ToggleFeatures.EnableOneDriveDeltaIncrementals {
  continue
}

if !delta.Reset {
  maps.Copy(exclucdedItems, excluded)
  continue
}

// code for finding/making collections for deleted folders
```

ashmrtn (approved):

meain:
A problem with this approach is that is the delta is is not present or if it is empty, we [do not pass in](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/connector/onedrive/collections.go#L195) the `oldPaths`. Wondering how we should proceed for this. What is the possibility that we have only partial metadata? Will that be caught by [this](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/operations/manifests.go#L41)?

meain:
A problem with this approach is that is the delta is is not present or if it is empty, we [do not pass in](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/connector/onedrive/collections.go#L195) the `oldPaths`. Wondering how we should proceed for this. What is the possibility that we have only partial metadata? Will that be caught by [this](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/operations/manifests.go#L41)?

/cc  @ashmrtn

ashmrtn:
> A problem with this approach is that is the delta is is not present or if it is empty, we [do not pass in](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/connector/onedrive/collections.go#L195) the `oldPaths`. Wondering how we should proceed for this. What is the possibility that we have only partial metadata? Will that be caught by [this](https://github.com/alcionai/corso/blob/297d704c394c8158951ee90f215f27c52c84974f/src/internal/operations/manifests.go#L41)?
> 
> Edit: Also, not sure how it would interact with use checking if the collection was present in prev backup [here](https://github.com/alcionai/corso/blob/012ec5f8881762612eaaf508e5ba21a8863e702f/src/internal/connector/onedrive/collections.go#L501).

Ah, this is an excellent. It goes sort of along the lines of something I was thinking about, but I'm not sure if there's a good solution for it with the code right now but we'll probably need to come up with something soon. The basis for the problem is the following:
other components assume we're doing an incremental backup because from their perspective things seem fine. However, GraphConnector finds a problem and that changes whether or not the other components should use base backups

We've been working around this with the `DoNotMerge` flag, but the problem you pointed out is that we may not always be able to guarantee that we have enough info to send to kopia wrapper where the merging happens.

Just brainstorming, it seems like the situations where we won't have enough information to pass all the collections to kopia wrapper like we need to are:
* having delta token but not folder map
* over-aggressive data pruning that may result in losing the folder map because the delta token was missing or invalid

For now I think it's ok for us to assume that we only really have to deal with expired tokens. I think Corso will mark a backup as failed if we couldn't persist one of the files. This gives us some time to come up with a better solution to this situation

meain:
@ashmrtn You had approved the PR, but I have since made some relatively significant changes and so requested a re review.

src/internal/connector/onedrive/collections.go
```diff
@@ -344,14 +344,59 @@ func (c *Collections) Get(
...
+			continue
+		}
+
+		// Set all folders in previous backup but not in the current
+		// one with state deleted
+		modifiedPaths := map[string]struct{}{}
+		for _, p := range c.CollectionMap {
+			modifiedPaths[p.FullPath().String()] = struct{}{}
+		}
+
```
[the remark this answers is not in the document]
meain: Take the following scenario which can occur if the original folder was deleted, but we create a new folder with the same path.

```
Existing backup:
id:folder1 path:/my-folder

Delta (we have a new collection at the same path):
id:folder2 path:/my-folder
```

In this scenario, we will already have a anew collection for that path, but with different id in the `CollectionMap`.

PR #2504, opened 2023-02-14
wrap up clues and fault additions to exchange

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/attachment.go
```diff
@@ -95,24 +94,26 @@ func uploadLargeAttachment(
...
 ) error {
-	ab := attachmentBytes(attachment)
-	size := int64(len(ab))
+	var (
+		bs   = attachmentBytes(attachment)
+		size = int64(len(bs))
+	)
 
-	session, err := uploader.uploadSession(ctx, *attachment.GetName(), size)
+	session, err := uploader.uploadSession(ctx, ptr.Val(attachment.GetName()), size)
```
ashmrtn, line 102: not this PR per se, but do we need to check for an empty attachment name?

ashmrtn (approved):

src/internal/connector/exchange/attachment.go
ryanfkeepers: Good question: I assume not, based on the prior code?  But can't say for sure.  @dadams39 would know better.

PR #2506, opened 2023-02-14
add clues/fault to sharepoint collections

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/sharepoint/collection.go
+++ b/src/internal/connector/sharepoint/collection.go
@@ -3,14 +3,14 @@ package sharepoint
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io"
 	"time"
 
+	"github.com/alcionai/clues"
 	absser "github.com/microsoft/kiota-abstractions-go/serialization"
 	kw "github.com/microsoft/kiota-serialization-json-go"
-	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/discovery/api"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	sapi "github.com/alcionai/corso/src/internal/connector/sharepoint/api"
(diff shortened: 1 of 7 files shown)


src/internal/connector/sharepoint/collection.go
```diff
@@ -217,126 +221,117 @@ func (sc *Collection) retrieveLists(
...
 	betaService := sc.betaService
 	if betaService == nil {
-		return metrics, fmt.Errorf("beta service not found in collection")
+		return metrics, clues.New("beta service required").WithClues(ctx)
 	}
 
 	pages, err := sapi.GetSitePages(ctx, betaService, sc.fullPath.ResourceOwner(), sc.jobs)
 	if err != nil {
-		return metrics, errors.Wrap(err, sc.fullPath.ResourceOwner())
+		return metrics, err
```
meain, line 287: Should we retain info about `sc.fullPath.ResourceOwner()`?

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -108,25 +120,30 @@ func collectLists(
...
+
 		dir, err := path.Builder{}.Append(tuple.name).
 			ToDataLayerSharePointPath(
 				tenantID,
 				siteID,
 				path.ListsCategory,
 				false)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to create collection path for site: %s", siteID)
+			errs.Add(clues.Wrap(err, "creating list collection path").WithClues(ctx))
```
meain, line 146: Should we retain info about `siteID`? Unless I missed this getting added anywhere else.

meain (approved):
Approving to unblock

src/internal/connector/sharepoint/collection.go
ryanfkeepers: That data should already be present in the context.

src/internal/connector/sharepoint/data_collections.go
ryanfkeepers: Since siteID is equal to the resource owner, that info is also already present in the context.

PR #2507, opened 2023-02-14
add clues/fault to sharepoint api

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/sharepoint/api/pages.go
```diff
@@ -97,18 +97,21 @@ func FetchPages(ctx context.Context, bs *discover.BetaService, siteID string) ([
...
-			temp := Tuple{pid, pid}
-
-			if entry.GetName() != nil {
-				temp.Name = *entry.GetName()
+			var (
+				pid  = *entry.GetId()
+				temp = NameID{pid, pid}
+			)
+
+			name, ok := ptr.ValOK(entry.GetName())
```
meain, line 109: I'm not sure if `ValOK` improves the ergonomics here by much. I personally feel like previous approach of checking for nil was a bit more obvious to a person reading the code.

meain (approved):

src/internal/connector/sharepoint/api/pages.go
ryanfkeepers: Two way door; lets bring it up in weekly discussion.  Personally, I think it matches the map lookup format nicely.  Eg `v, ok := myMap[key]` vs `v, ok := ptr.ValOK(ptr)`.  (wouldn't it be nice to be able to optionally extend return values in normal funcs, for cases like this?)

Issue #1547, opened 2022-11-17
Optionally reduce new graph connector data enumeration


ryanfkeepers:
## Issue
Currently, graph connector enumerates either all sites or all users, depending on the data selection.  There are many cases where only a finite (often single) user or site needs to be retrieved.  In these cases, a new graph connector should _optionally_ accept a list of user or site ids, and only retrieve those specified resources.

vkamra:
#1994 is related

ashmrtn:
merged fix to reduce enumeration of users. Unassigning for now as other work has come up

gmatev:
Is this still relevant?

ashmrtn:
still relevant, we should change when enumeration of sites is done. Right now it's done when we create a GraphConnector instance for a SharePoint backup. This could lead to enumerating sites multiple times if multiple sites are being backed up

ryanfkeepers:
closing: work was done within other tickets

PR #2508, opened 2023-02-14
Consolidate logic to see if Exchange/OneDrive user exists

ashmrtn:
## Description

Use the check to see if a user has access to OneDrive/Exchange to also see if the user exists. This does not stop the population of GraphConnector users every time an instance is created, but does move us one step closer to that

Manually tested the following cases:
* backup "*"
* backup single existing user without error
* backup multiple existing users without error
* error when attempting to backup user that doesn't exist in the same domain
* error when attempting to backup user that doesn't exist in a different domain

Sample error output (progress disabled):
```text
No backups available
Error: 1 error occurred:
        * Failed to run Exchange backup for user foo@other.onmicrosoft.com: doing backup: producing backup data collections: resource owner [foo@other.onmicrosoft.com] not found within tenant
```

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #1547 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/discovery/discovery.go
```diff
@@ -43,6 +45,10 @@ func Users(ctx context.Context, ga getAller, errs *fault.Errors) ([]models.Usera
 func User(ctx context.Context, gwi getWithInfoer, userID string) (models.Userable, *api.UserInfo, error) {
 	u, err := gwi.GetByID(ctx, userID)
 	if err != nil {
+		if graph.IsErrUserNotFound(err) {
+			return nil, nil, fmt.Errorf("resource owner [%s] not found within tenant", userID)
```
ryanfkeepers, line 49: Nit: time to get in the habit of using clues
```suggestion
			return nil, nil, clues.Wrap(err, "resource owner not found within tenant").With("resource_owner", userID)
```

ryanfkeepers (approved):

src/internal/connector/discovery/discovery.go
ashmrtn: with the suggested change the output on the CLI when corso is run with a user that does exist is
```text
Error: 1 error occurred:
        * Failed to run Exchange backup for user foo@other.onmicrosoft.com: doing backup: producing backup data collections: resource owner not found within tenant: getting user: maximum retries or unretryable:  Request_ResourceNotFound Resource 'foo@other.onmicrosoft.com' does not exist or one of its queried reference-property objects are not present.: error status code received from the API
```

Is this something we're comfortable with?

ryanfkeepers: Me?  Yes.  Part of this migration is to have a centralized source for error interpretation; and that's still forthcoming.  But it does mean we'll improve in the future.  In the meantime, even though this is unsightly, it's understandable.

@gmatev might not share my opinion.

PR #2510, opened 2023-02-14
add fault.tracker for error additions

ryanfkeepers:
## Description

Realized we had a race condition: in an async
runtime it's possible for an errs.Err() to be
returned by multiple functions, even though that
Err() was only sourced by one of them.  The
addition of a tracker contains the returned
error into the scope of that func so that only
the error produced in the current iteration is
returned.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -108,7 +108,10 @@ func (c Users) GetAll(ctx context.Context, errs *fault.Errors) ([]models.Userabl
 		return nil, clues.Wrap(err, "creating users iterator").WithClues(ctx).With(graph.ErrData(err)...)
 	}
 
-	us := make([]models.Userable, 0)
+	var (
+		us = make([]models.Userable, 0)
+		et = errs.Tracker()
+	)
 
 	iterator := func(item any) bool {
 		if errs.Err() != nil {
@@ -117,7 +120,7 @@ func (c Users) GetAll(ctx context.Context, errs *fault.Errors) ([]models.Userabl
 
 		u, err := validateUser(item)
 		if err != nil {
-			errs.Add(clues.Wrap(err, "validating user").WithClues(ctx).With(graph.ErrData(err)...))
+			et.Add(clues.Wrap(err, "validating user").WithClues(ctx).With(graph.ErrData(err)...))
 		} else {
 			us = append(us, u)
 		}
@@ -129,7 +132,7 @@ func (c Users) GetAll(ctx context.Context, errs *fault.Errors) ([]models.Userabl
 		return nil, clues.Wrap(err, "iterating all users").WithClues(ctx).With(graph.ErrData(err)...)
 	}
 
-	return us, errs.Err()
+	return us, et.Err()
 }
 
 func (c Users) GetByID(ctx context.Context, userID string) (models.Userable, error) {
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -114,7 +114,7 @@ func (c Users) GetAll(ctx context.Context, errs *fault.Errors) ([]models.Userabl
 	)
 
 	iterator := func(item any) bool {
-		if errs.Err() != nil {
+		if et.Err() != nil {
 			return false
 		}
 
--- a/src/internal/connector/exchange/data_collections.go
+++ b/src/internal/connector/exchange/data_collections.go
@@ -178,6 +178,7 @@ func DataCollections(
 	var (
 		user        = selector.DiscreteOwner
 		collections = []data.BackupCollection{}
+		et          = errs.Tracker()
 	)
 
 	cdps, err := parseMetadataCollections(ctx, metadata, errs)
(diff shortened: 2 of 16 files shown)


src/internal/connector/discovery/api/users.go
```diff
@@ -108,16 +108,19 @@ func (c Users) GetAll(ctx context.Context, errs *fault.Errors) ([]models.Userabl
 		return nil, clues.Wrap(err, "creating users iterator").WithClues(ctx).With(graph.ErrData(err)...)
 	}
 
-	us := make([]models.Userable, 0)
+	var (
+		us = make([]models.Userable, 0)
+		et = errs.Tracker()
```
ashmrtn, line 113: I know it's kind of sixes, but thoughts on passing in a tracker vs. making a new one in a function?

src/internal/connector/exchange/data_collections.go
```diff
@@ -178,6 +178,7 @@ func DataCollections(
 	var (
 		user        = selector.DiscreteOwner
 		collections = []data.BackupCollection{}
+		et          = errs.Tracker()
```
ashmrtn, line 181: Greetings human, I come from outer space 👽

ashmrtn (approved):
may need a minor fix in tracker `Add` logic

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 ### Added
 
+- New Flag `--on-error` which controls recoverable error handling behavior.  Accepts `--on-error fail` to fail out immediately, and `--on-error continue` (default) to proceed with the operation.
```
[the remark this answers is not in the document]
ryanfkeepers: It is.  I've been informed that forcing- or even defaulting to- fail fast isn't an acceptable system state.  There are other ways around the issue in the short term (ie: forcing hard failures in necessary locations).  But at large, this is what we're sticking with.

src/internal/connector/discovery/api/users.go
ryanfkeepers: It's not sixes, and would nullify the purpose of tracker.  If it were sixes, we'd be able to stick with the fault.Errors at large.  But that will cause the code to confuse global with local errors.  Good for failing out as fast as feasibly possible (honestly not a real concern, imo), then in trade we risk mass duplication of errors, which I think will do more harm overall.

By forcing new tracker generation in each func, we can ensure that only local errors are returned, thus protecting us against duplicate return values.

src/internal/connector/exchange/data_collections.go
ryanfkeepers: ~[phone home](https://thumbs.gfycat.com/BothZanyAmurratsnake-max-1mb.gif)

src/pkg/fault/fault.go
```diff
@@ -124,3 +124,48 @@ func (e *Errors) addErr(err error) *Errors {
...
+
+func (e *tracker) Add(err error) {
+	if err == nil {
+		return
+	}
+
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	if e.errs.Err() == nil && e.errs.failFast {
```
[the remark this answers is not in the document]
ryanfkeepers: Ah, good catch.  Technically it's a change between whether current tracks the first or the last error.  But there is a bug, you're right.  It should be `e.current == nil && e.errs.failFast`.

CHANGELOG.md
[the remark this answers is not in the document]
ashmrtn: then before merge this we need to be certain we're forcing hard failures when we need to. Otherwise we could end up with backups being used as bases for incrementals when they shouldn't be

src/internal/connector/discovery/api/users.go
ashmrtn: ah, yea sorry, I didn't explain that clearly. I meant passing a tracker instance down 😅 

If it were a tracker instance passed in then maybe down the road we could do something like define an interface that both `fault.Errors` and the `fault.Tracker` implement and just pass that around. Though after saying that I'm not sure if that would make things more or less error prone... 🤔 idk I guess probably not the best idea now that I've given it more thought haha

for example:
```go
type bus interface {
  Err() error
  Add(err error)
}

func foo(errs bus) error {
  // do stuff
  if err != nil {
    errs.Add(err)
  }

  return errs.Err()
}
```

CHANGELOG.md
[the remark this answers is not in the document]
ryanfkeepers: I'm not aware of all those critical points of failure.  Could I get your help with that?

src/internal/connector/discovery/api/users.go
ryanfkeepers: > maybe down the road we could do something like define an interface that both fault.Errors and the fault.Tracker implement and just pass that around

I started things off with an interface like that.  Ultimately it turned out that the interface was almost exactly the fault.Errors interface, not a subset, and without a dependency to mock there wasn't any point to the abstraction.

It's important to note that a Tracker should be spawned for _each nested point of recoverable errors_.  Ie:
```go
func turtles(errs *fault.Errors) error {
  trk := errs.Tracker()
  if err := allTheWay(errs); err != nil {
    trk.Add(err)
  }
  return  trk.Err()
}

func allTheWay(errs *fault.Errors) error {
  trk := errs.Tracker()
  if err := down(errs); err != nil {
    trk.Add(err)
  }
  return  trk.Err()
}

func down(errs *fault.Errors) error {
  trk := errs.Tracker()
  if err := etc(errs); err != nil {
    trk.Add(err)
  }
  return  trk.Err()
}
```

PR #2512, opened 2023-02-14
add clues & fault to onedrive collections

ryanfkeepers:

- [ ] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spatialcurrent/go-lazy/pkg/lazy"
 
+	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
@@ -82,7 +83,7 @@ type Collection struct {
 type itemReaderFunc func(
 	hc *http.Client,
 	item models.DriveItemable,
-) (itemInfo details.ItemInfo, itemData io.ReadCloser, err error)
+) (details.ItemInfo, io.ReadCloser, error)
 
 // itemMetaReaderFunc returns a reader for the metadata of the
 // specified item
@@ -166,7 +167,7 @@ func (oc *Collection) Items(
 	ctx context.Context,
 	errs *fault.Errors, // TODO: currently unused while onedrive isn't up to date with clues/fault
 ) <-chan data.Stream {
-	go oc.populateItems(ctx)
+	go oc.populateItems(ctx, errs)
 	return oc.data
 }
 
@@ -240,23 +241,22 @@ func (od *Item) ModTime() time.Time {
 
 // populateItems iterates through items added to the collection
 // and uses the collection `itemReader` to read the item
-func (oc *Collection) populateItems(ctx context.Context) {
+func (oc *Collection) populateItems(ctx context.Context, errs *fault.Errors) {
 	var (
-		errs       error
 		byteCount  int64
 		itemsRead  int64
 		dirsRead   int64
 		itemsFound int64
 		dirsFound  int64
 		wg         sync.WaitGroup
-		m          sync.Mutex
+		et         = errs.Tracker()
 	)
 
 	// Retrieve the OneDrive folder path to set later in
 	// `details.OneDriveInfo`
 	parentPathString, err := path.GetDriveFolderPath(oc.folderPath)
 	if err != nil {
-		oc.reportAsCompleted(ctx, 0, 0, 0, err)
+		oc.reportAsCompleted(ctx, 0, 0, 0, clues.Wrap(err, "getting drive path").WithClues(ctx))
 		return
 	}
 
@@ -271,15 +271,8 @@ func (oc *Collection) populateItems(ctx context.Context) {
 	semaphoreCh := make(chan struct{}, urlPrefetchChannelBufferSize)
 	defer close(semaphoreCh)
 
-	errUpdater := func(id string, err error) {
-		m.Lock()
-		// TODO: Label(fault.LabelForceNoBackupCreation)
-		errs = support.WrapAndAppend(id, err, errs)
-		m.Unlock()
-	}
-
 	for _, item := range oc.driveItems {
-		if oc.ctrl.FailFast && errs != nil {
+		if et.Err() != nil {
 			break
 		}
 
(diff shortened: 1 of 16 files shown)


src/internal/operations/restore.go
```diff
@@ -354,16 +354,17 @@ func formatDetailsForRestoration(
...
 		fdsPaths = fds.Paths()
 		paths    = make([]path.Path, len(fdsPaths))
+		et       = errs.Tracker()
 	)
 
 	for i := range fdsPaths {
-		if errs.Err() != nil {
-			return nil, errs.Err()
+		if et.Err() != nil {
+			break
```
neha-Gupta1, line 362: Just to understand-  'break' is introduced to executed below block?
https://github.com/alcionai/corso/blob/88b5df728bfeda07933a936a5ab0e3485b208d6c/src/internal/operations/restore.go#L385

src/internal/connector/onedrive/collections.go
```diff
@@ -480,25 +482,29 @@ func (c *Collections) UpdateCollections(
...
-			err := clues.New("no parent reference").With("item_id", *item.GetId())
-			if item.GetName() != nil {
-				err = err.With("item_name", *item.GetName())
-			}
-
-			return err
+			et.Add(clues.New("item missing parent reference").
+				WithClues(ictx).
+				With("item_id", itemID, "item_name", ptr.Val(item.GetName())))
+			continue
```
neha-Gupta1, line 496: just curious to know about this pattern- why was the continue added, when we know in the next iteration it will go
inside
 ```
if et.Err() != nil {
			break
		}
```
block and break the for loop and return from function.

File: src/internal/operations/restore.go (hunk 1 above)
Comment (ryanfkeepers) [reply src/internal/operations/restore.go]: Not necessarily.  It's sixes either way.

File: src/internal/connector/onedrive/collections.go (hunk 2 above)
Comment (ryanfkeepers) [reply src/internal/connector/onedrive/collections.go:543]: It will not necessarily break the loop.  `et.Err()` is only non-nil if we're running `failFast: true`.  Otherwise, this loop will continue even though we added an error.

**Comment by ryanfkeepers** [comment]
@neha-Gupta1 @meain rebased to main and ready for final review, please and thank you!

File: src/internal/connector/onedrive/item_test.go [hunk 3]
```diff
@@ -53,30 +42,10 @@ func (suite *ItemIntegrationSuite) SetupSuite() {
...
-	// require.Greaterf(t, len(spDrives), 0, "site %s does not have a drive", suite.site)
-
-	// // Pick the first drive
-	// suite.siteDriveID = *spDrives[0].GetId()
-
+	suite.service = loadTestService(t)
 	suite.user = tester.SecondaryM365UserID(t)
 
-	pager, err := PagerForSource(OneDriveSource, suite, suite.user, nil)
+	pager, err := PagerForSource(OneDriveSource, suite.service, suite.user, nil)
```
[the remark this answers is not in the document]
ryanfkeepers: That change is not within scope of this PR.  We can save it for a different one.

src/internal/connector/onedrive/permission.go
```diff
@@ -103,9 +103,13 @@ func createRestoreFoldersWithPermissions(
...
-func getChildPermissions(childPermissions, parentPermissions []UserPermission) ([]UserPermission, []UserPermission) {
-	addedPermissions := []UserPermission{}
-	removedPermissions := []UserPermission{}
+func getChildPermissions(
+	childPermissions, parentPermissions []UserPermission,
+) ([]UserPermission, []UserPermission) {
+	var (
+		addedPermissions   = []UserPermission{}
+		removedPermissions = []UserPermission{}
+	)
```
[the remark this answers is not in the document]
ryanfkeepers: Should already be handled.

src/internal/connector/onedrive/item.go
```diff
@@ -35,7 +33,12 @@ func getDriveItem(
 	srv graph.Servicer,
 	driveID, itemID string,
 ) (models.DriveItemable, error) {
-	return srv.Client().DrivesById(driveID).ItemsById(itemID).Get(ctx, nil)
+	di, err := srv.Client().DrivesById(driveID).ItemsById(itemID).Get(ctx, nil)
+	if err != nil {
+		return nil, clues.Wrap(err, "getting item").WithClues(ctx).With(graph.ErrData(err)...)
```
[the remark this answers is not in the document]
ryanfkeepers: much later down the line, yes.

src/pkg/logger/logger.go
```diff
@@ -247,7 +247,14 @@ func Seed(ctx context.Context, lvl, logfile string) (context.Context, *zap.Sugar
...
 	if l == nil {
-		zsl := singleton(level, defaultLogLocation())
+		logfile := os.Getenv("CORSO_LOG_FILE")
+
+		if len(logfile) == 0 {
+			logfile = defaultLogLocation()
+		}
+
+		zsl := singleton(level, logfile)
+
```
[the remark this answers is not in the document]
ryanfkeepers: I don't think minor fixes like this necessitate their own PR.  I could, if wanted, make a branch of this with only this change in it, and then fold that change into this one.  But the difference, in this case, is so minor that it ends up as toil with little gain.

If this were a larger change, and especially if future PRs depended on it, i'd be in agreement.

src/internal/connector/onedrive/collection.go
```diff
@@ -271,37 +271,36 @@ func (oc *Collection) populateItems(ctx context.Context) {
...
 
-	errUpdater := func(id string, err error) {
-		m.Lock()
-		errs = support.WrapAndAppend(id, err, errs)
-		m.Unlock()
-	}
-
 	for _, item := range oc.driveItems {
-		if oc.ctrl.FailFast && errs != nil {
+		if et.Err() != nil {
```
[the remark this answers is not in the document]
ryanfkeepers: Resolved in sync: failFast is getting handled here.

src/internal/connector/onedrive/drive.go
```diff
@@ -89,30 +88,31 @@ func drives(
...
-				detailedError := support.ConnectorStackErrorTraceWrap(err, "").Error()
-				if strings.Contains(detailedError, userMysiteURLNotFound) ||
-					strings.Contains(detailedError, userMysiteURLNotFoundMsg) ||
-					strings.Contains(detailedError, userMysiteNotFound) ||
-					strings.Contains(detailedError, userMysiteNotFoundMsg) {
+				errMsg := support.ConnectorStackErrorTraceWrap(err, "").Error()
+				if strings.Contains(errMsg, userMysiteURLNotFound) ||
+					strings.Contains(errMsg, userMysiteURLNotFoundMsg) ||
+					strings.Contains(errMsg, userMysiteNotFound) ||
+					strings.Contains(errMsg, userMysiteNotFoundMsg) {
```
dadams39, line 105: Probably should make this a function

dadams39 (approved):
One nit

src/internal/connector/onedrive/drive.go
ryanfkeepers: actually, I did!  It's in one of the coming PRs.

PR #2516, opened 2023-02-15
Fix tight-loop when channel is closed

vkamra:
## Description

TODO: Still in Draft because this is not the complete fix.

CPU profiling indicated there is a tight loop in the `observe.MessageWithCompletion` method.

This is because the for-loop has no exit condition *and* channels that are closed do not block (https://dave.cheney.net/2013/04/30/curious-channels).

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/observe/observe.go
+++ b/src/internal/observe/observe.go
@@ -199,7 +199,10 @@ func MessageWithCompletion(
 			select {
 			case <-contxt.Done():
 				bar.SetTotal(-1, true)
-			case <-ci:
+			case _, ok := <-ci:
+				if !ok {
+					return
+				}
 				// We don't care whether the channel was signalled or closed
 				// Use either one as an indication that the bar is done
 				bar.SetTotal(-1, true)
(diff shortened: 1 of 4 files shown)


src/internal/observe/observe.go
```diff
@@ -199,7 +199,10 @@ func MessageWithCompletion(
 			select {
 			case <-contxt.Done():
 				bar.SetTotal(-1, true)
-			case <-ci:
+			case _, ok := <-ci:
+				if !ok {
```
vkamra, line 203: This should probably move to *after* we call `SetTotal`. 

Also need to handle this correctly for when the context is cancelled

vkamra:
Closing this one. #2523 has the fix

Issue #2517, opened 2023-02-15
Store root in metadata for backup


meain:
Store path of root in OneDrive backup metadata for paths.

Issue #2117, opened 2023-01-11
OneDrive delta token-based incrementals


ashmrtn:
Tracking issue for enabling delta token-based incrementals in OneDrive

- [x] #1535 
- [x] #2120
- [x] #2242
- [x] #2243
- [x] #2121
- [x] #2122
- [x] #2123
- [x] #2124
- [x] #2125
- [ ] #2333
- [x] #2459
- [x] #2532
- [x] #2559
- [x] #2754
- [x] #2759

---

Testing
- [x] #2264
- [x] unit tests with mocks
- [x] #2755
- [ ] ~#2772~ (tracked under https://github.com/alcionai/corso/issues/2409)

---

Refactoring
- [x] #2517

---

Bugs
- [x] #2558

---

Improvements

- [ ] ~#2739~ (moved to later)

ashmrtn:
need some integration tests at the operations level prior to merging

PR #2518, opened 2023-02-15
Deletion related fixes for OneDrive delta incrementals

meain:
## Description

- parentReference for deleted items will not have path.
- GC was not responding with deleted state
- Deleted items are not streamed by kopia

Borrows some changes from https://github.com/alcionai/corso/pull/2503/files as that might be delayed with on merge. Will rebase that if this gets merged fist.

- [ ] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* https://github.com/alcionai/corso/issues/2117
* fixes https://github.com/alcionai/corso/issues/2517

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -225,8 +225,11 @@ func (gc *GraphConnector) OneDriveDataCollections(
 		maps.Copy(allExcludes, excludes)
 	}
 
-	for range collections {
-		gc.incrementAwaitingMessages()
+	for _, c := range collections {
+		if c.State() != data.DeletedState {
```
meain, line 229: ref: https://github.com/alcionai/corso/blob/main/src/internal/connector/data_collections.go#L77

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1244,11 +1283,14 @@ func (suite *OneDriveCollectionsSuite) TestGet() {
 							driveItem("folder", "folder", driveBasePath1, "root", false, true, false),
 							driveItem("file", "file", driveBasePath1+"/folder", "folder", true, false, false),
 						},
-						deltaLink: &empty,
+						deltaLink: &empty, // not sure if this is a valid scenario
```
ashmrtn, line 1286: fair point, it's probably not. I think I probably added it cause there's a check in other code to only persist delta token stuff if `len(deltaLink) > 0` 😅

src/internal/connector/onedrive/collections.go
```diff
@@ -434,24 +436,53 @@ func (c *Collections) UpdateCollections(
...
+				fmt.Sprintf(rootDrivePattern, driveID),
+				c.tenant,
+				c.resourceOwner,
+				c.source,
+			)
+			if err != nil {
+				return err
+			}
+
+			updatePath(newPaths, *item.GetId(), rootPath.String())
```
ashmrtn, line 453: Just checking, this is called to add the folder to the paths map? If so, could you add a comment to this line saying that. It's a little confusing off the cuff because I wouldn't expect the path of the root folder to change

ashmrtn (approved):

src/internal/connector/onedrive/collections.go
meain: Yup, that is it. I'll add a comment.

src/internal/connector/onedrive/collections_test.go
meain: I'll keep the comment there just for reference.

PR #2523, opened 2023-02-15
centralize observe channel listener handling

ryanfkeepers:
## Description

To ensure that the observe channel handling
doesn't accidentally spawn unkillable routines,
adds a centralized channel listener func for
standard channel management in observe
progress bars.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/internal/observe/observe.go
+++ b/src/internal/observe/observe.go
@@ -433,15 +433,19 @@ func waitAndCloseBar(bar *mpb.Bar, log func()) func() {
 
 var nop = func() {}
 
-// listen runs a
-func listen(ctx context.Context, ci <-chan struct{}, onEnd, onInc func()) {
+// listen handles reading, and exiting, from a channel.  It assumes the
+// caller will run it inside a goroutine (ex: go listen(...)).
+// On context timeout or channel close, the loop exits.
+// onEnd() is called on both ctx.Done() and channel close.  onInc is
+// called on every channel read except when closing.
+func listen(ctx context.Context, ch <-chan struct{}, onEnd, onInc func()) {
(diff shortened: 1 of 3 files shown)


ntolia:
@ryanfkeepers Please add a changelog entry. This is a high-impact change for CLI users.

vkamra:
@ryanfkeepers - changes look good (pending CHANGELOG request from @ntolia )

What can we do in terms of a regression test here. Is there a modification to `TestObserveProgressWithChannelClosed` that would have caught this?

src/internal/observe/observe.go
```diff
@@ -469,7 +431,26 @@ func waitAndCloseBar(bar *mpb.Bar, log func()) func() {
 // other funcs
 // ---------------------------------------------------------------------------
 
-const Bullet = "∙"
+var nop = func() {}
+
+// listen runs a
```
vkamra, line 436: Update comment

src/internal/observe/observe.go
```diff
@@ -469,7 +431,26 @@ func waitAndCloseBar(bar *mpb.Bar, log func()) func() {
...
+func listen(ctx context.Context, ci <-chan struct{}, onEnd, onInc func()) {
```
vkamra, line 437: RE: my comment about a regression test. I think we should at minimum add a unit test for this. Kudos for pulling this logic out - makes it testable!

ryanfkeepers: done, tests have been added

Issue #2529, opened 2023-02-15
[Bug]: panic in Corso


ntolia:
### What happened?

This is likely invalid configuration on my end. Will narrow that down but Corso should not panic with invalid config.

### Corso Version?

Corso v0.3.0

### Where are you running Corso?

Docker, macOS, arm64

### Relevant log output

```shell
Connecting to repository:      …
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x87b054]

goroutine 1514 [running]:
github.com/kopia/kopia/internal/cache.(*contentCacheImpl).fetchBlobInternal(0x4005a720c0, {0x3e52318, 0x40008880c0}, {0x4000057d00, 0x36}, 0x400076e2e1?)
	/go/pkg/mod/github.com/kopia/kopia@v0.12.2-0.20230123092305-e5387cec0acb/internal/cache/content_cache.go:105 +0x74
github.com/kopia/kopia/internal/cache.(*contentCacheImpl).getContentFromFullBlob(0x4005a720c0, {0x3e52318, 0x40008880c0}, {0x4000057d00, 0x36}, 0x20, 0x294, 0x0?)
	/go/pkg/mod/github.com/kopia/kopia@v0.12.2-0.20230123092305-e5387cec0acb/internal/cache/content_cache.go:76 +0x144
github.com/kopia/kopia/internal/cache.(*contentCacheImpl).GetContent(0x3e59d50?, {0x3e52318?, 0x40008880c0?}, {0x4005de9da0?, 0x106d?}, {0x4000057d00?, 0x4000696a58?}, 0x3e59d50?, 0x4005b9efa0?, 0x0?)
	/go/pkg/mod/github.com/kopia/kopia@v0.12.2-0.20230123092305-e5387cec0acb/internal/cache/content_cache.go:56 +0x48
github.com/kopia/kopia/repo/content.(*SharedManager).getContentDataReadLocked(0x400017e000, {0x3e52318, 0x40008880c0}, 0x0, {0x3e59d50, 0x4005b9efa0}, 0x3e52318?)
	/go/pkg/mod/github.com/kopia/kopia@v0.12.2-0.20230123092305-e5387cec0acb/repo/content/content_manager_lock_free.go:117 +0x29c
[...]

PR #2530, opened 2023-02-16
standardize panic recovery

ryanfkeepers:

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #2529

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/common/crash/crash.go
+++ b/src/internal/common/crash/crash.go
@@ -1,0 +1,54 @@
+package crash
+
+import (
+	"context"
+	"fmt"
+	"runtime"
+	"runtime/debug"
+
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/pkg/logger"
+)
+
+// Recovery provides a deferrable func that can be called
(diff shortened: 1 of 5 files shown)


meain (approved):
Approving to unblock

src/internal/common/crash/crash.go
```diff
@@ -0,0 +1,54 @@
...
+// to recover from, and log context about, crashes.
+// If an error is returned, then a panic recovery occurred.
+//
+// Call it as follows:
+//
+//	defer func() {
+//			if crErr := crash.Recovery(ctx, recover()); crErr != nil {
+//				err = crErr // err needs to be a named return variable
+//			}
+//		}()
```
meain, line 24: tiny tiny nit:
```suggestion
//	defer func() {
//		if crErr := crash.Recovery(ctx, recover()); crErr != nil {
//			err = crErr // err needs to be a named return variable
//		}
//	}()
```

meain (approved):
Forgot to add review comment

PR #2533, opened 2023-02-16
OneDrive test framework

ashmrtn:
## Description

Major changes:
* OneDrive files now store the lookup key in the first few bytes of the
  file and then the file data. This allows the test framework to
  continue working even when we switch to using item IDs as names
* Create a framework to create the data necessary for testing. This
  framework uses something akin to the builder pattern and automatically
  adds files to the collection if they should appear in that version.
  For example, trying to add a metadata file to a backup version that
  didn't have metadata files will result in a noop. Adding files will
  also add the metadata to the aux file set

Long term this will hopefully make ensuring compatibility between OneDrive
versions easier because only a few places in the code need to be updated
to support another version. It also allows easy ranges over different
versions via for-loop

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

## Issue(s)

* #1535

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector_onedrive_test.go
+++ b/src/internal/connector/graph_connector_onedrive_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/alcionai/corso/src/internal/connector/graph"
@@ -17,11 +18,24 @@ import (
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
-func getTestMetaJSON(t *testing.T, user string, roles []string) []byte {
+func getMetadata(fileName, user string, roles []string) onedrive.Metadata {
+	if len(user) == 0 || len(roles) == 0 {
+		return onedrive.Metadata{FileName: fileName}
+	}
+
 	id := base64.StdEncoding.EncodeToString([]byte(user + strings.Join(roles, "+")))
-	testMeta := onedrive.Metadata{Permissions: []onedrive.UserPermission{
-		{ID: id, Roles: roles, Email: user},
-	}}
+	testMeta := onedrive.Metadata{
+		FileName: fileName,
+		Permissions: []onedrive.UserPermission{
+			{ID: id, Roles: roles, Email: user},
+		},
+	}
+
+	return testMeta
+}
+
+func getTestMetaJSON(t *testing.T, fileName, user string, roles []string) []byte {
+	testMeta := getMetadata(fileName, user, roles)
 
 	testMetaJSON, err := json.Marshal(testMeta)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -439,185 +531,220 @@ func (suite *GraphConnectorOneDriveIntegrationSuite) TestPermissionsRestoreAndBa
...
+						RestorePermissions: true,
+						ToggleFeatures:     control.Toggles{EnablePermissionsBackup: true},
+					},
+				)
+			})
+		}
 	}
 }
 
+// TODO(ashmrtn): What this test is supposed to do needs investigated. It
```
meain, line 746: Woops, this one is on me. This was supposed to test the situation where user has not provided `--restore-permissions`, but as you mentioned, it does not look like that is what it is doing.

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -139,13 +185,150 @@ var (
...
+			name+onedrive.DataFileSuffix,
+			fileData))
+
+		metadata := onedriveMetadata(
+			c.t,
+			"",
+			name+onedrive.MetaFileSuffix,
+			user,
+			roles)
+		c.items = append(c.items, metadata)
```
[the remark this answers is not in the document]
ashmrtn: for the moment we need to add to both `items` and `aux`. `items` is currently the only one used to check results after a backup is done. We can rework that at some point, but I'd like to do that in a different PR

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -139,13 +185,150 @@ var (
...
+		c.aux = append(c.aux, metadata)
+
+	default:
```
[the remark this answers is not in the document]
ashmrtn: I was going to have this for when we start saving file names in metadata files. I should remove it from this PR though since it's not actually relevant. I probably added it by mistake 😅

src/internal/connector/graph_connector_onedrive_test.go
meain: I will pick up this one in another PR and fix it  in case you don't wanna get to it in this one.

meain (approved):

src/internal/connector/graph_connector_onedrive_test.go
ashmrtn: fixing in another PR sounds great. If you don't mind picking it up I would appreciate it. If you don't want to though I can take a stab at it

meain:
Thanks for reworking this btw. This is gonna make life much easier :D

Issue #1761, opened 2022-12-09
Emit 'Corso Start' as the CLI starts on any operation


gmatev:
Goal would be to have an event that allow to determine if users are running into issues. `Corso Start` without a corresponding subsequent event would indicate a problem.

vkamra:
@neha-Gupta1 - Follow up on this after we do #1762. This would involve firing the Corso Start event for each CLI operation.

neha-Gupta1:
Do we want to add this code to 'CLI' package or the `pkg` package? 
Adding it to CLI will provide us events only when actions done from Corso's CLI.

neha-Gupta1:
Completed in https://github.com/alcionai/corso/pull/2534

PR #2534, opened 2023-02-16
add corso start event at start of cli command

neha-Gupta1:
## Description

Trigger an event at start of every CLI command. 

## Type of change

- [x] :sunflower: Feature

## Issue(s)
* https://github.com/alcionai/corso/issues/1761

## Test Plan

- [ ] :muscle: Manual


ryanfkeepers:
This is a lot of boilerplate, and won't necessarily scale or maintain well.  Perhaps we could centralize it in the pre-run command, somehow? https://github.com/alcionai/corso/blob/main/src/cli/cli.go#L40

neha-Gupta1:
> This is a lot of boilerplate, and won't necessarily scale or maintain well. Perhaps we could centralize it in the pre-run command, somehow? https://github.com/alcionai/corso/blob/main/src/cli/cli.go#L40

@ryanfkeepers 
Agreed. This is adding more boilerplate. I can add this code to pre-run. But then we would be required to fetch config info using `viper` in the prerun function too. If its okay to do that I will go forward and implement the way you suggested.

src/cli/cli.go
```diff
@@ -55,8 +56,22 @@ func preRun(cc *cobra.Command, args []string) error {
...
+	}
+
+	avoidTheseCommands = []string{
+		"help", "init",
+	}
+
+	if !slices.Contains(avoidTheseCommands, cc.Use) {
+		s, acct, repoid, err := config.GetStorageRepoAndAccount(ctx)
+		if err != nil {
+			log.Debug(ctx, "Error while triggering event for ", cc.Use)
```
ryanfkeepers, line 71: This should exit by returning an error, not log.  It means we're unable to retrieve the configuration, which makes the whole command un-runnable.

Also, `log.Debug()` doesn't take in a `ctx`.  Just a message.  The context is already embedded by `logger.Ctx(ctx)` above.  `log.Debug(ctx)` Will use the context struct as the log message.

src/cli/config/config.go
```diff
@@ -178,13 +180,21 @@ func Read(ctx context.Context) error {
 
 // WriteRepoConfig currently just persists corso config to the config file
 // It does not check for conflicts or existing data.
-func WriteRepoConfig(ctx context.Context, s3Config storage.S3Config, m365Config account.M365Config) error {
-	return writeRepoConfigWithViper(GetViper(ctx), s3Config, m365Config)
+func WriteRepoConfig(ctx context.Context,
```
ryanfkeepers, line 183: Style standards: the first parameter should take its own line, not stay inlined.
```suggestion
func WriteRepoConfig(
    ctx context.Context,
```

src/cli/utils/utils.go
```diff
@@ -136,3 +140,21 @@ func splitFoldersIntoContainsAndPrefix(folders []string) ([]string, []string) {
...
+	ctx context.Context,
+	s storage.Storage,
+	tenID string,
+	data map[string]any,
+	repoID string,
+	opts control.Options,
+) {
+	bus, err := events.NewBus(ctx, s, tenID, opts)
+	if err != nil {
+		logger.Ctx(ctx).Debugw("analytics event failure", "err", err)
```
ryanfkeepers, line 155: Lets do info here, since most users don't run with debug logging.
```suggestion
		logger.Ctx(ctx).Infow("analytics event failure", "err", err)
```

src/cli/config/config.go
```diff
@@ -178,13 +180,21 @@ func Read(ctx context.Context) error {
...
+	s3Config storage.S3Config,
+	m365Config account.M365Config,
+	repo repository.Repository,
```
ryanfkeepers, line 186: Since we only need the repo ID, this should only accept the string we need, not the full repository.  Always try to keep the func footprint as small as possible.
```suggestion
	repoID string,
```

src/cli/config/config.go
```diff
@@ -209,14 +220,24 @@ func writeRepoConfigWithViper(vpr *viper.Viper, s3Config storage.S3Config, m365C
 	return nil
 }
 
+// GetStorageAndAccount creates a storage and account instance by mediating all the possible
+// data sources (config file, env vars, flag overrides) and the config file.
+func GetStorageRepoAndAccount(ctx context.Context) (storage.Storage, account.Account, string, error) {
```
ryanfkeepers, line 225: Instead of a new func, can we introduce a struct like this?
```go
type configRepoDetails struct {
    Storage storage.Storage
    Account account.Account
    RepoID string
}
```
Then return the struct in `GetStorageAndAccount` (maybe rename it to GetConfigRepoDetails).  That way we maintain flexibility and extensibility without additional funcs or 4+ value returns.

Note that I think we should _return_ this struct when reading the config, but not _require_ it when writing the config with `WriteRepoConfig`

src/cli/config/config.go
```diff
@@ -209,14 +220,24 @@ func writeRepoConfigWithViper(vpr *viper.Viper, s3Config storage.S3Config, m365C
...
+
 // GetStorageAndAccount creates a storage and account instance by mediating all the possible
 // data sources (config file, env vars, flag overrides) and the config file.
 func GetStorageAndAccount(
 	ctx context.Context,
 	readFromFile bool,
 	overrides map[string]string,
-) (storage.Storage, account.Account, error) {
-	return getStorageAndAccountWithViper(GetViper(ctx), readFromFile, overrides)
+) (storage.Storage,
```
ryanfkeepers, line 235: Style standards: similar to the func declaration, multi-line return values should give each param a new line.
```suggestion
) (
    storage.Storage,
```

src/cli/repo/s3.go
```diff
@@ -137,7 +140,7 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
 
 	Infof(ctx, "Initialized a S3 repository within bucket %s.", s3Cfg.Bucket)
 
-	if err = config.WriteRepoConfig(ctx, s3Cfg, m365); err != nil {
+	if err = config.WriteRepoConfig(ctx, s3Cfg, m365, r); err != nil {
```
ryanfkeepers, line 143: Following on the other comment about only taking in a string instead of a repository.
```suggestion
	if err = config.WriteRepoConfig(ctx, s3Cfg, m365, r.GetID()); err != nil {
```

ryanfkeepers (commented):
Looking good overall.  The centralization is a much improved pattern, and will help long term maintenance a lot.  Thank you for the pivot. 

Lots of nits and suggestions for improvements here.  Give it a second pass and I'm sure we'll get it merged tomorrow.

src/cli/utils/utils.go
neha-Gupta1: The function sending event - https://github.com/alcionai/corso/blob/a5ecc8dd7ea1295a15b9fe15ac4042669f6f6583/src/internal/events/events.go#L150
uses  `Debug` so it would be on similar line if kept debug.

ryanfkeepers: Ah: we originally did that because we weren't logging to a file. Now that we're logging to a file, I think it's fine for both to be infos.

src/cli/config/config.go
```diff
@@ -39,6 +38,13 @@ var (
 	displayDefaultFP   = filepath.Join("$HOME", ".corso.toml")
 )
 
+// RepoDetails config read using viper
```
ryanfkeepers, line 41:
```suggestion
// RepoDetails holds the repository configuration retrieved from
// the .corso.toml configuration file.
```

src/cli/repo/s3.go
```diff
@@ -109,22 +109,31 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
...
 	}
 
-	s3Cfg, err := s.S3Config()
+	// SendStartCorsoEvent uses distict ID as tenant ID because repoID is still not generated
+	utils.SendStartCorsoEvent(
+		ctx,
+		cfg.Storage,
+		cfg.Account.ID(),
+		map[string]any{"command": "init repo"},
+		cfg.Account.ID(),
```
ryanfkeepers, line 123: shouldn't this be the repo ID?
```suggestion
		cfg.RepoID,
```

neha-Gupta1: For `repo init` command the repoID will be generated after few initialisation. And since we want this event to be the first thing happening, had to make this command as exception. And had to use Account ID as the unique ID here.

ryanfkeepers: Hmm.  Okay, lets move ahead with this.  I"ll talk to Georgi and see what value will work best for his end of things.  Thanks!

PR #2544, opened 2023-02-16
Print api response in logs for 4xx errors

meain:
## Description

Print api response in logs for 4xx errors from GC. More info for things like 400 Bad Request could be useful.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph/service.go
```diff
@@ -307,9 +307,16 @@ func (handler *LoggingMiddleware) Intercept(
...
+		} else if resp.StatusCode/100 == 4 {
+			respDump, _ := httputil.DumpResponse(resp, true)
+			logger.Ctx(ctx).Infow(
+				"graph api error",
+				"status", resp.Status,
+				"method", req.Method,
+				"url", req.URL,
+				"response", string(respDump),
+			)
+		} else if resp.StatusCode/100 != 2 {
```
ryanfkeepers, line 319: What is this solving that the above conditional (toggling all api logging) doesn't?

src/internal/connector/graph/service.go
```diff
@@ -307,9 +307,16 @@ func (handler *LoggingMiddleware) Intercept(
...
-
-		if resp.StatusCode != http.StatusTooManyRequests && (resp.StatusCode/100) != 2 {
+		} else if resp.StatusCode/100 == 4 {
+			respDump, _ := httputil.DumpResponse(resp, true)
+			logger.Ctx(ctx).Infow(
+				"graph api error",
+				"status", resp.Status,
+				"method", req.Method,
+				"url", req.URL,
+				"response", string(respDump),
```
ryanfkeepers, line 317: Preference is to go the simpler route and dump the response on all non-2xx, non-throttling responses.

ryanfkeepers (approved):
Approving to unblock, but I don't see the value here when we already have the additional logging flags for getting the same information.  Keeping this to a flag helps maintain less chatty logs for standard runs.

src/internal/connector/graph/service.go
meain: This was done primarily to make sure we get some extra information about why requests failed in ark without having to enable debug logging which can be really noisy. The primary status code that I think logging resopnse could help is just 400 (not necessarily 4xx) as having the response might help us figure out what is causing a failure like in the case of possible "restricted directories". Let me know what you think.

ryanfkeepers: mm, I still think we'll get all this information out of the new structured errors.  But don't let me stop you.

If we're only looking for 400, we should explicitly catch that; not all 4xx.
```
else if resp.Status == http.StatusBadRequest {
```

PR #2545, opened 2023-02-16
More fine grained error messages for retries

meain:
## Description

This should help in figuring out more retry failures.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #<issue>

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/retry_middleware.go
+++ b/src/internal/connector/graph/retry_middleware.go
@@ -6,7 +6,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/alcionai/corso/src/internal/connector/support"
+	"github.com/alcionai/clues"
 	backoff "github.com/cenkalti/backoff/v4"
 	khttp "github.com/microsoft/kiota-http-go"
 )
@@ -50,7 +50,10 @@ func (middleware RetryHandler) retryRequest(
 
 		response, err := pipeline.Next(req, middlewareIndex)
 		if err != nil && !IsErrTimeout(err) {
-			return response, support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
+			return response, clues.Stack(err).
+				WithClues(ctx).
+				With("retry_count", executionCount).
+				With(ErrData(err)...)
 		}
 
 		return middleware.retryRequest(ctx,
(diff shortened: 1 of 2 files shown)


src/internal/connector/graph/retry_middleware.go
```diff
@@ -50,7 +51,11 @@ func (middleware RetryHandler) retryRequest(
 
 		response, err := pipeline.Next(req, middlewareIndex)
 		if err != nil && !IsErrTimeout(err) {
-			return response, support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
+			return response,
+				support.ConnectorStackErrorTraceWrap(
+					err,
+					fmt.Sprintf("unretryable error after %d retries", executionCount),
```
ryanfkeepers, line 57: I was reviewing an error message the other day and realized that this message will get injected, unhelpfully, within many different error strings.  I think we'll end with better error messages by not wrapping a message here at all.  Otherwise we end up with nearly every error saying, `foo: bar: baz: unretryable or failed: qux`, where `qux` normally contains details about the rest failure already.
```suggestion
				clues.Stack(err).WithClues(ctx).With("retry_count", executionCount).WithAll(graph.ErrData(err)...)
```

src/internal/connector/graph/service.go
```diff
@@ -327,7 +327,7 @@ func (middleware RetryHandler) Intercept(
 
 	response, err := pipeline.Next(req, middlewareIndex)
 	if err != nil && !IsErrTimeout(err) {
-		return response, support.ConnectorStackErrorTraceWrap(err, "maximum retries or unretryable")
+		return response, support.ConnectorStackErrorTraceWrap(err, "unretryable error")
```
ryanfkeepers, line 330: Same here.
```suggestion
		return response, clues.Stack(err).WithClues(ctx).WithAll(graph.ErrData(err)...)
```

ryanfkeepers (approved):
Approving to unblock, would like to see the suggested changes.

src/internal/connector/graph/retry_middleware.go
meain: This makes sense, I wanted just expose the retry count was unnecessarily convoluting things.

Issue #2550, opened 2023-02-16
Incremental backups fail if base didn't backup anything


ashmrtn:
In some situations it appears that we can find no items to backup for some service/data category. This leads to a situation where incremental backups attempt to use this empty backup as a base and fail when they cannot find the prefix folder for the given data category in kopia (e.x. `/tenant/exchange/user/contacts` for a contacts backup)

## Resolution
The solution is multi-part due to the current state of things
1. handle errors from kopia if the directory prefix is missing in the base (this allows us to move forward)
2. have corso always persist at least the directory prefix. This allows us to make some assumptions about backups in the future and we may be able to enforce stricter behavior for (1) once we have some reasonable assurance that semi-recent backups will definitely have the directory prefix even if it has no content

ashmrtn:
mitigation merged. PR in-flight to ensure these "prefix" directories always exist in a backup

ashmrtn:
pending merge of #2553

ashmrtn:
complete since #2929 merged

PR #2551, opened 2023-02-16
Add log messages about metadata saved/used

ashmrtn:
## Description

Record the number of entries in the saved metadata. This could help catch odd situations like not finding anything to backup.

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2550

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/service_iterators.go
```diff
@@ -185,6 +190,13 @@ func filterContainersAndFillCollections(
 		graph.NewMetadataEntry(graph.PreviousPathFileName, currPaths),
 	}
 
+	logger.Ctx(ctx).Infow(
+		"persisted metadata",
+		"num_paths_entries",
+		len(currPaths),
+		"num_deltas_entries",
+		len(deltaURLs))
```
ryanfkeepers, line 198: Nit: can I push for keeping key-value pairs on the same line?  Kinda reads like a map that way, which is easy to grok.
```suggestion
		"num_paths_entries", len(currPaths),
		"num_deltas_entries", len(deltaURLs))
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -50,6 +50,11 @@ func filterContainersAndFillCollections(
 		tombstones = makeTombstones(dps)
 	)
 
+	logger.Ctx(ctx).Infow(
+		"previous metadata",
+		"num_complete_items",
+		len(dps))
```
ryanfkeepers, line 56: Nit: messages should always refer to the current position in processing.  `logger.Ctx(ctx)` wraps `clues.In(ctx)`, so this log will naturally contain a lot more structured data than just the attached count.  That'll make it easy to lose the correlation between the log message and the structured data.
```suggestion
		"filling collections",
		"metadata_count",
		len(dps))
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -185,6 +190,13 @@ func filterContainersAndFillCollections(
 		graph.NewMetadataEntry(graph.PreviousPathFileName, currPaths),
 	}
 
+	logger.Ctx(ctx).Infow(
+		"persisted metadata",
```
ryanfkeepers, line 194: nit: since we haven't done any persistence yet.
```suggestion
		"adding metadata collections",
```

ryanfkeepers (approved):

PR #2552, opened 2023-02-16
log and continue if kopia can't find expected base snapshot directory

ashmrtn:
## Description

This allows corso to continue making a snapshot even if the expected directory in the base is not found. The fact that the entry wasn't found will be logged

This should be safe so long as backups that had any sort of error, including "non-catastrophic" errors are marked as failed. Marking those as failed ensures that corso is using a matched set of metadata and data in the base snapshot rather than a partial snapshot

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2550

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -908,6 +908,11 @@ func inflateBaseTree(
 
 		ent, err := snapshotfs.GetNestedEntry(ctx, dir, pathElems)
 		if err != nil {
+			if isErrEntryNotFound(err) {
+				logger.Ctx(ctx).Infow("base snapshot missing subtree", "error", err)
+				continue
+			}
+
 			return errors.Wrapf(err, "snapshot %s getting subtree root", snap.ID)
 		}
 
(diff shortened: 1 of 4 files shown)


ryanfkeepers (commented):
> This should be safe so long as backups that had any sort of error, including "non-catastrophic" errors are marked as failed.

I don't think this is an expectation we can support.  From my understanding, best effort support behavior is table stakes for us.  @vkamra should pitch in here.

ashmrtn:
>I don't think this is an expectation we can support

Long-term I agree. I would see this as more of an intermediate step. Right now there's the possibility of backups that are missing these prefix directories and we need to provide some path forward for them if they are chosen as bases. We can use this to accomplish that

Combined with #2553, we can ensure that we'll be able to handle bases missing prefix directories and new backups will include prefix directories. After some amount of time, we should be able to tighten this up and error out instead of continuing because any (semi-)recent versions of corso will be creating snapshots that meet the assumptions we've made

ryanfkeepers (approved):

PR #2553, opened 2023-02-16
Always make collections for prefix directories

ashmrtn:
## Description

Use empty collections to ensure the prefix directories are available in
kopia no matter what user data we find.

This helps ensure we have some set of things we can assume about
non-failed backups including:
  1. They have a valid snapshot ID for data
  2. They have a valid snapshot ID for backup details

Eventually, this will allow us to say that backups using a base that
doesn't have a prefix directory is invalid in some way

SharePoint tests are disabled because lists and pages ignore the none
target in selectors

- [ ] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2550

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/collections.go
+++ b/src/internal/connector/graph/collections.go
@@ -1,0 +1,78 @@
+package graph
+
+import (
+	"context"
+
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/connector/support"
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+var _ data.BackupCollection = emptyCollection{}
+
+type emptyCollection struct {
+	p  path.Path
+	su support.StatusUpdater
+}
+
+func (c emptyCollection) Items(ctx context.Context, errs *fault.Errors) <-chan data.Stream {
+	res := make(chan data.Stream)
+	close(res)
+
+	s := support.CreateStatus(ctx, support.Backup, 0, support.CollectionMetrics{}, nil, "")
+	c.su(s)
+
+	return res
+}
+
+func (c emptyCollection) FullPath() path.Path {
+	return c.p
+}
+
+func (c emptyCollection) PreviousPath() path.Path {
+	return c.p
+}
+
+func (c emptyCollection) State() data.CollectionState {
+	// This assumes we won't change the prefix path. Could probably use MovedState
+	// as well if we do need to change things around.
+	return data.NotMovedState
+}
+
+func (c emptyCollection) DoNotMergeItems() bool {
+	return false
+}
+
+func BaseCollections(
+	tenant, user string,
+	service path.ServiceType,
+	categories map[path.CategoryType]struct{},
+	su support.StatusUpdater,
+	errs *fault.Errors,
+) ([]data.BackupCollection, *fault.Errors) {
+	res := []data.BackupCollection{}
+
+	for cat := range categories {
+		p, err := path.Builder{}.Append("tmp").ToDataLayerPath(tenant, user, service, cat, false)
+		if err != nil {
+			// Shouldn't happen.
+			errs.Fail(clues.Wrap(err, "making path").WithAll("service", service, "category", cat))
+			continue
+		}
+
+		p, err = p.Dir()
+		if err != nil {
+			// Shouldn't happen.
+			errs.Fail(clues.Wrap(err, "getting base prefix").WithAll("serivce", service, "category", cat))
+			continue
+		}
+
+		// Pop off the last path element because we just want the prefix.
+		res = append(res, emptyCollection{p: p, su: su})
+	}
+
+	return res, errs
+}
--- a/src/internal/connector/graph/collections.go
+++ b/src/internal/connector/graph/collections.go
@@ -82,5 +82,9 @@ func BaseCollections(
 		res = append(res, emptyCollection{p: p, su: su})
 	}
 
+	if len(errs) > 0 {
+		return res, clues.Stack(errs...)
+	}
+
 	return res, nil
 }
--- a/src/internal/connector/graph/collections.go
+++ b/src/internal/connector/graph/collections.go
@@ -63,7 +63,7 @@ func BaseCollections(
 			// Shouldn't happen.
 			errs = append(
 				errs,
-				clues.Wrap(err, "making path").WithAll("service", service, "category", cat))
+				clues.Wrap(err, "making path").With("service", service, "category", cat))
 
 			continue
 		}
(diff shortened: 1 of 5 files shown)


src/internal/connector/graph/collections.go
```diff
@@ -0,0 +1,78 @@
...
+	su support.StatusUpdater,
+	errs *fault.Errors,
+) ([]data.BackupCollection, *fault.Errors) {
+	res := []data.BackupCollection{}
+
+	for cat := range categories {
+		p, err := path.Builder{}.Append("tmp").ToDataLayerPath(tenant, user, service, cat, false)
+		if err != nil {
+			// Shouldn't happen.
+			errs.Fail(clues.Wrap(err, "making path").WithAll("service", service, "category", cat))
```
ryanfkeepers, line 62: This isn't an appropriate use of fault.  If you want to hard fail, then you should return an error without accepting a fault.Errors parameter.  This standard helps save us from exchanging the error return values all throughout the code with fault.Errors, and likewise being confused in all cases about whether we should err or Fail().  If we need to ensure failFast, then that needs to be controlled from the top down, not bottom up.

ryanfkeepers (changes requested):
Fault handling needs refactoring.

src/internal/connector/graph/collections.go
ashmrtn: I think there's maybe a few things going on here that have gotten a bit tangled up with everything going on
1. we need a way to coalesce errors. I could use multierror here but it sounds like we were moving away from that because it didn't have the layout we wanted. Ideally we'd return all possible errors we could here so that we could make the most progress debugging in one go
2. we need a way to specify "this makes stuff fail" and "this error is alright to ignore" when it's not something that was passed in top-down. This is especially important if we have errors from a loop or similar
3. we need clear semantics around whether error are returned/checked for in the propagated `fault.Errors()` or if we're returning `error`

(3) has been a bit confusing to me since we seem to want all errors to end up in `fault.Error` so we can use that to see if we've failed. In parallel though we're also returning `error` and I'm never sure if we should be checking that value to see if a function has some issue we can't recover from or if we should check `fault.Err()` for it. Since there's multiple functions that have both `fault.Error` and a returned `error` it would be good to clarify this behavior

ashmrtn: actually, I take some of that back, the clues package suffices for (1). It's been a long week lol

ryanfkeepers: >  Ideally we'd return all possible errors we could here so that we could make the most progress debugging in one go

That's what `errs.Add()` is for.

> we need a way to specify "this makes stuff fail" and "this error is alright to ignore" when it's not something that was passed in top-down. This is especially important if we have errors from a loop or similar

I have ideas around this for v2 handling.  Initially, the best way to handle this right now is to simply return an error if you want to force a failure.  I've yet to see any loop where we need to handle both cases simultaneously, and this change doesn't strike me as a deviation.

> we need clear semantics around whether error are returned/checked for in the propagated fault.Errors() or if we're returning error

I've tried to make these semantics as clear as possible in the fault package examples.  If there's still confusion after reading that document, I'm more than happy to review where it fell short and make amendments there.

ashmrtn: >That's what errs.Add() is for.

>Initially, the best way to handle this right now is to simply return an error if you want to force a failure.

The above two together seem like an anti-pattern. We're saying that we need to add the error to both `fault.Errors` and return it from the function? Isn't that going to cause all of the code we have to have to check both the return value and `errs.Err()` to see if we need to exit?

We still seem to be lacking a way to both coalesce errors and return them to cause a failure all in one go. Instead we've split that across two things (returned value and in/out param `errs`) which seems like we'll get bugs from forgetting to check one or the other

ryanfkeepers: Good questions.

> We're saying that we need to add the error to both fault.Errors and return it from the function?

Pretty much.  Consider an analog when using multierrs:
```go
func do() error {
  var errs error
  for range stuff {
    if failfast && errs != nil {
      break
    }
    if err := foo(); err != nil {
      errs = multierr.Wrap(errs, err)
    }  
  }
  return errs
}
```
This is the same behavior: aggregate the looped errors, break if we need to fail out early, return any aggregated errors.  The primary difference between multierr and fault is that the multierror version isn't able to return nil if all errors were recoverable.

> Isn't that going to cause all of the code we have to have to check both the return value and errs.Err() to see if we need to exit?

The goal is to only check `errs.Err()` when working with a loop that has recoverable failures (ie: when we need to check on failing fast or recovering), or as the final check in the process (when we return a fault.Errors).  In all other cases, I think it's important to maintain standard `error` handling.  That way we stay as close as possible to the native behavior of the language.

> We still seem to be lacking a way to both coalesce errors and return them to cause a failure all in one go.

I don't believe we are missing that control.  Errors get coalesced with `Add(err)`.  Functions that call Add() should return `errs.Err()`.  The fault library handles the failFast decision about whether an Add() counts as a Fail().  What's left to the local implementation is whether the error is recoverable or not in the first place (ie, whether it should call Add() or just return an error).  

Going back to the multierr example, the fault contemporary is:
```go
func do(errs *fault.Errors) error {
  for range stuff {
    // fault handles failFast checks
    if errs.Err() != nil {
      break
    }
    if err := foo(); err != nil {
      // this is a recoverable error, therefore Add()
      // if non-recoverable, we'd expect `return err`
      errs.Add(err)
    }  
  }
  // return an err, iff fault considered this a Fail() case. 
  return errs.Err()
}
```
Upstream from here, one of two cases must be true: the returned error is non-nil, and we treat that like any other error return; or error is nil, and processing either recovered (callers can check the aggregated errors at the end of the run) or no error occurred.

ryanfkeepers: This PR is also relevant to this conversation: https://github.com/alcionai/corso/pull/2510

ryanfkeepers: resolving following sync chat

src/internal/connector/graph/collections.go
```diff
@@ -51,28 +51,36 @@ func BaseCollections(
...
+
 			continue
 		}
 
 		// Pop off the last path element because we just want the prefix.
 		res = append(res, emptyCollection{p: p, su: su})
 	}
 
-	return res, errs
+	return res, clues.Stack(errs...)
```
ryanfkeepers, line 85: this will always return a non-nil value, unfortunately.  So you'll want to check for errors separately.
```suggestion
        if len(errs) > 0 {
            return clues.Stack(errs...)
        }
        
	return res, nil
```

We'll need an `ErrorOrNil()` func we want to collapse that in the future.  Not my favorite pattern, but it's doable.

ryanfkeepers (approved):

ashmrtn:
replaced by #2929

Issue #2548, opened 2023-02-16
Extract and report errors recorded in kopia snapshot


ashmrtn:
Kopia does not directly expose some errors that occur when creating a snapshot. Instead, these errors are recorded in the snapshot itself and can be accessed later. For example, ignored errors count (these can include items that had reader errors during upload etc.)

The actual content of these errors should be extracted from kopia, recorded in corso if needed, logged, and/or displayed to the user. This will aid in debugging and giving the user the information they need if things go wrong

ashmrtn:
logged in #2598

PR #2555, opened 2023-02-17
Hook into error callback in kopia

ashmrtn:
## Description

Using this over extracting errors from the snapshot allows us to see all errors. The errors that each directory tracks are truncated when the directory is serialized for persistence if there're too many.

With the current `fault.Errors` behavior this cases the backup to fail if there's any errors
kopia ignored. This is consistent with behavior in the backup op that checks for any errors.

Unclear what the CLI output and log messages will be for these errors as they are grouped
into all other errors reported through the fault package

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2548

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper_test.go
+++ b/src/internal/kopia/wrapper_test.go
@@ -483,7 +483,7 @@ func (suite *KopiaIntegrationSuite) TestBackupCollections_ReaderError() {
 		tags,
 		true,
 		fault.New(true))
-	require.NoError(t, err)
+	require.Error(t, err)
 
 	assert.Equal(t, 0, stats.ErrorCount)
 	assert.Equal(t, 5, stats.TotalFileCount)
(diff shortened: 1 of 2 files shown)


src/internal/kopia/wrapper_test.go
```diff
@@ -483,7 +483,7 @@ func (suite *KopiaIntegrationSuite) TestBackupCollections_ReaderError() {
 		tags,
 		true,
 		fault.New(true))
-	require.NoError(t, err)
+	require.Error(t, err)
```
ashmrtn, line 486: updated because the fault is set to fail fast

vkamra (approved):

PR #2568, opened 2023-02-18
write test logs to stdout

ryanfkeepers:
stderr logging writes all our logs to the terminal (ie, what shows up in github workflows) instead of the logging file artifact.  On failures with thousands of lines of logs, this makes searching for the error source nearly impossible.  This change ensures we have clean outputs in github workflows, and logs stored in artifact files for extensive review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,7 @@ jobs:
           CORSO_M365_TEST_USER_ID: ${{ secrets.CORSO_M365_TEST_USER_ID }}
           CORSO_SECONDARY_M365_TEST_USER_ID: ${{ secrets.CORSO_SECONDARY_M365_TEST_USER_ID }}
           CORSO_PASSPHRASE: ${{ secrets.INTEGRATION_TEST_CORSO_PASSPHRASE }}
-          CORSO_LOG_FILE: stderr
+          CORSO_LOG_FILE: ./src/testlog/testlogging.log
           LOG_GRAPH_REQUESTS: true
         run: |
           set -euo pipefail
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -208,7 +208,7 @@ jobs:
           CORSO_M365_TEST_USER_ID: ${{ secrets.CORSO_M365_TEST_USER_ID }}
           CORSO_SECONDARY_M365_TEST_USER_ID: ${{ secrets.CORSO_SECONDARY_M365_TEST_USER_ID }}
           CORSO_PASSPHRASE: ${{ secrets.INTEGRATION_TEST_CORSO_PASSPHRASE }}
-          CORSO_LOG_FILE: stderr
```
meain, line 211: I'm not sure if this changes how the output is formatted in CI. I think we still log out everything if it fails. See line 221.

ryanfkeepers: Whoops, right you are. switched that to a file.

.github/workflows/load_test.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           CORSO_LOAD_TESTS: true 
-          CORSO_LOG_FILE: stderr
+          CORSO_LOG_FILE: ./src/testlog/testlogging.log
```
meain, line 56: The path that gets uploaded here is different.
```suggestion
          CORSO_LOG_FILE: ./src/test_results/testlogging.log
```

ryanfkeepers: arg.  That's what I get for updating things during a meeting.  Good catch.

meain (approved):

PR #2591, opened 2023-02-21
remove stats.Errs

ryanfkeepers:
## Description

Now that fault errors is in place, we don't need
the operation stats errs.

- [x] :no_entry: No 

## Type of change

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/stats/stats.go
```diff
@@ -15,12 +15,6 @@ type ReadWrites struct {
 	ResourceOwners int   `json:"resourceOwners,omitempty"`
 }
 
-// Errs tracks the aggregation of errors that occurred during a process.
-type Errs struct {
```
ashmrtn, line 19: I'm all for removing dead code but are we sure this is safe to do? If this struct is getting serialized into the backup model then won't removing it break deserialization? It seems like either we'd need a custom deserializer or we'd need to keep the field around if this is deserialized

ryanfkeepers: Extra elements don't break serialization.  They do get sliced out.  But we're not actually utilizing this value for anything other than the error counts on backup list (if that's even serializing properly in the first place; i haven't checked specifically, but errors in general aren't serializing correctly).  Technically a breaking change, but a benign one.  I do need to add a Changelog for it, though.

ashmrtn: That seems... a bit scary. I also forgot about that behavior :/ Do we assume we'll never want to be more strict about unmarshalling (there's a way to have it error on unknown fields)?

ashmrtn (approved):

src/internal/stats/stats.go
ryanfkeepers: Error marshalling/unmarshalling is getting solved in #2708.

ashmrtn: ah, not just unmarshalling for errors. But for any type. Golangs json package has an option to throw errors if it encounters unknown fields

PR #2596, opened 2023-02-21
GC: OneDrive/SharePoint WebURL Fix

### LINKED ISSUE #1747: SharePoint Libraries: missing sharepoint ItemInfo
## Issue
Not all items in a sharepoint library backup contain a `SharepointIds` drive value.  This slices out the driveID and webURL from the details.SharePointInfo.  See line: https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/item.go#L140

## Acceptance Criteria
- [ ] All SharePointInfo structs must have a populated `Owner` and `WebURL` value.  If discovery is not available (ie: the data isn't present in the driveItemable), backfill these values by passing down the data from the backup creation request.

### Carried forward from 2064
SharePoint's drive permissions can require certain permissions prior to the upload of documents or directories to a protected location. By default, `_catalog` directories do not allow for upload.

Ensure 
Add more metadata to backup to identify restricted directories

dadams39:
## Description
Changes: 
- Use of `ptr` package to ensure safe dereferencing of pointers
- Update to query to ensure that `sharepointIds` are part of the response
- Alternate method created to obtain originating sharepoint site url 

- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [x] :bug: Bugfix

## Issue(s)

* closes #1747<issue>

## Test Plan

- [x] :muscle: Manual


dadams39:
Tested within our tenant on every site. There were no incidents of a SharePointInfo being created without the resource owner or web URL not being populated within the tenant.

src/internal/connector/onedrive/item.go
```diff
@@ -344,3 +341,36 @@ func driveItemWriter(
...
+
+	r := adtl[desiredKey]
+	point, ok := r.(*string)
+
+	if !ok {
+		return url
+	}
+
+	value := ptr.Val(point)
+	if len(value) == 0 {
```
ryanfkeepers, line 368:
```suggestion
	value, ok := ptr.ValOK(point)
	if !ok {
```

src/internal/connector/onedrive/item.go
```diff
@@ -344,3 +341,36 @@ func driveItemWriter(
...
 }
+
+// constructWebURL helper function for recreating the webURL
+// for the originating SharePoint site. Uses additional data map
+// from a models.DriveItemable that possesses a downloadURL within the map.
+// Returns "" if map nil or key is not present.
+func constructWebURL(adtl map[string]any) string {
+	var (
+		desiredKey = "@microsoft.graph.downloadUrl"
+		del        = `/_layouts`
```
ryanfkeepers, line 352: nit: `del` reads like `delete`
```suggestion
		sep        = `/_layouts`
```

ryanfkeepers (approved):

src/internal/connector/onedrive/item.go
ashmrtn: can we be certain that the stuff we're working with will always have `_layouts` in it? This seems like a fragile part of the code that could break -- possibly silently -- without warning :/

Are there other options we have for retrieving this information? For example, is it available higher up the stack and can it be passed down?

dadams39: Great question! 

This is not the fundamental way web URLs are retrieved. This is an alternative. The primary way is found here:

https://github.com/alcionai/corso/blob/7b650ebe8de4641a67472615a6a56c3f68973355/src/internal/connector/onedrive/item.go#L288

With the other fix, the `sharepointIds`, the URL should not be empty. In the peculiar corner case that it is, the download URL possesses the information.

ashmrtn: hm, alright. We should keep an eye on this in the future though

PR #2603, opened 2023-02-21
force no backup model persistence

ryanfkeepers:
## Description

Under certain conditions (failure to get an item,
or failure to merge details) we want to prevent
the persistence of a Corso backup model.  This
allows a snapshot to exist, aiding assisted
incremental backups.  But prevents the use of
the backup in delta incrementals and in restores.

- [x] :no_entry: No 

## Type of change

- [x] :bug: Bugfix

## Issue(s)

* #1970

## Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -451,6 +451,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 func (oc *Collection) reportAsCompleted(ctx context.Context, itemsFound, itemsRead int, byteCount int64, errs error) {
 	close(oc.data)
 
+	// TODO: add Label(fault.LabelForceNoBackupCreation) to errs
 	status := support.CreateStatus(ctx, support.Backup,
 		1, // num folders (always 1)
 		support.CollectionMetrics{
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -273,6 +273,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 	errUpdater := func(id string, err error) {
 		m.Lock()
+		// TODO: Label(fault.LabelForceNoBackupCreation)
 		errs = support.WrapAndAppend(id, err, errs)
 		m.Unlock()
 	}
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -410,13 +410,16 @@ func (w Wrapper) RestoreMultipleItems(
 
 		ds, err := getItemStream(ctx, itemPath, snapshotRoot, bcounter)
 		if err != nil {
-			et.Add(err)
+			et.Add(clues.Stack(err).Label(fault.LabelForceNoBackupCreation))
 			continue
 		}
 
 		parentPath, err := itemPath.Dir()
 		if err != nil {
-			et.Add(clues.Wrap(err, "making directory collection").WithClues(ctx))
+			et.Add(clues.Wrap(err, "making directory collection").
+				WithClues(ctx).
+				Label(fault.LabelForceNoBackupCreation))
+
 			continue
 		}
 
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -126,24 +126,33 @@ func (c Contacts) EnumerateContainers(
 			With("options_fields", fields)
 	}
 
+	et := errs.Tracker()
 	builder := service.Client().
 		UsersById(userID).
 		ContactFoldersById(baseDirID).
 		ChildFolders()
 
 	for {
+		if et.Err() != nil {
+			break
+		}
+
 		resp, err := builder.Get(ctx, ofcf)
 		if err != nil {
 			return clues.Stack(err).WithClues(ctx).With(graph.ErrData(err)...)
 		}
 
 		for _, fold := range resp.GetValue() {
-			if errs.Err() != nil {
-				return errs.Err()
+			if et.Err() != nil {
+				break
 			}
 
 			if err := checkIDAndName(fold); err != nil {
-				errs.Add(clues.Stack(err).WithClues(ctx).With(graph.ErrData(err)...))
+				et.Add(clues.Stack(err).
+					WithClues(ctx).
+					With(graph.ErrData(err)...).
+					Label(fault.LabelForceNoBackupCreation))
+
 				continue
 			}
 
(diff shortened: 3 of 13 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -410,13 +410,16 @@ func (w Wrapper) RestoreMultipleItems(
 
 		ds, err := getItemStream(ctx, itemPath, snapshotRoot, bcounter)
 		if err != nil {
-			et.Add(err)
+			et.Add(clues.Stack(err).Label(fault.LabelForceNoBackupCreation))
```
ashmrtn, line 413: I don't think this is strictly necessary since it's mostly on the restore path

src/internal/connector/onedrive/collection.go
```diff
@@ -273,6 +273,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 
 	errUpdater := func(id string, err error) {
 		m.Lock()
+		// TODO: Label(fault.LabelForceNoBackupCreation)
```
ashmrtn, line 276: this needs implemented?

ashmrtn (changes requested):
also need to check for errors when enumerating items (sorry, I guess this is the thing we missed when we chatted. I think we were thinking of different things at the time

src/internal/kopia/wrapper.go
ryanfkeepers: StreamStore calls this to retrieve the backup details that it streams back when merging backups.  see: https://github.com/alcionai/corso/blob/main/src/internal/streamstore/streamstore.go#L115

src/internal/connector/onedrive/collection.go
ryanfkeepers: We're not passing in errs at this point.  That gets implemented in... two PRs from now?

ashmrtn: doesn't not having some check for onedrive make it unsafe though? I know we don't have incrementals for onedrive yet, but we're already persisting enough metadata for it to technically use a previous backup as a base

src/internal/kopia/wrapper.go
ashmrtn: mmmmm I get that, but is there a way to catch the error closer to the details merge code? It seems odd to place it here and then ignore it for the restore path :/

ryanfkeepers: I don't think it's odd.  That's the nature of labels.  Consumers don't need to concern themselves with the complete set of values.  In the meantime, the labeling i/o matches here.  If we caught it upstream from the adder then we'd have to investigate added errors and append labels post-hoc.  Much messier, and much more error prone.

src/internal/connector/onedrive/collection.go
ryanfkeepers: It is unsafe.  But no more unsafe than we already are with a lack of error tracking in the first place.  We don't have a mechanism for review here until fault gets put into place.  In the meantime, GC support will mark the error as a failure, which should cascade into an early exit all the same.

ashmrtn: >But no more unsafe than we already are with a lack of error tracking in the first place.

We're fail-fast right now so errors don't cause a backup model to be created avoiding the issue

>In the meantime, GC support will mark the error as a failure, which should cascade into an early exit all the same.

Where is this happening? Is GC supposed to be adding a label for that? Or are we assuming that there will be some error in the fault.Errors that causes the failure? If we switch from fail-fast will GC still be generating that failure?

src/internal/kopia/wrapper.go
ashmrtn: eh sure, kind of an implementation detail I suppose

src/internal/connector/onedrive/collection.go
ryanfkeepers: > Where is this happening?

As per the state of this PR: 
* error aggregation with multierrs: https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/collection.go#L374
* error reporting in statusUpdater: https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/collection.go#L460
* error return on non-zero gc.status errors: https://github.com/alcionai/corso/blob/main/src/internal/operations/backup.go#L279
* readErr exits before backing up the model: https://github.com/alcionai/corso/blob/main/src/internal/operations/backup.go#L625

ashmrtn: mkay, thanks for the references. Can we add a TODO or a comment to the final part to ensure we don't remove that prior to getting OneDrive fault handling setup?

Also, do we have some way to test this is actually doing what we think it should be if there's a failure? Would be really good to see it in action prior to merging

ashmrtn (approved):

src/internal/connector/onedrive/collection.go
ryanfkeepers: Done.  And no testing that I can easily manufacture.  We're kinda stuck here between complete e2e (no way to force targeted failures) and needing to be able to mock every layer (which we don't have coverage for yet).  I could hack something up for a manual show, if wanted.  But that's not exactly resilient testing.

ashmrtn: if something manual isn't super difficult I think that would be worth it to ensure things are at least working on some call path

ryanfkeepers: @ashmrtn 

Added a forced `et.Add(clues.New("forced failure").Label(fault.LabelForceNoBackupCreation))`.  

```
17:47:00.47	INFO	completed backup; conditional error forcing exit without model persistence	{"error": "forced failure", "errorVerbose": "forced failure\n\t/Users/keepers/go/src/github.com/alcionai/corso/src/internal/connector/exchange/api/mail.go:219", "results": {"bytesRead":930866,"bytesUploaded":949300,"itemsRead":41,"itemsWritten":22,"resourceOwners":1,"startedAt":"2023-02-21T17:46:19.23712-07:00","completedAt":"2023-02-21T17:47:00.475452-07:00","backupID":"696b60f3-6b16-4d33-99c2-5f0f2b950dc3"}}

No backups available
Error: 1 error occurred:
	* Failed to run Exchange backup for user Rfinders@8qzvrj.onmicrosoft.com: forced backup: forced failure
```

ashmrtn: excellent thanks!

ryanfkeepers: You're very welcome!

PR #2606, opened 2023-02-22
don't handle meta files in details

ryanfkeepers:
## Description

Remove the display and interaction with .meta
files from details entries.

## Type of change

- [x] :bug: Bugfix

## Test Plan

- [x] :zap: Unit test


src/internal/connector/onedrive/collection.go
```diff
@@ -415,6 +415,7 @@ func (oc *Collection) populateItems(ctx context.Context) {
 					ItemName:   itemInfo.OneDrive.ItemName,
 					DriveName:  itemInfo.OneDrive.DriveName,
 					ItemType:   itemInfo.OneDrive.ItemType,
+					IsMeta:     true,
```
meain, line 418: The filenames are pretty stable, they will always have one of `.data`, `.meta` or `.dirmeta` suffixes. Should we just make use of that instead of adding new filed in the driveInfo? This might be useful long term though.

src/pkg/backup/details/details.go
```diff
@@ -89,21 +92,48 @@ func (dm DetailsModel) Paths() []string {
...
-		res = append(res, &dm.Entries[i])
+		res = append(res, &ent)
 	}
 
 	return res
 }
 
+// SliceMetaFiles returns a copy of the Details with all of the
+// .meta files sliced out of the entries.
+func (dm DetailsModel) SliceMetaFiles() DetailsModel {
```
meain, line 113: nit: `SliceMetaFiles` seems like an odd name. How about `GetDataFiles` or `FilterDataFiles`.

src/pkg/backup/details/details.go
```diff
@@ -89,21 +92,48 @@ func (dm DetailsModel) Paths() []string {
...
+		if !ent.isMetaFile() {
+			d2.Entries = append(d2.Entries, ent)
+		}
+	}
+
+	return d2
+}
+
+func (de DetailsEntry) isMetaFile() bool {
+	// TODO: this is hazardous.  legitimate files could be suffixed .meta.
```
meain, line 128: We need to do this based on the backup version. For backup version 0, there is never any meta files. All files, even the ones that end with `.meta` are data files and for all other versions, if the suffix is `.meta` or `.dirmeta`, they are metadata files.

src/pkg/backup/details/details_test.go
```diff
@@ -224,6 +225,63 @@ var pathItemsTable = []struct {
 		expectRepoRefs:     []string{"abcde", "12345"},
 		expectLocationRefs: []string{"locationref", "locationref2"},
 	},
+	{
+		name: "multiple entries with meta file",
+		ents: []DetailsEntry{
+			{
+				RepoRef:     "abcde",
```
meain, line 232: It might not affect this test, but we should always have a suffix of `.data`, `.meta` or `.dirmeta` for the locationRef for any newer backups.

meain (changes requested):

src/internal/connector/onedrive/collection.go
ryanfkeepers: Stability isn't the concern.  Cross-contamination is my worry.  If we back-up a file named `foo.meta` that has nothing to do with our meta scheme, it could cause problems.  Sure, right now it'll be called `foo.meta.data`, but if we end up changing that naming scheme in the future, this might still catch us by surprise.

And/or we can drop it when we move to storing files by ID.

src/pkg/backup/details/details_test.go
ryanfkeepers: Let's not cross wires, and come back to that in a different PR.

src/pkg/backup/details/details.go
meain: My concern is that if we have files from older backup with a suffix of `.meta` in the original file name, it can end up getting treated as a metadata file even though it is not.

src/pkg/backup/details/details_test.go
meain: What I meant by newer backups is that any backups done with the latest version of the code.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Show owner information when doing backup list in json format
 
+### Fixed
+- .Meta files no longer appear in the backup details.
```
meain, line 14:
```suggestion
- .meta files no longer appear in the backup details.
```

src/pkg/backup/details/details.go
```diff
@@ -89,21 +92,48 @@ func (dm DetailsModel) Paths() []string {
...
+	return strings.HasSuffix(de.RepoRef, ".meta") ||
```
meain, line 129: We should also check for dirmeta.

src/pkg/backup/details/details_test.go
ryanfkeepers: Understood.  What I mean is that this test can be brought up to date in a different PR.

Also, now that I'm re-reading your comment, `locationRef` should _not_ contain those values.  It's entirely a folder reference, and contains no file data.  It should be unaware of both data and meta categorization.

src/pkg/backup/details/details.go
ryanfkeepers: Agreed, thus the todo.  I can add a known issue to the changelog, if that helps.  Also, not saying v2 needs to be a long follow.  You should probably tail this PR with the change immediately.  But I think this is safe enough for the immediate case, and it's too late for me to push that next part.

src/pkg/backup/details/details.go
```diff
@@ -35,10 +36,12 @@ type DetailsModel struct {
 // Print writes the DetailModel Entries to StdOut, in the format
 // requested by the caller.
 func (dm DetailsModel) PrintEntries(ctx context.Context) {
+	sl := dm.FilterMetaFiles()
```
meain, line 39: We should do this filter in OneDrive before we sent the `DetailsModel` to `PrintEntries` as it will otherwise affect others like Exchange which does not even have the idea of `.meta` files.

meain:
I have a follow up to this https://github.com/alcionai/corso/pull/2610 which fixes how we mark an item as a metadata item.

src/pkg/repository/repository.go
```diff
@@ -358,6 +361,20 @@ func (r repository) BackupDetails(
 		return nil, nil, errs.Fail(err)
 	}
 
+	// Retroactively fill in isMeta information for items in older
+	// backup versions without that info
+	// version.Restore2 introduces the IsMeta flag, so only v1 needs a check.
+	if b.Version == version.Restore1DataAndMetaFiles {
```
meain, line 367: Current version in prod is version 2, the change being it has name in meta. You should be able to see this if you check the version number of the current backups. This is why I had > and < checks.

meain: Also, if you see `Restore2IsMetaMarker` is current set to 3. So we need both 1 and 2 to run this.

src/internal/version/restore.go
```diff
@@ -0,0 +1,20 @@
...
+
+const (
+	// Restore1DataAndMetaFiles is the corso backup format version
+	// in which we split from storing just the data to storing both
+	// the data and metadata in two files.
+	Restore1DataAndMetaFiles = 1
+	// Restore2IsMetaMarker is a small improvement on
+	// VersionWithDataAndMetaFiles, but has a marker IsMeta which
+	// specifies if the file is a meta file or a data file.
+	Restore2IsMetaMarker = 3
```
meain, line 13:
```suggestion
	Restore3IsMetaMarker = 3
```

`RestoreCurrNameInMeta` is technically 2, but holding off for now.

meain:
@ashmrtn If you do have time, do take a look at the versioning related changes here.

meain (approved):
Seems to be good now. I would suggest waiting for Ash to take a look at it as well. There is a typecheck issue in link related to the file moves. You might wanna take a look at that as well. Approving it to unblock.

src/internal/version/restore.go
```diff
@@ -0,0 +1,20 @@
+package version
+
+import "math"
+
+const (
```
ashmrtn, line 5: I like having the comments on these. A few small non-blocking nits though:
* comments don't note that they actually apply to OneDrive stuff
* splitting restore and backup versions is a bit confusing. They're actually related to each other but having them different like this makes me think they refer to different things

ashmrtn (approved):

src/internal/version/restore.go
ryanfkeepers: > comments don't note that they actually apply to OneDrive stuff

Added

> splitting restore and backup versions is a bit confusing. They're actually related to each other but having them different like this makes me think they refer to different things

They were already split in the code, this is just migration.  We can make a tech-debt follow-up to unify them if needed.

ryanfkeepers: Ah, as soon as I looked at the code I grokked what you were saying.  Sorry, my brain is all a scatter today.  I've made a pass at the locations and naming scheme for the labels.  Should hopefully fit what you're looking for.

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -237,7 +232,7 @@ func (c *onedriveCollection) withPermissions(
 ) *onedriveCollection {
 	// These versions didn't store permissions for the folder or didn't store them
 	// in the folder's collection.
-	if c.backupVersion < 3 {
+	if c.backupVersion < version.OneDriveXIncludesPermissions {
```
ryanfkeepers, line 235: This took me waaaaaaaaaaaaaay too long to find.

PR #2618, opened 2023-02-22
OneDrive backup cleanup

ashmrtn:
## Description

General cleanup/consolidation of code for getting OneDrive collections during backup. Few minor changes to behavior:
* always return a collection for the root folder even if it's empty
* add items to the exclude list only after they've been added to a collection. This gives a better chance of preserving the item even if we encounter an error parsing paths or something. If there's a failure, the old version of the item is most likely to be preserved (this varies based on whether the item is returned in multiple delta results during the backup)

- [x] :no_entry: No 

## Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

## Issue(s)

* #2447

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/collections.go
```diff
@@ -69,7 +69,7 @@ type Collections struct {
 
 	// collectionMap allows lookup of the data.BackupCollection
 	// for a OneDrive folder
-	CollectionMap map[string]data.BackupCollection
+	CollectionMap map[string]*Collection
```
meain, line 72: This was a really good idea 👍🏼

src/internal/connector/onedrive/collections.go
```diff
@@ -687,81 +687,52 @@ func (c *Collections) UpdateCollections(
...
-					c.service,
-					c.statusUpdater,
-					c.source,
-					c.ctrl,
-					invalidPrevDelta,
-				)
-
-				c.CollectionMap[collectionID] = col
-				c.NumContainers++
+				return clues.New("item seen before parent folder").With("item_id", itemID)
```
meain, line 695: I think we could add the parent's id as well here.
```suggestion
				return clues.New("item seen before parent folder").With("item_id", itemID, "collection_id", collectionID)
```

src/internal/connector/onedrive/collections.go
```diff
@@ -687,81 +687,52 @@ func (c *Collections) UpdateCollections(
...
+				// Always add a file to the excluded list. The file may have been
+				// renamed/moved/modified, so we still have to drop the
+				// original one and download a fresh copy.
+				excluded[itemID+DataFileSuffix] = struct{}{}
+				excluded[itemID+MetaFileSuffix] = struct{}{}
+			}
+
 		default:
-			return errors.Errorf("item type not supported. item name : %s", *item.GetName())
+			return clues.New("item type not supported.").With("item_id", itemID)
```
meain, line 735:
```suggestion
			return clues.New("item type not supported").With("item_id", itemID)
```

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1236,7 +1258,9 @@ func (suite *OneDriveCollectionsSuite) TestGet() {
 			prevFolderPaths: map[string]map[string]string{
 				driveID1: {"root": rootFolderPath1},
 			},
-			expectedCollections: map[string]map[data.CollectionState][]string{},
+			expectedCollections: map[string]map[data.CollectionState][]string{
+				expectedPath1(""): {data.NotMovedState: {}},
```
meain, line 1262: I know it is this way in some of the existing code too, but there is actually `rootFolderPath1` which can be used for root. There are a few instances of this.

```suggestion
				rootFolderPath1: {data.NotMovedState: {}},
```

meain (approved):
Few nits, but LGTM.

ashmrtn:
@meain sorry to bother you for this, but can you take another look? I've switched some stuff around because some SharePoint tests were unhappy we were always returning the root when the selector didn't match it and working around it in that specific test felt a bit hacky

The new changes factor some functions out of `UpdateCollections`, adds the root collection in the same block that other collections are added, and handles deletions at the start of `UpdateCollections`

src/internal/connector/onedrive/collections.go
```diff
@@ -502,277 +606,151 @@ func (c *Collections) UpdateCollections(
...
+		if item.GetDeleted() != nil {
+			if err := c.handleDelete(
+				itemID,
+				driveID,
+				oldPaths,
+				newPaths,
+				isFolder,
+				excluded,
+			); err != nil {
+				return clues.Wrap(err, "marking item deleted")
```
meain, line 622:
```suggestion
				return clues.Wrap(err, "marking item deleted").With("item_id", itemID)
```

src/internal/connector/onedrive/collections.go
```diff
@@ -469,23 +464,132 @@ func updateCollectionPaths(
...
+		}
+
+		collectionPath, err = collectionPath.Append(name, false)
+		if err != nil {
+			return nil, clues.Wrap(err, "making non-root folder path").With("item_id", itemID)
+		}
+	}
+
+	return collectionPath, nil
+}
```
meain, line 591:
```suggestion
func (c *Collections) getCollectionPath(
	driveID string,
	item models.DriveItemable,
) (path.Path, error) {
	var (
		collectionPathStr string
		itemID            = ptr.Val(item.GetId())
		isRoot            = item.GetRoot() != nil
	)

	if isRoot {
		collectionPathStr = fmt.Sprintf(rootDrivePattern, driveID)
	} else {
		if item.GetParentReference() == nil ||
			item.GetParentReference().GetPath() == nil {
			err := clues.New("no parent reference").With("item_id", itemID)
			if item.GetName() != nil {
				err = err.With("item_name", ptr.Val(item.GetName()))
			}

			return nil, err
		}

		collectionPathStr = ptr.Val(item.GetParentReference().GetPath())
	}

	collectionPath, err := GetCanonicalPath(
		collectionPathStr,
		c.tenant,
		c.resourceOwner,
		c.source,
	)
	if err != nil {
		return nil, clues.Wrap(err, "making item path").
			With("item_id", ptr.Val(item.GetId()))
	}

	if isRoot {
		return collectionPath, nil
	}

	name := ptr.Val(item.GetName())
	if len(name) == 0 {
		return nil, clues.New("folder with empty name").With("item_id", itemID)
	}

	collectionPath, err = collectionPath.Append(name, false)
	if err != nil {
		return nil, clues.Wrap(err, "making non-root folder path").With("item_id", itemID)
	}

	return collectionPath, nil
}
```

src/internal/connector/onedrive/collections.go
```diff
@@ -502,277 +606,151 @@ func (c *Collections) UpdateCollections(
...
-			itemPath, err = collectionPath.Append(name, !isFolder)
-			if err != nil {
-				return err
-			}
-		}
-
 		// Skip items that don't match the folder selectors we were given.
-		if shouldSkipDrive(ctx, itemPath, c.matcher, driveName) &&
-			shouldSkipDrive(ctx, collectionPath, c.matcher, driveName) {
+		if shouldSkipDrive(ctx, collectionPath, c.matcher, driveName) {
```
meain, line 634: Should `shouldSkipDrive` check happen before we delete item(call `handleDelete`)? We could add a test for this.

src/internal/connector/onedrive/collections.go
```diff
@@ -469,23 +464,132 @@ func updateCollectionPaths(
...
+			item.GetParentReference().GetPath() == nil {
+			err := clues.New("no parent reference").With("item_id", itemID)
+			if item.GetName() != nil {
+				err = err.With("item_name", ptr.Val(item.GetName()))
+			}
+
+			return nil, err
+		}
+
+		collectionPathStr = ptr.Val(item.GetParentReference().GetPath())
```
meain, line 564: Just wondering if we should just append the name here itself instead of after creating the path.

ashmrtn: yea, I debated about that for a bit too. I can add that if you feel strongly about it. The reason I ended up using `path.Append` below instead is we know at that point how to add items to a path because we control the path package. If we append the folder element here, we're making assumptions about what the format of the path is from Graph (mostly path separators). I know they won't change this cause it will break a lot of people's code, but part of me kind of likes to reduce assumptions made about stuff we don't have direct control over 😅

src/internal/connector/onedrive/collections.go
ashmrtn: I'm not sure that would give the behavior we want either :/ Unfortunately when items are deleted they have their parent reference set to the root folder. This is why we need the exclude list instead of just adding a deleted item entry in the collection for the item. Since they all end up in the root folder and we don't have information about paths from the previous backup the `shouldSkipDrive` check isn't particularly useful

I guess on the bright side, `shouldSkipDrive` really only comes into play for CI testing. Since the code was originally written we've moved away from allowing CLI users to backup individual folders/items and instead require backing up the entire service. This makes me a little less worried about edge cases I suppose

src/internal/connector/onedrive/collections.go
ashmrtn: ooo nice suggestion, I like it 😄

ashmrtn (commented):
thanks for the re-review

PR #2623, opened 2023-02-23
Wrap all assert/require error assertion funcs

ryanfkeepers:
Since structured error data from clues is not added to error message logging from assert and require,
we're slicing out data that is valuable to debugging. This adds a package in tester which wraps the error assertion funcs so that structured error data is appended to all failure outputs.

Changes, outside of the new file (assert_wrapper.go) are 100% find&replace.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1970

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/tester/aw/assert_wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+
+// MustNotErrIs wraps require.NotErrorIs()
+func MustNotErrIs(t require.TestingT, err, target error, etc ...any) {
+	require.NotErrorIs(t, err, target, addClues(err, etc))
+}
+
+// MustErrAs wraps require.ErrorAs()
+func MustErrAs(t require.TestingT, err, target error, etc ...any) {
+	require.ErrorAs(t, err, target, addClues(err, etc))
+}
```
dadams39, line 38: Require is a more intuitive name.

dadams39 (changes requested):

src/internal/tester/aw/assert_wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// this file is subfoldered to avoid cyclic dependencies
+
+// ---------------------------------------------------------------------------
+// require wrappers
+// ---------------------------------------------------------------------------
+
+// MustErr wraps require.Error()
+func MustErr(t require.TestingT, err error, etc ...any) {
```
ashmrtn, line 16: I feel like we need to be careful doing stuff like this. Testify has a large surface area when it comes to functions and this only covers the `Error` cases, not the `Errorf` cases which expands the surface area even more. Is there some other way we can get the information we need? For example, does testify print the errors with the `%+v` specifier and can we leverage that? Overall, having to wrap test functions to get the info we need feels like a bit of an anti-pattern

ryanfkeepers: Yeah, I'm not particularly happy about it, either.  But we don't have much tooling to work with here.  Fwiw, here's testify's Error func:
```go
func Error(t TestingT, err error, msgAndArgs ...interface{}) bool {
	if err == nil {
		if h, ok := t.(tHelper); ok {
			h.Helper()
		}
		return Fail(t, "An error is expected but got nil.", msgAndArgs...)
	}

	return true
}
```
So it _does_ include `%+v`.  But I'm not certain this is the best way to display the data involved.  Without some preprocessing trigger (ex: `assert.Error(t, clues.Dump(err))`) we'd end up interpolating structured data throughout stack trace lines.  Maybe that's not the worst thing, if the caller is already requesting additional formatting.  It does deviate from standard expectations about stack trace design.  Much easier compliance that way, though.  I'll ask around for additional advice and noodle on the solution.

ryanfkeepers: Hmm, I think the most idiomatic way to do this is to append the data to the assert call.  Less fancy, but more proper.  going to close this PR and redo in a different branch.

PR #2624, opened 2023-02-23
1970 11 tracker fails redo

ryanfkeepers:
#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #1970

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -180,7 +180,7 @@ func (c Contacts) EnumerateContainers(
 		builder = users.NewItemContactFoldersItemChildFoldersRequestBuilder(link, service.Adapter())
 	}
 
-	return errs.Err()
+	return et.Err()
 }
 
 // ---------------------------------------------------------------------------
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -126,24 +126,33 @@ func (c Contacts) EnumerateContainers(
 			With("options_fields", fields)
 	}
 
+	et := errs.Tracker()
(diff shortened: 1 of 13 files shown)


src/internal/connector/exchange/api/events.go
```diff
@@ -178,7 +191,11 @@ func (c Events) EnumerateContainers(
...
 			if err := fn(temp); err != nil {
-				errs.Add(clues.Stack(err).WithClues(fctx).With(graph.ErrData(err)...))
+				et.Add(clues.Stack(err).
+					WithClues(fctx).
+					With(graph.ErrData(err)...).
+					Label(fault.LabelForceNoBackupCreation))
+
 				continue
 			}
 		}
```
ashmrtn, line 201: can't comment directly, but I think line 194/211 below needs to be `return et.Err()`

src/internal/connector/exchange/api/contacts.go
```diff
@@ -154,7 +163,11 @@ func (c Contacts) EnumerateContainers(
...
 			if err := fn(temp); err != nil {
-				errs.Add(clues.Stack(err).WithClues(fctx).With(graph.ErrData(err)...))
+				et.Add(clues.Stack(err).
+					WithClues(fctx).
+					With(graph.ErrData(err)...).
+					Label(fault.LabelForceNoBackupCreation))
+
 				continue
 			}
 		}
```
ashmrtn, line 173: can't comment directly but I think line 170/183 needs to be `return et.Err()`

ashmrtn (changes requested):

ashmrtn (approved):

PR #2496, opened 2023-02-14
Change to storing files by ID instead of display name

ashmrtn:
## Description

Change to storing OneDrive files by ID instead of their OneDrive display name. This allows delta token-based incrementals to use an exclude list to remove previously backed up items from a backup during hierarchy merging

Also updates the following:
* selectors to match on file display name instead of kopia file name
* ShortRefs for OneDrive files to update when the file display name is updated

## Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [x] :no_entry: No 

## Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

## Issue(s)

* closes #1535

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

Issue #2333, opened 2023-01-31
proper base selection when features are toggled via flags


ashmrtn:

ashmrtn:
somewhat related to #2261

meain:
@ashmrtn Do we have to consider this for incrementals and keep it associated to https://github.com/alcionai/corso/issues/2117 ? Permissions should be always available once we release it.

That said, we can keep this open if you want to use this issue to track using failed backups as base for future backups.

meain:
>  ... we provide something like a `--no-permissions` flag. I don't know if we want that flag ...

I don't think we will be keeping around a flag like that. @gmatev , feel free to correct me if I am wrong.

Issue #2361, opened 2023-02-01
Add another way to force a file to be marked as uncached in kopia


ashmrtn:
Mod time alone is not sufficient to allow Corso to get new data into kopia, especially if we want to start picking up data we weren't previously backing up. This ties into how kopia [decides a StreamingFile is cached](https://github.com/kopia/kopia/blob/b0c1e6d883ae65b5a6f75df89c84541163e72d32/snapshot/snapshotfs/upload.go#L667). Adding the ability to control one or more of these fields besides ModTime will give us the ability to force data updates on a per-item basis as needed

Another thing that would be good to have is to not have to read the previous version of an item to determine if we should send the updated version of the item to kopia. That will reduce the number of reads kopia does during a backup and keep the backup logic simple

ashmrtn:
if we don't want to add another field to the file entry struct that kopia currently uses, hijacking the `Owners` field for `StreamingFile` is probably our best bet. It would give us access to two `uint32`s and is already used in the comparison for cached files

PR #2642, opened 2023-02-24
Enable OneDrive incrementals, disable incrementals for permission backup

ashmrtn:
Permissions backup requires more work, so we only support delta incrementals for OneDrive if that's not enabled.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #2333

Pending merge of
* #2650
* #2496

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -292,12 +292,20 @@ func (op *BackupOperation) do(
 // checker to see if conditions are correct for incremental backup behavior such as
 // retrieving metadata like delta tokens and previous paths.
 func useIncrementalBackup(sel selectors.Selector, opts control.Options) bool {
-	// Delta-based incrementals currently only supported for Exchange
-	if sel.Service != selectors.ServiceExchange {
+	enabled := !opts.ToggleFeatures.DisableIncrementals
+
+	switch sel.Service {
+	case selectors.ServiceExchange:
+		return enabled
+
+	case selectors.ServiceOneDrive:
+		// TODO(ashmrtn): Remove the && part once we support permissions and
+		// incrementals.
+		return enabled && !opts.ToggleFeatures.EnablePermissionsBackup
+
+	default:
 		return false
 	}
-
-	return !opts.ToggleFeatures.DisableIncrementals
 }
 
 // ---------------------------------------------------------------------------
@@ -458,9 +466,7 @@ func consumeBackupDataCollections(
 		ctx,
 		bases,
 		cs,
-		// TODO(ashmrtn): When we're ready to enable incremental backups for
-		// OneDrive replace this with `excludes`.
-		nil,
+		excludes,
 		tags,
 		isIncremental,
 		errs)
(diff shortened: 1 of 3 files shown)


ryanfkeepers (approved):

src/internal/operations/backup.go
```diff
@@ -294,8 +294,10 @@ func (op *BackupOperation) do(
 // checker to see if conditions are correct for incremental backup behavior such as
 // retrieving metadata like delta tokens and previous paths.
 func useIncrementalBackup(sel selectors.Selector, opts control.Options) bool {
-	// TODO(meain): remove this once we stabilize delta incrementals for OneDrive
-	if sel.Service == selectors.ServiceOneDrive {
+	// TODO(meain): remove this once we stabilize delta incrementals for OneDrive.
+	// Updated version will still need to check if permissions backup is enabled
+	// though because we can't support that and delta incrementals right now.
+	if sel.Service == selectors.ServiceOneDrive && !opts.ToggleFeatures.EnablePermissionsBackup {
```
ryanfkeepers, line 300: Does this need any customer facing documentation?  Or, at least, a "known issues" entry?

ashmrtn: probably yes, though I'm also not exactly sure how this is planned to be released right now. It's based on a PR that adds a separate flag for OneDrive, but I'm not sure if we'll release the extra flag or not

meain: We should add an entry to "known issues" section about this in changelog though.

src/internal/operations/backup.go
```diff
@@ -292,12 +292,20 @@ func (op *BackupOperation) do(
...
+	enabled := !opts.ToggleFeatures.DisableIncrementals
+
+	switch sel.Service {
+	case selectors.ServiceExchange:
+		return enabled
+
+	case selectors.ServiceOneDrive:
+		// TODO(ashmrtn): Remove the && part once we support permissions and
+		// incrementals.
+		return enabled && !opts.ToggleFeatures.EnablePermissionsBackup
```
meain, line 304: Permissions backup should not be a problem now.

```suggestion
	case selectors.ServiceOneDrive:
		return enabled
```

meain: There is still one caveat even though delta for permissions now works. If they try to run a delta update without permissions, we will run into issues where only half of the items will have permissions.

ashmrtn: hmmm this does raise a broader question of how we'd handle it normally though. I don't think there's really a good way for us to do that right now. Either we're going to have to keep the permissions backups and no-permissions backups separate and not the no-permissions backups act as bases for the permissions backups or we're always going to have to force backing up permissions. Otherwise I just don't think there's a safe way to handle incrementals with both of them

re: eventually enabling permissions backup when we have incrementals, this may push us towards adding something else that lets us tell kopia a file isn't cached (#2361). At least that would solve things the first time around... maybe?

meain: Flag for enabling/disabling permissions backup is only temporary, once it is outside of experimental, permissions are always backed up.

ashmrtn: are we going to allow users to pass a flag saying not to backup permissions? If we do then we're kind of back to square one I think :/

meain: As of now yes, but we will be removing the option to not backup permissions before we "release" permissions backup.

ashmrtn: alright, I think we have a reasonable path forward then. I think something like #2361 will allow us to pull in permissions for a backup that doesn't have them so we can still do incrementals as well. Once permissions is enabled for good we won't have to worry about which base we select because they'll always be the same

PR #2650, opened 2023-02-24
Exclude metadata files from the set that we merge during incrementals

### LINKED ISSUE #2498: Do not add `.meta` files to backup details
As of now, the details file for OneDrive contains both `.data` and `.meta` files, but it is not necessary to store that information in details file. We should be able to figure out that the backup contains `.meta` files by looking at the backup version and augmenting the list in [formatDetailsForRestore](https://github.com/alcionai/corso/blob/43e5ccdb5bf03add030304510eb25a4559961d65/src/internal/operations/restore.go#L342). We can do this by dropping the need for info for metadata.

Prev discussion: https://github.com/alcionai/corso/pull/2473#issuecomment-1428398365

ashmrtn:
Metadata files don't need to appear in backup
details, which means they also shouldn't be
merged during incremental backups. If we
expect them to be merged then we'll fail the
item count check for the merge process

This excludes items with metadata extensions
from merging by not adding the to the list
of items to merge

Also make it so that metadata items don't
create backup details entries in the first
place

Manually tested to ensure
restore(backup(onedrive)) still works as
expected
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* closes #2498

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -22,6 +22,7 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/internal/connector/onedrive"
 	"github.com/alcionai/corso/src/internal/data"
 	D "github.com/alcionai/corso/src/internal/diagnostics"
 	"github.com/alcionai/corso/src/pkg/backup/details"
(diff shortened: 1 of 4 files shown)


src/internal/kopia/upload.go
```diff
@@ -448,16 +449,21 @@ func streamBaseEntries(
...
 
-		// All items have item info in the base backup. However, we need to make
-		// sure we have enough metadata to find those entries. To do that we add the
-		// item to progress and having progress aggregate everything for later.
-		d := &itemDetails{
-			info:         nil,
-			repoPath:     itemPath,
-			prevPath:     prevItemPath,
-			locationPath: locationPath,
+		// Meta files aren't in backup details since it's the set of items the user
```
meain, line 452: Should we bump up the backup version?

src/pkg/repository/repository.go
```diff
@@ -367,10 +366,7 @@ func (r repository) BackupDetails(
 	if b.Version >= version.OneDrive1DataAndMetaFiles && b.Version < version.OneDrive3IsMetaMarker {
```
meain, line 366: If we do bump up the version, we can use this to limit this loop to only ones under the version.

src/internal/kopia/upload.go
```diff
@@ -448,16 +449,21 @@ func streamBaseEntries(
...
+		// sees.
+		if !onedrive.IsMetaFile(itemPath.String()) {
```
meain, line 454: Just checking the extension feels a bit flaky to me. We should at least check if it a OneDrive file going through this check. Exchange will probably not have `.meta` as a file extension, but we should probably do that check.

Ideal would be to have something like `entity.IsAuxiliary()`, an extra item added to the interface, similar to `IsMeta` on `details.OneDriveInfo`, which we can use and *extend* in the future.

src/internal/connector/graph/metadata/metadata.go
```diff
@@ -0,0 +1,16 @@
+package metadata
```
ashmrtn, line 1: not sure what this package should be called or where this code should live exactly. We can't put it in the `graph` package because that creates an import cycle. Happy to take suggestions on a better name/location

meain: Maybe in `internal/connector`, but this seems OK too.

meain (approved):

src/internal/connector/graph/metadata/metadata.go
ashmrtn: I think `internal/connector` is not possible because it will create an import cycle between `onedrive` and `connector` :/

Issue #2577, opened 2023-02-20
Print out data in clues on error


meain:
As of now, even though we do record extra data and stacktrace with `clues`, we don't output any of them when we error out in the  corso CLI.

ryanfkeepers:
In the interim to a fix, these values should be available in logs.

ryanfkeepers:
related: https://github.com/alcionai/corso/issues/1970

PR #2659, opened 2023-02-27
Log item information when we are not able to download an item

meain:
This will help us pin point the file which caused issues and can be
used to debug the specific file.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -372,7 +372,7 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
-						logger.Ctx(ctx).With("error", err.Error()).Infow("unable to download item")
+						logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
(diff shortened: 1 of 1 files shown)


meain:
https://github.com/alcionai/corso/issues/2577 could help us in the future.

src/internal/connector/onedrive/collection.go
```diff
@@ -370,6 +372,7 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
+						logger.Ctx(ctx).With("error", err.Error()).Infow("unable to download item")
```
ryanfkeepers, line 375: Nit: avoid statements like "unable to".  This should be an error log, which will already make explicit that we were unable to / failed to do the thing.
```suggestion
						logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
```

ryanfkeepers (approved):

Issue #2673, opened 2023-02-28
Enable permissions backup for groups and applications


meain:
As of now we only backup permissions for users. Groups and applications can also have permissions as mentioned in [gratedToV2 field docs](https://learn.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0#properties).

> For user type permissions, the details of the users and applications for this permission.

Issue #2621, opened 2023-02-23
Save and restore permissions using id instead of email


meain:
It is possible that the email of a user can change, or not available(if security group) and so it is better to use id of the user instead of the email to restore permissions.

https://learn.microsoft.com/en-us/graph/api/driveitem-invite?view=graph-rest-1.0&tabs=http
https://learn.microsoft.com/en-us/graph/api/resources/driverecipient?view=graph-rest-1.0

PR #2674, opened 2023-02-28
Switch to using ids for storing permissions

meain:
This also enable backing up and restoring permissions for groups and applications.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/2621
* fixes https://github.com/alcionai/corso/issues/2673

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector_onedrive_test.go
+++ b/src/internal/connector/graph_connector_onedrive_test.go
@@ -22,7 +22,7 @@ import (
 
 // For any version post this(inclusive), we expect to be using IDs for
 // permission instead of email
-const versionPermissionSwitchedToId = version.OneDrive4DirIncludesPermissions
+const versionPermissionSwitchedToID = version.OneDrive4DirIncludesPermissions
 
 func getMetadata(fileName string, perm permData, permUseID bool) onedrive.Metadata {
 	if len(perm.user) == 0 || len(perm.roles) == 0 {
@@ -200,7 +200,7 @@ func (c *onedriveCollection) withFile(name string, fileData []byte, perm permDat
 			"",
 			name+onedrive.MetaFileSuffix,
 			perm,
-			c.backupVersion >= versionPermissionSwitchedToId)
+			c.backupVersion >= versionPermissionSwitchedToID)
 		c.items = append(c.items, metadata)
 		c.aux = append(c.aux, metadata)
 
@@ -224,7 +224,7 @@ func (c *onedriveCollection) withFolder(name string, perm permData) *onedriveCol
 				"",
 				name+onedrive.DirMetaFileSuffix,
 				perm,
-				c.backupVersion >= versionPermissionSwitchedToId))
+				c.backupVersion >= versionPermissionSwitchedToID))
 
 	default:
 		assert.FailNowf(c.t, "bad backup version", "version %d", c.backupVersion)
@@ -253,7 +253,7 @@ func (c *onedriveCollection) withPermissions(perm permData) *onedriveCollection 
 		name,
 		name+onedrive.DirMetaFileSuffix,
 		perm,
-		c.backupVersion >= versionPermissionSwitchedToId)
+		c.backupVersion >= versionPermissionSwitchedToID)
 
 	c.items = append(c.items, metadata)
 	c.aux = append(c.aux, metadata)
--- a/src/internal/connector/graph_connector_onedrive_test.go
+++ b/src/internal/connector/graph_connector_onedrive_test.go
@@ -11,14 +11,12 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
-	"github.com/alcionai/corso/src/internal/connector/discovery"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/onedrive"
 	"github.com/alcionai/corso/src/internal/tester"
 	"github.com/alcionai/corso/src/internal/version"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/control"
-	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
@@ -120,25 +118,13 @@ func (suite *GraphConnectorOneDriveIntegrationSuite) SetupSuite() {
 	suite.secondaryUser = tester.SecondaryM365UserID(suite.T())
 	suite.acct = tester.NewM365Account(suite.T())
 
-	// Not using m365.Users to avoid cyclic imports
-	users, err := discovery.Users(ctx, suite.connector.Owners.Users(), fault.New(true))
-	require.NoError(suite.T(), err, "fetching users")
+	user, err := suite.connector.Owners.Users().GetByID(ctx, suite.user)
+	require.NoErrorf(suite.T(), err, "fetching user %s", suite.user)
+	suite.userID = *user.GetId()
 
-	for _, u := range users {
-		if *u.GetUserPrincipalName() == suite.user {
-			suite.userID = *u.GetId()
-		} else if *u.GetUserPrincipalName() == suite.secondaryUser {
-			suite.secondaryUserID = *u.GetId()
-		}
-	}
-
-	if suite.userID == "" {
-		require.FailNowf(suite.T(), "unable to find user id", "user %d", suite.user)
-	}
-
-	if suite.secondaryUserID == "" {
-		require.FailNowf(suite.T(), "unable to find user id", "user %d", suite.secondaryUser)
-	}
+	secondaryUser, err := suite.connector.Owners.Users().GetByID(ctx, suite.secondaryUser)
+	require.NoErrorf(suite.T(), err, "fetching user %s", suite.secondaryUser)
+	suite.secondaryUserID = *secondaryUser.GetId()
 
 	tester.LogTimeOfTest(suite.T())
 }
--- a/src/internal/connector/graph_connector_onedrive_test.go
+++ b/src/internal/connector/graph_connector_onedrive_test.go
@@ -20,6 +20,10 @@ import (
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
+// For any version post this(inclusive), we expect to be using IDs for
+// permission instead of email
+const versionPermissionSwitchedToId = version.OneDrive4DirIncludesPermissions
+
 func getMetadata(fileName string, perm permData, permUseID bool) onedrive.Metadata {
 	if len(perm.user) == 0 || len(perm.roles) == 0 {
 		return onedrive.Metadata{FileName: fileName}
@@ -173,12 +177,7 @@ func (c onedriveCollection) collection() colInfo {
 	}
 }
 
-func (c *onedriveCollection) withFile(
-	name string,
-	fileData []byte,
-	perm permData,
-	permUseID bool,
-) *onedriveCollection {
+func (c *onedriveCollection) withFile(name string, fileData []byte, perm permData) *onedriveCollection {
 	switch c.backupVersion {
 	case 0:
 		// Lookups will occur using the most recent version of things so we need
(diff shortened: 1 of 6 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -195,9 +195,13 @@ func (oc Collection) DoNotMergeItems() bool {
...
 // OneDrive item.
 type UserPermission struct {
-	ID         string     `json:"id,omitempty"`
-	Roles      []string   `json:"role,omitempty"`
-	Email      string     `json:"email,omitempty"`
+	ID       string   `json:"id,omitempty"`
+	Roles    []string `json:"role,omitempty"`
+	Email    string   `json:"email,omitempty"` // DEPRECATED: Replaced with UserID in newer backups
+	EntityID string   `json:"entityId,omitempty"`
+	// TODO(meain): Should we be storing the type of entity? It is not
```
meain, line 202: Any opinions?

src/internal/connector/onedrive/item.go
```diff
@@ -265,10 +266,29 @@ func filterUserPermissions(perms []models.Permissionable) []UserPermission {
...
 		}
 
+		entityID := ""
+		if gv2.GetUser() != nil {
+			entityID = *gv2.GetUser().GetId()
+		} else if gv2.GetGroup() != nil {
+			// TODO(meain): Add test for Groups
+			entityID = *gv2.GetGroup().GetId()
+		} else if gv2.GetApplication() != nil {
+			// TODO(meain): Applications permissions restore not
```
meain, line 276: I'm still trying to see how we can give access to permissions for an application. This code should ideally work, but I don't know how to do this in the UI, or if it even is possible.

Let me know if you have any idea around this.

meain: I'm also partly tempted to add device id as that is a possibility(because of the [datatype](https://learn.microsoft.com/en-us/graph/api/resources/sharepointidentityset?view=graph-rest-1.0)) even though I could not find any reference to being able to set permissions to device.

src/internal/connector/onedrive/permission.go
```diff
@@ -210,7 +210,14 @@ func restorePermissions(
...
 
 		rec := models.NewDriveRecipient()
-		rec.SetEmail(&p.Email)
+		if p.EntityID != "" {
+			rec.SetObjectId(&p.EntityID)
+		} else {
+			// Previous versions used to only store email for a
+			// permissions. Use that if id is not found.
+			rec.SetEmail(&p.Email)
+		}
```
meain, line 219: Should just not bother with this as permissions is still in experimental?

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -110,6 +120,26 @@ func (suite *GraphConnectorOneDriveIntegrationSuite) SetupSuite() {
...
+	users, err := discovery.Users(ctx, suite.connector.Owners.Users(), fault.New(true))
+	require.NoError(suite.T(), err, "fetching users")
+
+	for _, u := range users {
+		if *u.GetUserPrincipalName() == suite.user {
+			suite.userID = *u.GetId()
+		} else if *u.GetUserPrincipalName() == suite.secondaryUser {
+			suite.secondaryUserID = *u.GetId()
+		}
+	}
```
ashmrtn, line 133: I think you can just use `suite.connector.Owners.Users().GetByID(ctx, <user name>)` to get a single `models.Userable` graph SDK struct for a given user ID or principal name (PN). PNs look a lot like emails and I think all the users we have right now have the same email and PN

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -110,6 +120,26 @@ func (suite *GraphConnectorOneDriveIntegrationSuite) SetupSuite() {
...
+
+	if suite.userID == "" {
+		require.FailNowf(suite.T(), "unable to find user id", "user %d", suite.user)
+	}
+
+	if suite.secondaryUserID == "" {
+		require.FailNowf(suite.T(), "unable to find user id", "user %d", suite.secondaryUser)
+	}
```
ashmrtn, line 141: nit: you can use `require.NotEmpty()` for these checks
```suggestion
	require.NotEmptyf(suite.T(), suite.userID, "unable to find user id", "user %d", suite.user)
	}

	require.NoteEmptyf(suite.T(), suite.secondaryUserID, "unable to find user id", "user %d", suite.secondaryUser)
```

src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -278,21 +309,21 @@ type onedriveTest struct {
...
-func testDataForInfo(t *testing.T, cols []onedriveColInfo, backupVersion int) []colInfo {
+func testDataForInfo(t *testing.T, cols []onedriveColInfo, backupVersion int, permUseID bool) []colInfo {
 	var res []colInfo
 
 	for _, c := range cols {
 		onedriveCol := newOneDriveCollection(t, c.pathElements, backupVersion)
 
 		for _, f := range c.files {
-			onedriveCol.withFile(f.name, f.data, f.perms.user, f.perms.roles)
+			onedriveCol.withFile(f.name, f.data, f.perms, permUseID)
```
ashmrtn, line 319: nit: if we're always going to have either all emails or all IDs in the permission info for backups then you can just make `permUseID` a field in the `onedriveCollection` struct like backup version is


edit: actually since it generally seems like it's set based on the backup version you could roll that into the onedriveCollection helper functions and not pass an extra parameter at all

src/internal/connector/onedrive/collection.go
```diff
@@ -197,7 +197,8 @@ func (oc Collection) DoNotMergeItems() bool {
 type UserPermission struct {
 	ID         string     `json:"id,omitempty"`
 	Roles      []string   `json:"role,omitempty"`
-	Email      string     `json:"email,omitempty"`
+	Email      string     `json:"email,omitempty"` // DEPRECATED: Replaced with UserID in newer backups
```
ashmrtn, line 200: nice, I like the comment and good to know we won't see JSON deserialization errors due to not having the field

src/internal/connector/onedrive/item.go
ashmrtn: hmmm I assume you didn't happen to find any documentation on how/if those fields can be set? It seems a little weird to be able to populate them for permissions and it's possible they're there just because the `sharePointIdentitySet` struct is being reused for permissions..?

Looking at the code though are we certain only one of the fields will be set?

I guess we have a few options:
1. don't persist anything but user or group IDs, collection non-catastrophic errors and/or log if there's data for the other fields. Logging/errors will at least let us know if we need to expand further. Not having (currently empty) fields won't impede future JSON deserialization as it's equivalent to having an empty value and `omitempty`
2. collect and persist all fields, rely on JSON `omitempty` tag to keep us from serializing a bunch of stuff that isn't there

First one may require more changes later on. Second one requires possible restore updated, but we could presumably roll those out at a later date if needed

src/internal/connector/onedrive/permission.go
ashmrtn: it seems like it would be easier to guarantee a backup would have one or the other. Is it possible to share items with someone that won't have an ID?

@vkamra do we have a gmail account or something we can attempt sharing with to see if we'll get an ID in the returned permission?

src/internal/connector/onedrive/item.go
```diff
@@ -265,10 +266,28 @@ func filterUserPermissions(perms []models.Permissionable) []UserPermission {
...
+		if entityID == "" {
+			// This should ideally not be hit
+			continue
+		}
+
 		up = append(up, UserPermission{
 			ID:         ptr.Val(p.GetId()),
 			Roles:      roles,
-			Email:      *p.GetGrantedToV2().GetUser().GetAdditionalData()["email"].(*string),
+			EntityID:   entityID,
```
ashmrtn, line 290: is overloading this field alright for restores? For example, does Graph API need to know if we're sending it a user ID or a group ID?

src/internal/connector/onedrive/permission.go
meain: > it seems like it would be easier to guarantee a backup would have one or the other. Is it possible to share items with someone that won't have an ID?

Ahh, you are right. It totally can have permissions for any random email outside of the system which will not have id. Posting a sample response here.

``` json
{
    "id": "c68a61e2-73fd-4f7d-85c9-fa4d920d6de6",
    "roles": [
        "write"
    ],
    "hasPassword": false,
    "grantedToIdentitiesV2": [
        {
            "user": {
                "@odata.type": "#microsoft.graph.sharePointIdentity",
                "displayName": "email@provider.com",
                "email": "email@provider.com"
            }
        }
    ],
    "grantedToIdentities": [
        {
            "user": {
                "displayName": "email@provider.com",
                "email": "email@provider.com"
            }
        }
    ],
    "link": {
        "scope": "users",
        "type": "edit",
        "webUrl": "https://1...2-my.sharepoint.com/:u:/g/personal/joh...ft_com/Ef...nw",
        "preventsDownload": false
    }
}
```

meain: It sends an email to you when the file is shared, clicking that open a web ui where it asks you for email. When you enter the email, you get a code sent to that account which you can use to open it.

src/internal/connector/onedrive/item.go
meain: Yup, I am able to restore both users and groups with just id. The [docs](https://learn.microsoft.com/en-us/graph/api/resources/driverecipient?view=graph-rest-1.0) also mention the field in restore as `objectId` and nothing specific.

src/internal/connector/graph_connector_onedrive_test.go
meain: > actually since it generally seems like it's set based on the backup version you could roll that into the onedriveCollection helper functions and not pass an extra parameter at all

~I was kinda debating this. This is not really based on version as there is no specific version(no new version created for this change) which switches it from email to id, plus both email and id should work with all version. We probably should run tests with both id and email for all versions but decided to not do that so that to double the test count.~

On second thought, I think it is OK to just base it off of the version.

src/internal/connector/onedrive/item.go
meain: > I assume you didn't happen to find any documentation on how/if those fields can be set?

I have not been able to find anything yet. I know application is possible from the fact that it mentions the id could be from user/application in the [docs for the field](https://learn.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0#properties).

>  it's possible they're there just because the sharePointIdentitySet struct is being reused for permissions

That is my best guess.

> are we certain only one of the fields will be set?

I have not found anything mentioned specifically in docs, but from the experiments that I did, it seems so. Even the docs examples, it does not have multiple set at any point.


> I guess we have a few options

Both seem like good options, though I am a bit more leaning towards the first one:
- It will let us catch and react when error happens(this comes with the problem that we will get errors)
- We are not storing unnecessary information

meain: I was able to find some documentation on how application level permission can be set. This seems to be only for SharePoint though, and not for OneDrive IIUC.
https://devblogs.microsoft.com/microsoft365dev/controlling-app-access-on-specific-sharepoint-site-collections/

src/internal/connector/onedrive/permission.go
meain: Closing this as this is specific to link shared which we are not handling as of now.

ashmrtn: agree with punting all of the details on this until if/when we handle link shares. From the code though, it seems like it may workout alright anyway though? If the ID isn't populated then we'll pull the email and send that when we restore permissions which should (in theory) trigger the whole workflow you described above?

ashmrtn (approved):

src/internal/connector/onedrive/permission.go
meain: Yup, that will work. The only difference is that link shares are in `grantedToIdentitiesV2` instead of `grantedToV2` which is what we are currently backing up.

Just for context, the primary reason that I did not originally work on link shares was because we could not create the same links as before and that becomes pointless when the link was shared externally. Even with link shares to emails, it would end up emailing the person about the new link(for every item with a custom permission).

Issue #2675, opened 2023-02-28
SDK BackupDetails production


ryanfkeepers:
## Issue
The repository BackupDetails produces a different result than the details consumed by the CLI or the Selectors reduce.  We can't unify them entirely, but we can close the gaps.

## Acceptance Criteria
1. Filter .meta and .dirmeta files and empty folders (post-meta file removal) from the repository.
2. Add a flag or command to produce all details in the CLI.  This set cannot be filtered by selectors (at this time).  May want the addition to be hidden.

PR #2677, opened 2023-02-28
filter empty paths from details entries

ryanfkeepers:
repository.BackupDetails now filters its results
to not contain meta or dirmeta items, and to
not contain empty folders, where empty is a
count of zero items after the removal of all
meta items.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #2675

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Corso-generated .meta files and permissions no longer appear in the backup details.
+- repo.BackupDetails from the SDK now filter out meta and dirmeta files, and empty folders, by standard.
 
 ### Known Issues
 - Folders and Calendars containing zero items or subfolders are not included in the backup.
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -305,6 +305,84 @@ func (d *Details) addFolder(folder folderEntry) {
 	})
 }
 
+// FilterMetaFiles returns a new Details struct with a copy of the DetailsModel
+// that has had all .meta and .dirmeta files stripped out.
+func (d *Details) FilterMetaFiles() *Details {
+	return &Details{
+		DetailsModel: d.DetailsModel.FilterMetaFiles(),
+	}
+}
+
+// FilterEmptyContainers returns a new Details struct all empty (ie: containing no
+// items) stripped out.  If meta files have not been filtered out already, they
+// will continue to count as a "populated" container.
+func (d *Details) FilterEmptyContainers() *Details {
+	type entCount struct {
+		ent       DetailsEntry
+		itemCount int
+	}
+
+	var (
+		// shortRef: entCount
+		srec  = map[string]entCount{}
+		items = []DetailsEntry{}
+	)
+
+	// split the entries into items and folders.
+	// folders are stored in a map by their shortRef for lookup.
+	for _, ent := range d.Entries {
+		if ent.Folder == nil {
+			items = append(items, ent)
+		} else {
+			srec[ent.ShortRef] = entCount{ent, 0}
+		}
+	}
+
+	// for every item, add a count to the owning folder.
+	// this assumes item parentRef == folder shortRef.
+	for _, ent := range items {
+		if len(ent.ParentRef) == 0 {
+			continue
+		}
+
+		ec := srec[ent.ParentRef]
+		ec.itemCount++
+		srec[ent.ParentRef] = ec
+
+		// to maintain a hierarchical count so that we don't
+		// slice parent folders, this loop walks the tree upward
+		// by parent ref, adding one count to each parent up
+		// to the root.
+		parentRef := ec.ent.ParentRef
+		parentCount := 0
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Corso-generated .meta files and permissions no longer appear in the backup details.
+- repo.BackupDetails from the SDK now filter out meta and dirmeta files, and empty folders, by standard.
```
zackrossman, line 15: Do we want to leave "OneDrive files ending in `.meta` or `.dirmeta` are omitted from details and restores." in the known issues section?

src/pkg/backup/details/details.go
```diff
@@ -305,6 +305,84 @@ func (d *Details) addFolder(folder folderEntry) {
...
+	// for every item, add a count to the owning folder.
+	// this assumes item parentRef == folder shortRef.
+	for _, ent := range items {
+		if len(ent.ParentRef) == 0 {
+			continue
+		}
+
+		ec := srec[ent.ParentRef]
+		ec.itemCount++
+		srec[ent.ParentRef] = ec
```
zackrossman, line 350: nit: replace this chunk with `srec[ent.ParentRef]++`

CHANGELOG.md
ryanfkeepers: Yes, both conditions are true.  We are filtering out our meta files.  And also, we might filter out meta files that we did not create, unexpectedly.

src/pkg/backup/details/details.go
```diff
@@ -305,6 +305,84 @@ func (d *Details) addFolder(folder folderEntry) {
...
+		// to the root.
+		parentRef := ec.ent.ParentRef
+		parentCount := 0
+
+		for len(parentRef) > 0 && parentCount == 0 {
+			ec := srec[parentRef]
+
+			// minor optimization: if the parentCount is already
+			// >zero, then all of its parents are guaranteed >zero.
+			parentCount = ec.itemCount
```
zackrossman, line 364: nit: further optimization, if parentCount > 0 can we skip the next couple of lines? As far as this PR is concerned, all that we care about is that parentCount > 0

zackrossman (approved):
LGTM but please get approval from a Corso engineer as well

src/pkg/backup/details/details.go
ryanfkeepers: no can do: https://go.dev/play/p/-hK9Qdy9tIb

src/pkg/backup/details/details.go
ryanfkeepers: I'm not particularly concerned with O(k) optimizations.  Frankly, the above is already a bit excessive.  But since it's closer to O(n), I figured I'd plug it in.

src/pkg/backup/details/details.go
```diff
@@ -305,6 +305,84 @@ func (d *Details) addFolder(folder folderEntry) {
...
+func (d *Details) FilterMetaFiles() *Details {
+	return &Details{
+		DetailsModel: d.DetailsModel.FilterMetaFiles(),
+	}
+}
+
+// FilterEmptyContainers returns a new Details struct all empty (ie: containing no
+// items) stripped out.  If meta files have not been filtered out already, they
+// will continue to count as a "populated" container.
+func (d *Details) FilterEmptyContainers() *Details {
```
vkamra, line 319: Was trying to see if there's a better way to do this instead of going through all folders/items again. We're already scanning this data structure multiple times.

Where are empty folders actually being added to details? Should we just stop doing that till we have a solution to handle restore of empty directories?

ryanfkeepers: > Where are empty folders actually being added to details? Should we just stop doing that till we have a solution to handle restore of empty directories?

If we can swallow the performance cost in the short term, this will also act as a good location to feature-flag the progression towards backing up and restoring empty folders.  Preventing their upload in the first place is effectively just a rollback on Ash's prior work, and a step back from our eventual goal.

vkamra: I might be overthinking it - can you push a small bench test that checks the overhead of running the two filter operations on a details model with 300K items. I can run it on some of the test systems I'm using for benchmarking (not our MacBook Pro's)

ashmrtn: >Where are empty folders actually being added to details? Should we just stop doing that till we have a solution to handle restore of empty directories?

Mostly these are coming from OneDrive as empty Exchange folders shouldn't create details entries. For OneDrive it's much easier to handle things if we don't exclude empty folders due to the fact that we see folders before items and a weird interaction between selector filtering and Graph API item ordering. I would expect to only see empty folders if backup permissions was enabled in OneDrive though. Otherwise empty folders may be made in kopia but there shouldn't be any items in backup details that cause a folder entry to exist in backup details for that folder. We're seeing empty folders purely because we filter some things out of backup details

ashmrtn: actually, I stand corrected, corso will save empty permissions files when backup permissions is not enabled. That would result in empty folders being present in the backup details

PR #2688, opened 2023-03-01
Graph ResyncRequired error can also manifest as resyncRequired

meain:
The only change is that the first letter is now small.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -24,6 +24,7 @@ const (
 	errCodeItemNotFoundShort           = "itemNotFound"
 	errCodeEmailFolderNotFound         = "ErrorSyncFolderNotFound"
 	errCodeResyncRequired              = "ResyncRequired"
+	errCodeResyncRequiredAlt           = "resyncRequired"
 	errCodeSyncFolderNotFound          = "ErrorSyncFolderNotFound"
 	errCodeSyncStateNotFound           = "SyncStateNotFound"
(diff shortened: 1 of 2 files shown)


src/internal/connector/graph/errors.go
```diff
@@ -79,7 +80,7 @@ func IsErrInvalidDelta(err error) bool {
 		return true
 	}
 
-	if hasErrorCode(err, errCodeSyncStateNotFound, errCodeResyncRequired) {
+	if hasErrorCode(err, errCodeSyncStateNotFound, errCodeResyncRequired, errCodeResyncRequired2) {
```
meain, line 83: We could maybe lowercase and compare it inside `hasErrorCode`, but I was not sure if we wanted to take that route.

ryanfkeepers (approved):
Approved, but, a pitch: what if we `ToLower()`'d all the values instead?  Would protect us from this at large in the future.

meain:
@ryanfkeepers I was thinking on similar lines with [this](https://github.com/alcionai/corso/pull/2688#discussion_r1121333984), but was being paranoid. Let me check it to lowercase and check.

PR #2690, opened 2023-03-01
Corso: Exchange: Details: add parentpath 

### LINKED ISSUE #2352: Add ParentPath to `corso backup details exchange`
This already exists in the OneDrive display so we should enable an analogous view for Exchange. It is pretty difficult to navigate the the list of emails without some folder context.

dadams39:
Parent Path field added to  `Exchange` Backup Details 
---
- Only Exchange mail will display the path for  backup details. 

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* closes #2352<issue>

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -267,7 +267,7 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 			}
 
 			info.Size = int64(len(data))
-			info.Path = strings.Join(col.fullPath.Folders(), "/")
+			info.ParentPath = strings.Join(col.fullPath.Folders(), "/")
 
 			col.data <- &Stream{
 				id:      id,
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -266,6 +267,7 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 			}
 
 			info.Size = int64(len(data))
+			info.Path = strings.Join(col.fullPath.Folders(), "/")
 
 			col.data <- &Stream{
 				id:      id,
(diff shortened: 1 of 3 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -266,6 +267,7 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 			}
 
 			info.Size = int64(len(data))
+			info.Path = strings.Join(col.fullPath.Folders(), "/")
```
ryanfkeepers, line 270: path package has a join func that you should use here.

src/pkg/backup/details/details.go
```diff
@@ -516,6 +516,7 @@ type ExchangeInfo struct {
 	ItemType    ItemType  `json:"itemType,omitempty"`
 	Sender      string    `json:"sender,omitempty"`
 	Subject     string    `json:"subject,omitempty"`
+	Path        string    `json:"path,omitempty"`
```
ryanfkeepers, line 519: Nit: the term 'path' is overloaded, and should be namespaced in usage.
```suggestion
	ParentPath        string    `json:"parentPath,omitempty"`
```

src/pkg/backup/details/details.go
```diff
@@ -538,7 +539,7 @@ func (i ExchangeInfo) Headers() []string {
 		return []string{"Contact Name"}
 
 	case ExchangeMail:
-		return []string{"Sender", "Subject", "Received"}
+		return []string{"Sender", "Parent Path", "Subject", "Received"}
```
ryanfkeepers, line 542: End users will probably respond more naturally to "Folder" than to "path", since they don't think of their mailbox as a directory structure.
```suggestion
		return []string{"Sender", "Folder", "Subject", "Received"}
```

ryanfkeepers (approved):
A couple small changes needed; but good enough to unblock.

src/internal/connector/exchange/exchange_data_collection.go
dadams39: https://github.com/alcionai/corso/blob/main/src/pkg/path/path.go#L590 is not a better use case for the above. If we decide the application is standard enough to create a helper function, we should include it in a follow-up PR.

src/pkg/backup/details/details.go
dadams39: Resolved with [8762275](https://github.com/alcionai/corso/pull/2690/commits/876227596f30eb12936b1aecd037ca680acbf0c5)

src/pkg/backup/details/details.go
dadams39: Resolved with [8762275](https://github.com/alcionai/corso/pull/2690/commits/876227596f30eb12936b1aecd037ca680acbf0c5)

PR #2700, opened 2023-03-03
Log if file tagged as malware by graph

meain:
Graph exposes a "malware" filed in the delta response. This just logs that if we fail to download an item.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -410,7 +410,12 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
-						logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+						if item.GetMalware() != nil {
+							logger.Ctx(ctx).With("error", err.Error(), "malware", true).Error("downloading item")
+						} else {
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -410,7 +410,12 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
-						logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+						if item.GetMalware() != nil {
+							logger.Ctx(ctx).With("error", err.Error(), "malware", true).Error("downloading item")
```
meain, line 414: Didn't wanna add the malware key if it was not explicit present, but let me know if we should always log it.

ryanfkeepers: Could be useful to have this knowledge outside of errors.  I'd suggest adding it in the `clues.Add()` on 378 (whether or not the item is malware).

ryanfkeepers (approved):
one small nit

Issue #2701, opened 2023-03-03
Skip items flagged as malware


ryanfkeepers:
## Issue
Corso fails to back up items which are flagged as malware.

## Acceptance Criteria
- [x] Skip downloading items that are flagged as malware.
- [x] Track and persist skipped items.
- [x] Expose skipped item heuristics to end users.

PR #2702, opened 2023-03-04
avoid downloading known malware

ryanfkeepers:
If graph flags an item as malware,
avoid attempting to download it.
Only applies to drive items.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2701

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test


ntolia:
@ryanfkeepers Can we count and expose the number of malware files found and skipped? In particular, only display that X objects were not backed up because of security issues? As we see on Discord, this would be very valuable to users.

ryanfkeepers:
> @ryanfkeepers Can we count and expose the number of malware files found and skipped? In particular, only display that X objects were not backed up because of security issues? As we see on Discord, this would be very valuable to users.

@ntolia That change is tracked in the ticket.  It's not included in this PR for a couple reasons: 1, to keep the change small; 2, to separate the fix from other enhancements; and 3, because our code flow isn't set up to easily expand on metrics like that, so I don't know the scope of the change yet.  Likely a light lift, but will touch more files than might be expected.  It will arrive as a fast follow to this.

ntolia:
> > @ryanfkeepers Can we count and expose the number of malware files found and skipped? In particular, only display that X objects were not backed up because of security issues? As we see on Discord, this would be very valuable to users.
> 
> @ntolia That change is tracked in the ticket. It's not included in this PR for a couple reasons: 1, to keep the change small; 2, to separate the fix from other enhancements; and 3, because our code flow isn't set up to easily expand on metrics like that, so I don't know the scope of the change yet. Likely a light lift, but will touch more files than might be expected. It will arrive as a fast follow to this.

Thank you. I had not seen the ticket but the above makes sense. I just wanted to make sure that this didn't get dropped as given the value folks have already seen from this, I wanted to make sure we don't move on without letting users know what happened.

src/internal/connector/onedrive/collections.go
```diff
@@ -608,6 +608,13 @@ func (c *Collections) UpdateCollections(
 			break
 		}
 
+		if item.GetMalware() != nil {
+			// TODO: track the item as skipped; logging alone might
+			// slice out the data from tracking.
+			logger.Ctx(ctx).Infow("malware detected", "malware_description", item.GetMalware().GetDescription())
```
meain, line 614: From the [docs](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/resources/malware?view=odsp-graph-online), malware does not have any additional fields within it, unless this is a field filled in by the sdk. I did not try desc for malware but other fields with this method seemed to return nil.

meain (commented):
We can probably remove the logging that I introduced in a previous commit now that we will no longer be downloading items that are flagged as malware.

src/internal/connector/onedrive/collections.go
ryanfkeepers: https://learn.microsoft.com/en-us/graph/api/resources/malware?view=graph-rest-1.0

src/internal/connector/onedrive/collections_test.go
```diff
@@ -724,6 +724,34 @@ func (suite *OneDriveCollectionsSuite) TestUpdateCollections() {
 			},
 			expectedExcludes: map[string]struct{}{},
 		},
+		{
```
ashmrtn, line 727: you could probably leave a test in one or the other test functions. They overlap a lot and we should probably merge `TestUpdateCollections` with `TestGet` at some point since the latter is a superset of functionality of the former 😅

ryanfkeepers:
> We can probably remove the logging that I introduced in a previous commit now that we will no longer be downloading items that are flagged as malware.

On the other hand, keeping it around might come in handy under any circumstance that we bypass this check. Let's leave it in for now, and we can always revisit.

ashmrtn (approved):

PR #2703, opened 2023-03-04
backup malware catch

ryanfkeepers:
assuming it's possible for a graph
item to skip the malware detection,
we still want to catch and handle
400's from attempted malware
downloads downstream.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #2701

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/internal/common/ptr/pointer.go
+++ b/src/internal/common/ptr/pointer.go
@@ -43,7 +43,3 @@ func OrNow(t *time.Time) time.Time {
 
 	return *t
 }
-
-func To[T any](t T) *T {
-	return &t
-}
--- a/src/internal/common/ptr/pointer.go
+++ b/src/internal/common/ptr/pointer.go
@@ -43,3 +43,7 @@ func OrNow(t *time.Time) time.Time {
 
 	return *t
 }
+
+func To[T any](t T) *T {
+	return &t
+}
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -416,13 +416,13 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
-						if item.GetMalware() != nil {
-							logger.Ctx(ctx).With("error", err.Error(), "malware", true).Error("downloading item")
+						if clues.HasLabel(err, graph.LabelsMalware) {
+							logger.Ctx(ctx).Infow("malware item", clues.InErr(err).Slice()...)
 						} else {
 							logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+							el.AddRecoverable(clues.Stack(err).WithClues(ctx).Label(fault.LabelForceNoBackupCreation))
 						}
 
-						el.AddRecoverable(clues.Stack(err).WithClues(ctx).Label(fault.LabelForceNoBackupCreation))
 						return nil, err
 					}
 
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -22,6 +22,7 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/graph/metadata"
 	"github.com/alcionai/corso/src/internal/data"
 	D "github.com/alcionai/corso/src/internal/diagnostics"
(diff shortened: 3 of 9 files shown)


src/internal/common/ptr/pointer.go
```diff
@@ -43,3 +43,7 @@ func OrNow(t *time.Time) time.Time {
 
 	return *t
 }
+
+func To[T any](t T) *T {
```
meain, line 47: nit: This does not seem to be used anywhere.

meain (approved):
LGTM, but it would be good to not have it count towards ignored errors as mentioned [here](https://github.com/alcionai/corso/pull/2703/files#r1125184723)

src/internal/connector/onedrive/collection.go
```diff
@@ -410,8 +410,13 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 
 					// check for errors following retries
 					if err != nil {
-						logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+						if clues.HasLabel(err, graph.LabelsMalware) {
+							logger.Ctx(ctx).Infow("malware item", clues.InErr(err).Slice()...)
+						} else {
+							logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+						}
 						el.AddRecoverable(clues.Stack(err).WithClues(ctx).Label(fault.LabelForceNoBackupCreation))
```
ashmrtn, line 418: do we want to only mark backups as not completing if they don't have malware? Otherwise we'll continue to see backup failures if we somehow miss some of the malware checks prior to attempting to download the item's data

src/internal/kopia/upload.go
```diff
@@ -262,6 +263,14 @@ func (cp *corsoProgress) CachedFile(fname string, size int64) {
 // during the upload process. This could be from reading a file or something
 // else.
 func (cp *corsoProgress) Error(relpath string, err error, isIgnored bool) {
+	// The malware case is an artifact of being unable to skip the item
+	// if we catch detection at a late enough stage in collection enumeration.
+	// This is our next point of error handling, where we can identify and
+	// skip over the case.
+	if clues.HasLabel(err, graph.LabelsMalware) {
```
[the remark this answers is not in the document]
ashmrtn: I'm not really comfortable with modifying the returned stats from kopia post-backup as it could lead to hard to debug problems in the future. I think we're kind of conflating the information kopia returns with the information we'd like to persist in Corso and display to users later on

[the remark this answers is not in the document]
ryanfkeepers: Right.  I was hoping less for a post-stat modification, and more for a way to tell kopia "this isn't an error, don't worry about it, skip the file and move on" in the first place.  But since the error source is embedded in the lazy reader, I'm having trouble finding a good way to do that.

src/internal/connector/onedrive/collection.go
ryanfkeepers: Yeah, fair catch.

src/internal/common/ptr/pointer.go
ryanfkeepers: removed

src/internal/kopia/upload.go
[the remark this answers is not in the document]
ashmrtn: I don't think there is a way to do that and again, I don't think we really should be doing stuff like that. Kopia is really meant to deal with static file systems, so any error getting file data is a big issue in that context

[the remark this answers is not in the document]
ashmrtn: yes, I get that not handling this in kopia wrapper may mean we need to rethink how we handle errors in other layers of the stack. Adding that handling to kopia wrapper though seems like it leaks stuff from other layers into kopia which will complicate stuff down the road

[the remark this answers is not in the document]
ryanfkeepers: Totally with you.  Worth a check.  Hmm, I'll just have to find another way around the issue.

src/internal/kopia/upload.go
```diff
@@ -141,6 +142,18 @@ type corsoProgress struct {
 	mu         sync.RWMutex
 	totalBytes int64
 	errs       *fault.Bus
+	// expectedIgnoredErrors is a count of error cases caught in the Error wrapper
+	// which are well known and actually ignorable.  At the end of a run, if the
+	// manifest ignored error count is equal to this count, then everything is good.
+	expectedIgnoredErrors int
```
ryanfkeepers, line 148: @ashmrtn How's this approach for tallying ignored errors?

src/internal/kopia/upload.go
```diff
@@ -262,6 +275,15 @@ func (cp *corsoProgress) CachedFile(fname string, size int64) {
 // during the upload process. This could be from reading a file or something
 // else.
 func (cp *corsoProgress) Error(relpath string, err error, isIgnored bool) {
+	// The malware case is an artifact of being unable to skip the item
+	// if we catch detection at a late enough stage in collection enumeration.
+	// This is our next point of error handling, where we can identify and
+	// skip over the case.
+	if clues.HasLabel(err, graph.LabelsMalware) {
```
ashmrtn, line 282: would prefer a more generic label tbh. Checking for malware specifically means we're also going to have to check for all the other error labels we may end up needing

ashmrtn (approved):

Issue #2459, opened 2023-02-09
propagate permissions change downstream on delta change


meain:
Permissions change to a folder only results in the folder and its parent in the delta query and none of the downstream folder/items. We will have to query and update permissions for all of them. We might not have to query graph for new permissions as we can use the ones that is present on the folder, but we will end up having to update the backup details. Make sure items under folder with reduced permissions are handed properly.

PR #2705, opened 2023-03-06
Improve  handling of permissions using sharing field

meain:
Permissions are only stored if SharingMode is custom( "sharing" field is present). This will help with delta incrementals as well.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/2459

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -58,7 +58,6 @@ type SharingMode int
 const (
 	SharingModeCustom = SharingMode(iota)
 	SharingModeInherited
-	SharingModeEmpty
 )
 
 // Collection represents a set of OneDrive objects retrieved from M365
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -53,6 +53,14 @@ var (
 	_ data.StreamModTime    = &metadataItem{}
 )
 
+type SharingMode int
+
+const (
+	SharingModeInherited = SharingMode(iota)
+	SharingModeEmpty
+	SharingModeCustom
+)
+
 // Collection represents a set of OneDrive objects retrieved from M365
 type Collection struct {
 	// configured to handle large item downloads
@@ -213,7 +221,12 @@ type UserPermission struct {
 // ItemMeta contains metadata about the Item. It gets stored in a
 // separate file in kopia
 type Metadata struct {
-	FileName    string           `json:"filename,omitempty"`
+	FileName string `json:"filename,omitempty"`
+	// SharingMode denotes what the current mode of sharing is for the object.
+	// - inherited: permissions same as parent permissions (no "shared" in delta)
+	// - empty: remove all permissions (empty "shared" in delta)
+	// - custom: use Permissions to set correct permissions ("shared" has value in delta)
+	SharingMode SharingMode      `json:"permissionMode,omitempty"` // TODO(meain): SharingModeInherited is removed by omitempty
 	Permissions []UserPermission `json:"permissions,omitempty"`
 }
 
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -223,7 +223,6 @@ type Metadata struct {
 	FileName string `json:"filename,omitempty"`
 	// SharingMode denotes what the current mode of sharing is for the object.
 	// - inherited: permissions same as parent permissions (no "shared" in delta)
-	// - empty: remove all permissions (empty "shared" in delta)
 	// - custom: use Permissions to set correct permissions ("shared" has value in delta)
 	SharingMode SharingMode      `json:"permissionMode,omitempty"`
 	Permissions []UserPermission `json:"permissions,omitempty"`
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -56,7 +56,8 @@ var (
 type SharingMode int
 
 const (
-	SharingModeInherited = SharingMode(iota)
+	SharingModeUnknown = SharingMode(iota) // For older versions. This will be treated similar to cutsom
+	SharingModeInherited
 	SharingModeEmpty
 	SharingModeCustom
 )
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -56,10 +56,9 @@ var (
 type SharingMode int
 
 const (
-	SharingModeUnknown = SharingMode(iota) // For older versions. This will be treated similar to cutsom
+	SharingModeCustom = SharingMode(iota)
 	SharingModeInherited
 	SharingModeEmpty
-	SharingModeCustom
 )
 
 // Collection represents a set of OneDrive objects retrieved from M365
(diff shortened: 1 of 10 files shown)


meain:
The PR should be good to go if the design is good.

src/internal/connector/onedrive/drive.go
```diff
@@ -170,6 +170,9 @@ func defaultItemPager(
 		link,
 		[]string{
 			"content.downloadUrl",
+			// https://learn.microsoft.com/en-us/onedrive/developer/rest-api
+			// /concepts/scan-guidance#scanning-permissions-hierarchies
+			"oneDrive.sharedChanged",
```
ryanfkeepers, line 175: Does this cover sharepoint as well?

src/internal/connector/onedrive/permission.go
```diff
@@ -113,58 +102,52 @@ func createRestoreFoldersWithPermissions(
...
-// getChildPermissions is to filter out permissions present in the
-// parent from the ones that are available for child. This is
-// necessary as we store the nested permissions in the child. We
-// cannot avoid storing the nested permissions as it is possible that
-// a file in a folder can remove the nested permission that is present
-// on itself.
-func getChildPermissions(
-	childPermissions, parentPermissions []UserPermission,
+func getPermissionDiff(
+	after, before []UserPermission,
```
ryanfkeepers, line 112: My OCD is so confused.  Should before come before after?  Or should it maintain alphabetical order?

my brain!

ryanfkeepers (approved):
Some nits/questions, but generally looks good

src/internal/connector/onedrive/collection.go
```diff
@@ -53,6 +53,13 @@ var (
 	_ data.StreamModTime    = &metadataItem{}
 )
 
+type SharingMode int
+
+const (
+	SharingModeCustom = SharingMode(iota)
+	SharingModeInherited
+)
```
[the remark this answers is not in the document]
meain: - I don't have a strong opinion on `string` vs `int`. I choose `int` mainly because it is lesser bytes (plus iota felt more like using enums)
- I don't mind switching to strings. As you mentioned, they are much more readable. **Although, any opinion on using `TitleCase` instead of `snake_case`?**
- About `SharingModeUnknown`, I initially had it, but removed as treating older backups without permissions as `SharingModeCustom` is the "right" choice anyways. Plus, the "omitempty" in the struct will drop `SharingModeCustom` even as of now.
- Another possibility is to just convert it to a boolean as there are only two options. I created it as an "enum" as we had more options(`SharingModeEmpty`) initially. I don't think we will need more options here, but I just didn't want to take a risk.

src/internal/connector/onedrive/collection.go
```diff
@@ -212,7 +219,12 @@ type UserPermission struct {
 // ItemMeta contains metadata about the Item. It gets stored in a
 // separate file in kopia
 type Metadata struct {
-	FileName    string           `json:"filename,omitempty"`
+	FileName string `json:"filename,omitempty"`
+	// SharingMode denotes what the current mode of sharing is for the object.
+	// - inherited: permissions same as parent permissions (no "shared" in delta)
+	// - empty: remove all permissions (empty "shared" in delta)
```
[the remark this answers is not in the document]
meain: `empty` here refers to empty from graph side. On an related note, I forgot to update the comment. We now treat  "empty" same as custom but with no permissions.

src/internal/connector/onedrive/drive.go
meain: Good questions. I did not check it for SharePoint as we still don't have permissions for SharePoint. This was added for `@microsoft.graph.sharedChanged`, but we are not actively using it as of now.

src/internal/connector/onedrive/permission.go
meain: Haha, thanks for bringing this up. I had it in my todo list to change, but missed it. I too think should be `before` and then `after`.

src/internal/connector/onedrive/collection.go
```diff
@@ -53,6 +53,13 @@ var (
 	_ data.StreamModTime    = &metadataItem{}
 )
 
+type SharingMode int
```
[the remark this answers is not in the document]
meain: It is used in test in `graph_connector_onedrive_test.go`

src/internal/connector/onedrive/drive.go
meain: I'm removing it for now, will add it to the TODO to be picked up later.

src/internal/connector/onedrive/collection.go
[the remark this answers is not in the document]
ashmrtn: >I don't think we will need more options here, but I just didn't want to take a risk

seems reasonable

>They tend to be better for readability

This is a limitation of golang, not the approach of using ints for enums. Also Abin has a good point that the number of bytes can matter. We haven't really been worrying about this too much to this point (after all, we're using JSON encoding for a lot of things...), but in storage and networking it's pretty common to try to minimize the number of bytes that need to be stored/sent as much as possible. That's why enums in those codebases often end up as integers instead of strings

Overall, I'm hesitant to switch persisted enum values to strings (I'm a bit less worried about non-persisted enums as we can change those however much we want since they don't affect backwards compatibility). Int enums have nice default values that we can associate with things and add meaning to. For example, having 0 be the "Unknown" (or whatever default we need it to be) value works well to catch things that weren't populated to begin with or things that weren't set explicitly. Strings don't have the equivalent as having a "meaningful" name for it would imply `const UnknownType = "unknown_type"` which is different from `""` that will be returned for unpopulated values. Having `const UnknownType = ""` or `const LegacyType = ""` is equally unhelpful if we assume we're just printing these values as an empty string in the log is unhelpful; putting us right back in the situation we said string enums would help us avoid

src/internal/connector/onedrive/collections.go
```diff
@@ -622,6 +622,14 @@ func (c *Collections) UpdateCollections(
...
 		)
 
+		// TODO(meain): Use `@microsoft.graph.sharedChanged` in
+		// item.GetAdditionalData() to optimize fetching
+		// permissions. When permissions change, this flags is
+		// present, if only data changes, it is not present.  Have to
+		// add `oneDrive.sharedChanged` in `$select` in delta
+		// https://learn.microsoft.com/en-us/onedrive/developer/rest-api
+		// /concepts/scan-guidance#scanning-permissions-hierarchies
+
```
ashmrtn, line 632: is this TODO stale?

meain: Nope, this is still pending. I need to do a few tests to be certain of the behaviour. Will add it as a followup later.

src/internal/connector/onedrive/collection.go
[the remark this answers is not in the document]
meain: > This is a limitation of golang, not the approach of using ints for enums.

I too share a similar opinion.

I think I'll stick with int for now, plus we can always write `Stringer`s if necessary.

Issue #2708, opened 2023-03-06
Per-item failure tracking


ryanfkeepers:
## Issue
Corso needs to track its runtime errors with a serializable struct so that error cases can be formalized into per-item references with information that is valuable to end-users. 

## Proposal
```go
// in pkg fault.  eg: fault.Item{}
type Item struct {
  // deduplication identifier; the ID of the backing item.  
  ID string `json:"id"`
  // a human-readable reference: file/container name, email, etc
  Name string `json:"name"`
  // the name of the container inside which this item is stored.
  // may be empty in certain cases (eg: resourceOwner)
  ContainerName string `json:"containerName"`
  // the id of the container inside which this item is stored.
  // may be empty in certain cases (eg: resourceOwner)
  ContainerID string `json:"containerID"`
  // one of: file | container | resourceOwner
  Kind string `json:"kind"`
  // Error() of the causal error, or a sentinel if this is the
  // source of the error.  In case of ID collisions, the first
  // item takes priority.
  Cause string `json:"cause"`
}

func (fi *Item) Error() string {
  // => "processing <Kind>"
}
```

## Acceptance Criteria
- [x] Failed items should generate fault.Item structs as errors.
- [x] Operations consumers (ie: cli, sdk) should have a means of serializing items errors following an operation Run().
- [x] fault.Items should serialize into the backup metadata.
- [x] users should be able to retrieve the list of failed items from a backup by ID.
- [x] users should be able to review errors, item failures, and skips from the `backup list --backup x` command.

PR #2710, opened 2023-03-06
introduce fault items

ryanfkeepers:
Adds the item struct to the fault package for tracking serializable and dedupliatable error sources.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/fault/item.go
+++ b/src/pkg/fault/item.go
@@ -8,6 +8,8 @@ const (
 	ResourceOwnerType itemType = "resource_owner"
 )
 
+var _ error = &Item{}
+
 // Item contains a concrete reference to a thing that failed
 // during processing.  The categorization of the item is determined
 // by its Type: file, container, or reourceOwner.
@@ -54,7 +56,11 @@ func (i *Item) Error() string {
 	return string("processing " + i.Type)
 }
 
-// ContainerErr constructs a Container-type Item.
+// ---------------------------------------------------------------------------
+// Constructors
+// ---------------------------------------------------------------------------
+
+// ContainerErr produces a Container-type Item.
 func ContainerErr(cause error, id, name, containerID, containerName string) *Item {
 	return &Item{
 		ID:            id,
--- a/src/pkg/fault/item.go
+++ b/src/pkg/fault/item.go
@@ -127,7 +127,7 @@ type Skipped struct {
 	item Item
 }
 
-// String complies with the stringer interface. func (s *Skipped) String() string {
+// String complies with the stringer interface.
 func (s *Skipped) String() string {
 	if s == nil {
 		return "<nil>"
--- a/src/pkg/fault/item.go
+++ b/src/pkg/fault/item.go
@@ -8,17 +8,27 @@ const (
 	ItemKindResourceOwner itemKind = "resource_owner"
 )
 
-// in pkg fault.  eg: fault.Item{}
+// Item contains a concrete reference to a thing that failed
+// during processing.  The categorization of the item is determined
+// by its Kind: file, container, or reourceOwner.
+//
+// Item is compliant with the error interface so that it can be
+// aggregated with the fault bus, and deserialized using the
+// errors.As() func.  The idea is that fault,Items, during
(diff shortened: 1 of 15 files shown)


src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
+package fault
+
+type itemKind string
```
dadams39, line 3: I understand that we already have `itemType`. However, it does seem to work in this connotation. There are other things that may be more useful:
- class
- division
- group

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+
+const (
+	ItemKindFile          itemKind = "file"
+	ItemKindContainer     itemKind = "container"
+	ItemKindResourceOwner itemKind = "resource_owner"
```
dadams39, line 8: It's unclear why `Kind` is required within the constant name. 
- `ItemFile`
- `ItemContainer`

flows better or `GroupFile`, `CatFile`

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// in pkg fault.  eg: fault.Item{}
```
dadams39, line 11: more comments needed for the struct

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+type Item struct {
+	// deduplication identifier; the ID of the item under scrutiny.
```
dadams39, line 13: `ID of the observed item`

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+	ID string `json:"id"`
+
+	// a human-readable reference: file/container name, email, etc
+	Name string `json:"name"`
+
+	// the name and id of the container inside which this item is
+	// stored.  May be empty if the item is not stored within a
+	// container (ex: resourceOwner)
```
dadams39, line 21: This is comment is unclear.

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+	ContainerName string `json:"containerName"`
+	ContainerID   string `json:"containerID"`
+
+	// tracks the type of item represented by this entry.
+	Kind itemKind `json:"kind"`
```
dadams39, line 26: Classification?

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Error() of the causal error, or a sentinel if this is the
+	// source of the error.  In case of ID collisions, the first
+	// item takes priority.
+	Cause string `json:"cause"`
```
dadams39, line 31: Love this

src/pkg/fault/item.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if len(i.Kind) == 0 {
+		return "processing item of unknown kind"
+	}
+
+	return string("processing " + i.Kind)
+}
+
+// ContainerErr produces a Container-kind Item.
```
dadams39, line 47: produces or returns?

src/pkg/fault/item.go
ryanfkeepers: Just a reference to its underlying type, and a way to namespace it from other similarly named consts.

src/pkg/fault/item.go
ryanfkeepers: * `class` too overloaded a term in software
* `division` makes me think that this is a divider, which it is not.
* `group` could work, but imo it reads identically to `kind`.

src/pkg/fault/item.go
ryanfkeepers: constructs 😁

src/pkg/fault/item.go
ryanfkeepers: swapped `kind` for `type`.

dadams39 (approved):

PR #2712, opened 2023-03-06
add item skips to fault

ryanfkeepers:
extends the fault package to track skiped items
as well as erroneous items.  Skipped items are
assumed to be permanently unprocessable due
to well-known conditions.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/fault/fault.go
+++ b/src/pkg/fault/fault.go
@@ -140,8 +140,16 @@ func (e *Bus) addRecoverableErr(err error) *Bus {
 	return e
 }
 
-// AddSkip appends the skipped item to the slice of skipped
-// items (ie: bus.skipped).
+// AddSkip appends a record of a Skipped item to the fault bus.
+// Importantly, skipped items are not the same as recoverable
+// errors.  An item should only be skipped under the following
+// conditions.  All other cases should be handled as errors.
+// 1. The conditions for skipping the item are well-known and
+// well-documented.  End users need to be able to understand
+// both the conditions and identifications of skips.
+// 2. Skipping avoids a permanent and consistent failure.  If
+// the underlying reason is transient or otherwise recoverable,
+// the item should not be skipped.
 func (e *Bus) AddSkip(s *Skipped) *Bus {
 	if s == nil {
 		return e
@@ -251,6 +259,16 @@ func (e *localBus) AddRecoverable(err error) {
 	e.bus.AddRecoverable(err)
 }
 
+// AddSkip appends a record of a Skipped item to the local bus.
+// Importantly, skipped items are not the same as recoverable
+// errors.  An item should only be skipped under the following
+// conditions.  All other cases should be handled as errors.
+// 1. The conditions for skipping the item are well-known and
+// well-documented.  End users need to be able to understand
+// both the conditions and identifications of skips.
+// 2. Skipping avoids a permanent and consistent failure.  If
+// the underlying reason is transient or otherwise recoverable,
+// the item should not be skipped.
 func (e *localBus) AddSkip(s *Skipped) {
 	if s == nil {
 		return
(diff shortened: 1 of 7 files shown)


src/pkg/fault/item.go
```diff
@@ -83,3 +83,71 @@ func OwnerErr(cause error, id, name string) *Item {
...
 }
+
+// ---------------------------------------------------------------------------
+// Skipped Items
+// ---------------------------------------------------------------------------
+
+type skipCause string
+
+const SkipMalware skipCause = "malware_detected"
+
```
dadams39, line 94: Should there be another cause for unknown or unresolvable?

ryanfkeepers: Nope.  We shouldn't be skipping unknown cases.  Those should just be errors.  Skips should only occur for well-known, well documented conditions.

ryanfkeepers:
> For the most part, everything looks good. Looking at future concerns, it feels like for HTTP timeouts or credentials, we would want the ability to skip based on those items and come back to them later.

@dadams39 I'll add a comment to clarify this, but: skipping should only be used for well known, permanently unrecoverable situations.  Not to be confused with recoverable errors.  Transient failures (ex: timeouts), unknown causes, things of that nature should not create skips.  Those are just errors.

src/pkg/fault/fault.go
```diff
@@ -216,6 +259,27 @@ func (e *localBus) AddRecoverable(err error) {
 	e.bus.AddRecoverable(err)
 }
 
+// AddSkip appends a record of a Skipped item to the local bus.
+// Importantly, skipped items are not the same as recoverable
+// errors.  An item should only be skipped under the following
+// conditions.  All other cases should be handled as errors.
+// 1. The conditions for skipping the item are well-known and
+// well-documented.  End users need to be able to understand
+// both the conditions and identifications of skips.
```
dadams39, line 268: Perhaps, a statement like: `well-known` issues can be found in the /docs/Well-known.md` or something similar.

src/pkg/fault/item.go
```diff
@@ -10,17 +10,7 @@ const (
...
-// during processing.  The categorization of the item is determined
-// by its Type: file, container, or reourceOwner.
-//
-// Item is compliant with the error interface so that it can be
-// aggregated with the fault bus, and deserialized using the
-// errors.As() func.  The idea is that fault,Items, during
-// processing, will get packed into bus.AddRecoverable (or failure)
-// as part of standard error handling, and later deserialized
-// by the end user (cli or sdk) for surfacing human-readable and
-// identifiable points of failure.
```
dadams39, line 23: Where did the description the struct go?

src/pkg/fault/fault.go
```diff
@@ -24,6 +24,11 @@ type Bus struct {
 	// we'd expect to see 1 error added to this slice.
 	recoverable []error
 
+	// skipped is the accumulation of skipped items.  Skipped items
+	// are not errors themselves, but instead represent some permanent
+	// inability to process an item, due to a well-known cause.
```
dadams39, line 29: Instead of redefining Skip in several locations, point back to the details in some way.
`@see /src/pkf/fault/fault.go:Skip` or whatever the idiomatic go way is for comments.

dadams39 (approved):
Mostly nits surrounding comments. No blockers.

src/pkg/fault/fault.go
ryanfkeepers: I'll look into where we might want to document these cases.

src/pkg/fault/item.go
ryanfkeepers: huh, no idea

src/pkg/fault/fault.go
ryanfkeepers: Eh, that's not very user friendly.  I would not, myself, want to look up other comments by reference.  The idiomatic way is to have the necessary information available where it's needed.

And we can always adjust this in the future, if wanted.

PR #2717, opened 2023-03-07
Only set repo configuration on init

vkamra:
Setting the repository policy calls `policy.SetPolicy` in Kopia which involves repository scan. This is an
expensive operation if the repository has a lot of snapshots.
 
This commit changes Corso behavior to only call this on repository init and not on every connect.

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/conn.go
+++ b/src/internal/kopia/conn.go
@@ -90,13 +90,22 @@ func (w *conn) Initialize(ctx context.Context) error {
 		return clues.Wrap(err, "initialzing repo").WithClues(ctx)
 	}
 
-	return w.commonConnect(
+	err = w.commonConnect(
 		ctx,
 		cfg.KopiaCfgDir,
 		bst,
 		cfg.CorsoPassphrase,
 		defaultCompressor,
 	)
+	if err != nil {
+		return err
+	}
+
+	if err := w.setDefaultConfigValues(ctx); err != nil {
+		return clues.Stack(err).WithClues(ctx)
+	}
+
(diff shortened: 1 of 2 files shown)


src/internal/kopia/conn.go
```diff
@@ -90,13 +90,22 @@ func (w *conn) Initialize(ctx context.Context) error {
...
 		bst,
 		cfg.CorsoPassphrase,
 		defaultCompressor,
 	)
+	if err != nil {
+		return err
+	}
+
+	if err := w.setDefaultConfigValues(ctx); err != nil {
+		return clues.Stack(err).WithClues(ctx)
```
ryanfkeepers, line 105: Since you passed in ctx, you can assume that the called function sets `WithClues(ctx)` for you.
```suggestion
		return err
```

ryanfkeepers (approved):

ashmrtn:
we may need to revisit this later. It now means we can't update the config values at all because we never try to set them once the repo is created

vkamra:
> we may need to revisit this later. It now means we can't update the config values at all because we never try to set them once the repo is created

Yep you're right. Didn't want to design that yet till we had a use case - but we could trigger "on-demand" as needed (either as a connect option or a new CLI option)

Issue #2697, opened 2023-03-02
Permissions for a folder are not restored if it does not contain any files


meain:
This is probably a regression from #2604. As of now, if a folder does not contain any files (empty or just folders), the permissions for it are not restored. This seems to be because we restore only collections with items and we restore permissions only for the collections that we explicitly restore.

PR #2718, opened 2023-03-07
Augment paths for OneDrive restore

meain:
This adds the missing .dirmeta files which are needed to make sure that permissions are being restored properly.

By adding a dirmeta entry for each folder that is part of the items, we make sure that all folders get their own collection which in turn will help us properly restore permissions for all folders.

**This also makes it so that meta items are not returned from older backups.**

*Although the newer version can fetch permissions file using the Fetch API, we still have to populate the items as it cannot fetch items from previous directories which is needed to make sure we can restore permissions correctly.*

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/2697

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -599,3 +599,71 @@ func getMetadata(metar io.ReadCloser) (Metadata, error) {
 
 	return meta, nil
 }
+
+// Augment restore path to add extra files(meta) needed for restore as
+// well as do any other ordering operations on the paths
+func AugmentRestorePaths(backupVersion int, paths []path.Path) ([]path.Path, error) {
+	colPaths := map[string]path.Path{}
+	for _, p := range paths {
+		for {
+			np, err := p.Dir()
+			if err != nil {
+				return nil, err
+			}
+
+			onedrivePath, err := path.ToOneDrivePath(np)
+			if err != nil {
+				return nil, err
+			}
+
+			if len(onedrivePath.Folders) == 0 {
+				break
+			}
+
+			_, found := colPaths[np.String()]
+			if !found {
+				colPaths[np.String()] = np
+			}
+			p = np
+		}
+	}
+
+	// Adds dirmeta files as we need to make sure collections for all
+	// directories involved are created and not just the final one. No
+	// need to add `.meta` files (metadata for files) as they will
+	// anyways be looked up automatically.
+	// TODO: Stop populating .dirmeta for newer versions once we can
+	// get files from parent directory via `Fetch` in a collection.
+	// As of now look up metadata for parent directories from a
+	// collection.
+	for _, p := range colPaths {
+		el := p.Elements()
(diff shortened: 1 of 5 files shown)


src/internal/connector/onedrive/restore_test.go
```diff
@@ -0,0 +1,152 @@
...
+		name    string
+		version int
+		input   []string
+		output  []string
+	}{
+		{
+			name:    "no change v0",
+			version: 0,
+			input: []string{
+				"file.txt.data",
```
ashmrtn, line 33: just for my paranoia can you add an additional file that has no `.data` suffix to this test case and the v0 test case below? That would help check that
1. the code doesn't depend on the `.data` suffix in some way
2. we give it input without the `.data` suffix which is what we'll get in v0 backups in most cases (unless the user has something with `.data`)

This is totally one of those "what ways do the tests assume the code may be broken" sort of things that I get paranoid about 😅

src/internal/connector/onedrive/restore_test.go
```diff
@@ -0,0 +1,152 @@
...
+				p, err := path.FromDataLayerPath(base+ps, true)
+				require.NoError(t, err, "creating path")
+
+				outPaths = append(outPaths, p)
+			}
+
+			actual, err := AugmentRestorePaths(test.version, inPaths)
+			require.NoError(t, err, "augmenting paths")
+
+			assert.Equal(t, outPaths, actual, "augmented paths")
```
ashmrtn, line 149: nit: can you add a comment here saying that ordering is important in the output. Otherwise it may get mistakenly changed to `ElementsMatch` by someone who needs to update the tests down the road

ashmrtn (approved):

src/internal/connector/onedrive/restore.go
```diff
@@ -599,3 +599,71 @@ func getMetadata(metar io.ReadCloser) (Metadata, error) {
...
+			}
+
+			if len(onedrivePath.Folders) == 0 {
+				break
+			}
+
+			_, found := colPaths[np.String()]
+			if !found {
+				colPaths[np.String()] = np
+			}
```
ryanfkeepers, line 626: Nit/curiosity: any reason not to let this clobber?
```suggestion
			colPaths[np.String()] = np
```

src/internal/connector/onedrive/restore_test.go
meain: Ahh, good catch!

src/internal/connector/onedrive/restore.go
meain: You are right, it  doesn't matter. I think this is an artifact from me testing stuff out.

Issue #2709, opened 2023-03-06
Create different item types for sharepoint data categories


ashmrtn:
SharePoint should follow the pattern in Exchange and have different `ItemType`s defined for things like Lists, Pages, and Libraries. This will allow switching behavior based on the item type and also verifying that only items of the expected type are being handled

Possible issues:
* dealing with previous backups that used a single item type for all details items. Since SharePoint is considered experimental still it may be alright to break backwards compatibility

PR #2720, opened 2023-03-07
GC: SharePoint: ItemType Decouple

dadams39:
ItemTypes for SharePoint were `SharePointItem` only. This did not allow for data to be separated by category. 
Newly created categories:
- SharePointLibrary
- SharePointList
- SharePointPage

Breaks connectivity. All previous items are available during `Corso List` command. However, when using the details command with the identified backup ID, the return is:
```
no items match the specified selectors
```
---

- [x] 📝 : Will need documentation

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* related to #2709<issue>

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -403,7 +403,9 @@ const (
 	ExchangeEvent
 	ExchangeMail
 
-	SharePointItem ItemType = iota + 100
+	SharePointLibrary ItemType = iota + 100
+	SharePointList    ItemType = iota + 100
+	SharePointPage    ItemType = iota + 100
 
 	OneDriveItem ItemType = iota + 200
 
@@ -416,7 +418,7 @@ func UpdateItem(item *ItemInfo, repoPath path.Path) error {
 	var updatePath func(path.Path) error
 
 	switch item.infoType() {
-	case SharePointItem:
+	case SharePointLibrary:
 		updatePath = item.SharePoint.UpdateParentPath
 	case OneDriveItem:
 		updatePath = item.OneDrive.UpdateParentPath
--- a/src/internal/connector/sharepoint/api/pages.go
+++ b/src/internal/connector/sharepoint/api/pages.go
@@ -263,7 +263,7 @@ func PageInfo(page models.SitePageable, size int64) *details.SharePointInfo {
 	)
 
 	return &details.SharePointInfo{
-		ItemType: details.SharePointItem,
+		ItemType: details.SharePointPage,
 		ItemName: name,
 		Created:  created,
 		Modified: modified,
--- a/src/internal/connector/sharepoint/listInfo.go
+++ b/src/internal/connector/sharepoint/listInfo.go
@@ -18,7 +18,7 @@ func sharePointListInfo(lst models.Listable, size int64) *details.SharePointInfo
 	)
 
 	return &details.SharePointInfo{
-		ItemType: details.SharePointItem,
+		ItemType: details.SharePointList,
 		ItemName: name,
 		Created:  created,
 		Modified: modified,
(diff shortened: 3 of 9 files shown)


src/pkg/backup/details/details.go
```diff
@@ -403,7 +403,9 @@ const (
 	ExchangeEvent
 	ExchangeMail
 
-	SharePointItem ItemType = iota + 100
+	SharePointLibrary ItemType = iota + 100
+	SharePointList    ItemType = iota + 100
+	SharePointPage    ItemType = iota + 100
```
ryanfkeepers, line 408: Well, apparently I didn't fully grok iota when I set this up.  The current change will update both sharepoint (a benign breaking change) _and_ onedrive (an actual breaking change).  see: https://go.dev/play/p/0kmJBLzJ7ME

We should probably update this whole declaration set to the following (with some additional comments for future clarity):


```suggestion
// Item typing is enumerated by service (hundredth digit) and data type (ones digit).
// Ex: exchange is 00x where x is the data type.  Sharepoint is 10x, and etc.
// Every item info struct should get its own hundredth enumeration entry.
// Every item category for that service should get its own entry (even if differences
// between types aren't apparent on initial implementation, this future-proofs
// against breaking changes).  
const (
	UnknownType ItemType = iota // 0, global unknown value

	ExchangeContact
	ExchangeEvent
	ExchangeMail
)

const (
        SharePointLibrary ItemType = iota + 100
	SharePointList    ItemType
	SharePointPage    ItemType
)
	
// 205 is a result of prior bad use of iota.
// if we introduce future oneDrive types, they
// should increment from here.  eg: 205, 206...
const OneDriveItem ItemType = 205

// 306 is a result of prior bad use o iota.
// if we introduce future folder types, they
// should increment from here.  eg: 306, 307...
//
// The next additional service should start
// enumerating at iota+400
const FolderItem ItemType = 306
```

ashmrtn: as I posted elsewhere, I think we may be over-complicating this a bit. For things that already have numbers assigned, we need to keep those numbers (with the exception of sharepoint as this PR updates it). Otherwise, the assignment of numbers to services is a bit immaterial imo. The reason for enums is so that we don't have to remember random numbers

Sure, not having contiguous numbers for data types in a service makes things a bit less pretty to read if anyone happens to dump the raw JSON, but honestly I'm not sure it's even possible to print the JSON in Corso without filtering out some fields. You could print it using kopia, but at that point you're so far down the stack that having to do a little number translation is the least of your worries. But having a rule of "append new data types to the end of the enum set" and having a comment telling people not to rearrange the order of the enum set feels harder to accidentally misuse

ashmrtn (approved):
apart from numbering that may need to be fixed to ensure OD and folders keeps working (didn't dig into it too much) LGTM. Debate about how we should actually do the numbering isn't technically blocking, just about setting expectations for future use

src/pkg/backup/details/details.go
ryanfkeepers: Totally fine with me.  Lets move it to concrete ints and drop the iota?

ashmrtn: having the iota is still fine as it makes appending new entries to the list dead simple. We just need to make sure we don't disturb prior values that need to be fixed. We could just
1. fixup SP numbers and leave their declarations where they are
2. for future additions, just add them to after the folder declaration

ryanfkeepers: Specifying an int const isn't dead simple?

ashmrtn: it can be, but I'm always afraid I'll accidentally type the wrong number and not notice (happens more than I'd like 😅 )

dadams39: Resovled with [0ccde14](https://github.com/alcionai/corso/pull/2720/commits/0ccde1491fe68459b3becb816cc3022d02e7c589)

PR #2723, opened 2023-03-07
refactor streamstore for generic use

ryanfkeepers:
Refactors the stream store so that it can comply with multiple different writers and readers beyond just the details.Details struct.  This will allow us to use the stream store to ingest large lists of errors fault items without burdening the backup manifest.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/streamstore/details.go
```diff
@@ -0,0 +1,159 @@
+package streamstore
```
ryanfkeepers, line 1: Moved here from streamstore.go.  Some changes involved, a bit of code tidying, and then compliance with the new interfaces.

ryanfkeepers: Decided to split that into its own PR.  See https://github.com/alcionai/corso/pull/2725 first.

vkamra (approved):

src/internal/streamstore/streamstore.go
```diff
@@ -12,14 +12,37 @@ import (
...
+// Streamer is the core interface for all types of data streamed to and
+// from the store.
+type Streamer interface {
+	Writer
+	Reader
+	Delete(context.Context, string) error
+}
+
+type Reader interface {
+	Read(context.Context, string, Unmarshaller, *fault.Bus) error
```
ashmrtn, line 28: I can't quite put my finger on why, but the fact that `Read` also performs the unmarshalling makes me uncomfortable. Maybe it's because it seems like it will make debugging unmarshall errors a bit harder (need to check whether it was an error unmarshalling the data or before trying to unmarshall the data). Maybe it's because it feels like it will make streaming data bit more difficult (i.e. does all the stream processing now have to be part of the `Unmarshaller` callback?)

PR #2733, opened 2023-03-08
GC: OneDrive: Efficient Drive Name Update

### LINKED ISSUE #2732: GC: OneDrive: Reduce calls for Drive Update
Information obtained regarding drive metadata was removed due to information to inefficiency. Metadata is still required. Reintroduce method to obtain the metadata without each potential thread having to reobtain the information. 

----
- Related:
- https://github.com/alcionai/corso/pull/2727
- #2632

dadams39:
Updates the call for `populateItems()` for OneDrive. Reduces the overall amount of calls made to obtain the Drive Name. 
---
### Brief Description
- DriveName is saved with a collection field driveMap. 
- The driveMap is a map where the key is the M365ID and the value is the display name for the drive
- During `populateItems()`, the driveName the map is sufficient for OneDrive but requires an update for SharePoint. During SharePoint process, the map is updated as additional driveLibraries are referenced. 

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix
#### Issue(s)

* closes  #2732<issue>

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -354,18 +354,7 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				"backup_item_size", itemSize,
 			)
 
-			if oc.source == SharePointSource {
-				l.Lock()
-
-				pr, err = updateParentReference(ctx, oc.service, item.GetParentReference(), driveMap, l)
-				if err != nil {
-					el.AddRecoverable(clues.Wrap(err, "getting parent reference").Label(fault.LabelForceNoBackupCreation))
-					return
-				}
-			} else {
-				pr = updateParentReferenceOneDrive(item.GetParentReference(), oc.driveName)
-			}
-
+			pr = updateParentReferenceOneDrive(item.GetParentReference(), oc.driveName)
 			item.SetParentReference(pr)
 
 			isFile := item.GetFile() != nil
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -336,7 +336,6 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				itemID       = ptr.Val(item.GetId())
 				itemName     = ptr.Val(item.GetName())
 				itemSize     = ptr.Val(item.GetSize())
-				pr           models.ItemReferenceable
 				itemInfo     details.ItemInfo
 				itemMeta     io.ReadCloser
 				itemMetaSize int
@@ -350,8 +349,7 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				"backup_item_size", itemSize,
 			)
 
-			pr = updateParentReferenceOneDrive(item.GetParentReference(), oc.driveName)
-			item.SetParentReference(pr)
+			item.SetParentReference(setName(item.GetParentReference(), oc.driveName))
 
 			isFile := item.GetFile() != nil
 
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -340,10 +340,12 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				itemID       = ptr.Val(item.GetId())
 				itemName     = ptr.Val(item.GetName())
 				itemSize     = ptr.Val(item.GetSize())
+				pr           models.ItemReferenceable
 				itemInfo     details.ItemInfo
 				itemMeta     io.ReadCloser
 				itemMetaSize int
 				metaSuffix   string
+				err          error
 			)
 
 			ctx = clues.Add(ctx,
@@ -352,12 +354,16 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				"backup_item_size", itemSize,
 			)
 
-			l.Lock()
+			if oc.source == SharePointSource {
+				l.Lock()
 
-			pr, err := updateParentReference(ctx, oc.service, item.GetParentReference(), driveMap, l)
-			if err != nil {
-				el.AddRecoverable(clues.Wrap(err, "getting parent reference").Label(fault.LabelForceNoBackupCreation))
-				return
+				pr, err = updateParentReference(ctx, oc.service, item.GetParentReference(), driveMap, l)
+				if err != nil {
+					el.AddRecoverable(clues.Wrap(err, "getting parent reference").Label(fault.LabelForceNoBackupCreation))
+					return
+				}
+			} else {
+				pr = updateParentReferenceOneDrive(item.GetParentReference(), oc.driveName)
 			}
 
 			item.SetParentReference(pr)
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -290,15 +290,13 @@ func (od *metadataItem) ModTime() time.Time {
 // and uses the collection `itemReader` to read the item
 func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 	var (
-		byteCount   int64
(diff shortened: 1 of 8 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -65,7 +65,10 @@ type Collection struct {
 	folderPath path.Path
 	// M365 IDs of file items within this collection
 	driveItems map[string]models.DriveItemable
-	// M365 ID of the drive this collection was created from
+	// Map of M365 Drive/Names associated with collection
+	// key M365ID: Value DriveName
+	driveMap map[string]string
```
vkamra, line 70: A collection always belongs to the same Drive - shouldn't this just be a `driveName` string. 

Also - should we just query and persist this in the `Collections` struct? Collections is also 1:1 with a Drive

vkamra (changes requested):
Changes requested

src/internal/connector/onedrive/collection.go
vkamra: We should just query this metadata when we call `drives`

vkamra: https://github.com/alcionai/corso/blob/main/src/internal/connector/onedrive/collections.go#L275

dadams39: The collection does not always belong to the same Drive. `sharePoint` has several different drives associated.

vkamra: I understand - but does each collection belong to multiple drives?

dadams39: Personal OneDrives are 1:1
SharePoint  OneDrives are 1 collection to Many
In some cases, it is one. But sometimes it can be more than that.

dadams39: With the changes applied, there is one additional call per existing drive for SharePoint. 

For OneDrive, there are no additional calls.
I tested by running this with the `Corso` binary and viewing the logs that were added each time the call is requested.

vkamra: > For a site, a developer can choose to create an additional document library which will result in an additional drive. For OneDrive, a user can create an additional document library; however, it is not discovered during backup. These newly created drives are part of the shared space which we do not have coverage for as of yet.

I understand this - but a collection (which is a folder in a drive) - only belongs to a single drive - correct?

dadams39: Commit [db8bd41](https://github.com/alcionai/corso/pull/2733/commits/db8bd41783d48e9fe7912911344b5241822e799e) creates an efficient pathway for `OneDrive` or any other features where we know there is only one available document library.

src/internal/connector/onedrive/item.go
```diff
@@ -410,33 +411,81 @@ func constructWebURL(adtl map[string]any) string {
...
+
+		drives[driveID] = name
+		mu.Unlock()
+		logger.Ctx(ctx).Infof("Drive Name not in given map: Retrieved: %s", name)
+	}
+
+	orig.SetName(&name)
+
+	return orig, nil
+}
```
vkamra, line 478: It's not clear to me why we're calling `fetchDriveName` here? Don't we already have the drive name in the driveMap?

vkamra: @dadams39 and I synced 1:1 on this. We should be able to just stash the driveName in the `onedrive.Collection` and use it issuing any other queries.

vkamra (changes requested):
@dadams39 and I synced 1:1 on this to clarify what I was saying. A OneDrive collection always represents a folder in a specific drive (in either onedrive or sharepoint cases) and therefore we should be able to stash the driveName in the collection and use it where needed (without any other queries).

@dadams39 - we also need a regression test for this PR (maybe update `TestItemReader_oneDrive` to check that the ItemInfo includes the drivename.

src/internal/connector/onedrive/collection.go
dadams39: There is a bug within the `UpdateCollections`. Working to resolve

dadams39: Issue #2740 filed

src/internal/connector/onedrive/item.go
dadams39: A bug within `UpdateCollections` caused incorrect population of collections when multiple drives were involved.  Fixes within [f87066e](https://github.com/alcionai/corso/pull/2733/commits/f87066ec3f9062f3ee02f159c15c820c83b1b99d)

src/internal/connector/onedrive/item.go
```diff
@@ -416,33 +416,15 @@ func constructWebURL(adtl map[string]any) string {
...
-// 	drive, err := service.Client().DrivesById(driveID).Get(ctx, options)
-// 	if err != nil {
-// 		return nil, graph.Stack(ctx, err)
-// 	}
-
-// 	orig.SetName(drive.GetName())
-
-// 	return orig, nil
-// }
+func updateParentReferenceOneDrive(
```
ryanfkeepers, line 419: Nit: naming
```suggestion
func updateDriveItemParentRef(
```

src/internal/connector/onedrive/collection.go
```diff
@@ -72,8 +72,11 @@ type Collection struct {
 	folderPath path.Path
 	// M365 IDs of file items within this collection
 	driveItems map[string]models.DriveItemable
-	// M365 ID of the drive this collection was created from
-	driveID        string
+
+	// Primary M365 ID of the drive this collection was created from
+	driveID string
+	// Display Name of the associated drive
+	driveName      string
```
ryanfkeepers, line 79: quick check, are the driveName and driveID getting added to the clues ctx anywhere?  Quick search on my current branch doesn't turn up anything.  We should probably add those somewhere.  Might come in handy for future debugging.

ryanfkeepers (approved):

ashmrtn (approved):
lgtm with a few nits assuming it merges after #2760

src/internal/connector/onedrive/collections.go
```diff
@@ -341,27 +348,32 @@ func (c *Collections) Get(
...
 
 		if !delta.Reset {
-			maps.Copy(excludedItems, excluded)
+			eidi, ok := excludedItems[driveID]
+			if !ok {
+				eidi = map[string]struct{}{}
+			}
+
+			maps.Copy(eidi, excluded)
+			excludedItems[driveID] = eidi
```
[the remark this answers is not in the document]
ryanfkeepers: done

src/internal/connector/onedrive/collections.go
```diff
@@ -352,23 +357,31 @@ func (c *Collections) Get(
 				c.source)
 			if err != nil {
 				return nil, nil,
-					clues.Wrap(err, "making exclude prefix for drive").WithClues(ctx).With("drive_id", driveID)
+					clues.Wrap(err, "making exclude prefix").WithClues(ctx)
```
ashmrtn, line 360: so many context switches ;)
```suggestion
					clues.Wrap(err, "making exclude prefix").WithClues(ictx)
```

src/internal/connector/onedrive/collections_test.go
```diff
@@ -1860,6 +1892,10 @@ func (suite *OneDriveCollectionsSuite) TestGet() {
 			c.drivePagerFunc = drivePagerFunc
 			c.itemPagerFunc = itemPagerFunc
 
+			for driveID := range test.items {
+				c.CollectionMap[driveID] = map[string]*Collection{}
+			}
+
```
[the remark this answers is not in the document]
ryanfkeepers: Ah yeah, good catch. That was a leftover from trying to fix a different panic.

src/internal/connector/onedrive/collections.go
ryanfkeepers: so many rebase collisions! 🎉 

(thanks for the catch)

PR #2734, opened 2023-03-08
quick refactor for item metadata extensibility

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 	"github.com/pkg/errors"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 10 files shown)


src/pkg/backup/backup.go
```diff
@@ -49,9 +49,6 @@ type Backup struct {
 	// stats are embedded so that the values appear as top-level properties
 	stats.ReadWrites
 	stats.StartAndEndTime
-
-	// Deprecated
-	stats.Errs // replaced with fault.Errors
```
vkamra, line 54: I'm assuming this doesn't cause unmarshal of older serialized backup models to fail. I think the default behavior is that missing fields are just ignored

src/internal/connector/graph/errors.go
```diff
@@ -313,3 +314,32 @@ func appendIf(a []any, k string, v *string) []any {
 
 	return append(a, k, *v)
 }
+
+// MalwareInfo gathers potentially useful information about a malware infected
+// drive item, and aggregates that data into a map.
+func MalwareInfo(item models.DriveItemable) map[string]any {
```
vkamra, line 320: Minor - use const for `created_by` and other fields. Add a simple unit test to validate this method

vkamra (approved):

src/pkg/backup/backup.go
ryanfkeepers: Yup, shouldn't cause any serialization errors.  Worst case is that old backups lose the count on `Completed (X errors)`

src/internal/connector/graph/errors.go
ryanfkeepers: done

PR #2744, opened 2023-03-09
store fault errors produced during backup

ryanfkeepers:
Now that streamstore has a controller for consuming fault errors, we can store the errors results following a backup.
---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2708

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -123,6 +123,7 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		opStats      backupStats
 		startTime    = time.Now()
 		detailsStore = streamstore.NewDetails(op.kopia, op.account.ID(), op.Selectors.PathService())
+		errorsStore  = streamstore.NewFaultErrors(op.kopia, op.account.ID(), op.Selectors.PathService())
 	)
 
 	op.Results.BackupID = model.StableID(uuid.NewString())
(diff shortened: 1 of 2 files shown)


src/internal/operations/backup.go
```diff
@@ -655,30 +657,37 @@ func (op *BackupOperation) persistResults(
...
 	}
 
-	detailsID, err := detailsStore.Write(ctx, backupDetails, op.Errors)
+	detailsID, err := detailsStore.Write(ctx, backupDetails, errs)
 	if err != nil {
-		return clues.Wrap(err, "creating backupDetails model").WithClues(ctx)
+		return clues.Wrap(err, "creating backupDetails persistence").WithClues(ctx)
+	}
+
+	errorsID, err := errorsStore.Write(ctx, errs.Errors(), errs)
```
ashmrtn, line 677: both writing and possibly adding to `errs` is probably a bad idea. What if we serialized things as we went along and somehow we ended up adding to `errs` in the call to persist `errs`?

ashmrtn (approved):

src/internal/operations/backup.go
ryanfkeepers: `errs.Errors()` is a wholly separate snapshot (not kopia) of data relative to the errs Bus.  Worst case scenario is we write to `errs` within the streamStore, and don't persist that error.  Not too different from your standard "persist an error with a client that can error" problem.

ashmrtn: >errs.Errors() is a wholly separate snapshot (not kopia) of data relative to the errs Bus.

That assumes a very specific implementation of `errs.Errors()` (specifically the current one that makes a snapshot of the already existing errors). Separating where errors are collected during the call from what errors are being serialized is a more future-proof solution because changing how `errs.Errors()` is marshalled won't cause unexpected failures here

PR #2745, opened 2023-03-09
remove getDetailsIDByBackup func

ryanfkeepers:
Quck cleanup to make future changes more manageable. There's no need for an interface that retrieves exported values from concrete structs.  Maintaining this pattern will cause future toil, especially as we add more fields to be exported.  The toil can be reduced by expecting callers to get the base item and extract the needed values themselves.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/common.go
+++ b/src/internal/operations/common.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/clues"
 	"github.com/alcionai/corso/src/internal/model"
 	"github.com/alcionai/corso/src/internal/streamstore"
 	"github.com/alcionai/corso/src/pkg/backup"
@@ -20,18 +21,23 @@ func getBackupAndDetailsFromID(
 	detailsStore streamstore.Reader,
 	errs *fault.Bus,
 ) (*backup.Backup, *details.Details, error) {
-	dID, bup, err := ms.GetDetailsIDFromBackupID(ctx, backupID)
+	bup, err := ms.GetBackup(ctx, backupID)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "getting backup details ID")
 	}
 
 	var (
-		deets details.Details
-		umt   = details.UnmarshalTo(&deets)
+		deets     details.Details
+		umt       = details.UnmarshalTo(&deets)
+		detailsID = bup.DetailsID
 	)
 
-	if err := detailsStore.Read(ctx, dID, umt, errs); err != nil {
-		return nil, nil, errors.Wrap(err, "getting backup details data")
+	if len(detailsID) == 0 {
+		return bup, nil, clues.New("no details in backup").WithClues(ctx)
+	}
+
+	if err := detailsStore.Read(ctx, detailsID, umt, errs); err != nil {
+		return nil, nil, errors.Wrap(err, "reading backup details")
 	}
 
 	return bup, &deets, nil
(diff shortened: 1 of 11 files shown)


src/internal/operations/common.go
```diff
@@ -20,18 +21,23 @@ func getBackupAndDetailsFromID(
...
 	)
 
-	if err := detailsStore.Read(ctx, dID, umt, errs); err != nil {
-		return nil, nil, errors.Wrap(err, "getting backup details data")
+	if len(detailsID) == 0 {
+		return bup, nil, clues.New("no details in backup").WithClues(ctx)
+	}
+
+	if err := detailsStore.Read(ctx, detailsID, umt, errs); err != nil {
+		return nil, nil, errors.Wrap(err, "reading backup details")
```
ashmrtn, line 40: nit: should be clues wrap?
```suggestion
		return nil, nil, clues.Wrap(err, "reading backup details")
```

src/pkg/repository/repository.go
```diff
@@ -338,27 +338,35 @@ func (r repository) BackupsByTag(ctx context.Context, fs ...store.FilterOption)
...
-	dID, b, err := sw.GetDetailsIDFromBackupID(ctx, model.StableID(backupID))
+	b, err := sw.GetBackup(ctx, model.StableID(backupID))
 	if err != nil {
 		return nil, nil, errs.Fail(err)
 	}
 
+	detailsID = b.DetailsID
+
+	if len(detailsID) == 0 {
+		return nil, b, errs.Fail(clues.New("no details in backup").WithClues(ctx))
```
ashmrtn, line 360: do we want backup ID here too?

ashmrtn (approved):

src/internal/operations/common.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/clues"
```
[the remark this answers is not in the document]
ryanfkeepers: jeez, I know.  This used to catch it for me.  What gives?

src/internal/operations/common.go
```diff
@@ -20,18 +21,23 @@ func getBackupAndDetailsFromID(
...
-		umt   = details.UnmarshalTo(&deets)
+		deets     details.Details
+		umt       = details.UnmarshalTo(&deets)
+		detailsID = bup.DetailsID
 	)
 
-	if err := detailsStore.Read(ctx, dID, umt, errs); err != nil {
-		return nil, nil, errors.Wrap(err, "getting backup details data")
+	if len(detailsID) == 0 {
+		return bup, nil, clues.New("no details in backup").WithClues(ctx)
```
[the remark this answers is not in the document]
ryanfkeepers: Since the backupID (and a ctx) gets passed in by a caller, this code assumes that it's already stashed in the CTX.  I'll double check that it is, just in case.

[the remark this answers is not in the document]
ryanfkeepers: Yup: it's there

src/internal/operations/common.go
ryanfkeepers: sixes, since we aren't adding structured data, it isn't necessary one way or the other.  And I still haven't made an "everything should be clues" pass on the code.

src/pkg/repository/repository.go
ryanfkeepers: done

PR #2747, opened 2023-03-09
add printable interface to items and skips

ryanfkeepers:
Extends the fault items and fault skipped structs with print.Printable interface compliance.  This will allow us to print those values to the terminal from a fault.Errors struct.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2708

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/fault/fault.go
+++ b/src/pkg/fault/fault.go
@@ -7,9 +7,10 @@ import (
 	"io"
 	"sync"
 
-	"github.com/alcionai/corso/src/cli/print"
 	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
+
+	"github.com/alcionai/corso/src/cli/print"
 )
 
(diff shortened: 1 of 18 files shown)


src/pkg/fault/fault.go
```diff
@@ -1,11 +1,13 @@
...
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"io"
 	"sync"
 
+	"github.com/alcionai/corso/src/cli/print"
 	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
```
ashmrtn, line 12:
```suggestion
	"golang.org/x/exp/maps"
	"golang.org/x/exp/slices"

	"github.com/alcionai/corso/src/cli/print"
```

ashmrtn (approved):

PR #2749, opened 2023-03-09
Fixup OneDrive selectors for different file names

ashmrtn:
Update selectors to match on the OneDrive/SharePoint
file name specified in backup details instead of the
path. Path name matching is disabled completely as the
name in kopia is never displayed to the user

Plan is to merge this PR into #1535 and then merge into
main when that's ready so this all goes in at once

Manually tested restoring a OneDrive file by file name
and had no trouble

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #1535

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/onedrive.go
+++ b/src/pkg/selectors/onedrive.go
@@ -3,6 +3,8 @@ package selectors
 import (
 	"context"
 
+	"github.com/alcionai/clues"
+
 	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/fault"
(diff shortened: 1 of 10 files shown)


src/pkg/selectors/scopes.go
```diff
@@ -351,7 +351,11 @@ func reduce[T scopeT, C categoryT](
 			continue
 		}
 
-		pv := dc.pathValues(repoPath, *ent)
+		pv, err := dc.pathValues(repoPath, *ent)
+		if err != nil {
+			el.AddRecoverable(clues.Wrap(err, "getting path values").WithClues(ctx))
```
ryanfkeepers, line 356: it occurs to me that we don't have any specific item identification for errors added in this loop.  I can't comment up on that line, but could you add something like this on line 311?
```go
ictx := clues.Add(ctx, "short_ref", ent.ShortRef)
```
and then propagate that inner context within the other calls.
```suggestion
			el.AddRecoverable(clues.Wrap(err, "getting path values").WithClues(ictx))
```

src/pkg/selectors/sharepoint.go
```diff
@@ -476,30 +478,44 @@ func (c sharePointCategory) isLeaf() bool {
...
+
 	case SharePointList, SharePointListItem:
 		folderCat, itemCat = SharePointList, SharePointListItem
+
 	case SharePointPage, SharePointPageFolder:
 		folderCat, itemCat = SharePointPageFolder, SharePointPage
+
 	default:
-		return map[categorizer][]string{}
+		return nil, clues.New("bad sharePointCategory").With("category", c)
```
ryanfkeepers, line 506: nit
```suggestion
		return nil, clues.New("unrecognized sharePointCategory").With("category", c)
```

ryanfkeepers (approved):
only nits, lgtm

src/pkg/selectors/onedrive.go
```diff
@@ -376,20 +378,27 @@ func (c oneDriveCategory) isLeaf() bool {
...
+		return nil, clues.New("no OneDrive ItemInfo in details")
+	}
+
 	// Ignore `drives/<driveID>/root:` for folder comparison
 	rFld := path.Builder{}.Append(repo.Folders()...).PopFront().PopFront().PopFront().String()
 
 	result := map[categorizer][]string{
 		OneDriveFolder: {rFld},
-		OneDriveItem:   {repo.Item(), ent.ShortRef},
+		OneDriveItem:   {ent.OneDrive.ItemName, ent.ShortRef},
```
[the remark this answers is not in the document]
ashmrtn: `ItemName` [looks like](https://github.com/alcionai/corso/commit/c4e5915983e2098fd4de01131a400bacf222d585#diff-cd4ad81a441fe5b1a43835fb95436f2a797ad246312beb00704340a57d810c65R126) it's been populated from the very beginning so if it's empty it's probably an error somewhere rather than a legacy feature

PR #2760, opened 2023-03-10
Scope excluded items by prefix

### LINKED ISSUE #2759: Scope OneDrive excluded files to each drive
#2740 shows that we can have items with the same ID in different drives. While we haven't confirmed that non-root items can have the same IDs, a good change to make would be to scope OneDrive items to exclude from an incremental backup to the drive they belong to. This would prevent accidentally "deleting" files from incremental backups if they are in a different drive and just happen to have the same ID

This can be accomplished by expanding the kopia exclude list to take a `map[string]map[string]struct{}` and use the first map instance to have folder prefixes. The kopia logic should select the map with the longest prefix for a given directory

On the OneDrive side, additions to the map will need updated to include a partial prefix for the path. The prefix should include at least up until the drive ID

ashmrtn:
In kopia select the longest prefix's exclude set. Has undeterministic
behavior if there are somehow prefixes of the same length.

In OneDrive, add a prefix that contains the drive ID to all excludes.
This makes incrementals safe even if two items in different drives
somehow have the same ID.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* closes #2759

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collections.go
+++ b/src/internal/connector/onedrive/collections.go
@@ -345,7 +345,18 @@ func (c *Collections) Get(
 			numDeltas)
 
 		if !delta.Reset {
-			excludedItems[driveID] = excluded
+			p, err := GetCanonicalPath(
+				fmt.Sprintf(rootDrivePattern, driveID),
+				c.tenant,
+				c.resourceOwner,
+				c.source)
+			if err != nil {
+				return nil, nil,
+					clues.Wrap(err, "making exclude prefix for drive").WithClues(ctx).With("drive_id", driveID)
+			}
+
+			excludedItems[p.String()] = excluded
+
 			continue
 		}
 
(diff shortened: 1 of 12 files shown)


src/internal/connector/onedrive/collections.go
```diff
@@ -349,7 +345,18 @@ func (c *Collections) Get(
...
+				fmt.Sprintf(rootDrivePattern, driveID),
+				c.tenant,
+				c.resourceOwner,
+				c.source)
+			if err != nil {
+				return nil, nil,
+					clues.Wrap(err, "making exclude prefix for drive").WithClues(ctx).With("drive_id", driveID)
+			}
+
+			excludedItems[p.String()] = excluded
```
ryanfkeepers, line 358: I'm using `driveID` here in my PR.  Assuming all the ids in excluded are unique, could the path creation here get shortened to just that?

ryanfkeepers (approved):

src/internal/connector/onedrive/collections.go
ashmrtn: in theory yes, but that won't play well with kopia because it would turn into a `strings.Contains` calls and that makes me very nervous 😅

PR #2761, opened 2023-03-10
Fix check in sanity test

vkamra:
CLI output was changed from `Completed (0 Errors)` to `Completed`

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] :robot: Test

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/sanity-test.yaml
+++ b/.github/workflows/sanity-test.yaml
@@ -108,12 +108,12 @@ jobs:
           --user "${CORSO_M365_TEST_USER_ID}" \
            --hide-progress 2>&1 | tee $TEST_RESULT/backup_exchange.txt 
 
-          if ! grep -q 'Completed (0 errors)' $TEST_RESULT/backup_exchange.txt
+          if ! grep -q 'Completed' $TEST_RESULT/backup_exchange.txt
           then
               echo "backup was not successfull"
               exit 1
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity-test.yaml
```diff
@@ -108,12 +108,12 @@ jobs:
           --user "${CORSO_M365_TEST_USER_ID}" \
            --hide-progress 2>&1 | tee $TEST_RESULT/backup_exchange.txt 
 
-          if ! grep -q 'Completed (0 errors)' $TEST_RESULT/backup_exchange.txt
+          if ! grep -q 'Completed' $TEST_RESULT/backup_exchange.txt
```
ryanfkeepers, line 111: I think we should check on the presence of errors here.  `Completed (99 errors)` is probably not a case we want to pass.
```suggestion
          if grep -q 'errors' $TEST_RESULT/backup_exchange.txt
```

ryanfkeepers (approved):
approved to unblock

.github/workflows/sanity-test.yaml
vkamra: I think a better way to go here is going to be to switch to `--json` output and parse the structured output. Going to file a ticket for that and we can address this follow up there.

Issue #2757, opened 2023-03-10
GC: Libraries: Multi-Drive Support


dadams39:
`SharePoint.Libaries` supports multiple document libraries. 
## Backup
- [x] Ability to backup all libraries
- [x] Ability to backup a targeted library
## Restore
The restore workflow for Graph Connector currently has one folder where all items are restored. 
The current workflow needs to be tested to ensure that the output behavior is correct. 
### PreRequesite
- Drive name to become part of the folder naming convention
### Desired behavior
- Restored drives from both drives should be saved in one folder within the main drive, `Documents`. 
- The Corso Restore folder should possess a folder for each drive

PR #2774, opened 2023-03-13
rename libraries -> libraryFolder

ryanfkeepers:
The selector moniker "libraries" is incorrect.  It should refer to the library folder, since it only
matches on directory structures within a given
drive.  The 'library' is analogous to the drive
itself, and will need a separate selector of its
own.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #2757

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/selectors/sharepoint.go
```diff
@@ -390,14 +390,14 @@ const (
...
-	SharePointListItem    sharePointCategory = "SharePointListItem"
-	SharePointLibrary     sharePointCategory = "SharePointLibrary"
-	SharePointLibraryItem sharePointCategory = "SharePointLibraryItem"
-	SharePointPageFolder  sharePointCategory = "SharePointPageFolder"
-	SharePointPage        sharePointCategory = "SharePointPage"
+	SharePointWebURL        sharePointCategory = "SharePointWebURL"
+	SharePointSite          sharePointCategory = "SharePointSite"
+	SharePointList          sharePointCategory = "SharePointList"
+	SharePointListItem      sharePointCategory = "SharePointListItem"
+	SharePointLibraryFolder sharePointCategory = "SharePointLibraryFolder"
```
ashmrtn, line 397: is this part of the selector persisted or used for metrics? If it is, do we need to worry about backwards compatibility for it?

If we can't change the enum value that's alright as long as we know the correspondence. Changing the enum name will at least help

ashmrtn (approved):

PR #2783, opened 2023-03-13
2757 cli 2

ryanfkeepers:
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2757

#### Test Plan

- [x] :zap: Unit test


src/cli/utils/sharepoint_test.go
```diff
@@ -180,13 +180,6 @@ func (suite *SharePointUtilsSuite) TestIncludeSharePointRestoreDataSelectors() {
 			},
 			expectIncludeLen: 2,
 		},
-		{
```
vkamra, line 183: @ryanfkeepers - why is this removed?

vkamra:
@ryanfkeepers - LGTM. There's no PR description but assuming that's because it's a PR train. I used the commit descriptions for review. Does need lint fixes.

vkamra (approved):

src/cli/utils/sharepoint_test.go
ryanfkeepers: It isn't actually covered within that test.  I plugged it in with the last PR, then forgot to take it out when I realized it was the wrong test.

PR #2791, opened 2023-03-14
Get OneDrive item with getM365

meain:
Add support for onedrive in getM365

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/cmd/getM365/onedrive/get_item.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(out)
+
+	return nil
+}
+
+func serialize(data item) (string, error) {
```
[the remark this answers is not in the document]
meain: That does a simple json marshalling, I think we'll need more than just that.

src/internal/connector/graph/service.go
```diff
@@ -29,6 +29,7 @@ import (
 const (
 	logGraphRequestsEnvKey    = "LOG_GRAPH_REQUESTS"
 	log2xxGraphRequestsEnvKey = "LOG_2XX_GRAPH_REQUESTS"
+	log2xxGraphResponseEnvKey = "LOG_2XX_GRAPH_RESPONSES"
```
meain, line 32: I've added an option to log the response for 2xx entries as well. This can be a separate toggle that is enabled only when needed, but could come in handy for debugging failing of follow up requests like in the case of mail attachment fetch failure, but the request to get the mail succeeds.

src/cmd/getM365/onedrive/get_item.go
```diff
@@ -0,0 +1,221 @@
...
+
+		it.Data.Size = len(content)
+	}
+
+	out, err := serialize(it)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(out)
```
ryanfkeepers, line 130: prefer print or log instead of fmt.

ryanfkeepers (approved):

src/cmd/getM365/onedrive/get_item.go
```diff
@@ -37,132 +36,119 @@ var (
...
-	it.Permissions = perms
-
-	if item != nil {
-		content, err := getDriveItemContent(item)
-		if err != nil {
-			return err
-		}
-
-		it.Data.Size = len(content)
-	}
```
meain: While we can get size from `item.GetSize()`, this call was to ensure that we are able to download the file. I'm adding this back.

meain:
@ryanfkeepers Thanks for cleaning up the code. Also, the reason for not using `serializeObject` initially was to avoid it printing json strings for values. I've fixed it by re-encoding it to `json.RawMessage`.

PR #2793, opened 2023-03-14
append ToCore to all errors tests

ryanfkeepers:
In order to retrieve all clues structured error data in tests, we need to extract it from the error using the clues library.

This change appends `clues.ToCore(err)` to all
variations of `assert.NoError(t, err)`.  The only
other changes are those necessary to preserve
linting, or to produce an error variable for the
ToCore call.

---

- [x] :no_entry: No

#### Type of change

- [x] :robot: Test

#### Issue(s)

* #1970

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange_e2e_test.go
+++ b/src/cli/backup/exchange_e2e_test.go
@@ -429,7 +429,7 @@ func (suite *PreparedBackupExchangeE2ESuite) TestExchangeListCmd_badID() {
 
 			// run the command
 			err := cmd.ExecuteContext(ctx)
-			require.NoError(t, err, clues.ToCore(err))
+			require.Error(t, err, clues.ToCore(err))
 		})
 	}
 }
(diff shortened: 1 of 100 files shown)


ashmrtn (approved):
is it possible to make a lint rule for this? Do we want a lint rule for this?

ryanfkeepers:
> is it possible to make a lint rule for this? Do we want a lint rule for this?

I don't think a linter is necessary.  ie: it won't _harm_ anything to not have a core err.  And some tests might not even benefit from it.

ryanfkeepers:
> Right now people will have to figure out via review feedback (assuming reviewer knows) or by looking at other tests

I suppose what I mean to say is that I don't view this as a pattern with requires enforcement.  Seems you have a strong belief to the contrary.  I can add it in on a follow-up PR.

PR #2796, opened 2023-03-14
ShortRef uniqueness fixup

ashmrtn:
Ensure OneDrive/SharePoint ShortRefs are unique
after munging by generating with
displayName/kopiaFileName instead of the other
way around. Doing it this way allows us to
piggyback on OneDrive's requirement that all
files and subfolders in a given folder must be
unique

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #1535

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -313,7 +313,19 @@ func (d *Details) add(
 			filename = info.SharePoint.ItemName
 		}
 
-		entry.ShortRef = p.ToBuilder().Append(filename).ShortRef()
+		// Make the new path be all of the display names and then the M365 item ID.
+		// This ensures the path will be unique, thus ensuring the ShortRef will be
+		// unique.
+		//
+		// If we just appended the file display name to the path then it's possible
+		// the user could have a folder in the parent directory with a display name
+		// equal to the M365 ID of this file and have a file in the folder with a
+		// display name the same as this file's. That would cause a ShortRef
+		// collision because the generated paths for the file in the theoretical
+		// folder and this item would be the same.
+		elements := p.Elements()
+		elements = append(append(elements[:len(elements)-1], filename), p.Item())
+		entry.ShortRef = path.Builder{}.Append(elements...).ShortRef()
 	}
 
 	d.Entries = append(d.Entries, entry)
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -313,18 +313,17 @@ func (d *Details) add(
 			filename = info.SharePoint.ItemName
 		}
 
-		// Make the new path be all of the display names and then the M365 item ID.
+		// Make the new path contain all display names and then the M365 item ID.
 		// This ensures the path will be unique, thus ensuring the ShortRef will be
 		// unique.
 		//
-		// If we just appended the file display name to the path then it's possible
-		// the user could have a folder in the parent directory with a display name
-		// equal to the M365 ID of this file and have a file in the folder with a
-		// display name the same as this file's. That would cause a ShortRef
-		// collision because the generated paths for the file in the theoretical
-		// folder and this item would be the same.
+		// If we appended the file's display name to the path then it's possible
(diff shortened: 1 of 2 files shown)


src/pkg/backup/details/details.go
```diff
@@ -313,7 +313,19 @@ func (d *Details) add(
 			filename = info.SharePoint.ItemName
 		}
 
-		entry.ShortRef = p.ToBuilder().Append(filename).ShortRef()
+		// Make the new path be all of the display names and then the M365 item ID.
```
ryanfkeepers, line 316:
```suggestion
		// Make the new path contain all of the display names plus the M365 item ID.
```

src/pkg/backup/details/details.go
```diff
@@ -313,7 +313,19 @@ func (d *Details) add(
...
+		// This ensures the path will be unique, thus ensuring the ShortRef will be
+		// unique.
+		//
+		// If we just appended the file display name to the path then it's possible
```
ryanfkeepers, line 320:
```suggestion
		// If we appended the file's display name to the path then it's possible
```

src/pkg/backup/details/details.go
```diff
@@ -313,7 +313,19 @@ func (d *Details) add(
...
+		// the user could have a folder in the parent directory with a display name
+		// equal to the M365 ID of this file and have a file in the folder with a
+		// display name the same as this file's. That would cause a ShortRef
+		// collision because the generated paths for the file in the theoretical
+		// folder and this item would be the same.
```
ryanfkeepers, line 325:
```suggestion
		// for a folder in the parent directory to have the same display name
		// as the M365 ID of this file, and also have a file in the folder with a
		// display name that matches the file's. That would result in duplicate
		// ShortRefs, which we can't allow.
```

src/pkg/backup/details/details_test.go
```diff
@@ -469,6 +484,64 @@ func (suite *DetailsUnitSuite) TestDetails_Add_ShortRefs() {
...
+	)
+
+	require.NoError(t, b.Add(
+		itemPath.String(),
+		"deadbeef",
+		itemPath.ToBuilder().Dir().String(),
+		itemPath.String(),
+		false,
+		info,
+	))
```
ryanfkeepers, line 535: To prep for the upcoming change:
```suggestion
        err :=  b.Add(
		itemPath.String(),
		"deadbeef",
		itemPath.ToBuilder().Dir().String(),
		itemPath.String(),
		false,
		info)
	require.NoError(t, err)
```

src/pkg/backup/details/details.go
```diff
@@ -313,7 +313,19 @@ func (d *Details) add(
...
+		elements := p.Elements()
+		elements = append(append(elements[:len(elements)-1], filename), p.Item())
```
ryanfkeepers, line 327: double append not needed
```suggestion
		elements = append(elements[:len(elements)-1], filename, p.Item())
```

ryanfkeepers (approved):
So the basic assumption is that displayName+ID is sufficiently unique, whereas the ID alone isn't?  I'm good with that, just making sure I grok.

ashmrtn:
>So the basic assumption is that displayName+ID is sufficiently unique, whereas the ID alone isn't? I'm good with that, just making sure I grok.

Actually 2 things:
1. we need to make sure ShortRef changes when a file is renamed (display name change) in OneDrive so SDK consumers don't see weird behavior
2. using `<kopia file name>/<display name>` as the final two path elements in the munged ShortRef can lead to collisions with user-generated folders. This would occur because the kopia file name isn't what's checked by OneDrive when the user makes a new item, leading to the situation where we have an item in a folder and then the user makes a subfolder (in the same folder) with display name equal to the item's ID and a subsubfolder with display name equal to the file's display name

PR #2804, opened 2023-03-15
Fix item re-download on unauthorized requests

meain:
We had a regression which caused item re-downloads on 401s from
initial delta URL to be silently skipped. This fixes that behavior.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"github.com/pkg/errors"
 	"github.com/spatialcurrent/go-lazy/pkg/lazy"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
@@ -83,6 +84,7 @@ type Collection struct {
 	source         driveSource
 	service        graph.Servicer
 	statusUpdater  support.StatusUpdater
+	itemGetter     itemGetterFunc
 	itemReader     itemReaderFunc
 	itemMetaReader itemMetaReaderFunc
 	ctrl           control.Options
@@ -98,6 +100,13 @@ type Collection struct {
 	// should only be true if the old delta token expired
 	doNotMergeItems bool
 }
+
+// itemGetterFunc gets an specified item
+type itemGetterFunc func(
+	ctx context.Context,
+	srv graph.Servicer,
+	driveID, itemID string,
+) (models.DriveItemable, error)
 
 // itemReadFunc returns a reader for the specified item
 type itemReaderFunc func(
@@ -146,8 +155,10 @@ func NewCollection(
 	// Allows tests to set a mock populator
 	switch source {
 	case SharePointSource:
+		c.itemGetter = getDriveItem
 		c.itemReader = sharePointItemReader
 	default:
+		c.itemGetter = getDriveItem
 		c.itemReader = oneDriveItemReader
 		c.itemMetaReader = oneDriveItemMetaReader
 	}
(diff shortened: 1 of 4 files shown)


src/internal/connector/onedrive/collection_test.go
```diff
@@ -388,6 +388,111 @@ func (suite *CollectionUnitTestSuite) TestCollectionReadError() {
...
+			defer flush()
+
+			var (
+				t          = suite.T()
+				testItemID = "fakeItemID"
+				collStatus = support.ConnectorOperationStatus{}
+				wg         = sync.WaitGroup{}
+			)
+
+			wg.Add(1)
```
vkamra, line 423: Do we need the wg?

src/internal/connector/onedrive/collection.go
```diff
@@ -35,6 +35,10 @@ const (
 	// TODO: Tune this later along with collectionChannelBufferSize
 	urlPrefetchChannelBufferSize = 5
 
+	// maxRetries specifies the number of times a file download should
+	// be retried
+	maxRetries = 3
```
vkamra, line 40:
```suggestion
	maxDownloadRetries = 3
```

src/internal/connector/onedrive/collection_test.go
meain: I guess it does not matter, but it is good it have. I mostly copied the skeleton from previous tests.

src/internal/connector/onedrive/collection.go
meain: Good call, fixed!

src/internal/connector/onedrive/collection.go
```diff
@@ -286,6 +300,58 @@ func (od *metadataItem) ModTime() time.Time {
...
+			el.AddSkip(fault.FileSkip(fault.SkipMalware, itemID, itemName, graph.MalwareInfo(item)))
+		} else {
+			logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
+			el.AddRecoverable(clues.Stack(err).WithClues(ctx).Label(fault.LabelForceNoBackupCreation))
+		}
+
+		// return err, not el.Err(), because the lazy reader needs to communicate to
+		// the data consumer that this item is unreadable, regardless of the fault state.
+		return nil, err
+	}
```
vkamra, line 350: Should leave this in `populateItems` because if you fail from `itemGetter` above on line 331 - you do want this logic to get invoked as well

vkamra (changes requested):
Looks good - a few changes requested

src/internal/connector/onedrive/collection.go
meain: No, I think the error that we provide there is better. Since we are only retrying for unauthorized `retrieving expired item` is a better error that this to wrap.

vkamra: The issue is that if you return on line 332 above - you will not add the error to the error bus (the logic below is skipped). 

```
			logger.Ctx(ctx).With("error", err.Error()).Error("downloading item")
			el.AddRecoverable(clues.Stack(err).WithClues(ctx).Label(fault.LabelForceNoBackupCreation))
```

meain: I think I'll keep in here itself to consolidate the logic, but I've reworked the logic to make sure we do properly handle it.

vkamra (approved):

PR #2809, opened 2023-03-15
add load tester suite

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E


src/internal/tester/suite.go
```diff
@@ -53,6 +53,7 @@ func NewUnitSuite(t *testing.T) *unitSuite {
 }
 
 type unitSuite struct {
+	//nolint:forbidigo
```
ashmrtn, line 56: these shouldn't be required as there's an exclude rule in .golangci.yaml file for it

src/pkg/repository/loadtest/repository_load_test.go
```diff
@@ -384,38 +384,52 @@ func normalizeCategorySet(t *testing.T, cats map[string]struct{}) []string {
 	return sl
 }
 
+/* ================================================
+*   A note on load test setup:
+*   Even though most of the code here is boiler-
+*   plate and could be easily compressed into a
+*   test matrix, we want to keep the suites separte
```
ashmrtn, line 391:
```suggestion
*   test matrix, we want to keep the suites separate
```

src/pkg/repository/loadtest/repository_load_test.go
```diff
@@ -384,38 +384,52 @@ func normalizeCategorySet(t *testing.T, cats map[string]struct{}) []string {
...
 }
 
-func TestRepositoryLoadTestExchangeSuite(t *testing.T) {
-	tester.RunOnAny(t, tester.CorsoLoadTests)
-	suite.Run(t, new(RepositoryLoadTestExchangeSuite))
+func TestLoadExchangeSuite(t *testing.T) {
+	suite.Run(t, &LoadExchangeSuite{
+		Suite: tester.NewLoadSuite(
+			t,
+			[][]string{tester.AWSStorageCredEnvs, tester.M365AcctCredEnvs},
```
ashmrtn, line 416: should we just make this part of the default check for the load suite in tester package?

ashmrtn (approved):

src/internal/tester/suite.go
ryanfkeepers: s'what I get for going by the ide and not the lint command.

src/pkg/repository/loadtest/repository_load_test.go
ryanfkeepers: not necessarily.  It's a nominal addition, and we may want this granular control for local testing or if/when we introduce non-microsoft platforms.

PR #2813, opened 2023-03-15
CI job to run unit tests without env vars

ashmrtn:
Ensures that new test suites won't accidentally break running `go test` when no/incorrect env vars are given.

Sets M365 users to bad variables so failures there will show up as well

Adds a dep for actions that also have `Test-Suite-Trusted` as a dep

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #2373

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/ci.yml
```diff
@@ -230,6 +230,57 @@ jobs:
           if-no-files-found: error
           retention-days: 14
 
+  Unit-Test-Suite:
+    needs: [Precheck, Checkout]
+    environment: Testing
+    runs-on: ubuntu-latest
+    if: needs.precheck.outputs.srcfileschanged == 'true'
```
ashmrtn, line 237: If I understand this line correctly, this will run for any PR whether or not it's from a fork. Do we need to restrict it again to non-fork PRs so that we don't have to worry about dealing with cache pollution?

.github/workflows/ci.yml
```diff
@@ -230,6 +230,57 @@ jobs:
...
+          go-version-file: src/go.mod
+
+      - run: mkdir testlog
+
+      # Install gotestfmt
+      - name: Set up gotestfmt
+        run: go install github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@latest
+
+      # run the tests
+      - name: Unit Tests
```
meain, line 256: Wondering if we could make `Test-Suite-Trusted` run non unit tests alone so as to not run them twice.

meain (approved):

PR #2816, opened 2023-03-15
Sanity test cleanups and additional info

ryanfkeepers:
updating sanity tests: fix the permissions check by not using reflect.DeepEquals, and various code cleanup.

---

- [x] :no_entry: No

#### Type of change

- [x] :robot: Test

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/sanity-test.yaml
+++ b/.github/workflows/sanity-test.yaml
@@ -76,8 +76,8 @@ jobs:
 
           if ! grep -q 'Initialized a S3 repository within bucket' $TEST_RESULT/initrepo.txt 
           then
-              echo "repo could not be initiated"
-              exit 1
+            echo "repo could not be initiated"
+            exit 1
           fi
           
           echo result="$prefix" >> $GITHUB_OUTPUT 
@@ -93,8 +93,8 @@ jobs:
 
           if ! grep -q 'Connected to S3 bucket' $TEST_RESULT/connect.txt 
           then
-              echo "repo could not be connected"
-              exit 1
+            echo "repo could not be connected"
+            exit 1
           fi
 
       # run the tests
@@ -103,21 +103,21 @@ jobs:
         run: |
           ./corso backup create exchange \
           --user "${CORSO_M365_TEST_USER_ID}" \
-           --hide-progress \
-           --json \
-           2>&1 | tee $TEST_RESULT/backup_exchange.txt 
+          --hide-progress \
+          --json \
+          2>&1 | tee $TEST_RESULT/backup_exchange.txt 
 
           resultjson=$(sed -e '1,/Completed Backups/d' $TEST_RESULT/backup_exchange.txt ) 
 
           if  [[ $( echo $resultjson | jq -r '.[0] | .errorCount') -ne 0 ]]; then
-              echo "backup was not successfull"
-              exit 1
+            echo "backup was not successful"
+            exit 1
           fi
 
           data=$( echo $resultjson | jq -r '.[0] | .id' )
           echo result=$data >> $GITHUB_OUTPUT
 
-      # list all exhange backups
+      # list all exchange backups
       - name: Backup exchange list test
         run: |
           set -euo pipefail
@@ -127,23 +127,23 @@ jobs:
 
           if ! grep -q  ${{ steps.exchange-test.outputs.result }} $TEST_RESULT/backup_exchange_list.txt 
           then
-              echo "listing of backup was not successfull"
-              exit 1
+            echo "listing of backup was not successful"
+            exit 1
           fi
 
-      # list the previous exhange backups
-      - name: Backup exchange list test
+      # list the previous exchange backups
+      - name: Backup exchange list single backup test
         run: |
           set -euo pipefail
           ./corso backup list exchange \
           --hide-progress \
           --backup "${{ steps.exchange-test.outputs.result }}" \
-          2>&1 | tee $TEST_RESULT/backup_exchange_list.txt
+          2>&1 | tee $TEST_RESULT/backup_exchange_list_single.txt
 
           if ! grep -q  ${{ steps.exchange-test.outputs.result }} $TEST_RESULT/backup_exchange_list.txt 
           then
-              echo "listing of backup was not successfull"
-              exit 1
+            echo "listing of backup was not successful"
+            exit 1
           fi
 
       # test exchange restore
@@ -165,7 +165,7 @@ jobs:
           set -euo pipefail
           ./sanityCheck
 
-      # test incremental backup exhange
+      # test incremental backup exchange
       - name: Backup exchange incremental
         id: exchange-incremental-test
         run: |
@@ -173,15 +173,14 @@ jobs:
           ./corso backup create exchange \
           --hide-progress \
           --user "${CORSO_M365_TEST_USER_ID}" \
-          --hide-progress \
           --json \
           2>&1 | tee $TEST_RESULT/backup_exchange_incremental.txt 
 
           resultjson=$(sed -e '1,/Completed Backups/d' $TEST_RESULT/backup_exchange_incremental.txt ) 
 
           if  [[ $( echo $resultjson | jq -r '.[0] | .errorCount') -ne 0 ]]; then
-              echo "backup was not successfull"
-              exit 1
+            echo "backup was not successful"
+            exit 1
           fi
 
           echo result=$( echo $resultjson | jq -r '.[0] | .id' ) >> $GITHUB_OUTPUT
@@ -216,15 +215,14 @@ jobs:
           ./corso backup create onedrive \
           --hide-progress \
           --user "${CORSO_M365_TEST_USER_ID}" \
-          --hide-progress  \
           --json \
           2>&1 | tee $TEST_RESULT/backup_onedrive.txt 
 
           resultjson=$(sed -e '1,/Completed Backups/d' $TEST_RESULT/backup_onedrive.txt ) 
 
           if  [[ $( echo $resultjson | jq -r '.[0] | .errorCount') -ne 0 ]]; then
-              echo "backup was not successfull"
-              exit 1
+            echo "backup was not successful"
+            exit 1
           fi
 
           data=$( echo $resultjson | jq -r '.[0] | .id' )
@@ -240,8 +238,8 @@ jobs:
 
           if ! grep -q  ${{ steps.onedrive-test.outputs.result }} $TEST_RESULT/backup_onedrive_list.txt 
           then
-              echo "listing of backup was not successfull"
-              exit 1
+            echo "listing of backup was not successful"
+            exit 1
           fi
 
       # list the previous onedrive backup
@@ -251,12 +249,12 @@ jobs:
           ./corso backup list onedrive \
           --hide-progress \
           --backup "${{ steps.onedrive-test.outputs.result }}" \
-          2>&1 | tee $TEST_RESULT/backup_onedrive_list.txt 
+          2>&1 | tee $TEST_RESULT/backup_onedrive_list_single.txt
 
           if ! grep -q  ${{ steps.onedrive-test.outputs.result }} $TEST_RESULT/backup_onedrive_list.txt 
           then
-              echo "listing of backup was not successfull"
-              exit 1
+            echo "listing of backup was not successful"
+            exit 1
           fi
 
       # test onedrive restore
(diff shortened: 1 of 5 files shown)


.github/workflows/sanity-test.yaml
```diff
@@ -115,12 +117,28 @@ jobs:
           data=$( echo $resultjson | jq -r '.[0] | .id' )
           echo result=$data >> $GITHUB_OUTPUT
 
-      # list the backup exhange
+      # list all exhange backups
```
neha-Gupta1, line 120: nit-
```suggestion
      # list all exchange backups
```

.github/workflows/sanity-test.yaml
```diff
@@ -115,12 +117,28 @@ jobs:
...
+          2>&1 | tee $TEST_RESULT/backup_exchange_list.txt
+
+          if ! grep -q  ${{ steps.exchange-test.outputs.result }} $TEST_RESULT/backup_exchange_list.txt 
+          then
+              echo "listing of backup was not successfull"
+              exit 1
+          fi
+
+      # list the previous exhange backups
+      - name: Backup exchange list test
```
neha-Gupta1, line 135: Is this test case added to test the `--backup` flag? And if it is so can we rename it to- 
 
```suggestion
      - name: List single exchange backup
```
or something similar

.github/workflows/sanity-test.yaml
```diff
@@ -238,8 +284,11 @@ jobs:
         run: |
           set -euo pipefail
           ./corso backup create onedrive \
+          --hide-progress \
```
neha-Gupta1, line 287: added this flag twice I believe

neha-Gupta1 (approved):
Thank you for all the refactoring. Since I am working on same files some re-basing coming my way!

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
+	}
+
+	if len(init) > 0 {
+		maps.Copy(result, init)
+	}
+
+	// check if permission are correct on folder
+	for _, p := range pcr.GetValue() {
+		var (
+			pid   = ptr.Val(p.GetId())
```
ashmrtn, line 443: I don't know if we can assume that restore will re-use the same permission IDs. It seems like OneDrive will try to reuse them if a permission for the same stuff already exists, but I don't think there's a guarantee. It may be better to compare role+grantedToV2

I think the above also depends a bit on how we want the compare to work. Since both role and grantedToV2 are both slices there's a possibility of checking if we got the same permission sets back from a "structure" perspective (i.e. same roles/grantedTo pairs) or just from a more broad perspective of "all the grantedTo entries that had role X are the same no matter what other roles may have been assigned in the same permission object." The latter would essentially merge grantedTo values across permission objects so that everything could be compared by role

src/pkg/logger/logger.go
```diff
@@ -104,31 +104,40 @@ func PreloadLoggingFlags() (string, string) {
...
-	if len(logfile) == 0 {
-		logfile = os.Getenv("CORSO_LOG_FILE")
+	if len(r) == 0 {
+		r = os.Getenv("CORSO_LOG_FILE")
 	}
 
-	if logfile == "-" {
-		logfile = "stdout"
+	if r == "-" {
+		r = "stdout"
```
ashmrtn, line 128: `stdout` and `stderr` should be consts since they are also consumed in this package to set the output of the logger

ashmrtn (changes requested):

.github/workflows/sanity-test.yaml
ryanfkeepers: yep, it's just the added --backup flag

.github/workflows/sanity-test.yaml
```diff
@@ -256,13 +305,27 @@ jobs:
...
           set -euo pipefail
           ./sanityCheck
+
+      # Upload the original go test output as an artifact for later review.
+      - name: Upload test log
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-log
+          path: src/testlog/*
```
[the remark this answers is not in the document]
ryanfkeepers: CORSO_LOG_FILE matches this path

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -2,99 +2,120 @@ package main
...
+	"golang.org/x/exp/slices"
+
+	"github.com/alcionai/corso/src/internal/common"
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 func main() {
+	ctx, log := logger.Seed(context.Background(), "info", logger.GetLogFile(""))
```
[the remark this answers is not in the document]
ryanfkeepers: Should no longer be necessary, with the other changes we have.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -2,99 +2,120 @@ package main
...
+			if itemName == folderName {
+				restoreFolder = v
 				continue
 			}
 
-			getAllSubFolder(client, testUser, r, name, messageCount)
+			getAllSubFolder(ctx, client, testUser, v, itemName, itemCount)
 
-			messageCount[name], _ = ptr.ValOK(r.GetTotalItemCount())
+			itemCount[itemName] = ptr.Val(v.GetTotalItemCount())
```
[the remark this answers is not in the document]
ryanfkeepers: That's something you should bring up with @neha-Gupta1 to fix in another PR.  I'm not introducing the pattern here.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
 	for _, child := range childFolder.GetValue() {
-		childDisplayName, _ := ptr.ValOK(child.GetDisplayName())
-
-		fullFolderName := parentFolder + "/" + childDisplayName
-
-		childTotalCount, _ := ptr.ValOK(child.GetTotalItemCount())
+		var (
+			childDisplayName = ptr.Val(child.GetDisplayName())
+			childTotalCount  = ptr.Val(child.GetTotalItemCount())
+			fullFolderName   = parentFolder + "/" + childDisplayName
```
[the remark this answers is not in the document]
ryanfkeepers: @neha-Gupta1 I'll fix this up, but for awareness ^

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
-			if startTime.Before(rStartTime) {
-				fmt.Printf("The restore folder %s was created after %s. Will skip check.", *driveItem.GetName(), folderName)
 				continue
 			}
 		}
 
 		// if it's a file check the size
 		if driveItem.GetFile() != nil {
-			file[*driveItem.GetName()] = *driveItem.GetSize()
+			file[itemName] = ptr.Val(driveItem.GetSize())
```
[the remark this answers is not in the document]
ryanfkeepers: I'm not sure either.  @neha-Gupta1 probably needs investigation in a follow-up.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
-				os.Exit(1)
-			}
-
-			// check if permission are correct on folder
-			for _, permission := range permission.GetValue() {
-				folderPermission[*driveItem.GetName()] = permission.GetRoles()
-			}
-
-			continue
+			folderPermission[itemID] = permissionsIn(ctx, client, driveID, itemID, folderPermission[itemID])
```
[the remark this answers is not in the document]
ryanfkeepers: don't see why not

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
+				fmt.Println("-  actual:", itemSize)
+				fmt.Println("Item:", itemName, itemID)
 				os.Exit(1)
 			}
 
 			continue
 		}
 
-		itemBuilder := client.DrivesById(driveID).ItemsById(*restoreData.GetId())
+		if item.GetFolder() == nil {
```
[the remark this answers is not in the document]
ryanfkeepers: I'm honestly not sure.  I suppose we could.

src/cmd/sanity_test/sanity_tests.go
ryanfkeepers: Leaving this for a follow-up. @neha-Gupta1

src/cmd/sanity_test/sanity_tests.go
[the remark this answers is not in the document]
ashmrtn: oops, actually I think I should have removed this comment. I did read more and see messages are stored by path in the recursive functions so I think this one is alright

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -2,99 +2,117 @@ package main
...
+			var (
+				itemID     = ptr.Val(v.GetId())
+				itemName   = ptr.Val(v.GetDisplayName())
+				ictx       = clues.Add(ctx, "item_id", itemID, "item_name", itemName)
+				folderTime = mustGetTimeFromName(ctx, itemName)
+			)
 
-			var rStartTime time.Time
+			// only test against folders within the test boundary time
+			if !errors.Is(err, common.ErrNoTimeString) && startTime.Before(folderTime) {
```
ashmrtn, line 98: I think this needs updated now that we have `mustGetTimeFromName`

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +123,333 @@ func checkEmailRestoration(
...
+		folderTime := mustGetTimeFromName(ictx, itemName)
 
-		restoreStartTime := strings.SplitAfter(*driveItem.GetName(), "Corso_Restore_")
-		if len(restoreStartTime) > 1 {
-			rStartTime, _ = time.Parse(time.RFC822, restoreStartTime[1])
-			if startTime.Before(rStartTime) {
-				fmt.Printf("The restore folder %s was created after %s. Will skip check.", *driveItem.GetName(), folderName)
-				continue
-			}
+		if !errors.Is(err, common.ErrNoTimeString) && startTime.Before(folderTime) {
```
ashmrtn, line 322: also needs updated I think

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +123,333 @@ func checkEmailRestoration(
...
+
+		slices.Sort(roles)
+
+		result[pid] = roles
+	}
+
+	return result
+}
+
+func mustGetTimeFromName(ctx context.Context, name string) time.Time {
```
ashmrtn, line 448: maybe we should just turn this into a function that returns a bool as to whether we should skip the folder or not? That would remove the remaining if-condition in the calling code, contain error handling to this, and make reading the other code a bit easier maybe?

ashmrtn (approved):
some minor nits remaining besides broader questions

src/cmd/sanity_test/sanity_tests.go
ryanfkeepers: yeah, not a bad idea.  Obviously my brain is only half into this.  Thanks for the checks

src/cmd/sanity_test/sanity_tests.go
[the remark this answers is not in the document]
neha-Gupta1: With what I understood from graph API - all the folders were existing in `root` folder. And so the implementation.
But since the test data is getting too huge and due concurrent test runs sanity tests break. So I will be creating custom data and validate only that in coming PR.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -105,260 +126,328 @@ func checkEmailRestoration(
...
-	folderPermission map[string][]string,
+	folderPermission map[string]map[string][]string,
 ) {
-	itemBuilder := client.DrivesById(driveID).ItemsById(restoreFolderID)
-
-	restoreResponses, err := itemBuilder.Children().Get(context.Background(), nil)
+	restored, err := client.
+		DrivesById(driveID).
+		ItemsById(restoreFolderID).
+		Children().
```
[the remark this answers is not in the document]
neha-Gupta1: Will take care of this in coming PR.

src/cmd/sanity_test/sanity_tests.go
neha-Gupta1: Will take care of this in coming PR.

neha-Gupta1: Will take care of this in coming PR.

neha-Gupta1: @ashmrtn - Checks that I will be doing for Onedrive in https://github.com/alcionai/corso/pull/2938
- if file- verify file size matches
- if folder - check permission(role+grantedToV2) on folder matches

Issue #2786, opened 2023-03-14
Sharepoint document libraries support: CLI fit & finish feedback


gmatev:
## corso backup create sharepoint
- [x] Usage is incorrect and should emphasize using web urls to specify the site 
   * `corso backup create sharepoint  --site <site url> | '*' [flags]`
- [x] Fix examples to use URL 
- [x] Remove `--web-url` flag - use dual `site ulr` or `site ID` similar to how this works for users
- [x] `--data` flag should be hidden default to `files` until `pages` and/or `lists` are introduced 
- [x] Remove Pre-release warning
- [x] Progress output 
   * For each library - `* library - <name of library>`. For example, `* library - Documents`
   * `Queuing items` entries should prefix with name of library - e.g. `Queuing items   /Documents 3/3` - more consistent with OneDrive 

## corso backup details sharepoint
- [x] Remove pre-release warning
- [x] Fix examples using `--site` with a UTR
- [x] Fix flags - model after OneDrive. Specifically fix ordering. Hide unsupported flags (e.g. pages)
```
Flags:
      --backup string                 ID of the backup to explore. (required)
      --site string              Select backup details by site url; accepts '*' to select all sites.
      --library string               Select backup details by Library name; defaults to '*'
      --folder                      Select backup details by folder
      --file                          Select backup details by file name or ID.      
      --file-created-after string     Select backup details for files created after this datetime.
      --file-created-before string    Select backup details for files created before this datetime.
      --file-modified-after string    Select backup details for files modified after this datetime.
      --file-modified-before string   Select backup details for files modified before this datetime.
  -h, --help                          help for sharepoint
```

#### Actual display 
- [x] Remove WebURL - repetitive and not useful since backups are one site at a time 
- [x] Rename Drive => Library
- [x] Parity with OneDrive (except for the Library field)

## corso backup details sharepoint
- [x] Remove pre-release warning
- [x] Fix examples - `--site` usage with WebURL. Make example of names more relatable (use the ones from OneDrive)
- [x] Add example with `--library` - use `Documents`
- [x] Fix flags to support the same selectors as `backup details`
- [x] Fix order to match the order for `backup details`
- [x] Fix usage text for flags to be consistent with working from OneDrive
- [x] Hide unsupported flags

ryanfkeepers:
> Remove --web-url flag - use dual site ulr or site ID similar to how this works for users

[...]

PR #2821, opened 2023-03-15
centralize details, restore flags

ryanfkeepers:
Centralizes details and restore flags for all 3
m365 services.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2786

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange.go
+++ b/src/cli/backup/exchange.go
@@ -25,28 +25,7 @@ import (
 
 // exchange bucket info from flags
 var (
-	backupID     string
 	exchangeData []string
-	user         []string
-
-	contact       []string
-	contactFolder []string
-	contactName   string
-
-	email               []string
-	emailFolder         []string
-	emailReceivedAfter  string
-	emailReceivedBefore string
-	emailSender         string
-	emailSubject        string
-
-	event             []string
-	eventCalendar     []string
-	eventOrganizer    string
-	eventRecurs       string
-	eventStartsAfter  string
-	eventStartsBefore string
-	eventSubject      string
 )
 
 const (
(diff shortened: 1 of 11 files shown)


gmatev:
@ryanfkeepers does this preserve the ability to control the flag order?

src/cli/utils/sharepoint.go
```diff
@@ -36,14 +42,65 @@ type SharePointOpts struct {
...
 
+// AddSharePointDetailsAndRestoreFlags adds flags that are common to both the
+// details and restore commands.
+func AddSharePointDetailsAndRestoreFlags(cmd *cobra.Command) {
+	fs := cmd.Flags()
+
+	fs.StringVar(
+		&Library,
+		LibraryFN, "",
+		"Select only this library. Default includes all libraries.")
```
gmatev, line 53:
```suggestion
		"Select only this library; defaults to all libraries")
```
Consistency with flag below

ryanfkeepers:
> does this preserve the ability to control the flag order?

I have no reason to believe it wouldn't.  I can run some checks for proof afterward.

gmatev:
the text tweaks look good to me as long as we have preserved the ordering of the flags. 
I'll let @ashmrtn commend on the other changes.

src/cli/backup/exchange.go
```diff
@@ -108,25 +87,20 @@ func addExchangeCommands(cmd *cobra.Command) *cobra.Command {
...
 			"Select one or more types of data to backup: "+dataEmail+", "+dataContacts+", or "+dataEvents)
 		options.AddOperationFlags(c)
 
 	case listCommand:
-		c, fs = utils.AddCommand(cmd, exchangeListCmd())
-
-		fs.StringVar(&backupID,
-			"backup", "",
-			"ID of the backup to retrieve.")
+		c, _ = utils.AddCommand(cmd, exchangeListCmd())
```
ashmrtn, line 99: seems like you could remove the flag set return value from `AddCommand`?

ashmrtn (approved):

ashmrtn:
maybe eventually we should not use global flags but use flags tied to an instance of a struct? This would allow us some parallelism in tests if we wanted it, though it would take a bit more work to setup

src/cli/backup/exchange.go
ryanfkeepers: it's used in other places, primarily to maintain `fs.SortFlags = false`.  We should keep it in the return set for now.  Can always revisit later.

ryanfkeepers:
> maybe eventually we should not use global flags but use flags tied to an instance of a struct? This would allow us some parallelism in tests if we wanted it, though it would take a bit more work to setup

Yep!  Just hasn't been a hard need as of yet.

PR #2841, opened 2023-03-16
track id and display name for resource owners

ryanfkeepers:
For display purposes, we want to show the site
weburl, not the ID.  however, we normally process
using the site ID.  To do both, we have to pass
both around, using the ID for queries and storage,
and the display name for cli output.

---

- [x] :no_entry: No

#### Issue(s)

* #2786

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -115,6 +115,7 @@ func (gc *GraphConnector) DataCollections(
 		colls, excludes, err := sharepoint.DataCollections(
 			ctx,
 			gc.itemClient,
+			gc.Sites[sels.DiscreteOwner], // siteID -> webURL lookup
 			sels,
 			gc.credentials,
 			gc.Service,
--- a/src/internal/connector/graph_connector.go
+++ b/src/internal/connector/graph_connector.go
@@ -41,7 +41,7 @@ type GraphConnector struct {
 	itemClient *http.Client // configured to handle large item downloads
 
 	tenant      string
-	Sites       map[string]string // key<???> value<???>
+	Sites       map[string]string // webURL -> siteID and siteID -> webURL
 	credentials account.M365Config
 
 	// wg is used to track completion of GC tasks
@@ -310,6 +310,7 @@ func getResources(
 		}
 
 		resources[k] = v
+		resources[v] = k
 
 		return true
 	}
--- a/src/internal/connector/sharepoint/data_collections.go
+++ b/src/internal/connector/sharepoint/data_collections.go
@@ -31,6 +31,7 @@ type statusUpdater interface {
 func DataCollections(
 	ctx context.Context,
 	itemClient *http.Client,
+	ownerName string,
 	selector selectors.Selector,
 	creds account.M365Config,
 	serv graph.Servicer,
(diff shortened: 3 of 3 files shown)


ryanfkeepers:
Before:
```
Discovering items to backup:      1s done
∙ libraries - 10rqc2.sharepoint.com,009d6803-aa00-4db7-8233-e2cbd7d9f503,5408143c-ebd0-4130-82e6-767375686853:      1s done
```

After:
```
Discovering items to backup:      1s done
∙ libraries - https://10rqc2.sharepoint.com/:      1s done
```

src/internal/observe/alias.go
```diff
@@ -0,0 +1,9 @@
+package observe
+
+// IDName is used to pass around a tuple representing
+// some id (such as a resource owner), and some display
+// name that end users would find more easy to consume.
+type IDName struct {
```
ashmrtn, line 6: I feel like from a scalability standpoint (i.e. how far can we push the design for different types/situations) we may be better off using an interface instead of a concrete struct for this and for the Selector/Name struct defined in the CLI layer.

Together, these structs are trying to represent:
* the human readable name for something,
* the ID of a resource so we can get it with Graph
* the other selectors for the resource

By making a set of concrete structs, we assume we can always squeeze all of the above down to what the structs define and we don't leave ourselves an easy way forward if they don't. It also ties generating all of that information to struct creation time (i.e. the struct needs to have those fields set immediately otherwise the struct ceases to be useful) and forces the code that creates the struct to know exactly how to generate the data. The latter especially spreads type-specific logic around the codebase instead of centralizing it (i.e. if both GC and CLI layers need some of the IDName stuff, they both need to know how to make it). We also have a bit of what feels like an anti-pattern where we make a struct and then continue to individually access and pass around separate struct values (i.e. call to `repo.NewBackup`). That spreads the handling of all of these separate things to other components as well

If we use an interface and have implementations for the interface then we can
* delay generating ID/Name values until the functions are called
* centralize the code that needs to know how to generate various values
* fine-tune each data type to have different outputs easily because there's only one place to change things

Just to spitball an interface design here (this need not necessarily be defined in this file, but since I'm already commenting here...)
```go
// idk, can't name things this morning
type ResourceNamer interface {
  // DisplayName returns a human-readable name for the resource
  DisplayName() string
  // ResourceID returns the external service ID for the resource
  ResourceID() string
}

type DisplaySelector interface {
  ResourceNamer
  // Selector returns the selector being used for this operation
  Selector() Selector
}
```

The observe package could use the `ResourceNamer` interface (so we don't give it access to more functions than it needs) and other things could take the `DisplaySelector` interface. That way we continue to pack the data together, but allow accessing all of it through well-defined interfaces

ryanfkeepers: Overall: agree that this is the wrong pattern here, I'm going to work on that.  Beyond that, general disagreements on the particulars.  Primarily that this is an attempt to represent selectors, and that interfaces are an effective solution to this situation.

What's needed is a display name for the resource owner, and I overreached in turning this into a struct.  Simplifying the implementation to the needs at hand will be the best way forward.

ryanfkeepers: @ashmrtn could you kindly check out the new iteration?

ashmrtn:
needs rebased before I can take another look

ashmrtn (approved):
works for now. Long-term, we shouldn't assume that GC has a map of all the sites though, just like we no long assume it has a map of all the users

ryanfkeepers:
> we shouldn't assume that GC has a map of all the sites though, just like we no long assume it has a map of all the users

We should be able to assume that it has the _current_ resource owner's information though.  Otherwise we can't pass validation on the discreteOwner being part of the tenant.  Needs attention for healthy longevity, but mostly as an implementation detail.

PR #2842, opened 2023-03-16
scopes typing rename, filter -> info

ryanfkeepers:
Renames a lot of the names in scopes that contain
the word "filter" with the word "info".  The basis here
is to reduce confusion in the libraries.  Filter as a term
is overloaded.  It currently means: 1/
the set of scopes that comparisons _must_ pass
to succeed; 2/ the pkg and  struct that handles
comparison logic; and 3/  the set of scopes that
compare values in details.ItemInfo instead of in
the item's storage or logical path.  The third
category is the primary overload here, and is
better off with a different name.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test


meain (approved):
Good call, it had kinda tripped me up in the beginning.

src/pkg/selectors/exchange.go
```diff
@@ -492,20 +493,18 @@ const (
...
-	ExchangeFilterContactName        exchangeCategory = "ExchangeFilterContactName"
-	ExchangeFilterEventOrganizer     exchangeCategory = "ExchangeFilterEventOrganizer"
-	ExchangeFilterEventRecurs        exchangeCategory = "ExchangeFilterEventRecurs"
-	ExchangeFilterEventStartsAfter   exchangeCategory = "ExchangeFilterEventStartsAfter"
-	ExchangeFilterEventStartsBefore  exchangeCategory = "ExchangeFilterEventStartsBefore"
-	ExchangeFilterEventSubject       exchangeCategory = "ExchangeFilterEventSubject"
-	// append new filter cats here
+
+	// data contained within details.ItemInfo
+	ExchangeInfoMailSender         exchangeCategory = "ExchangeInfoMailSender"
```
ashmrtn, line 498: just double checking: we never need to deserialize these values to display to the user right? Like we never show the selector that was used for a backup with backup list?

ryanfkeepers: Yes, this is entirely under the hood, and scoped to the current.  No end user display, no re-use after retrieval.

PR #2844, opened 2023-03-17
Update value of absoluteMaxDelay to be based on other constants

meain:
This will make sure this is always kept updated with the timeout changes that w make. This doe not fix the timeouts that we have due to context deadline limit though.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -22,16 +22,21 @@ import (
 )
 
 const (
-	logGraphRequestsEnvKey  = "LOG_GRAPH_REQUESTS"
-	numberOfRetries         = 3
-	retryAttemptHeader      = "Retry-Attempt"
-	retryAfterHeader        = "Retry-After"
-	defaultMaxRetries       = 3
-	defaultDelay            = 3 * time.Second
-	absoluteMaxDelaySeconds = 180
-	rateLimitHeader         = "RateLimit-Limit"
-	rateRemainingHeader     = "RateLimit-Remaining"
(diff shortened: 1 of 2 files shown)


src/internal/connector/graph/service.go
```diff
@@ -22,16 +22,21 @@ import (
...
+	retryAttemptHeader     = "Retry-Attempt"
+	retryAfterHeader       = "Retry-After"
+	defaultMaxRetries      = 3
+	defaultDelay           = 3 * time.Second
+	httpTimeout            = time.Minute * 3
+
+	// absoluteMaxDelay is used to define the maximum delay for all
+	// the retries. Extra 1 minute to account for any other delays
+	// outside of http.Timeout
+	absoluteMaxDelay = defaultMaxRetries*httpTimeout + 1*time.Minute
```
ryanfkeepers, line 35: what happens if someone sets max-retires to, say, 10?  Is this better off as a func instead of a const?

ryanfkeepers (approved):

src/internal/connector/graph/service.go
meain: I think we should still ideally compute it, unless we want to ceil to a max time duration.

meain:
We ended up removing this in https://github.com/alcionai/corso/pull/2888

PR #2848, opened 2023-03-17
fixes, logging improvements for fault.Errors

ryanfkeepers:
A last-second change in 2708 caused us to
pass along the wrong fault.Errors into backup
persistence, thus slicing the count of skipped
items.  That's been fixed, along with improved
end-of-operation logging of fault errors.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #2708

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E


src/pkg/backup/backup.go
```diff
@@ -89,6 +88,8 @@ func New(
 		}
 	}
 
+	fmt.Printf("\n-----\nskipped %+v %v %v\n-----\n", malware, notFound, otherSkips)
+
```
ashmrtn, line 92: guessing this should be removed as it's a debug statement?
```suggestion
```

src/pkg/fault/fault.go
```diff
@@ -291,6 +292,40 @@ func (e *Errors) PrintItems(ctx context.Context, ignoreErrors, ignoreSkips, igno
 	print.All(ctx, sl...)
 }
 
+// Log is a helper function that logs all entries in the Errors struct.
+func (e *Errors) Log(ctx context.Context, prefix string) {
```
ashmrtn, line 296: while I get this is useful, I feel it constrains other code a little bit as well. I think it would be more preferable to return a formatted output set (somehow) that could be used to print to anything rather than just a logger (or take some interface to print to?). For example, what if someone is debugging code where a log is not available and would like to print this info to stdout?

There's also minimal control over the formatting of the printed messages, so it may become difficult to combine this with other log messages down the road

ryanfkeepers: > For example, what if someone is debugging code where a log is not available and would like to print this info to stdout?

In that case couldn't that person print out the struct, or its properties, directly as needed?  Everything in the struct is exported.  Think of this more as boilerplate reduction than an accessor.

> There's also minimal control over the formatting of the printed messages, so it may become difficult to combine this with other log messages down the road

Two way door, I think?  Nothing here is persisted, so we can rework as necessary down the line.  Without having concrete needs for "other log messages" which would break this interface, I'm going to assume there's no immediate obstruction.

src/pkg/backup/backup.go
ryanfkeepers: ayup!

src/pkg/fault/fault.go
ashmrtn: >In that case couldn't that person print out the struct, or its properties, directly as needed?

Yes, but I'm trying to avoid death by thousand cuts here. Same goes for the other comment. We don't know right now how this code will be reused, but we've picked a pretty restrictive interface for using it. That means that if we need to reuse it in unexpected ways going forward either we need to work around the current limitations somehow, which can lead to code that is difficult to maintain, or we need to put in more work to change things so they have a nicer interface. If we can get a less restrictive interface off the bat then we can avoid some of the above

ryanfkeepers: > We don't know right now how this code will be reused, but we've picked a pretty restrictive interface for using it

I guess I don't see this as a problem.  I'm curious if the issue is simply the location?  Suppose I move this func to operations with the signature `LogErrors(ctx, Errors, prefix)`.  Do the same concerns hold up in that case?  That change is incidental to me- since I don't view this method as a critical part of the struct interface- but discarding the relationship might be more palatable for you?

ashmrtn: that would be alright to get this PR merged. You're right that part of what gives me pause with this PR is having this great new interface on a specific object that has all these restrictions and built-in assumptions to it. It breaks the idea of OOP where we generally try to avoid that and make stuff generic. Moving it to the Log code is a bit better because at least the restrictions on output are clearly visible (i.e. clearly it's only for the log code and not for Printf too). The fact that it needs to manually manage the fields of the object is still a bit meh as it's brittle (as opposed to asking the object to hand it what to output), but less concerning than the other way around

ryanfkeepers: Couldn't put the func into /logger, circular deps.  It's now a func in operations.

ashmrtn (approved):

PR #2852, opened 2023-03-17
add 10 day deadline to cli context

ryanfkeepers:
Ensures all CLI usage has a context deadline so that we
don't bump into the graph api 100 second deadline.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -147,8 +147,9 @@ func BuildCommandTree(cmd *cobra.Command) {
 
 // Handle builds and executes the cli processor.
 func Handle() {
+	tenDays := time.Now().Add(time.Hour * 24 * 10)
 	//nolint:forbidigo
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Hour*24*10))
+	ctx, cancel := context.WithDeadline(context.Background(), tenDays)
 	defer cancel()
 
 	ctx = config.Seed(ctx)
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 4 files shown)


vkamra:
> any luck with some manual tests on this even if they are manufactured by making excess Graph API calls to hit throttling limits?

Doing that now

vkamra:
This is not working as expected. The context used in Corso has a deadline that is 10 days out but I am seeing a 3 minute deadline in Graph (different from what we observed earlier when it was 100 seconds). Is the http timeout getting set to 3 minutes?

ashmrtn:
> Is the http timeout getting set to 3 minutes?

yes, it's [set when we create the HTTP client](https://github.com/alcionai/corso/blob/91328d9898093ec2a227151af956a7bfd4f4d739/src/internal/connector/graph/service.go#L202) that we reuse. It should be possible to make a new option to set it to a higher value?

src/cli/cli.go
```diff
@@ -147,7 +148,10 @@ func BuildCommandTree(cmd *cobra.Command) {
 // Handle builds and executes the cli processor.
 func Handle() {
 	//nolint:forbidigo
-	ctx := config.Seed(context.Background())
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Hour*24*10))
```
vkamra, line 151: Nit: Not blocking here - but `context.WithTimeout(context.Background(), time.Hour*24*10)` is equivalent and easier to parse (do check if I'm correct though)

vkamra (approved):

src/cli/cli.go
ryanfkeepers: not equivalent: Deadline takes in a time.Time, not a time.Duration (that's what I initially tried).

vkamra: @ryanfkeepers - to clarify - `WithTimeout` should take a duration.

ryanfkeepers: whoops, missed that, thanks

PR #2864, opened 2023-03-17
Add option to control parallelism getting items

ashmrtn:
This option is only consumed by Exchange right now and will control the number of items fetched in parallel per collection. Note that the total number of parallel item fetches may be higher is kopia is concurrently uploading multiple collections.

This PR does not expose this option at the CLI layer

Manually tested to ensure it
* uses the default if options value is 0
* uses the default if options value > default
* uses option value otherwise

There's another PR open for this but it's not expected to main. We still
want to make sure the change goes in though, hence this PR.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :robot: Test

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/exchange_data_collection.go
+++ b/src/internal/connector/exchange/exchange_data_collection.go
@@ -197,7 +197,21 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 	}
 
 	// Limit the max number of active requests to GC
-	semaphoreCh := make(chan struct{}, urlPrefetchChannelBufferSize)
+	fetchParallelism := col.ctrl.ItemFetchParallelism
+	if fetchParallelism == 0 || fetchParallelism > urlPrefetchChannelBufferSize {
+		fetchParallelism = urlPrefetchChannelBufferSize
+		logger.Ctx(ctx).Infow(
+			"fetch parallelism value not set or out of bounds, using default",
+			"default_parallelism",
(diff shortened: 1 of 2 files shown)


src/internal/connector/exchange/exchange_data_collection.go
```diff
@@ -197,7 +197,21 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 	}
 
 	// Limit the max number of active requests to GC
-	semaphoreCh := make(chan struct{}, urlPrefetchChannelBufferSize)
+	fetchParallelism := col.ctrl.ItemFetchParallelism
+	if fetchParallelism == 0 || fetchParallelism > urlPrefetchChannelBufferSize {
```
ryanfkeepers, line 201: Because you never know.
```suggestion
	if fetchParallelism < 1 || fetchParallelism > urlPrefetchChannelBufferSize {
```

ryanfkeepers (approved):
This parallelism occurs in a number of places.  Is this the last one remaining, or do they all need the same control?

ashmrtn:
>This parallelism occurs in a number of places. Is this the last one remaining, or do they all need the same control?

Eventually they'll all need the same control. For the moment, we were just targeting this item fetch, hence the narrow focus of the PR

PR #2865, opened 2023-03-17
cli: use withTimeout, not withDeadline

ryanfkeepers:
#### Type of change


src/internal/connector/graph/service.go
```diff
@@ -36,6 +36,7 @@ const (
 	rateLimitHeader           = "RateLimit-Limit"
 	rateRemainingHeader       = "RateLimit-Remaining"
 	rateResetHeader           = "RateLimit-Reset"
+	defaultHTTPClientTimeout  = 1 * time.Hour
```
meain, line 39: 1 hour feels a bit too long IMO, any reason why we have such a big timeout? I doubt any request would take >10-15min.

meain: Ahh, I see this change is already in and this is just restructuring the code.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix repo connect not working without a config file
 - Fix item re-download on expired links silently being skipped
 - Improved permissions backup and restore for OneDrive
+- CLI calls default to a 10-day context deadline to avoid letting graph api restrict requests to a 100 second deadline.
```
meain, line 19: Since the CHANGELOG will be user facing, not sure this is how we should phrase it.

meain (approved):

meain:
The new PR seem to cause context deadline issues, let's merge this in for now.

Issue #2330, opened 2023-01-31
[Bug]: `--endpoint` parameter does not accept fully qualified name


gmatev:
### What happened?

```
corso repo init s3 --endpoint=http://127.0.0.1:9000 --bucket foobar
Error: Failed to initialize a new S3 repository: initializing repo: unable to create client: Endpoint url cannot have fully qualified paths.
```

The behavior is unintuitive especially for a user trying to connect to an HTTP endpoint for testing. 

One of the following should happen:
* Option 1:
   * Fix inline help and error message to indicate that `http://` vs `https://` is controlled via the `--disable-tls` flag  
   * Should provide a specific example. Even the example today uses a fully qualified path including the protocol for the endpoint
* Option 2:
   *  Optionally accept `http:// | https://`. If not specified, current behavior (default to `https://` unless `disable-tls` is specified) applies
   * If prefix is `https://` make sure is is not specified in conjunction with `disable-tls` since the two are incompatible
   * If prefix is `http://`, `--disable-tls` is implied but can still be specified
* 

### Corso Version?

Corso v0.2.0

### Where are you running Corso?

OS: MacOS 12.5
Storage: Minio

ntolia:
@gmatev This behavior is inherited from kopia choices and we should just keep those options as-is. Today, by default, we try TLS. We can allow for both insecure TLS (`https` w/o cert validation) and no TLS via flags (`http`). This behavior is consistent across the flags and I don't know if adding a translation layer is worth it at this moment.

gmatev:
We should at least enhance the inline help for the flags to clarify (Option 1). We had multiple people trip on this until we got it right. 

Also someone :) once sent me a Kopia example to use with Minio which did include `--endpoint http://127.0.0.1:9000` so unfortunately it tends to be the natural inclination. Clarifying the error message and the help parameters should help.

ntolia:
@gmatev Yes, I agree with option 1 here. Also, I really wonder who that someone was 🤔.

neha-Gupta1:
`repo connect` command seems to have similar issue. Should I update the error message there too?

PR #2874, opened 2023-03-20
handle endpoint flag to check http:// or https://

neha-Gupta1:
add CLI check to handle http or https protocol.  

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/2330

#### Test Plan

- [ ] :muscle: Manual


src/cli/repo/s3.go
```diff
@@ -73,7 +74,7 @@ corso repo init s3 --bucket my-bucket
 corso repo init s3 --bucket my-bucket --prefix my-prefix
 
 # Create a new Corso repo in an S3 compliant storage provider
-corso repo init s3 --bucket my-bucket --endpoint https://my-s3-server-endpoint`
+corso repo init s3 --bucket my-bucket --endpoint my-s3-server-endpoint`
```
meain, line 77: Haha, good catch!

src/cli/repo/s3.go
```diff
@@ -128,6 +129,14 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
 		return Only(ctx, errors.Wrap(err, "Retrieving s3 configuration"))
 	}
 
+	if strings.Contains(s3Cfg.Endpoint, "http") {
```
meain, line 132: Maybe use `HasPrefix(..., "https?://")` instead of contains. `Contains` can be a bit fragile.

src/cli/repo/s3.go
```diff
@@ -128,6 +129,14 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
...
+		invalidEndpointErr := `endpoint doesn't support fullpath. 
+		pass --disable-tls flag to use http:// instead of default https://`
+
+		return Only(ctx,
+			errors.New(invalidEndpointErr))
```
meain, line 137: nit
```suggestion
		return Only(ctx, errors.New(invalidEndpointErr))
```

src/cli/repo/s3.go
neha-Gupta1: Yupp! also will it apply to `repo connect` too?

meain: Yup, it should be dropped from there too.

src/cli/repo/s3.go
```diff
@@ -129,12 +129,11 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
 		return Only(ctx, errors.Wrap(err, "Retrieving s3 configuration"))
 	}
 
-	if strings.Contains(s3Cfg.Endpoint, "http") {
-		invalidEndpointErr := `endpoint doesn't support fullpath. 
-		pass --disable-tls flag to use http:// instead of default https://`
+	if strings.HasPrefix(s3Cfg.Endpoint, "http") {
```
meain, line 132: While this might be OK, I feel like being more explicit here by including `://` might be a safer option.
```suggestion
	if strings.HasPrefix(s3Cfg.Endpoint, "http://") || strings.HasPrefix(s3Cfg.Endpoint, "https://") {
```

neha-Gupta1: Then I need to check for both http and https. Will do it to stay on safer side

src/cli/repo/s3.go
```diff
@@ -128,6 +129,13 @@ func initS3Cmd(cmd *cobra.Command, args []string) error {
 		return Only(ctx, errors.Wrap(err, "Retrieving s3 configuration"))
 	}
 
+	if strings.HasPrefix(s3Cfg.Endpoint, "http://") || strings.HasPrefix(s3Cfg.Endpoint, "https://") {
+		invalidEndpointErr := "endpoint doesn't support specifying protocol. " +
+			"pass --disable-tls flag to use http:// instead of default https://"
+
+		return Only(ctx, errors.New(invalidEndpointErr))
```
meain, line 136: IIUC, `Only` is supposed to print just the error without a help message, but somehow it seems to be doing this. It is unrelated to this PR, so don't have  to address in this one, but just mentioning to keep track.

meain (changes requested):

src/cli/repo/s3.go
neha-Gupta1: noted!!

meain (approved):
LGTM

Issue #2212, opened 2023-01-20
Corso Runtime Metrics


ryanfkeepers:
## Issue
Corso does not currently track granular metrics such as download/upload counts, request timing, or other significant occurrences (throttling responses, recoverable error frequency, etc).  The general tracking of things that are relevant to construct dashboards or other insights about, in terms of the broad scale behavior and efficiency of the application.

## Acceptance Criteria
- [x] Pick a library to produce/emit statistics.  Utilize our current Telemetry wrapper or event bus (though rudderstack might not be appropriate) if possible.
- [x] Identify which metrics we need to start tracking.  Communicate with SDK users to discover needs as necessary.
- [ ] We need a way to retrieve metrics post run, when telemetry hooks are insufficient for propagating those details.
- [x] Implement metrics tracking.

ryanfkeepers:
closing for now, we can revisit persistence and retrieval later

PR #2884, opened 2023-03-20
metrics PoC for github.com/armon/go-metrics

ryanfkeepers:
Adds a logger-oriented metrics tracker to aggregate statistics like requests-per-minute and average
http request time.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2212

#### Test Plan

- [x] :muscle: Manual


ryanfkeepers:
Average output, currently on a 1 minute interval:

```
16:18:10.07	WARN	[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.total_gc_runs': 13.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.num_goroutines': 292.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.alloc_bytes': 59980688.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.sys_bytes': 194733104.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.malloc_count': 4082038.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.free_count': 3895006.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.heap_objects': 187032.000
[2023-03-20 16:17:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.total_gc_pause_ns': 2041878.000
[2023-03-20 16:17:00 -0600 MDT][C] 'api_call': Count: 251 Sum: 251.000 LastUpdated: 2023-03-20 16:17:59.951317 -0600 MDT m=+109.898305918
[2023-03-20 16:17:00 -0600 MDT][S] 'api_call': Count: 246 Min: 85.000 Mean: 189.980 Max: 1598.000 Stddev: 120.772 Sum: 46735.000 LastUpdated: 2023-03-20 16:17:59.950246 -0600 MDT m=+109.897234418
[2023-03-20 16:17:00 -0600 MDT][S] 'corso.runtime.gc_pause_ns': Count: 7 Min: 92041.000 Mean: 164637.143 Max: 197458.000 Stddev: 37392.482 Sum: 1152460.000 LastUpdated: 2023-03-20 16:17:59.210719 -0600 MDT m=+109.157707043

16:19:10.06	WARN	[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.heap_objects': 258246.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.total_gc_pause_ns': 4967045.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.total_gc_runs': 22.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.num_goroutines': 111.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.alloc_bytes': 57080576.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.sys_bytes': 194733104.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.malloc_count': 7158732.000
[2023-03-20 16:18:00 -0600 MDT][G] 'corso.Ryans-MBP.runtime.free_count': 6900486.000
[2023-03-20 16:18:00 -0600 MDT][C] 'api_call': Count: 2052 Sum: 2052.000 LastUpdated: 2023-03-20 16:18:59.974432 -0600 MDT m=+169.921475793
[2023-03-20 16:18:00 -0600 MDT][S] 'api_call': Count: 2054 Min: 88.000 Mean: 112.256 Max: 1324.000 Stddev: 61.006 Sum: 230573.000 LastUpdated: 2023-03-20 16:18:59.973682 -0600 MDT m=+169.920725876
[2023-03-20 16:18:00 -0600 MDT][S] 'corso.runtime.gc_pause_ns': Count: 9 Min: 92583.000 Mean: 325018.556 Max: 1283792.000 Stddev: 363904.188 Sum: 2925167.000 LastUpdated: 2023-03-20 16:18:56.277573 -0600 MDT m=+166.224613459
```

src/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.19
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.2.0
 	github.com/alcionai/clues v0.0.0-20230314154528-c469e1adafb6
+	github.com/armon/go-metrics v0.4.0
```
ryanfkeepers, line 8:
```
go: found github.com/hashicorp/go-metrics in github.com/hashicorp/go-metrics v0.4.1
go: github.com/alcionai/corso/src/cli imports
        github.com/hashicorp/go-metrics: github.com/hashicorp/go-metrics@v0.4.1: parsing go.mod:
        module declares its path as: github.com/armon/go-metrics
                but was required as: github.com/hashicorp/go-metrics
```

/shrug

vkamra: Weird - and why isn't this `v0.4.1`?

ntolia: This has been fixed recently - https://github.com/hashicorp/go-metrics/pull/143

src/internal/events/events.go
```diff
@@ -159,3 +162,88 @@ func tenantHash(tenID string) string {
...
+	flush := func() {
+		close(stop)
+		sig.Stop()
+		gm.Shutdown()
+	}
+
+	return context.WithValue(ctx, sinkCtxKey, sink), flush
+}
+
+func dumpMetrics(ctx context.Context, stop <-chan struct{}, sig *metrics.InmemSignal) {
```
vkamra, line 212: So if I understand this correctly - it's sending a signal on reportInterval - which triggers the output you provided?

Why is `NewInmemSink` configured with a `reportInterval` if we need to implement this ourselves? Also - is there a way to control the output? The GC stats are useful but I don't think we want them by default

ryanfkeepers: > So if I understand this correctly - it's sending a signal on reportInterval - which triggers the output you provided?

Correct.

> Why is NewInmemSink configured with a reportInterval if we need to implement this ourselves?

Ah, good question.  The sink interval doesn't have any relationship with the dump interval.  The sink interval's purpose is to bucket statistics to a given time-frame.  For example, if you want to get per-minute metrics, but only want to publish findings every 10 minutes, the you'd construct a sink with `inmemSink(1_minute, at_least_10_minutes)`, and time the log dump for once every 10 minutes.  That way, every log report contains 10 stacks of the stats block: one for each interval minute.

> Also - is there a way to control the output?

I'll look again, but I don't recall seeing any controls there.

src/cli/cli.go
```diff
@@ -165,6 +167,14 @@ func Handle() {
 		_ = log.Sync() // flush all logs in the buffer
 	}()
 
+	// ----------------------------------------------------------------------------------------
+
+	// print to stderr for poc.  should use the logger.
+	ctx, flushMetrics := events.NewMetrics(ctx, logger.Writer{Ctx: ctx})
+	defer flushMetrics()
```
vkamra, line 174: As a follow-up - should move this into operations so SDK users can also get this output

src/internal/events/events.go
```diff
@@ -159,3 +162,88 @@ func tenantHash(tenID string) string {
...
+
+	if sink == nil || !ok {
+		return
+	}
+
+	sink.IncrCounter([]string{string(cat)}, 1)
+}
+
+// Since records the duration between the provided time and now, in millis.
+func Since(ctx context.Context, cat m, start time.Time) {
```
vkamra, line 240: Should we use the `metrics` pkg i.e. `metrics.MeasureSince()` instead?

vkamra (commented):
LGTM. Can approve once you're ready with final version to merge

src/internal/events/events.go
ryanfkeepers: It'll work for aggregating metrics, but it's a singleton global instance that deprives us of context-bounded tracking.  This could be good if, say for CI, we want to track all calls across all the CI runtime.  But not so good if we want to isolate the metrics count within each test context.

src/internal/events/events.go
```diff
@@ -159,3 +162,84 @@ func tenantHash(tenID string) string {
...
+		sig  = metrics.NewInmemSignal(sink, metrics.DefaultSignal, w)
+	)
+
+	cfg.EnableHostname = false
+	cfg.EnableRuntimeMetrics = false
+
+	gm, err := metrics.NewGlobal(cfg, sink)
+	if err != nil {
+		logger.CtxErr(ctx, err).Error("metrics bus constructor")
+		return ctx, func() { sig.Stop() }
```
ashmrtn, line 199: if we failed to make a metrics object, should we not stop the signal object immediately? Is it going to have anything to dump if we failed?
```suggestion
		logger.CtxErr(ctx, err).Error("metrics bus constructor")
		sig.Stop()
		
		return ctx, func() {}
```

ashmrtn: also, after looking at `dumpMetrics` should we handle shutting down the metrics signal from there? Presumably once `dumpMetrics` exits there's no other way for the caller to get a handle to things to actually output more metrics is there?

src/internal/events/events.go
```diff
@@ -159,3 +162,84 @@ func tenantHash(tenID string) string {
...
+	}
+}
+
+// Inc increments the given category by 1.
+func Inc(cat m, keys ...string) {
+	cats := append([]string{string(cat)}, keys...)
+	metrics.IncrCounter(cats, 1)
+}
+
+// Since records the duration between the provided time and now, in millis.
```
ashmrtn, line 241: I'm confused by this comment and how this function is called in `service.go`. If we're getting the duration between the passed in time and the current time and all of this is done in the `Intercept()` function, is it really netting us useful metrics? Won't it just be the few ms it takes to generate the original timestamp and then reach this function call?

src/internal/events/events.go
```diff
@@ -159,3 +162,84 @@ func tenantHash(tenID string) string {
...
+		return ctx, func() { sig.Stop() }
+	}
+
+	stop := make(chan struct{})
+	go dumpMetrics(ctx, stop, sig)
+
+	flush := func() {
+		close(stop)
+		sig.Stop()
+		gm.Shutdown()
```
ashmrtn, line 208: should `gm.Shutdown()` be called before the other two so we get the final data it may have output?
```suggestion
		gm.Shutdown()
		close(stop)
		sig.Stop()
```

src/internal/events/events.go
ryanfkeepers: > is it really netting us useful metrics? Won't it just be the few ms it takes to generate the original timestamp and then reach this function call?

The implementation is effectively:
```go
	start     := time.Now()
	resp, err := pipeline.Next(req, middlewareIndex)
	events.Since(start, events.APICall)
```
The sequence of events will work as follows:
1. start the timer.
2. call pipeline.Next.  This will call the downstream http RoundTripper (ie, make the http request), blocking until it gets resolved.
3. call Since() to mark the end of the timer.  The recorded duration is `now() - start`: ie, the difference in time between 1/ and 3/.

The result is a measurement of the http RoundTripper (and any other middleware stacked on this one, so it's important for this to sit at the top).

src/internal/events/events.go
ryanfkeepers: I don't think that'll guarantee us any additional data.  But I _can_ tell the system to make signal print out a final response before we close.

src/internal/events/events.go
ryanfkeepers: > should we not stop the signal object immediately
It's sixes either way, really.  The only way to leak a resource here would be for the caller to panic and not have defer'd the closer.  No harm in protecting against the leak, though, so I'll take the sig.Stop out of the func.

> also, after looking at dumpMetrics should we handle shutting down the metrics signal from there
We could.  Again, it's sixes overall.   I feel safer handling it directly in the controller

src/internal/events/events.go
ashmrtn: thanks for pointing that out, I skimmed over the `pipeline.Next()` call because I'm not used to stuff doing work in `var` blocks 😅

src/internal/events/events.go
ashmrtn: eh, idk, it just seems weird that we're stopping the thing that prints stuff before we stop the thing that makes the stuff we print. Maybe I'm just used to having to worry about ordering for stuff a lot though

src/internal/events/events.go
ryanfkeepers: hah, yeah, it does feel a little weird.

ashmrtn (approved):

Issue #2876, opened 2023-03-20
Handle 503 when backing up [large] SharePoint files


meain:
Large OneDrive backups sometimes end up with a few 503. It might be a good idea to bump up the retries(unless there is a better way around this).

---

This behaviour was seen when backing up SharePoint files. [Here](https://learn.microsoft.com/en-us/answers/questions/319882/unable-to-download-onenote-file-using-microsoft-gr?orderby=newest) is a thread from Microsoft forums which discusses about 503 when backing up large SharePoint files (~2GB). This seems to be reproducible with test tenants.

vkamra:
Fixed with #2888

meain:
@vkamra I was waiting to maybe add a skip for large OneNote items that give a 503

vkamra:
Ah yes let's please track and do that. I know I stole this ticket from you :) - can you open a new one please?

meain:
Create an issue at https://github.com/alcionai/corso/issues/2910

PR #2888, opened 2023-03-21
Handle 503 retries where the `Retry-After` header exceeds what the Graph SDK will wait for

vkamra:
This logic will only trigger if the Graph retry handler does not handle the 503 (it is invoked first).

We remove the check in Corso for `absoluteMaxDelay` because our retry config will not trigger that limit (180s).
This allows the retry to kick in if the 503 response contains a `Retry-After` value > 180s

This could help with issues like https://github.com/alcionai/corso/issues/2876

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #2876 

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/service.go
+++ b/src/internal/connector/graph/service.go
@@ -27,12 +27,10 @@ import (
 const (
 	logGraphRequestsEnvKey    = "LOG_GRAPH_REQUESTS"
 	log2xxGraphRequestsEnvKey = "LOG_2XX_GRAPH_REQUESTS"
-	numberOfRetries           = 3
 	retryAttemptHeader        = "Retry-Attempt"
 	retryAfterHeader          = "Retry-After"
 	defaultMaxRetries         = 3
 	defaultDelay              = 3 * time.Second
-	absoluteMaxDelaySeconds   = 180
 	rateLimitHeader           = "RateLimit-Limit"
 	rateRemainingHeader       = "RateLimit-Remaining"
 	rateResetHeader           = "RateLimit-Reset"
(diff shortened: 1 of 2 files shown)


gmatev (approved):

src/internal/connector/graph/service.go
```diff
@@ -30,7 +30,7 @@ const (
 	numberOfRetries           = 3
 	retryAttemptHeader        = "Retry-Attempt"
 	retryAfterHeader          = "Retry-After"
-	defaultMaxRetries         = 3
+	defaultMaxRetries         = 6
 	defaultDelay              = 3 * time.Second
 	absoluteMaxDelaySeconds   = 180
```
meain, line 35: We could maybe bump this up to account for the increase in number of retries.

src/internal/connector/graph/service.go
```diff
@@ -309,7 +307,7 @@ func (handler *LoggingMiddleware) Intercept(
 	msg := fmt.Sprintf("graph api error: %s", resp.Status)
 
 	// special case for supportability: log all throttling cases.
-	if resp.StatusCode == http.StatusTooManyRequests {
+	if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable {
```
meain, line 310: Letting it retry for any 500 or 503. The graph sdk retry does not happen for 503 (probably because of cumulativeDelay). For now, we blindly retry for 500 and 503 as they ideally should not be an issue from our end.

PR #2898, opened 2023-03-21
add meta files for sharepoint

### LINKED ISSUE #2896: SharePoint restores have incorrect filenames
Seems like merging https://github.com/alcionai/corso/pull/2496 has caused problems with SharePoint. SharePoint backups now store files with IDs instead of names. *The name information is not available at all as we don't store metadata for SharePoint*. This also causes the restores to restore with the item IDs instead of actual file names.

ryanfkeepers:
Adds the meta files for sharepoint. Permissions
are not currently backed up.  But this gives us
necessary data like the original file name.

Also fixes a bug with backupID reference in
sharepoint restore.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* fixes #2896

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -123,7 +123,6 @@ type itemMetaReaderFunc func(
 	driveID string,
 	item models.DriveItemable,
 	fetchPermissions bool,
-	errs *fault.Bus,
 ) (io.ReadCloser, int, error)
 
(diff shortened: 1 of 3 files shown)


ashmrtn (approved):
would feel much more comfortable with PR if we had some automated CI tests in GC or something

PR #2900, opened 2023-03-21
check subfolders email counts in exchange

neha-Gupta1:
check that no of emails in all subfolders is equivalent in restores. 

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual


src/cmd/sanity_test/sanity_tests.go
```diff
@@ -46,13 +46,19 @@ func main() {
...
+	startTime time.Time,
+) {
 	var (
 		messageCount  = make(map[string]int32)
 		restoreFolder models.MailFolderable
 	)
 
-	builder := client.UsersById(testUser).MailFolders()
+	user := client.UsersById(testUser)
+	builder := user.MailFolders()
```
ryanfkeepers, line 61: We should avoid builders and construct the call where needed.  There's not a lot of re-use going on here to benefit from.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -112,6 +119,76 @@ func checkEmailRestoration(client *msgraphsdk.GraphServiceClient, testUser, fold
...
+	}
+}
+
+func getAllSubFolder(
+	user *users.UserItemRequestBuilder,
+	r models.MailFolderable,
+	parentFolder string,
+	messageCount map[string]int32,
+) {
+	folder := user.MailFoldersById(*r.GetId())
```
ryanfkeepers, line 133: Prefer ptr.Val for graph object gets, and storing them at the top for re-use.
```suggestion
        var (
            folderID = ptr.Val(r.GetId())
        )

	folder := user.MailFoldersById(folderID)
```

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -112,6 +119,76 @@ func checkEmailRestoration(client *msgraphsdk.GraphServiceClient, testUser, fold
...
+
+	childFolder, err := folder.ChildFolders().Get(context.Background(), nil)
+	if err != nil {
+		fmt.Printf("Error getting the drive: %v\n", err)
+		os.Exit(1)
```
ryanfkeepers, line 138: Can this get rebased onto my other changes so that we can call `fatal()` here? (and minimize rebase collisions)

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -112,6 +119,76 @@ func checkEmailRestoration(client *msgraphsdk.GraphServiceClient, testUser, fold
 				*restore.GetTotalItemCount())
 			os.Exit(1)
 		}
+
+		checkAllSubFolder(user, restore, "", messageCount)
+	}
+}
+
+func getAllSubFolder(
```
ryanfkeepers, line 127: comments explaining what these functions do, if they test for anything, and what the conditions of the tests are, would be really helpful.

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -112,6 +119,76 @@ func checkEmailRestoration(client *msgraphsdk.GraphServiceClient, testUser, fold
...
+			fullFolderName = parentFolder + "/" + fullFolderName
+		}
+
+		if messageCount[fullFolderName] != *child.GetTotalItemCount() {
+			fmt.Println("Restore was not succesfull for: ",
+				fullFolderName,
+				"Folder count: ",
+				messageCount[fullFolderName],
+				"Restore count: ",
+				*child.GetTotalItemCount())
```
ryanfkeepers, line 184: NIt: code style standards
```suggestion
			fmt.Println(
				"Restore was not succesfull for: ", fullFolderName,
				"Folder count: ", messageCount[fullFolderName],
				"Restore count: ", *child.GetTotalItemCount())
```

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -112,6 +119,76 @@ func checkEmailRestoration(client *msgraphsdk.GraphServiceClient, testUser, fold
 				*restore.GetTotalItemCount())
 			os.Exit(1)
 		}
+
+		checkAllSubFolder(user, restore, "", messageCount)
+	}
+}
+
+func getAllSubFolder(
+	user *users.UserItemRequestBuilder,
```
ryanfkeepers, line 128: Accepting a user id here and building a call from that will result in less abstraction (smaller mental model) and allow us to shift these calls to an api layer more easily in the future.

ryanfkeepers (approved):
approving to unblock, all comments are nits/refinements

src/cmd/sanity_test/sanity_tests.go
neha-Gupta1: its a paginated call. on line - https://github.com/alcionai/corso/blob/7207be77155b90911596847c7b77bce5f0714610/src/cmd/sanity_test/sanity_tests.go#L101
new builder value is inserted. So initialising the builder before the `for` loop was required.

PR #2904, opened 2023-03-22
remove the last use of multierrs

### LINKED ISSUE #1970: Error processing and management

ryanfkeepers:
removes the last instance of multierrs in favor of the fault.Bus.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* closes #1970

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -216,9 +216,5 @@ func (cr *containerResolver) populatePaths(
 		}
 	}
 
-	if lastErr != nil {
-		return lastErr
-	}
-
-	return nil
+	return lastErr
 }
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
-	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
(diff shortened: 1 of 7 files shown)


src/internal/connector/exchange/container_resolver.go
```diff
@@ -193,16 +192,33 @@ func (cr *containerResolver) DestinationNameToID(dest string) string {
...
+			lastErr = err
 		}
 	}
 
-	return errs.ErrorOrNil()
+	if lastErr != nil {
+		return lastErr
+	}
+
+	return nil
```
meain, line 223: nit
```suggestion
	return lastErr
```

meain (approved):

PR #2906, opened 2023-03-22
Don't attempt to restore dirmeta files for newer backup versions

### LINKED ISSUE #2878: Don't attempt to load [dir]meta files in RestoreMultipleItems
Newer versions of Corso can load metadata files inline so we don't need to augment the path set during restore. However the logic for this was never updated

Action items:
* remove [dir]meta files from the path set during OneDrive restore for newer versions of Corso

> The paths are not present in the metadata anymore, but they do get added in `AugmentRestorePaths` as we need to inject directory paths inorder to make sure permissions for folders without any files are restored properly. ([ref](https://github.com/alcionai/corso/blob/d1ce661b8fa03805e57b21ea7abcfb75d01a47ac/src/internal/connector/onedrive/restore.go#L643))

ashmrtn:
Newer backup versions can fetch the dirmeta files
inline while restoring directories so there's no
need to add them here

Exclude backup versions 0 and > 4 from path
augmentation

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* closes #2878

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -615,9 +615,10 @@ func getMetadata(metar io.ReadCloser) (Metadata, error) {
 func AugmentRestorePaths(backupVersion int, paths []path.Path) ([]path.Path, error) {
 	// Nothing to do for versions where the directory metadata is stored in the
 	// directory it refers to as Corso will retrieve this data inline when
-	// restoring the collection for the directory.
-	if backupVersion >= version.OneDrive4DirIncludesPermissions {
-		return nil, nil
+	// restoring the collection for the directory. Path ordering does not matter
+	// for these versions.
+	if backupVersion == 0 || backupVersion >= version.OneDrive4DirIncludesPermissions {
+		return paths, nil
 	}
 
 	colPaths := map[string]path.Path{}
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/restore.go
```diff
@@ -613,6 +613,14 @@ func getMetadata(metar io.ReadCloser) (Metadata, error) {
 // Augment restore path to add extra files(meta) needed for restore as
 // well as do any other ordering operations on the paths
 func AugmentRestorePaths(backupVersion int, paths []path.Path) ([]path.Path, error) {
+	// Nothing to do for versions where the directory metadata is stored in the
+	// directory it refers to as Corso will retrieve this data inline when
+	// restoring the collection for the directory. Path ordering does not matter
+	// for these versions.
+	if backupVersion == 0 || backupVersion >= version.OneDrive4DirIncludesPermissions {
```
ashmrtn, line 620: just double checking cause I'm juggling a few things at the moment: it's safe to exclude version 0 and not sort the paths correct?

meain: It should be safe to exclude 0, but we can't exclude >= 4 . We need to augment paths for newer items because of https://github.com/alcionai/corso/issues/2697 .

meain:
I might have a fix for this in https://github.com/alcionai/corso/pull/2908

src/internal/connector/onedrive/restore.go
ashmrtn: good point, thanks for catching that and posting a PR of your own to fix things

ashmrtn:
closing in favor of #2908

PR #2907, opened 2023-03-22
Split some CI tests for OneDrive into nightly and regular CI

### LINKED ISSUE #2756: Move most OneDrive RestoreAndBackup-style tests to a nightly regression suite
The new testing framework for OneDrive RestoreAndBackup tests in the `src/internal/connector` package make it really easy to test that restores/backups give the expected results for older versions of data as well as the most recent version. However, these tests are starting to take a non-trivial amount of time to run in CI which could lead to throttling or other issues if we're not careful

Since these tests are also mostly regression tests, it would be best to move all but a test for the most recent backup version to a regression test GitHub Action workflow instead of running all of them all the time in CI. This would help keep CI runtimes in check while still providing some assurances that backwards compatibility remains

Ideally we would be able to use the same code to run both the most recent version and the nightly regression tests. We would just change what value the loop in the test starts at. [Possible](https://stackoverflow.com/a/40178776) ways to do this include env vars (though that may need to be tied into the testing env var system) or build constraints

ashmrtn:
Try to reduce the overall number of tests run during
regular CI by splitting most of the OneDrive
RestoreAndBackup tests into a new "nightly" test suite.
Regular CI will now only run against the latest backup
version to make sure nothing immediate broke

This should only be merged once we have a GitHub action
for nightly tests and some way of reporting the outcome
of the action

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #2756

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


vkamra:
LGTM. @neha-Gupta1 - once these go in, could you add a job that runs these nightly and on-demand (if needed). We can also connect your slack notification here so if it fails we get notified. This will help reduce the load on CI.

ryanfkeepers:
Per load vs nightly.  The load tests were originally intended to run on a nightly cadence.  Should we unify the two?  Run load and nightly on the same cadence, with the same github action?

ashmrtn:
>Should we unify the two? Run load and nightly on the same cadence, with the same github action? Might be easier than maintaining a fourth test category.

I think whether we unify or not should depend on how easy it is to get/find reporting on the unified set. We want a clear indicator of if the load tests are failing or the nightly (regression/longer running regular tests) are failing. The latter failing is an immediate problem to look into. Load tests failing is not great but maybe not the end of the world

ryanfkeepers:
> We want a clear indicator of if the load tests are failing or the nightly (regression/longer running regular tests) are failing.

Point.  How about this: keep the categories separate, but kick both of them off in the load_testing.yml?

> The latter failing is an immediate problem to look into. Load tests failing is not great but maybe not the end of the world

Only in the case that the load is the cause of failure, and not produced by some other fault that might be critical, but only identified in load testing.  Ideally, we'd aggregate alerts for load, nightly, and sanity tests in a spot where they are all received with equal concern.

ryanfkeepers (approved):

Issue #2913, opened 2023-03-22
[Bug]: Backup failure after valid user revokes license.


ryanfkeepers:
### What happened?

Repro steps:
1. Start with a licensed user account.
2. Run an exchange backup (this should succeed).
3. Revoke the exchange license for that user.
4. Run an incremental exchange backup, this should also succeed.
5. Run another incremental exchange backup.  This one should fail.

### Corso Version?

Corso v0.6.1

### Where are you running Corso?

any

### Relevant log output

```shell
n/a
```

ryanfkeepers:
Closing: license issue turned out to be a red herring.

PR #2916, opened 2023-03-22
additional logging when creating backup model

ryanfkeepers:
#### Type of change

#### Issue(s)

* #2913

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -978,27 +978,28 @@ func inflateBaseTree(
 	for _, subtreePath := range snap.SubtreePaths {
 		// We're starting from the root directory so don't need it in the path.
 		pathElems := encodeElements(subtreePath.PopFront().Elements()...)
+		ictx := clues.Add(ctx, "subtree_path_elems", pathElems)
 
-		ent, err := snapshotfs.GetNestedEntry(ctx, dir, pathElems)
+		ent, err := snapshotfs.GetNestedEntry(ictx, dir, pathElems)
 		if err != nil {
 			if isErrEntryNotFound(err) {
-				logger.CtxErr(ctx, err).Infow("base snapshot missing subtree")
+				logger.CtxErr(ictx, err).Infow("base snapshot missing subtree")
 				continue
 			}
 
-			return clues.Wrap(err, "getting subtree root").WithClues(ctx)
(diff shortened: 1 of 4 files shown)


src/internal/kopia/upload.go
```diff
@@ -960,14 +960,16 @@ func inflateBaseTree(
 		return nil
 	}
 
+	ctx = clues.Add(ctx, "snapshot_root_id", snap.ID)
```
ashmrtn, line 963: This is the ID of the base snapshot, not the ID of the root item in the base snapshot
```suggestion
	ctx = clues.Add(ctx, "snapshot_base_id", snap.ID)
```

src/internal/kopia/upload.go
```diff
@@ -976,35 +978,36 @@ func inflateBaseTree(
 	for _, subtreePath := range snap.SubtreePaths {
 		// We're starting from the root directory so don't need it in the path.
 		pathElems := encodeElements(subtreePath.PopFront().Elements()...)
+		ictx := clues.Add(ctx, "subtree_path_elems", pathElems)
```
ashmrtn, line 981: you probably want to use `subtreePath` else you'll get base64 encoded values
```suggestion
		ictx := clues.Add(ctx, "subtree_path_elems", subtreePath)
```

src/internal/operations/backup.go
```diff
@@ -671,21 +671,23 @@ func (op *BackupOperation) createBackupModels(
...
 		return clues.New("no backup details to record").WithClues(ctx)
 	}
 
-	err := sscw.Collect(ctx, streamstore.DetailsCollector(backupDetails))
+	ctx = clues.Add(ctx, "details_entry_count", len(deets.Entries))
+
+	err := sscw.Collect(ctx, streamstore.DetailsCollector(deets))
 	if err != nil {
-		return clues.Wrap(err, "creating backupDetails persistence").WithClues(ctx)
+		return clues.Wrap(err, "creating details persistence").WithClues(ctx)
```
ashmrtn, line 690:
```suggestion
		return clues.Wrap(err, "persisting details").WithClues(ctx)
```

ashmrtn (approved):

src/internal/operations/backup.go
ryanfkeepers: not quite, since this isn't actually the write action.  Just a collection producer.  But i did update the error to be more clear.

Issue #2825, opened 2023-03-15
Support specifying User ID instead of User UPN for backup operations


vkamra:
User UPNs can change (rename) and supporting User IDs will allow working around that.

Even if we do not expose this in the CLI - it would make sense to start resolving to User IDs internally and use those instead of UPNs. This future proofs if the user is renamed and will allow incremental backup chains to be maintained.

ryanfkeepers:
## Assumptions
* All storage will utilize the ID.
* All end-user display (eg, CLI) will prefer the owner display name.
* In cases where no display name exists (or is sensible), the ID will be used for both values.
* CLI and SDK interfaces can accept either display names, interchangeably.

## Proposal
Ensure we store two values in the operation.Backup and backup.Backup structs:
* ResourceOwner string (the ID)
* ResourceOwnerName string (the display name)

ResourceOwner already exists, and is taken from the selectors.DiscreteOwner value.  Ownership of display names doesn't align with the design of selectors; we're better off passing the display name value into the operation, and letting that struct maintain ownership.

## Challenges/Gotchas
### order of operations
Currently, the resource owner isn't validated until a backup is underway, when we construct a GC client.  If we want ids and names to be handled interchangeably, we'll need to move the gc client construction up earlier into the operation creation (instead of run) to gather both values early.

### listing backups
Suppose a user creates backup B1 for an owner with name N1.  The user then changes their name to N2.  Another backup creates B2.  B1 and B2 will list as having different owners.

### utilizing prior display names
No current issues, but a potential future gotcha.  If a situation arises where a user attempts to perform an action involving an out-of-date display name (ex: listing backups filtered by owner name), and we attempt to resolve that display name to the user ID beforehand, validation could fail due to the old owner name not appearing in the results.

### collisions over time
(note: Investigation pending.)  Is it possible for owner A to switch their name from N1 to N2, and then owner B switch their name from N3 to N1?  If so, what happens to corso lookups as a result?

### testing
Integration testing of owner name changes is implausible, since a stable test would need to avoid race conditions with other tests acting upon the same external resource.  Mocks are our best option for controlled testing.

## Owner ID in Storage Paths
Currently, all exchange and oneDrive data is stored using the user's PN (ie, their display name) as the ID.  To complete this change, we'll want to begin storing all files under the user's ID instead of their PN.  When we introduce this change, one of the following will be true.  Corso will need to...
[...]

PR #2921, opened 2023-03-22
introduce the resource owner name to op.backup

ryanfkeepers:
adds a resourceOwnerName property to the op
backup struct.  Also adds a resource owner id
and owner name to the backup.backup.  These
values will be used for both identification and for
end user display.  Values are not currently populated
reliably, nor are they used.  This change only
allows them to exist.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -35,10 +35,12 @@ import (
 type BackupOperation struct {
 	operation
 
-	ResourceOwner string             `json:"resourceOwner"`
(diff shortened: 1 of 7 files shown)


ashmrtn (approved):
high-level: would be good if you could have different values for the tests when setting name/ID, this can help find unexpected assumptions in the code

Issue #2910, opened 2023-03-22
Skip large OneNote files resulting in a 503


meain:
OneNote files on OneDrive and SharePoint are limited to 2GB max. One might still be able to edit it via the UI, but trying to access it via the API results in 503. We should skip these files instead of failing the backup.

Ref: https://support.microsoft.com/en-us/office/restrictions-and-limitations-in-onedrive-and-sharepoint-64883a5d-228e-48f5-b3d2-eb39e07630fa#onenotenotebooks

PR #2925, opened 2023-03-23
Skip large OneNote items that cannot be downloaded

meain:
OneNote files >2GB cannot be downloaded via the graph API. This change will gracefully skip them and prevent the backup from erroring out.

A OneNote file is represented by a Package(folder) which contains two file within it, `<file>.one` and `<file>.onetoc2`.  From what I can tell the `onetoc2` file is the metadata file and the `one` file is the one that contains data and can get big.

**The current implementation has a limitation(or feature depending on how you see it) in that we will end up doing a partial backup with just the `onetoc2` file but avoiding the big `one` file if that is big.**

I did not try to change the behaviour as even currently if for somehow we ended up with an error like 404 when trying to backup one of these files, we do skip it. I've added a comment in for now, let me know if we should go back and revisit the behaviour. *The main issue is the fact that we only hit this problem once it reaches kopia and removing it at that point is harder.*

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/2910

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enable compression for all data uploaded by kopia.
 - SharePoint --folder selectors correctly return items.
 - Fix Exchange cli args for filtering items
+- Skip huge OneNote items that Graph API prevents us from downloading
 
 ## [v0.6.1] (beta) - 2023-03-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SharePoint --folder selectors correctly return items.
 - Fix Exchange cli args for filtering items
 - Skip huge OneNote items that Graph API prevents us from downloading
+- Skip OneNote items bigger than 2GB (Graph API prevents us from downloading them)
 
 ## [v0.6.1] (beta) - 2023-03-21
 
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -99,7 +99,13 @@ type Collection struct {
 	// Specifies if it new, moved/rename or deleted
 	state data.CollectionState
 
-	kind collectionKind
+	// scope specifies what scope the items in a collection belongs
+	// to. This is primarily useful when dealing with a "package",
+	// like in the case of a OneNote file. A OneNote file is a
+	// collection with a package scope and multiple files in it. Most
+	// other collections have a scope of folder to indicate that the
+	// files within them belong to a folder.
+	scope collectionScope
 
 	// should only be true if the old delta token expired
 	doNotMergeItems bool
@@ -139,7 +145,7 @@ func NewCollection(
 	statusUpdater support.StatusUpdater,
 	source driveSource,
 	ctrlOpts control.Options,
-	colKind collectionKind,
+	colKind collectionScope,
 	doNotMergeItems bool,
 ) *Collection {
 	c := &Collection{
@@ -154,7 +160,7 @@ func NewCollection(
 		statusUpdater:   statusUpdater,
 		ctrl:            ctrlOpts,
 		state:           data.StateOf(prevPath, folderPath),
-		kind:            colKind,
+		scope:           colKind,
 		doNotMergeItems: doNotMergeItems,
 	}
 
@@ -352,7 +358,7 @@ func (oc *Collection) getDriveItemContent(
 		}
 
 		if clues.HasLabel(err, graph.LabelStatus(http.StatusNotFound)) || graph.IsErrDeletedInFlight(err) {
-			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipNotFound).Error("item not found")
+			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipNotFound).Info("item not found")
 			el.AddSkip(fault.FileSkip(fault.SkipNotFound, itemID, itemName, graph.ItemInfo(item)))
 
 			return nil, clues.Wrap(err, "downloading item").Label(graph.LabelsSkippable)
@@ -360,14 +366,14 @@ func (oc *Collection) getDriveItemContent(
 
 		// Skip big OneNote files as they can't be downloaded
 		if clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) &&
-			oc.kind == CollectionKindPackage && *item.GetSize() >= MaxOneNoteFileSize {
+			oc.scope == CollectionScopePackage && *item.GetSize() >= MaxOneNoteFileSize {
 			// FIXME: It is possible that in case of a OneNote file we
 			// will end up just backing up the `onetoc2` file without
 			// the one file which is the important part of the OneNote
 			// "item". This will have to be handled during the
 			// restore, or we have to handle it separately by somehow
 			// deleting the entire collection.
-			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipBigOneNote).Error("invalid file")
+			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipBigOneNote).Info("max OneNote file size exceeded")
 			el.AddSkip(fault.FileSkip(fault.SkipBigOneNote, itemID, itemName, graph.ItemInfo(item)))
 
 			return nil, clues.Wrap(err, "downloading item").Label(graph.LabelsSkippable)
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -145,7 +145,7 @@ func NewCollection(
 	statusUpdater support.StatusUpdater,
 	source driveSource,
 	ctrlOpts control.Options,
-	colKind collectionScope,
+	colScope collectionScope,
 	doNotMergeItems bool,
 ) *Collection {
 	c := &Collection{
@@ -160,7 +160,7 @@ func NewCollection(
 		statusUpdater:   statusUpdater,
 		ctrl:            ctrlOpts,
 		state:           data.StateOf(prevPath, folderPath),
-		scope:           colKind,
+		scope:           colScope,
 		doNotMergeItems: doNotMergeItems,
 	}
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Incorrectly moving subfolder hierarchy from a deleted folder to a new folder at the same path during OneDrive incremental backup.
 - Handle calendar events with no body.
 - Items not being deleted if they were created and deleted during item enumeration of a OneDrive backup.
+- Skip huge OneNote items that Graph API prevents us from downloading
```
ashmrtn, line 16: nit: should put the size of the OneNote items here and/or link to a place that has the issue better documented. That way users of Corso have a way to see if something is wrong or if it's expected

src/internal/connector/onedrive/collection.go
```diff
@@ -352,6 +359,21 @@ func (oc *Collection) getDriveItemContent(
 			return nil, clues.Wrap(err, "downloading item").Label(graph.LabelsSkippable)
 		}
 
+		// Skip big OneNote files as they can't be downloaded
+		if clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) &&
+			oc.kind == CollectionKindPackage && *item.GetSize() >= Size2GB {
+			// FIXME: It is possible that in case of a OneNote file we
+			// will end up just backing up the `onetoc2` file without
+			// the one file which is the important part of the OneNote
+			// "item". This will have to be handed during the restore,
```
ashmrtn, line 368:
```suggestion
			// "item". This will have to be handled during the restore,
```

src/internal/connector/onedrive/collection.go
```diff
@@ -42,6 +42,9 @@ const (
 	MetaFileSuffix    = ".meta"
 	DirMetaFileSuffix = ".dirmeta"
 	DataFileSuffix    = ".data"
+
+	// Used to compare in case of OneNote files
+	Size2GB = 2 * 1024 * 1024 * 1024
```
ashmrtn, line 47: nit: a slightly more descriptive name can make reading this var in code easier
```suggestion
	MaxOneNoteFileSize = 2 * 1024 * 1024 * 1024
```

src/internal/connector/onedrive/collection_test.go
```diff
@@ -597,3 +601,116 @@ func (suite *CollectionUnitTestSuite) TestCollectionPermissionBackupLatestModTim
...
+				return item, nil
+			}
+
+			_, err := col.getDriveItemContent(ctx, item, errs)
+			if test.err == nil {
+				assert.NoError(t, err, "no error")
+			} else {
+				assert.EqualError(t, err, clues.Wrap(test.err, "downloading item").Error(), "error")
+				assert.Equal(t, test.skippable, clues.HasLabel(err, graph.LabelsSkippable), "skippable")
+			}
```
ashmrtn, line 713: given that all the error messages are the same, I would suggest refactoring this and doing a few things:
* use `assert.ErrorAssertionFunc` to check if there was an error or not. That would change the struct declaration to be something like `errCheck: assert.[No]Error` and the call to use it like `test.errCheck(t, err)`
* compare labels instead of error strings
* fold comparison for skippable into labels check

With the above, the code could become something like
```go
// do stuff to get results to compare against
test.errCheck(t, err)

if err == nil {
  return // nothing left to check
}

// check label(s) for error
assert.Equal(t, <map[string]struct{} of labels>, err.Labels())
```

ashmrtn (commented):
apart from a few nits code looks good. Not sure if Keepers has input/thoughts about error reporting on the CLI etc. I'm not familiar with that code so can't give a good review on how this ties in with it

CHANGELOG.md
meain: Good call, added.

src/internal/connector/onedrive/collection_test.go
meain: Ahh, that looks much cleaner.

meain: This looks cleaner. I've updated the code to reflect most of the items suggested here. I had to retain specifying the errors so as to add the properly labels onto it for returning from the [itemReader](https://github.com/alcionai/corso/blob/f8f35b1b4c1336da65960a4ebd616673bc8f3045/src/internal/connector/onedrive/collection_test.go#L695) function.

src/internal/connector/onedrive/collection.go
```diff
@@ -352,6 +359,21 @@ func (oc *Collection) getDriveItemContent(
 			return nil, clues.Wrap(err, "downloading item").Label(graph.LabelsSkippable)
 		}
 
+		// Skip big OneNote files as they can't be downloaded
+		if clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) &&
+			oc.kind == CollectionKindPackage && *item.GetSize() >= MaxOneNoteFileSize {
```
ryanfkeepers, line 364: I'd prefer this composite check get moved as close to the original request as possible, and replaced with a clearer label like `graph.LabelsMaxFileSize`.  As a pattern, that ensures that condition identification stays centralized to the request, rather than democratized to all upstream catchers, and that the label to catch is clearly identifiable and readable.

src/internal/connector/onedrive/collection.go
```diff
@@ -352,6 +359,21 @@ func (oc *Collection) getDriveItemContent(
...
+			// FIXME: It is possible that in case of a OneNote file we
+			// will end up just backing up the `onetoc2` file without
+			// the one file which is the important part of the OneNote
+			// "item". This will have to be handled during the
+			// restore, or we have to handle it separately by somehow
+			// deleting the entire collection.
+			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipBigOneNote).Error("invalid file")
```
ryanfkeepers, line 371: Nit: skips should be Info level logs (same for the 404 on line 356).  They're known conditions that we're handling.  Not unexpected errors.
```suggestion
			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipBigOneNote).Info("max size file")
```

src/internal/connector/onedrive/collection.go
```diff
@@ -97,6 +100,8 @@ type Collection struct {
 	// Specifies if it new, moved/rename or deleted
 	state data.CollectionState
 
+	kind collectionKind
```
ryanfkeepers, line 103: per prior discussion on naming, and a want to follow [these naming conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md), `kind` implies a the body serialization structure, which isn't entirely appropriate for this case.   `type collectionType` would be more appropriate.  Or, noting the consts, maybe `scope collectionScope` would be better?

Also, this could use a header for explanation of its purpose.

src/internal/connector/onedrive/collection.go
meain: Firstly, about using the label `graph.LabelsMaxFileSize`. This 2GB max limit is only for OneNote files and does not affect other files(or they could have other limit) and so I don't think it is appropriate to label every file that crosses this limit with that label. Another issue is that we won't really have the size value of the item parsed at the point where we create the other labels as of now.

Another reason why this has to be moved to here instead of where the request happens is because we have to check if it is part of a `package` to determine if we should be skipping it or just failing outright.

src/internal/connector/onedrive/collection.go
meain: Ahh, let me make that change in logging to Info. I'm not sure about the "max size file" message though. This file size limit is specific to just OneNote. How about "max OneNote file size exceeded"?

src/pkg/fault/item.go
```diff
@@ -156,6 +156,13 @@ const (
 	// SkipNotFound identifies that a file was skipped because we could
 	// not find it when trying to download contents
 	SkipNotFound skipCause = "file_not_found"
+
+	// SkipBigOneNote identifies that a file was skipped because it
+	// was big OneNote file and we can only download OneNote files which
+	// are less that 2GB in size.
+	//nolint:lll
+	// https://support.microsoft.com/en-us/office/restrictions-and-limitations-in-onedrive-and-sharepoint-64883a5d-228e-48f5-b3d2-eb39e07630fa#onenotenotebooks
+	SkipBigOneNote skipCause = "big_one_note_file"
```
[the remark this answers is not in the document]
meain: I have the same concern for this as well. This size limit is specific for OneNote and I feel like using something generic like "max file size" might not be the right choice here as 1) other files don't share this limt, 2) other files might have separate limits.

I think there are few similar items about the messages/label. Let me know your thoughts on those and I can change if you still think they would be better as a generic messages/labels.

src/internal/connector/onedrive/collection.go
meain: Ahh, you are right. I wanted to use `type` as well, but since that is a keyword, don't think we will be able to use it. Let me switch to `scope`.

src/internal/connector/onedrive/collection.go
```diff
@@ -135,6 +146,7 @@ func NewCollection(
 	statusUpdater support.StatusUpdater,
 	source driveSource,
 	ctrlOpts control.Options,
+	colKind collectionScope,
```
ryanfkeepers, line 149: Nit: naming consistency.  Admittedly, I think I'd prefer `type` just so we don't overload or confuse terminology from selector scopes.  That's minor, though.
```suggestion
	colScope collectionScope,
```

src/internal/connector/onedrive/collections.go
```diff
@@ -743,6 +759,11 @@ func (c *Collections) UpdateCollections(
 				continue
 			}
 
+			collectionKind := CollectionScopeFolder
```
ryanfkeepers, line 762: Nit: naming consistency
```suggestion
			colScope := CollectionScopeFolder
```

src/internal/connector/onedrive/collection.go
meain: `type` would be ideal, but unfortunately we can't use that as it is a reserved keyword.

PR #2929, opened 2023-03-23
Always make empty collections for prefix directories

ryanfkeepers:
Use empty collections to ensure the prefix directories are available in
kopia no matter what user data we find.

This helps ensure we have some set of things we can assume about
non-failed backups including:

    They have a valid snapshot ID for data
    They have a valid snapshot ID for backup details

Eventually, this will allow us to say that backups using a base that
doesn't have a prefix directory is invalid in some way

SharePoint tests are disabled because lists and pages ignore the none
target in selectors

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #2550

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/data_collections.go
+++ b/src/internal/connector/exchange/data_collections.go
@@ -221,7 +221,7 @@ func DataCollections(
 			su,
 			errs)
 		if err != nil {
-			return collections, nil, err
+			return nil, nil, err
 		}
 
 		collections = append(collections, baseCols...)
--- a/src/internal/connector/exchange/data_collections.go
+++ b/src/internal/connector/exchange/data_collections.go
@@ -179,6 +179,7 @@ func DataCollections(
 		user        = selector.DiscreteOwner
 		collections = []data.BackupCollection{}
 		el          = errs.Local()
+		categories  = map[path.CategoryType]struct{}{}
 	)
 
 	cdps, err := parseMetadataCollections(ctx, metadata, errs)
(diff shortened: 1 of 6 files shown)


src/internal/connector/exchange/data_collections.go
```diff
@@ -205,9 +206,27 @@ func DataCollections(
...
+		baseCols, err := graph.BaseCollections(
+			ctx,
+			acct.AzureTenantID,
+			user,
+			path.ExchangeService,
+			categories,
+			su,
+			errs)
+		if err != nil {
+			return collections, nil, err
```
ashmrtn, line 224: I'm a little hesitant returning collections here. It makes it feel like this is a "maybe we'll do this, maybe we won't" which makes enforcing invariants about it down the road kind of useless

src/internal/connector/graph/collections.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			// Shouldn't happen.
+			err = clues.Wrap(err, "getting base prefix").WithClues(ictx)
+			el.AddRecoverable(err)
+			lastErr = err
+
+			continue
+		}
+
+		// Pop off the last path element because we just want the prefix.
```
ashmrtn, line 86: this comment should be on line 76

src/internal/connector/graph_connector_test.go
```diff
@@ -1054,3 +1054,136 @@ func (suite *GraphConnectorIntegrationSuite) TestRestoreAndBackup_largeMailAttac
...
+				if fullPath.Service() != test.service {
+					continue
+				}
+
+				assert.Empty(t, fullPath.Folders(), "non-prefix collection")
+				assert.NotEqual(t, col.State(), data.NewState, "prefix collection marked as new")
+				foundCategories = append(foundCategories, fullPath.Category().String())
+
+				t.Logf("looking at collection %s\n", fullPath)
+
```
ashmrtn, line 1178: nit: could remove this
```suggestion
```

ashmrtn (approved):
I'm wondering if we'll need to fixup the `RestoreAndBackup` comparison logic to ignore these collections

src/internal/connector/exchange/data_collections.go
ryanfkeepers: oh, yeah, totally with you.

src/internal/connector/graph/collections.go
ryanfkeepers: good catch

Issue #2795, opened 2023-03-14
Ensure duplicate interfaces are centralized


ryanfkeepers:
## Issue

Some interfaces tend to crop up in multiple places.  We'd be better off with single declarations and clear owners for each interface.  Interfaces should not all be owned by the same package.  Rather they should be owned by whomever owns the behavior being interfaced.  To avoid circular references, interfaces should be sub-packaged within the owner, similar to mock packages.

Examples: `backuper`, `restorer`.

PR #2931, opened 2023-03-23
Create interfaces for kopia functions

ashmrtn:
These interfaces were being defined in multiple
places so it seemed time to consolidate and
define them in one place only

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2795

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -22,6 +22,54 @@ import (
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
+type IncrementalBase struct {
+	*snapshot.Manifest
+	SubtreePaths []*path.Builder
+}
+
+// PrevRefs hold the repoRef and locationRef from the items
+// that need to be merged in from prior snapshots.
+type PrevRefs struct {
+	Repo     path.Path
+	Location path.Path
+}
+
+type Backuper interface {
+	// BackupCollections takes a set of collections and creates a kopia snapshot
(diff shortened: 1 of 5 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -22,6 +22,54 @@ import (
...
+}
+
+// PrevRefs hold the repoRef and locationRef from the items
+// that need to be merged in from prior snapshots.
+type PrevRefs struct {
+	Repo     path.Path
+	Location path.Path
+}
+
+type Backuper interface {
```
ryanfkeepers, line 37: Hah, funny enough, we're both making similar changes at the same time.  See: https://github.com/alcionai/corso/pull/2932

Thinking on the correct owner, so that these can be shared across libraries, thoughts on moving this to something like `/operations/interfaces/interfaces.go`?   Imo, it makes more sense that this is a pattern owned by operations, and that KW complies with.  In that way, we could swap out KW for another storage, or a mock, or etc.

ryanfkeepers (approved):

ryanfkeepers:
@ashmrtn Should this get closed/updated following the interface relocation to `operations/inject` from the other PR?

ashmrtn:
closing as it's covered by other PRs

PR #2941, opened 2023-03-24
move getDrives to onedrive/api, add mock

ryanfkeepers:
Moves the drive fetcher and pagers into the
onedrive/api package so that it can be easily re-
used by other packages.  Also moves the mocks
out from local testing and into a mock package
nested in the api pkg.

first of a couple PRs fixing up sharepoint testing
issues.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E


src/internal/connector/onedrive/api/drive.go
```diff
@@ -269,3 +271,76 @@ func (p *siteDrivePager) GetFolderIDByName(ctx context.Context, driveID, folderN
...
+
+	// Loop through all pages returned by Graph API.
+	for {
+		var (
+			err  error
+			page api.PageLinker
+		)
+
+		// Retry Loop for Drive retrieval. Request can timeout
+		for i := 0; i <= maxRetryCount; i++ {
```
ashmrtn, line 307: do we still need this retry logic or is it covered by middleware retry handler we now have?

src/internal/connector/onedrive/api/mock/drive.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type DrivePager struct {
+	ToReturn []PagerResult
+	GetIdx   int
+}
+
+func (p *DrivePager) GetPage(context.Context) (api.PageLinker, error) {
+	if len(p.ToReturn) <= p.GetIdx {
+		return nil, assert.AnError
```
ashmrtn, line 33: nit: for debugability maybe we should wrap this with a message of "out of bounds" or something? That will also give stack trace info. Or just return a clues error and forget the wrap call?
```suggestion
		return nil, clues.Wrap(assert.AnError, "index out of bounds")
```

ashmrtn (approved):

src/internal/connector/onedrive/api/drive.go
ryanfkeepers: I guess that depends on the pager implementation.  It *should* be an expectation we can carry through, but not a guarantee.  I'm fine removing this, but would prefer to do so in a follow-up PR.

src/internal/connector/onedrive/api/mock/drive.go
ryanfkeepers: Sure, why not?

PR #2943, opened 2023-03-24
fix sp backup-restore test

ryanfkeepers:
The sp backup-restore wants to validate the count of items that it restored after backup.  This PR manufactures a very brittle set of expectations to pass that test, and will need future love.  In the meantime, it also refactors onedrive restore for clarity and supportability.

Note that the site being used for the test is changing here.  This will get changed in github secrets as well, else we won't have our brittle environment to pass on.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E


src/internal/connector/onedrive/restore.go
```diff
@@ -217,129 +213,184 @@ func RestoreCollection(
 				continue
 			}
 
-			if backupVersion >= version.OneDrive1DataAndMetaFiles {
```
ryanfkeepers, line 220: This file got a refactoring so that I could have an easier time making assertions about the state of writes and counts.  Could still use a little more love, but now it should read more linearly for any given collection.

src/internal/operations/restore_test.go
```diff
@@ -218,38 +222,68 @@ func (suite *RestoreOpIntegrationSuite) SetupSuite() {
...
-	csel.DiscreteOwner = siteID
-	csel.Include(csel.LibraryFolders(selectors.Any()))
+	var (
+		siteID = tester.M365SiteID(t)
+		sites  = []string{siteID}
+		spsel  = selectors.NewSharePointBackup(sites)
+	)
+
+	spsel.DiscreteOwner = siteID
+	// assume a folder name "test" exists in the drive.
```
ashmrtn, line 241: I think we have a page somewhere that documents the minimum data set required to stand up CI on a new tenant. Can you add this there please?

src/internal/operations/restore_test.go
```diff
@@ -218,38 +222,68 @@ func (suite *RestoreOpIntegrationSuite) SetupSuite() {
...
-		// Discount metadata files (2 paths, 2 deltas) as
-		// they are not part of the data restored.
-		items: bo.Results.ItemsWritten - 4,
+		selectorResourceOwners: sites,
+		resourceOwner:          siteID,
+		backupID:               bo.Results.BackupID,
+		// Discount metadata files (1 delta, 1 prev path)
+		// assume only one folder, and therefore 1 dirmeta per drive
+		// assume only one file in each folder, and therefore 1 meta per drive
+		// all of these are not part of the data restored.
```
ashmrtn, line 285: The meta files are still used during restore (e.x. to source file names)
```suggestion
		// all of these are not directly part of the data restored.
```

src/internal/tester/config.go
```diff
@@ -112,36 +112,31 @@ func readTestConfig() (map[string]string, error) {
...
-	)
+		"AdeleV@10rqc2.onmicrosoft.com,LynneR@10rqc2.onmicrosoft.com")
 	fallbackTo(
 		testEnv,
 		TestCfgSiteID,
 		os.Getenv(EnvCorsoM365TestSiteID),
 		vpr.GetString(TestCfgSiteID),
-		"10rqc2.sharepoint.com,bb5d5b4a-e089-4e66-9868-9e263ecc635d,4fa3a2c0-fa81-4e6f-8e8b-1479a8927bc6",
-	)
+		"10rqc2.sharepoint.com,4892edf5-2ebf-46be-a6e5-a40b2cbf1c1a,38ab6d06-fc82-4417-af93-22d8733c22be")
```
ashmrtn, line 139: is the purge script setup to handle this SharePoint site?

src/internal/operations/restore_test.go
```diff
@@ -218,38 +222,68 @@ func (suite *RestoreOpIntegrationSuite) SetupSuite() {
 	require.NotEmpty(t, bo.Results.BackupID)
```
ashmrtn, line 222: nit: can you actually make separate subtests for each backup type? That will help with output in github actions. It's hard to find the output for the service we want if the setup fails because it has both assertions and log output all under one section and some services are chatty with the logs

src/internal/connector/onedrive/api/drive.go
```diff
@@ -203,75 +203,6 @@ func (p *siteDrivePager) ValuesIn(l api.PageLinker) ([]models.Driveable, error)
...
-
-		p.builder = mssites.NewItemDrivesRequestBuilder(link, p.gs.Adapter())
-	}
-
-	return empty, nil
-}
-
-// GetFolderIDByName is a helper function to retrieve the M365ID of a folder within a site document library.
-// Returns "" if the folder is not within the drive
-func (p *siteDrivePager) GetFolderIDByName(ctx context.Context, driveID, folderName string) (string, error) {
```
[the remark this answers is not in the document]
ryanfkeepers: This was part of a post-test cleanup, where the intent was to immediately delete the folders generated by the test restore.  The intent was good, but the implementation had some problems: it didn't support multi-drive deletion (drives outside of the standard Documents weren't getting cleaned up), and it was request heavy, which is not something we need more of in CI.  Since we have the cleanup script running, and we aren't using these calls elsewhere, there's no need to keep them around.

src/internal/connector/onedrive/restore.go
```diff
@@ -186,16 +186,12 @@ func RestoreCollection(
...
 		colMeta,
-		permissionIDMappings,
-	)
+		permissionIDMappings)
 	if err != nil {
 		return metrics, folderMetas, permissionIDMappings, clues.Wrap(err, "creating folders for restore")
 	}
 
-	var (
-		el    = errs.Local()
```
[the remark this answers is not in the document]
ryanfkeepers: It didn't get removed, just moved up to line 144.  The difference is immaterial, except for grouping more vars at the top of the func.

src/internal/connector/onedrive/restore.go
```diff
@@ -217,129 +213,189 @@ func RestoreCollection(
...
-					true,
-					itemInfo)
-				if err != nil {
-					// Not critical enough to need to stop restore operation.
-					logger.Ctx(ctx).Infow("accounting for restored item", "error", err)
-				}
-
-				metrics.Successes++
+			if skipped {
+				logger.Ctx(ctx).With("item_path", itemPath).Debug("did not restoring item")
```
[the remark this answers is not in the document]
ryanfkeepers: grammaring is hard

src/internal/connector/onedrive/restore.go
```diff
@@ -217,129 +213,189 @@ func RestoreCollection(
...
+		if err != nil {
+			return details.ItemInfo{}, false, clues.Wrap(err, "v1 restore")
+		}
+
+		return itemInfo, false, nil
+	}
+
+	// only v6+ backups from this point on
+
+	itemInfo, err := restoreV2File(
```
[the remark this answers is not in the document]
ryanfkeepers: 👍 I was wondering if there was reason for the delta between versions there.

src/internal/connector/onedrive/restore.go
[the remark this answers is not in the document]
ashmrtn: ahhhhh you're right. We should rename one of these functions lol. I was worried it got moved to `RestoreCollections` with an `s` which is not called by SharePoint restore -_-

src/internal/tester/config.go
ryanfkeepers: The purge script runs on whatever secret var we populate in our GH env.  I've set up the current site with the same environment expectations (two drives, test folder, one item each), so if the script is running on the current site successfully, we _should_ be able to swap between the two without issue.

src/internal/operations/restore_test.go
ryanfkeepers: Oh, good idea!  Can do.

ashmrtn (approved):

PR #2955, opened 2023-03-27
introduce id-name lookup maps

ryanfkeepers:
Adds two maps to resource-owner handling:
id-to-name and name-to-id.  Expectation is that
these maps will either get populated by a caller
as a pre-process before initializing the gc client, or
gc will (later pr) be able to look up the owner and
populate those maps itself.  The maps are
used to set the selector id and name for iface
compliance.  Only supported by exchange in this PR.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -13,6 +13,7 @@ import (
 	"github.com/alcionai/corso/src/cli/options"
 	. "github.com/alcionai/corso/src/cli/print"
 	"github.com/alcionai/corso/src/cli/utils"
+	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/internal/data"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/backup"
@@ -194,8 +195,7 @@ func runBackups(
 	r repository.Repository,
 	serviceName, resourceOwnerType string,
 	selectorSet []selectors.Selector,
-	resourceOwnersIDToName map[string]string,
-	resourceOwnersNameToID map[string]string,
+	ins common.IDNameSwapper,
 ) error {
 	var (
 		bIDs []string
@@ -208,7 +208,7 @@ func runBackups(
 			ictx  = clues.Add(ctx, "resource_owner", owner)
 		)
 
-		bo, err := r.NewBackup(ictx, discSel, resourceOwnersIDToName, resourceOwnersNameToID)
+		bo, err := r.NewBackup(ictx, discSel, ins)
 		if err != nil {
 			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
 			Errf(ictx, "%v\n", err)
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -194,6 +194,8 @@ func runBackups(
 	r repository.Repository,
 	serviceName, resourceOwnerType string,
 	selectorSet []selectors.Selector,
+	resourceOwnersIDToName map[string]string,
+	resourceOwnersNameToID map[string]string,
 ) error {
 	var (
 		bIDs []string
@@ -203,21 +205,21 @@ func runBackups(
 	for _, discSel := range selectorSet {
 		var (
 			owner = discSel.DiscreteOwner
-			bctx  = clues.Add(ctx, "resource_owner", owner)
+			ictx  = clues.Add(ctx, "resource_owner", owner)
 		)
 
-		bo, err := r.NewBackup(bctx, discSel)
+		bo, err := r.NewBackup(ictx, discSel, resourceOwnersIDToName, resourceOwnersNameToID)
 		if err != nil {
-			errs = append(errs, clues.Wrap(err, owner).WithClues(bctx))
-			Errf(bctx, "%v\n", err)
+			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
+			Errf(ictx, "%v\n", err)
 
 			continue
 		}
 
-		err = bo.Run(bctx)
+		err = bo.Run(ictx)
 		if err != nil {
-			errs = append(errs, clues.Wrap(err, owner).WithClues(bctx))
-			Errf(bctx, "%v\n", err)
+			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
+			Errf(ictx, "%v\n", err)
 
 			continue
 		}
--- a/src/internal/common/idname.go
+++ b/src/internal/common/idname.go
@@ -1,9 +1,47 @@
 package common
 
+import "golang.org/x/exp/maps"
+
 type IDNamer interface {
 	// the canonical id of the thing, generated and usable
 	//  by whichever system has ownership of it.
 	ID() string
 	// the human-readable name of the thing.
 	Name() string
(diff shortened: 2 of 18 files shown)


meain (commented):
Code changes LGTM, but I'll let `ashmrtn` take a look as well, mostly into the behaviour of `PopulateOwnerIDAndNamesFrom` and `getOwnerIDAndNameFrom`.

src/cli/backup/backup.go
```diff
@@ -194,6 +194,8 @@ func runBackups(
 	r repository.Repository,
 	serviceName, resourceOwnerType string,
 	selectorSet []selectors.Selector,
+	resourceOwnersIDToName map[string]string,
+	resourceOwnersNameToID map[string]string,
```
ashmrtn, line 198: since this encapsulates a complete idea (the ability to go from ID->PN and vice-versa) and is passed through multiple layers of the stack at least make it a struct with a well defined interface. Ideally it would be an interface with a struct as the concrete implementation so that we could easily swap it out later without depending on the specific struct details

If you make it just a struct, please don't export the fields and use functions instead. Again, this is so we don't come to rely on specific struct internals and can swap it out more easily in the future if we need to

src/internal/connector/graph_connector.go
```diff
@@ -49,9 +49,17 @@ type GraphConnector struct {
...
 	credentials account.M365Config
 
+	// TODO: remove in favor of the maps below.
+	Sites map[string]string // webURL -> siteID and siteID -> webURL
+
+	// maps of resource owner ids to names, and names to ids.
+	// not guaranteed to be populated, only here as a post-population
+	// reference for processes that choose to populate the values.
+	ResourceOwnerIDToName map[string]string
+	ResourceOwnerNameToID map[string]string
```
ashmrtn, line 61: do these need to be exported? I'm hesitant to say GC should guarantee that some data is pre-populated as that puts us in an awkward situation when it comes to repeating work etc. Having these public makes it harder to remove those guarantees because there's no way to hide changes in functionality (e.x. no function to change the implementation of behind the scenes)

src/internal/connector/graph_connector.go
```diff
@@ -108,6 +116,63 @@ func NewGraphConnector(
...
+	} else if i, ok := nameToID[owner]; ok {
+		return i, owner, nil
+	}
+
+	// TODO: look-up user by owner, either id or name,
+	// and populate with maps as a result.  Only
+	// return owner, owner as a very last resort.
+
+	return owner, owner, nil
+}
```
ashmrtn, line 174: this should be done by whatever interface/struct is responsible for keeping the maps. It should also probably be done prior to reaching GC

src/pkg/services/m365/m365.go
```diff
@@ -55,34 +55,29 @@ func Users(ctx context.Context, acct account.Account, errs *fault.Bus) ([]*User,
 	return ret, nil
 }
 
-func UserIDs(ctx context.Context, acct account.Account, errs *fault.Bus) ([]string, error) {
+// UsersMap retrieves all users in the tenant, and returns two maps: one id-to-principalName,
+// and one principalName-to-id.
+func UsersMap(
```
ashmrtn, line 60: do SDK consumers need all the results this function has been modified to return? It may be easier to make a new function rather than force SDK consumers to update their calls

src/internal/connector/graph_connector.go
ryanfkeepers: GC is the one responsible for constructing/storing the maps.  Per prior lookup, the eventual resolution of these changes is to more lazily defer owner lookup until this moment, rather than eagerly pull all known owners and filter down to the ones involved.

src/internal/common/idname.go
```diff
@@ -1,9 +1,47 @@
...
+	NameOf(id string) (string, bool)
+	IDs() []string
+	Names() []string
+}
+
+var _ IDNameSwapper = &IDsNames{}
+
+type IDsNames struct {
+	IDToName map[string]string
+	NameToID map[string]string
```
ashmrtn, line 24: please don't export these. People get tempted to use them in weird ways if they are and then we're tied to a specific implementation (cause there's only one implementation at the moment so why not just pass the concrete type instead of the interface right? ;) ). If you need to allow setting them use a constructor function

src/internal/connector/graph_connector.go
```diff
@@ -108,6 +116,59 @@ func NewGraphConnector(
...
+	}
+
+	gc.IDNameLookup = ins
+
+	if len(ins.IDs()) == 0 && len(ins.Names()) == 0 {
+		gc.IDNameLookup = common.IDsNames{
+			IDToName: map[string]string{id: name},
+			NameToID: map[string]string{name: id},
+		}
+	}
```
ashmrtn, line 150: either `ins` will be nil and needs created or it will be set to some value by earlier code
```suggestion
	if ins == nil {
		gc.IDNameLookup = common.NewIDNameSwapper()
		return "", "", nil
	}
	
	// move this to GC method
	id, name, err := getOwnerIDAndNameFrom(owner, ins)
	if err != nil {
		return "", "", errors.Wrap(err, "resolving resource owner details")
	}
	
	gc.IDNameLookup = ins
```

src/internal/connector/graph_connector.go
```diff
@@ -49,9 +50,16 @@ type GraphConnector struct {
...
-	Sites       map[string]string // webURL -> siteID and siteID -> webURL
 	credentials account.M365Config
 
+	// TODO: remove in favor of the maps below.
+	Sites map[string]string // webURL -> siteID and siteID -> webURL
+
+	// lookup for resource owner ids to names, and names to ids.
+	// not guaranteed to be populated, only here as a post-population
+	// reference for processes that choose to populate the values.
+	IDNameLookup common.IDNameSwapper
```
ashmrtn, line 61: still would prefer if this isn't exported so it's easier to change when it's initialized down the line

src/internal/connector/graph_connector_test.go
```diff
@@ -123,18 +124,138 @@ func (suite *GraphConnectorUnitSuite) TestUnionSiteIDsAndWebURLs() {
...
+			t := suite.T()
+
 			result, err := gc.UnionSiteIDsAndWebURLs(ctx, test.ids, test.urls, fault.New(true))
 			assert.NoError(t, err, clues.ToCore(err))
 			assert.ElementsMatch(t, test.expect, result)
 		})
 	}
 }
 
+func (suite *GraphConnectorUnitSuite) TestPopulateOwnerIDAndNamesFrom() {
```
ashmrtn, line 139: need another function to test when `ins` is nil. Pretty sure current code will NPE for that case

ashmrtn (approved):
need at least another test case and minor refactor of the GC function

src/internal/common/idname.go
ryanfkeepers: I'm overloading common here already.  Deferring this to a later cleanup step.

src/internal/connector/graph_connector_test.go
ryanfkeepers: aye, good call

Issue #1995, opened 2022-12-29
Move sharepoint site retrieval to connector/discovery


ryanfkeepers:
## Issue
The retrieval of users was moved from graph_connector to discovery to aid in exposing apis for sdk consumers.  Sharepoint sites (which are a category of resourceOwners similar to users) were being added to the graph_connector file around the same time, thus missing that migration.  It makes the most sense to co-locate both types of resource owners in the discovery package.

gmatev:
This probably needs to happen sooner rather than later since the site discovery is being consumed.

PR #2958, opened 2023-03-27
move sharepoint onto map-based lookup

ryanfkeepers:
Refactors the sharepoint site lookup to use the
id-to-name maps.  This has a momentary regression
that will get solved in the next PR: we no longer
match on weburl suffixes, and instead require
a complete match.

Also, migrates the sharepoint lookup code out
of GC and into Discovery.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825
* #1995
* #1547

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
@ashmrtn adding you to this (and future) PRs in the chain to keep things moving along.  Thank you for the reviews so far!

src/cli/backup/sharepoint.go
ashmrtn, line 1: There's a lot of little changes in here, do we have adequate test coverage to make sure things still work as expected?

src/internal/connector/data_collections.go
```diff
@@ -47,7 +47,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 		diagnostics.Index("service", sels.Service.String()))
 	defer end()
 
-	err := verifyBackupInputs(sels, gc.GetSiteIDs())
+	err := verifyBackupInputs(sels, gc.IDNameLookup.IDs())
```
ashmrtn, line 50: is this still all site IDs?

src/internal/connector/discovery/api/sites.go
```diff
@@ -0,0 +1,144 @@
...
+	service, err := c.service()
+	if err != nil {
+		return nil, err
+	}
+
+	var resp models.SiteCollectionResponseable
+
+	resp, err = service.Client().Sites().Get(ctx, nil)
+
+	if err != nil {
```
ashmrtn, line 46: nit: seems like this could be cleaned up a small amount
```suggestion
	resp, err := service.Client().Sites().Get(ctx, nil)
	if err != nil {
```

src/internal/connector/discovery/api/sites.go
```diff
@@ -0,0 +1,144 @@
...
+// identifiers that we handle with all users.
+// returns the item as a Siteable model.
+func validateSite(item any) (models.Siteable, error) {
+	m, ok := item.(models.Siteable)
+	if !ok {
+		return nil, clues.New(fmt.Sprintf("unexpected model: %T", item))
+	}
+
+	id, ok := ptr.ValOK(m.GetId())
+	if !ok || len(id) == 0 {
```
ashmrtn, line 118: this and below can just use `ptr.Val` and do a len check as it's equivalent and we're not giving different errors?
```suggestion
	id := ptr.Val(m.GetId())
	if len(id) == 0 {
```

src/internal/connector/discovery/api/sites.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, clues.New("missing ID")
+	}
+
+	url, ok := ptr.ValOK(m.GetWebUrl())
+	if !ok || len(url) == 0 {
+		return nil, clues.New("missing webURL").With("site_id", id) // TODO: pii
+	}
+
+	// personal (ie: oneDrive) sites have to be filtered out server-side.
+	if ok && strings.Contains(url, personalSitePath) {
```
ashmrtn, line 128: probably don't need to check `ok` here. If the input is an empty string we're still going to continue through this function because the `Contains` will be false

src/internal/connector/discovery/api/sites_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (suite *SitesUnitSuite) TestValidateSite() {
+	site := models.NewSite()
+	site.SetWebUrl(ptr.To("sharepoint.com/sites/foo"))
+	site.SetDisplayName(ptr.To("testsite"))
+	site.SetId(ptr.To("testID"))
+
+	tests := []struct {
+		name           string
+		args           interface{}
```
ashmrtn, line 31: did this get changed to `any` in newer golang versions?
```suggestion
		args           any
```

src/pkg/services/m365/m365.go
```diff
@@ -98,55 +112,56 @@ type Site struct {
...
-}
-
-// parseUser extracts information from `models.Userable` we care about
-func parseUser(item models.Userable) (*User, error) {
-	if item.GetUserPrincipalName() == nil {
-		return nil, clues.New("user missing principal name").
-			With("user_id", clues.Hide(ptr.Val(item.GetId())))
+	ins := common.IDsNames{
+		IDToName: make(map[string]string, len(sites)),
+		NameToID: make(map[string]string, len(sites)),
```
ashmrtn, line 158: this makes me nervous

src/pkg/services/m365/m365.go
```diff
@@ -98,55 +112,56 @@ type Site struct {
...
+		}
+
+		ret = append(ret, ps)
 	}
 
 	return ret, nil
 }
 
-// SiteURLs returns a list of SharePoint site WebURLs in the specified M365 tenant
-func SiteURLs(ctx context.Context, acct account.Account, errs *fault.Bus) ([]string, error) {
```
ashmrtn, line 119: double checking: nobody was using this right? Same with `SiteIDs` below

src/pkg/services/m365/m365.go
```diff
@@ -98,55 +112,56 @@ type Site struct {
...
-// SiteIDs returns a list of SharePoint sites IDs in the specified M365 tenant
-func SiteIDs(ctx context.Context, acct account.Account, errs *fault.Bus) ([]string, error) {
-	gc, err := connector.NewGraphConnector(ctx, graph.HTTPClient(graph.NoTimeout()), acct, connector.Sites, errs)
+// SitesMap retrieves all sites in the tenant, and returns two maps: one id-to-webURL,
+// and one webURL-to-id.
+func SitesMap(
+	ctx context.Context,
+	acct account.Account,
+	errs *fault.Bus,
+) (common.IDsNames, error) {
```
ashmrtn, line 150: do we intend SDK consumers to use this? If we do we may need to move `IDsNames` to a different package as it's in `internal` right now

src/internal/connector/data_collections.go
ryanfkeepers: Yep.  id-name guarantees should be complete by this stage.

src/pkg/services/m365/m365.go
ryanfkeepers: As far as I have researched, yes.

src/pkg/services/m365/m365.go
ryanfkeepers: Since consumers shouldn't need to _populate_ idsnames, only receive and use its interface (and pass it into interface-accepting funcs, if anything), it's okay for the struct to remain internal.

src/pkg/services/m365/m365.go
ryanfkeepers: Can you explain your nervousness, please?

ryanfkeepers:
> There's a lot of little changes in here, do we have adequate test coverage to make sure things still work as expected?

I feel pretty good about it.  Most of this PR is lift and shift.  What's new (exposing the interface in services/m365) is thin on logic and underpinned by code that already has coverage.

src/pkg/services/m365/m365.go
ashmrtn: I'm not really a fan of modifying internal field structs in packages that don't own said structs

ashmrtn (approved):
Please make sure to thoroughly test CLI functionality to make sure it still works as expected. We may not have quite as much testing around that as we found when we switched over the CLI's variables

src/pkg/services/m365/m365.go
ryanfkeepers: technically this is construction, not modification.  It'll turn into a real constructor on a later pass.

ryanfkeepers:
> Please make sure to thoroughly test CLI functionality to make sure it still works as expected. We may not have quite as much testing around that as we found when we switched over the CLI's variables

Already done throughout modifications and review.  Everything looks clean so far.

PR #2960, opened 2023-03-27
correct input aliasing

ryanfkeepers:
Updates and corrects input aliasing according to
the following rules (in priority order):
1. if the library name is usable, use it
2. if not, alias to the package name
3. if the package name is weird, alias sensibly
4. in case of collision, alias more distant imports
5. aliases should be consistent throughout

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1970

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/mockconnector/mock_data_message.go
+++ b/src/internal/connector/mockconnector/mock_data_message.go
@@ -6,8 +6,8 @@ import (
 	"testing"
 
 	"github.com/alcionai/clues"
-	serialization "github.com/microsoft/kiota-abstractions-go/serialization"
-	kioser "github.com/microsoft/kiota-serialization-json-go"
+	"github.com/microsoft/kiota-abstractions-go/serialization"
+	kjson "github.com/microsoft/kiota-serialization-json-go"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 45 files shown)


src/internal/connector/mockconnector/mock_data_message.go
```diff
@@ -5,13 +5,12 @@ import (
 	"fmt"
 	"testing"
 
-	absser "github.com/microsoft/kiota-abstractions-go/serialization"
-	js "github.com/microsoft/kiota-serialization-json-go"
+	"github.com/alcionai/clues"
+	serialization "github.com/microsoft/kiota-abstractions-go/serialization"
```
meain, line 9: Suggestion:

```suggestion
	kserialization "github.com/microsoft/kiota-abstractions-go/serialization"
```

src/internal/connector/mockconnector/mock_data_message.go
```diff
@@ -5,13 +5,12 @@ import (
...
+	kioser "github.com/microsoft/kiota-serialization-json-go"
```
meain, line 10: Suggestion:

```suggestion
	kjson "github.com/microsoft/kiota-serialization-json-go"
```

src/internal/connector/sharepoint/collection.go
```diff
@@ -7,13 +7,13 @@ import (
 	"time"
 
 	"github.com/alcionai/clues"
-	absser "github.com/microsoft/kiota-abstractions-go/serialization"
-	kw "github.com/microsoft/kiota-serialization-json-go"
+	"github.com/microsoft/kiota-abstractions-go/serialization"
+	kioser "github.com/microsoft/kiota-serialization-json-go"
```
meain, line 11: Same as previously mentioned if we are updating it.

meain (approved):
Just two suggestions, approving.

src/internal/connector/mockconnector/mock_data_message.go
ryanfkeepers: Oh, that shouldn't be aliased in the first place.

src/internal/connector/mockconnector/mock_data_message.go
ryanfkeepers: I dig it

PR #2967, opened 2023-03-28
Re add --restore-permissions flag

meain:
This seems to have accidentally removed in 285f8f0

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/restore/onedrive_test.go
+++ b/src/cli/restore/onedrive_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
+	"github.com/alcionai/corso/src/cli/utils"
 	"github.com/alcionai/corso/src/internal/tester"
 )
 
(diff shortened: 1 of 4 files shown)


src/cli/restore/onedrive_test.go
```diff
@@ -50,3 +51,36 @@ func (suite *OneDriveSuite) TestAddOneDriveCommands() {
...
+			require.NotNil(t, c)
+
+			cmd.SetArgs([]string{"onedrive", "--restore-permissions"})
+
+			actual := new(bytes.Buffer)
+			cmd.SetOut(actual)
+			cmd.SetErr(actual)
+			cmd.Execute()
+
+			assert.NotContains(t, actual.String(), "Error: unknown flag: --restore-permissions")
```
meain, line 83: Not sure if this test is too hacky. I don't know if there is a simpler way to test this with our current setup.

src/cli/restore/onedrive_test.go
```diff
@@ -50,3 +50,32 @@ func (suite *OneDriveSuite) TestAddOneDriveCommands() {
...
+
+			cmd := &cobra.Command{Use: test.use}
+
+			c := addOneDriveCommands(cmd)
+			require.NotNil(t, c)
+
+			cmd.SetArgs([]string{"onedrive", "--restore-permissions"})
+
+			err := cmd.Execute()
+			assert.NotEqual(t, err.Error(), `unknown flag: --restore-permissions`)
```
meain, line 78: Not sure if this test is too hacky. I don't know if there is a simpler way to test this with our current setup.

ryanfkeepers: I don't think you need to execute the command.  Since flags are stored in the command, once you call `addOneDriveCommands(cmd)`, you can get check for the flag from the command with something like https://pkg.go.dev/github.com/spf13/cobra#Command.Flag.  Or you could iterate through https://pkg.go.dev/github.com/spf13/cobra#Command.LocalFlags.

ryanfkeepers (approved):

src/cli/restore/onedrive_test.go
meain: Yeah, having tests for flags seems like simple enough, but pretty useful. Let me [add](https://github.com/alcionai/corso/issues/2975) a `tech-debt` issue for now.

PR #2969, opened 2023-03-28
restrict sharepoint test backups to test folders

ryanfkeepers:
To avoid over-production of data, and thus longer
CI times, any sharepoint integration test that can scope
library usage down to only the "test" folders should do so.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections_test.go
```diff
@@ -334,7 +335,7 @@ func (suite *ConnectorCreateSharePointCollectionIntegrationSuite) TestCreateShar
 	)
 
 	sel := selectors.NewSharePointBackup(siteIDs)
-	sel.Include(sel.LibraryFolders([]string{"foo"}, selectors.PrefixMatch()))
+	sel.Include(sel.LibraryFolders([]string{"zzazil"}, selectors.PrefixMatch()))
```
pandeyabs, line 338: What does this magic keyword do?

src/pkg/selectors/testdata/sharepoint.go
```diff
@@ -0,0 +1,9 @@
+package testdata
+
+import "github.com/alcionai/corso/src/pkg/selectors"
+
+// BackupFolderScope is the standard folder scope that should be used
+// in integration backups with sharepoint.
+func BackupFolderScope(sel *selectors.SharePointBackup) []selectors.SharePointScope {
```
ashmrtn, line 7: nit: should prefix this as `SharePointBackupFolderScope` as other services may need the same sort of function in the future

src/internal/connector/data_collections_test.go
ryanfkeepers: That magic keyword reminds me that I need to come back and investigate a change in that location.  Usually.  Eventually.

ryanfkeepers:
@pandeyabs @ashmrtn thanks for the comments.  This is ready for a final review.

pandeyabs (approved):

Issue #2707, opened 2023-03-06
CI Test Failure: Http transport connection broken


ryanfkeepers:
## Issue
We're seeing intermittent failures in both GH and local integration testing with erroneous responses from graph api.

Link to one such failed test: https://github.com/alcionai/corso/actions/runs/4298555994/jobs/7492799605

Error Output:
```
# GH:
Delete "https://graph.microsoft.com/v1.0/users/***/mailFolders/***": net/http: HTTP/1.x transport connection broken: malformed HTTP response "0"

# Local:
2023/03/03 17:26:21 Unsolicited response received on idle HTTP channel starting with "0\r\n\r\n"; err=<nil>
```

Response payload dump from linked error:
```
HTTP/1.1 204 No Content
Transfer-Encoding: chunked
Cache-Control: private
Client-Request-Id: e276360f-029a-456d-8879-e625bfacf1c0
Date: Wed, 01 Mar 2023 00:19:48 GMT
Request-Id: 8eb66e4a-4797-4623-96de-3f977020f273
Strict-Transport-Security: max-age=31536000
X-Ms-Ags-Diagnostic: {"ServerInfo":{"DataCenter":"East US","Slice":"E","Ring":"5","ScaleUnit":"005","RoleInstance":"MN1PEPF00002F6D"}}

0


```

meain:
https://groups.google.com/g/golang-nuts/c/ItC745n3MrI seems to be what we are running into. `Transfer-Encoding` is set to `chunked` in the graph response for pretty much all endpoints. Also looking at the previous request printed out by the CI logs, it looks like a 204 which might be the problem, maybe. ([related](https://github.com/golang/go/issues/4454)).

https://github.com/microsoftgraph/msgraph-sdk-java-core/issues/138 could also be related.

meain:
The problem seem to be that Microsoft server is sending `0\r\n\r\n` for `204 No Content` even thought it should be **no content**. *The random string is the end/empty marker for `Transfer-Encoding: chunked` which is somehow always set even if this request is not chunked.*

https://github.com/golang/go/issues/19895
https://www.rfc-editor.org/rfc/rfc9110.html#section-15.3.5-5

meain:
Good news and bad news.
Bad news: Don't think we can fix it on our end
Good new: This won't happen during actual backups as this only affects responses returning 204 which is only delete requests which we are not doing during corso backup

vkamra:
[...]

PR #2970, opened 2023-03-28
create a new http client on every delete

ryanfkeepers:
in order to avoid errors cascading from graph API's
populated 204 responses containing `0/r/n/r/n`,
in particular in response to deletes, this change
hacks in the creation of a unique http client for
each delete call.

This is dangerous to do, since we already know
that generating clients per calls can leak
consumption of system resources such as sockets.
This change presumes that runtime deletes are
minimally, if ever, called during standard backup
and restore.  The permission deletion in onedrive
is the stand-out exception that might need some
additional investigation.
---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2707

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/onedrive/permission.go
```diff
@@ -194,7 +198,17 @@ func restorePermissions(
 	permAdded, permRemoved := diffPermissions(currentPermissions, meta.Permissions, permissionIDMappings)
 
 	for _, p := range permRemoved {
-		err := service.Client().
+		// deletes require unique http clients
+		// https://github.com/alcionai/corso/issues/2707
+		// this is bad citizenship, and could end up consuming a lot of
+		// system resources if servicers leak client connections (sockets, etc).
+		a, err := graph.CreateAdapter(m365.AzureTenantID, m365.AzureClientID, m365.AzureClientSecret)
```
ryanfkeepers, line 205: @meain I'm a bit concerned about this point, since it feels like we might accumulate a lot of calls during a restore runtime.  Do you know any expectations about how frequently this could get invoked?

src/internal/connector/onedrive/permission.go
```diff
@@ -173,6 +176,7 @@ func diffPermissions(
 // the necessary permissions on onedrive objects.
 func restorePermissions(
 	ctx context.Context,
+	m365 account.M365Config,
```
meain, line 179: nit: (just for consistency)
```suggestion
	creds account.M365Config,
```

src/internal/connector/onedrive/permission.go
```diff
@@ -89,6 +90,7 @@ func getCollectionMetadata(
 // Passing nil for the permissions results in just creating the folder(s).
 func createRestoreFoldersWithPermissions(
 	ctx context.Context,
+	creds account.M365Config,
```
meain, line 93: Having  to pass down `cdes` seems unncessary. Should we just add a method into `graph.Services` which gets us a new client from the available adapter?

meain (changes requested):

src/internal/connector/onedrive/permission.go
meain: I feel like this should be pretty limited in most normal cases.

src/internal/connector/onedrive/permission.go
ryanfkeepers: That was the original goal.  But the http client in the adapter is unexported.  We'd have to store the adapter's original http client (and, preferably, configuration) separately in the graph.Service.  It's an option, maybe one we'll return to, but this felt like a more direct solution in the meantime.

meain (approved):

PR #2983, opened 2023-03-29
match sites with a url path suffix

ryanfkeepers:
Adds suffix matching to the site owner id/name
lookup process.  This allows consumers to query
for a site by only its path (ex: "/sites/foo") and still end up with a well formed id, name tuple.  One
gotcha is that this requires the lookup maps to
be populated.  If a lookup map is not passed in
with a suffix matcher, the fallback lookup will fail.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph_connector.go
+++ b/src/internal/connector/graph_connector.go
@@ -236,8 +236,10 @@ type getOwnerIDAndNamer interface {
 // (PrincipalName for users, WebURL for sites).
 //
 // Consumers are allowed to pass in a path suffix (eg: /sites/foo) as a site
-// owner, but only if they also pass in a nameToID map.  A nil map will cascade
-// to the fallback, which will fail for having a malformed id value.
+// owner, but only if they also pass in a nameToID map.  A nil map will fallback
+// to a lookup by id using the suffix, which will fail for being malformed.
+// Suffix matching follows filter PathSuffix rules: only complete elements match,
+// case-independant, leading/tailing slashes are optional.
 func (r resourceClient) getOwnerIDAndNameFrom(
 	ctx context.Context,
 	discovery api.Client,
@@ -261,7 +263,7 @@ func (r resourceClient) getOwnerIDAndNameFrom(
 
 	// check if the provided owner is a suffix of a weburl in the lookup map
 	if r.enum == Sites {
-		url, _, ok := filters.PathSuffix([]string{owner}).CompareAny(ins.Names()...)
+		url, ok := filters.PathSuffix([]string{owner}).CompareAny(ins.Names()...)
 		if ok {
 			id, _ := ins.IDOf(url)
 			return id, url, nil
--- a/src/internal/connector/graph_connector.go
+++ b/src/internal/connector/graph_connector.go
@@ -18,6 +18,7 @@ import (
 	"github.com/alcionai/corso/src/internal/operations/inject"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/filters"
 )
 
 // ---------------------------------------------------------------------------
@@ -233,6 +234,10 @@ type getOwnerIDAndNamer interface {
 // api to fetch the user or site using the owner value. This fallback assumes
 // that the owner is a well formed ID or display name of appropriate design
 // (PrincipalName for users, WebURL for sites).
+//
+// Consumers are allowed to pass in a path suffix (eg: /sites/foo) as a site
+// owner, but only if they also pass in a nameToID map.  A nil map will cascade
+// to the fallback, which will fail for having a malformed id value.
 func (r resourceClient) getOwnerIDAndNameFrom(
 	ctx context.Context,
 	discovery api.Client,
@@ -254,9 +259,14 @@ func (r resourceClient) getOwnerIDAndNameFrom(
 		err      error
 	)
 
-	// if r.enum == Sites {
-	// TODO: check all suffixes in nameToID
-	// }
+	// check if the provided owner is a suffix of a weburl in the lookup map
+	if r.enum == Sites {
+		url, _, ok := filters.PathSuffix([]string{owner}).CompareAny(ins.Names()...)
+		if ok {
+			id, _ := ins.IDOf(url)
+			return id, url, nil
+		}
+	}
 
 	id, name, err = r.getter.GetIDAndName(ctx, owner)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


src/internal/connector/graph_connector.go
```diff
@@ -233,6 +234,10 @@ type getOwnerIDAndNamer interface {
 // api to fetch the user or site using the owner value. This fallback assumes
 // that the owner is a well formed ID or display name of appropriate design
 // (PrincipalName for users, WebURL for sites).
+//
+// Consumers are allowed to pass in a path suffix (eg: /sites/foo) as a site
+// owner, but only if they also pass in a nameToID map.  A nil map will cascade
+// to the fallback, which will fail for having a malformed id value.
```
ashmrtn, line 240: I have a few concerns with where/how this is implemented and the comments
* it's not clear from the comment alone how much the suffix encompasses as it never states what the SP URL is like to begin with. For example, is `foo` a valid suffix?
* it's not clear this will work in all cases (maybe partly due to the above) as it's not clear if it will allows matching on multiple sites. There's no logic to select the longest match so final results may actually be indeterminate. For example, if the suffix is `foo`, what if there is `sites/foo` and `sites/bar/foo` (idk if `sites/bar/foo` is valid as I don't know as much about SP but seems like we should have answers to this if we want to add this feature or we need more definition as to what a suffix is)?
* I'm hesitant to implement this in GraphConnector's init logic (which iirc is where `getOwnerIDAndNameFrom` is called). Waiting until this point makes it difficult to enforce invariants about the state of data which can make reasoning about things difficult. I would suggest that if we want to allow this we require that some higher layer fixup stuff such that we have a well-formed, fully qualified URL by the time we reach this function (and an instance of GC being created for use during backup). This also avoids weirdness about the fallback failing and generally makes things more consistent

src/internal/connector/graph_connector_test.go
```diff
@@ -249,6 +250,18 @@ func (suite *GraphConnectorUnitSuite) TestPopulateOwnerIDAndNamesFrom() {
 			expectName: name,
 			expectErr:  require.NoError,
 		},
+		{
+			name:  "site suffix lookup",
```
ashmrtn, line 254: should have a test to show that it consistently selects the same ("correct"?) site when there are multiple sites that have the same suffix

src/pkg/filters/filters.go
```diff
@@ -359,21 +359,23 @@ func newSliceFilter(c comparator, targets, normTargets []string, negate bool) Fi
...
+// If nothing matches, returns ("", -1, false)
 //
 // Note that, as a gotcha, CompareAny can resolve truthily
 // for both the standard and negated versions of a filter.
 // Ex: consider the input CompareAny(true, false), which
 // will return true for both Equals(true) and NotEquals(true),
 // because at least one element matches for both filters.
-func (f Filter) CompareAny(inputs ...string) bool {
-	for _, in := range inputs {
+func (f Filter) CompareAny(inputs ...string) (string, int, bool) {
```
ashmrtn, line 371: I'm very hesitant to return indexes. It seems like an excellent footgun as it makes the assumption that the caller has a stable slice/array for the input which is something very subtle that can be missed. For example, having something that converts an underlying map type to a slice to return either the keys or the values could return incorrect results if the following were used

```go
sites := // map of ids -> names
_, idx, matched := filter.CompareAny(maps.Values(sites)...)

// Get the "correct" site ID.
siteID := maps.Keys(sites)[idx]
```

Ideally someone would realize the fact that this wouldn't work, but it's something very subtle and can't be easily enforced by a linter. The compiler won't catch it either so it's pretty easy to miss

src/internal/connector/graph_connector_test.go
ryanfkeepers: Hmm, we're not making any guarantees there (outside of basic correctness) in the first place.  Ordering matters.  The result will be consistent for a given array, but not for a given set of values.  I'm going to punt this to a follow-up PR where I can look at longest/shortest match control more generically.

src/pkg/filters/filters.go
ryanfkeepers: nbd, will remove; it was an afterthought anyway.

src/internal/connector/graph_connector.go
ryanfkeepers: 1: updated the comment for clarity.  2: going to check in with georgi on multiple matches.  My assumption is that we'll want to include all; but I hadn't thought of that up front, so it might require a second pass at the logic here.  3:

> init logic

Nitpicky: this isn't init so much as optional validation/population.  You can use GC without performing these calls.  Still, per location, I think this is the appropriate place for implementation.  After all, the intent of this logic is to validate and enforce expected preconditions.  Not to consume other layer's enforcement of them.

> higher layer fixup stuff such that we have a well-formed, fully qualified URL by the time we reach this function

The core problem here is that we can't construct a fully qualified url without knowing the hostname of the tenant.  [To do that](https://goodworkaround.com/2022/02/09/finding-azure-ad-domain-from-tenant-id/) not only requires a graph call, but a new permission we aren't currently requiring (might need it in the future, though, so I'll look into that as a request).  I see that as an iteration we can make, more than a requirement for this change.

ashmrtn: > >init logic
>
>Nitpicky: this isn't init so much as optional validation/population. You can use GC without performing these calls. Still, per location, I think this is the appropriate place for implementation. After all, the intent of this logic is to validate and enforce expected preconditions. Not to consume other layer's enforcement of them.

My concern is a bit more general than just the function I commented on. I'm more concerned with the entire initialization flow. Sure, the individual function may not be called, but if I were try to write down the state that GC ended up in after initialization I would have trouble doing that, and that complexity invites bugs because it's difficult to reason what state things are in and/or miss considering a case. If I were to take a stab at it, it would maybe be something like (including this change)
* populates name and ID from IDNamer if given
* if IDNamer not given, attempts to get it from Graph
* may throw an error, but only if IDNamer wasn't given and Graph didn't have it (for non-prefix users and sites)
* may *not* throw an error if IDNamer *was* given and a site prefix *was* given
* may throw an error if IDNamer *wasn't* given and a site prefix *was* given

ryanfkeepers: Well good news!  We're dropping suffix matching support for v0.  I'm going to close this PR and make a cleanup in its place.

ryanfkeepers:
Closing: replaced with https://github.com/alcionai/corso/pull/3082

PR #2990, opened 2023-03-30
cli naming readability and clarification

ryanfkeepers:
Appends FV to the shared values populated by
cobra flags.  This helps ensure those values are
not confused by other, like-named values used as
func params or local vars.  Also amends some
other varname issues.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test


src/cli/backup/exchange.go
```diff
@@ -148,7 +148,7 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	if err := validateExchangeBackupCreateFlags(utils.User, utils.CategoryData); err != nil {
+	if err := validateExchangeBackupCreateFlags(utils.UserFV, utils.CategoryDataFV); err != nil {
```
pandeyabs, line 151: noob q: What's FV?

pandeyabs (approved):
lgtm!

src/cli/backup/exchange.go
ryanfkeepers: Flag Value.  Not the best namespacing, but I've bumped into the "is this the foo populated by the flag, or the foo that I passed in as a param" bug a couple times now.  So having _something_ as a naming convention to separate values could identify those mixups.

PR #2991, opened 2023-03-30
Add GetUserInfo to services/m365

ryanfkeepers:
Adds an interface for GetUserInfo, which gets
corso-specific metadata for a single user.  Also
does some refactoring around discovery for
better interface consistency, both as variables
and as access layers.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/discovery/discovery.go
```diff
@@ -20,13 +19,17 @@ type getter interface {
...
 	GetInfo(context.Context, string) (*api.UserInfo, error)
 }
 
 type getWithInfoer interface {
 	getter
-	getInfoer
+	GetInfoer
+}
+
+type getAller interface {
```
pandeyabs, line 31: noob question: why is this getAller and not GetAller like line #22 above? Is this a naming pattern?

pandeyabs (approved):
Looks good!

src/internal/connector/discovery/discovery.go
ryanfkeepers: There's no immediate need to export it.  In case you haven't looked into golang var exporting yet: https://medium.com/golangspec/exported-identifiers-in-go-518e93cc98af

PR #3002, opened 2023-03-31
Incrementals blog pt1

serverless-mom:
Our Incremental release on Incremental backups :) will PR Part 2 shortly
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


gmatev:
@serverless-mom @ashmrtn the post and the mechanics of the PR are fine, but I would have loved to have at least some of the meat around the topic in this Part 1. In my opinion there is a lot of background info and not enough on the mechanics of incrementals. 

Is there a way to reduce the "Backup layout" or the "Terminology section" in favor of including some of the steps that follow? Another alternative would be delaying one of these sections to Part2 where it can precede some of the steps where it is relevant.

If none of that works or resonates, feel free to go with what you already have. 

In either case, it might be worthwhile to add a into section that will outline both parts of the blog and will give a map to the reader of what they will learn in Part 1 and what they will learn in Part 2. This may also mitigate some of my concerns from above.

serverless-mom:
@gmatev what if we just PR both part 1 and part 2 at the same time? There's good reasons not to post a single post that's quite so long, but we could have both parts go up on the same day, so that you're not left hanging when you finish part 1

PR #3003, opened 2023-03-31
move parallelism logs/checks upstream

ryanfkeepers:
The fetch paralellism checks and logs occur on
every item streamed from GC.  This is a bit chatty,
and has been moved upstream in the process for
a more centralized behavior.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -47,8 +47,8 @@ func (gc *GraphConnector) ProduceBackupCollections(
 	defer end()
 
 	// Limit the max number of active requests to graph from this collection.
-	ctrlOpts.ItemFetchParallelism = graph.Parallelism(sels.PathService()).
-		ItemOverride(ctx, ctrlOpts.ItemFetchParallelism)
+	ctrlOpts.Parallelism.ItemFetch = graph.Parallelism(sels.PathService()).
+		ItemOverride(ctx, ctrlOpts.Parallelism.ItemFetch)
 
 	err := verifyBackupInputs(sels, gc.IDNameLookup.IDs())
 	if err != nil {
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -19,7 +19,6 @@ import (
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/fault"
-	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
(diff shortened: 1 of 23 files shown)


src/internal/connector/graph/consts.go
```diff
@@ -25,3 +25,10 @@ const (
 	// given endpoint.
 	PreviousPathFileName = "previouspath"
 )
+
+// ---------------------------------------------------------------------------
+// Runtime Configuration
+// ---------------------------------------------------------------------------
+
+// URLItemFetchParallelism controls the worker pool size for streaming items.
+const URLItemFetchParallelism = 4
```
ashmrtn, line 34: this loses context about what service it's for. I assume OneDrive may have different parallelism needs than Exchange

ashmrtn: maybe missing a push? I still see the generic name from the previous commit

ryanfkeepers: heh, I expect people to be slow in response, so I'll say "it's done!" and then go work on the change.  😳

src/internal/connector/data_collections.go
```diff
@@ -46,6 +46,10 @@ func (gc *GraphConnector) ProduceBackupCollections(
 		diagnostics.Index("service", sels.Service.String()))
 	defer end()
 
+	// Limit the max number of active requests to graph
```
ashmrtn, line 49: nit: make clear this is not concurrency control across collections (yet?)

```suggestion
	// Limit the max number of active requests to graph from this collection
```

src/internal/connector/graph/consts.go
```diff
@@ -25,3 +32,87 @@ const (
...
 )
+
+// ---------------------------------------------------------------------------
+// Runtime Configuration
+// ---------------------------------------------------------------------------
+
+type parallelism struct {
+	collectionPool int
+	itemFetch      int
+	prefetchPool   int
```
ashmrtn, line 43: These need comments for what they're used for

src/internal/connector/graph/consts.go
```diff
@@ -25,3 +32,87 @@ const (
...
+	return p.collectionPool
+}
+
+func (p parallelism) CollectionPoolOverride(ctx context.Context, override int) int {
+	logger.Ctx(ctx).Infow(
+		"collection pool parallelism",
+		"default_parallelism", p.itemFetch,
+		"requested_paralellism", override)
+
+	if override < 1 || (p.collectionPool > 0 && override > p.collectionPool) {
```
ashmrtn, line 56: I know it's a simple comparison, but I kind of feel like we should have a separate helper function for this. It would make implementing minor changes to stuff a bit easier in the long run

ashmrtn (approved):

src/internal/connector/graph/consts.go
```diff
@@ -25,3 +32,87 @@ const (
...
+		collectionPool: 4,
+		itemFetch:      4,
+		prefetchPool:   4,
+	},
+	path.OneDriveService: {
+		collectionPool: 5,
+		itemFetch:      4,
+		prefetchPool:   5,
+	},
+	path.SharePointService: {
```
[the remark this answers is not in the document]
ryanfkeepers: I think the assumption is that library parameters will use the onedrive params, and that sharepoint entirely controls lists/page parallelism.

src/internal/connector/data_collections.go
```diff
@@ -46,6 +46,10 @@ func (gc *GraphConnector) ProduceBackupCollections(
...
+	ctrlOpts.ItemFetchParallelism = graph.Parallelism(sels.PathService()).
```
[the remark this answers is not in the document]
ryanfkeepers: I don't really see a case where we'd want to revert the value.  Do you?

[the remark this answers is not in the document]
ashmrtn: it's no so much about reverting, but making sure code for one part of the system doesn't pollute the values that code for another part of the system may see. To take an example, say we have backup for all SharePoint data types enabled. If we continue to have only a single control option then the different data types will fight for which value this is set to and also what value was passed into subsequent calls (i.e. accidentally use the output of OneDrive parameters as input for SP list parameters check because things are executed serially). If we scope it to downstream only (i.e. by returning a copy of the struct with updated values like with adding kv-pairs to context) then setting it for different services/data types can have different results which is what we'd want

Of course, we could add separate variables for each service/data type in the control options, but it seems more likely that could grow without bound and cause other problems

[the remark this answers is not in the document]
ryanfkeepers: Yeah, point.  Might make a tech debt for this.

PR #3005, opened 2023-03-31
introduce fail-after-recovery option

ryanfkeepers:
While fail-fast and best-effort make for good
categories at the extreme ends of error handling, 
we keep finding outselves wanting to operate
in a middle ground.  This change introduces a new
error handling category: FailAfterRecovery.  This
option tells corso to complete as much of its
process as it can, even if it recovers from errors.
But at the end of processing, if it recovered from
any errors, an error is returned for the operation.

This behavior is the new failure handling default,
instead of failFast.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/cli/options/options.go
+++ b/src/cli/options/options.go
@@ -11,7 +11,10 @@ import (
 func Control() control.Options {
 	opt := control.Defaults()
 
-	opt.FailFast = fastFail
+	if fastFail {
+		opt.FailureHandling = control.FailFast
+	}
+
 	opt.DisableMetrics = noStats
 	opt.RestorePermissions = restorePermissions
 	opt.SkipReduce = skipReduce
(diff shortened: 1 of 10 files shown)


src/cli/options/options.go
```diff
@@ -11,7 +11,10 @@ import (
 func Control() control.Options {
 	opt := control.Defaults()
 
-	opt.FailFast = fastFail
+	if fastFail {
+		opt.FailureHandling = control.FailFast
```
vkamra, line 15: What are our defaults for backup and restore?

For both it should be `failAfterRecovery`

ryanfkeepers: It is: see https://github.com/alcionai/corso/pull/3005/files#diff-8d31208fb3c9d64612a8181dd37f8031eaa1797a0a0c252829c939864df94ac5R32

vkamra (approved):

PR #3009, opened 2023-04-01
clean up exchangePurge output

ryanfkeepers:
Get contact counts isn't working. Just comments out the create for now.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual


ryanfkeepers:
@gmatev uncommented the contact creation and removed the non-functional get-count func.  This is just some minor display improvements now.  Would you mind giving it a once-over, please?

src/cmd/purge/scripts/exchangePurge.ps1
```diff
@@ -536,6 +543,12 @@ function Purge-Contacts {
         
         }
     }
+
+    # Create one seed contact which will have recent create date and will not be swept
+    # This is needed since tests rely on some contact data being present
+
+    Write-Host "`nCreating seed contact" 
+    Create-Contact
```
gmatev, line 551: @ryanfkeepers the previous location of this was actually safer since the new contact was in place before the contacts were purged. Prevents a case where a test that relies on at least 1 contact being present executes in the timing window where all contacts were purged and the new one is not there.

gmatev: I can fix this for you and will approve it

gmatev (approved):

src/cmd/purge/scripts/exchangePurge.ps1
gmatev: Done

ryanfkeepers: much appreciated

Issue #2976, opened 2023-03-29
Optimize permission diff computation in OneDrive


meain:
Currently we fetch the latest set of permissions from Graph inorder to get the current state of permissions for an item. We can avoid this by computing this data with the information that we have with us locally. This could save us one request(5 credits) per permissions restore.

https://github.com/alcionai/corso/blob/91e0a2783104b7f9ed49d2a5b88fb8d790432a13/src/internal/connector/onedrive/permission.go#L188

PR #3022, opened 2023-04-04
Compute parent permissions from available metadata

meain:
Previously we were querying Graph to get the current permissions and then using that to compute the permission diff. This changes that to use the metadata that we have available locally to do the computation avoiding a graph call to get permissions.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/2976

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/permission.go
+++ b/src/internal/connector/onedrive/permission.go
@@ -131,10 +131,7 @@ func isSame(first, second []string) bool {
 	return slices.Equal(first, second)
 }
 
-func diffPermissions(
-	before, after []UserPermission,
-	permissionIDMappings map[string]string,
-) ([]UserPermission, []UserPermission) {
+func diffPermissions(before, after []UserPermission) ([]UserPermission, []UserPermission) {
 	var (
 		added   = []UserPermission{}
 		removed = []UserPermission{}
@@ -168,7 +165,6 @@ func diffPermissions(
 		}
 
 		if !found {
-			pp.ID = permissionIDMappings[pp.ID]
 			removed = append(removed, pp)
 		}
 	}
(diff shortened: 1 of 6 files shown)


src/internal/connector/onedrive/permission.go
```diff
@@ -167,30 +177,54 @@ func diffPermissions(
...
+		onedrivePath, err := path.ToOneDrivePath(parent)
+		if err != nil {
+			return Metadata{}, clues.New("get parent path")
+		}
+
+		if len(onedrivePath.Folders) == 0 {
+			return Metadata{}, nil
+		}
+
+		meta, ok = folderMetas[parent.String()]
```
ashmrtn, line 206: judging by this we're assuming we've already restored the permissions on all ancestor folders correct? Can you add a comment on this function to say that please?

src/internal/connector/onedrive/permission.go
```diff
@@ -253,11 +295,74 @@ func RestorePermissions(
...
+			permissionIDMappings[p.ID] = ptr.Val(np.GetValue()[0].GetId())
+		}
 	}
 
 	return nil
 }
+
+// RestorePermissions takes in the permissions that were added and the
+// removed(ones present in parent but not in child) and adds/removes
+// the necessary permissions on onedrive objects.
```
ashmrtn, line 313: Comment could use revision as now we're just taking in the item metadata and computing the added/removed permission sets

src/internal/connector/onedrive/permission.go
```diff
@@ -253,11 +295,74 @@ func RestorePermissions(
...
+	}
+
+	permAdded, permRemoved := diffPermissions(parentPermissions.Permissions, meta.Permissions)
+
+	return updatePermissions(ctx, creds, service, driveID, itemID, permAdded, permRemoved, true, permissionIDMappings)
+}
+
+// SetPermissions is similar to RestorePermissions, but fetches the
+// current permissions from graph inorder to update the permissions on
+// an item. This is necessary in tests(any currently only used in
```
ashmrtn, line 343:
```suggestion
// an item. This is necessary in tests (and currently only used in
```

src/internal/connector/onedrive/permission.go
```diff
@@ -167,30 +177,54 @@ func diffPermissions(
...
+
+func updatePermissions(
 	ctx context.Context,
 	creds account.M365Config,
 	service graph.Servicer,
 	driveID string,
 	itemID string,
-	meta Metadata,
+	permAdded, permRemoved []UserPermission,
+	usePermissionIDMappings bool,
```
ashmrtn, line 224: add a comment to the function for why we take in a bool flag (appears it's so we can reuse this code for tests?)

src/internal/connector/onedrive/permission_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	for _, test := range table {
+		suite.Run(test.name, func() {
+			_, flush := tester.NewContext()
+			defer flush()
+
+			t := suite.T()
+
+			m, err := computeParentPermissions(pth, test.parentPerms)
```
ashmrtn, line 119: I don't remember off the top of my head, can `computeParentPermissions` be called on what was the root directory in the backup? If so, we should make a test case for that

ashmrtn (approved):

src/internal/connector/onedrive/permission.go
meain: I've removed it altogether, it now just computes the mapping from the set to be removed and passes that into the function.

src/internal/connector/onedrive/permission_test.go
meain: I won't be called on the root dir. The first dir it will be called will be a dir in the root dir and not the root dir itself. Let me add a test for that though.

meain:
The test is catching some issue with the permissions backup. Let me look into what is going on.

PR #3024, opened 2023-04-04
No warn logs

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :muscle: Manual


src/.golangci.yml
```diff
@@ -38,10 +38,12 @@ linters-settings:
       # Don't allow use of testify suite directly. Use one of the wrappers from
       # tester/suite.go instead. Use an ignore lower down to exclude packages
       # that result in import cycles if they try to use the wrapper.
-      - 'suite\.Suite(# tests should use one of the Suite wrappers in tester package )?'
+      - 'suite\.Suite(# tests should use one of the Suite wrappers in tester package)?'
       # All errors should be constructed and wrapped with the clues package.
       # String formatting should be avoided in favor of structured errors (ie: err.With(k, v)).
       - '(errors|fmt)\.(New|Stack|Wrap|Error)f?\((# error handling should use clues pkg)?'
+      # Avoid Warn-level logging in favor of Info or Error.
```
pandeyabs, line 45: Is this a corso wide design decision? May I know why we are avoiding warn level logging? :)

ryanfkeepers: It is, you can find the design reference [here in the code](https://github.com/alcionai/corso/blob/main/src/pkg/logger/example_logger_test.go#L88).

Long and short of it: the "maybe badness" of warn level introduces unnecessary confusion and decision making.  Logging only needs three empirical categories: 1/ things we don't always need to know (debug).  2/ things we always need to know (info).  3/ things we need to be immediately aware of or alert on. (error).  90% of warns are category 2, 10% are category 3, it's more effective to put them there.

pandeyabs: Thanks @ryanfkeepers. The context helps! The design choice makes sense to me. We could eliminate the capability of adding warn logs altogether, just to make it airtight.

pandeyabs: fair point

pandeyabs:
I'm unable to approve this PR. @meain @neha-Gupta1  could you do it? :)

Issue #2987, opened 2023-03-30
Unable to build corso on Windows


meain:
In the [PR to add metrics](https://github.com/alcionai/corso/pull/2884), we seem to have introduced [`syscall.Kill`](https://github.com/alcionai/corso/blame/03edc603da3e6255405bebb4392b45d1018d28f0/src/internal/events/events.go#L238) which is not available on Windows. As of now, a temporary fix to avoid use of syscall on Windows is done via https://github.com/alcionai/corso/pull/2986 to unblock CI, but we have to properly fix this before next release.

An example failing run can be found [here](https://github.com/alcionai/corso/actions/runs/4560534347/jobs/8045919974#step:1:29).

pandeyabs:
This is mitigated, although metrics on Windows doesn't work currently. I'll open a new issue for that.

PR #3035, opened 2023-04-05
[Draft] [Do not push] Use DisplayMetrics instead of dumping metrics using syscalls ( #2987) 

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* [#<issue>](https://github.com/alcionai/corso/issues/2987)

#### Test Plan

- [ ] :muscle: Manual
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/events/events_signal_windows.go
+++ b/src/internal/events/events_signal_windows.go
@@ -2,10 +2,13 @@ package events
 
 import (
 	"context"
-
 	"github.com/alcionai/corso/src/pkg/logger"
+	"golang.org/x/sys/windows"
+	"signal"
+	"syscall"
 )
 
 func signalDump(ctx context.Context) {
-	logger.Ctx(ctx).Warn("cannot send signal on Windows")
+	//logger.Ctx(ctx).Warn("cannot send signal on Windows")
+	windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, syscall.Getpid())
 }
--- a/src/internal/events/events_signal_windows.go
+++ b/src/internal/events/events_signal_windows.go
@@ -8,6 +8,8 @@ import (
 )
 
 func signalDump(ctx context.Context) {
-	//logger.Ctx(ctx).Warn("cannot send signal on Windows")
-	windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, uint32(syscall.Getpid()))
+	err := windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, uint32(syscall.Getpid()))
+	if err != nil {
+		logger.CtxErr(ctx, err).Error("metrics interval signal")
+	}
 }
--- a/src/internal/events/events_signal_windows.go
+++ b/src/internal/events/events_signal_windows.go
@@ -2,13 +2,12 @@ package events
 
 import (
 	"context"
-	"github.com/alcionai/corso/src/pkg/logger"
-	"golang.org/x/sys/windows"
-	"signal"
 	"syscall"
+
+	"golang.org/x/sys/windows"
 )
 
 func signalDump(ctx context.Context) {
 	//logger.Ctx(ctx).Warn("cannot send signal on Windows")
-	windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, syscall.Getpid())
+	windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, uint32(syscall.Getpid()))
 }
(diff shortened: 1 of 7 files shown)


src/internal/events/events_signal_windows.go
```diff
@@ -2,10 +2,14 @@ package events
...
 	"context"
+	"syscall"
 
-	"github.com/alcionai/corso/src/pkg/logger"
+	"golang.org/x/sys/windows"
 )
 
 func signalDump(ctx context.Context) {
-	logger.Ctx(ctx).Warn("cannot send signal on Windows")
+	err := windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, uint32(syscall.Getpid()))
```
pandeyabs, line 11: Here are all the options i explored to add ctrl-break support for windows:
1. [signal.Notify()](https://pkg.go.dev/os/signal#pkg-types) -- Doesn't work for us, since we don't have access to InmemSignal.sigCh( private). 
2. [process.signal() ](https://github.com/golang/go/commit/f86f9a3038eb6db513a0ea36bc2af7a13b005e99) -- Unfortunately, the authors reverted this commit as it was breaking their tools. It's also based off GenerateConsoleCtrlEvent. 
3. x/sys [windows.GenerateConsoleCtrlEvent()](https://cs.opensource.google/go/x/sys/+/refs/tags/v0.7.0:windows/zsyscall_windows.go;l=2000) -- This compiles fine on Windows. I need some help testing the metrics PoC with this change on both windows/unix. I'll check with you tomorrow @ryanfkeepers

pandeyabs: @ryanfkeepers I will add unit tests to emit SIGUSR1/SIGBREAK signals using the sys.unix/windows APIs and test whether we can receive these signals on a dummy channel. 

Any other thoughts?

ryanfkeepers: That sounds reasonable.  Following deployment I think we should do at least one manual test on each container (a single backup with a log file review is sufficient) to ensure baseline functionality looks as expected.  But that's a follow-up outside of this PR.

pandeyabs: Could you tell me how to emit metrics during manual tests? I couldn't get it to work.

meain: >  Following deployment I think we should do at least one manual test on each container (a single backup with a log file review is sufficient) to ensure baseline functionality looks as expected. But that's a follow-up outside of this PR.

This can probably go in the [sanity-tests](https://github.com/alcionai/corso/blob/a2e669a3a40cef12da3eb1938f459e5a97b867a7/.github/workflows/sanity-test.yaml).

meain: > Could you tell me how to emit metrics during manual tests? 

You shouldn't have to do anything, the log files would contain the metrics lines.

meain: Also, does the unit test to ensure we are able to emit a signal useful here? We will just be testing go internal IMO. 🤔

pandeyabs: Some updates after doing unit + manual testing on windows:

1. We are able to dump metrics on windows now but with an unwanted side effect. While corso is running, user cannot use ^C to terminate it, because we are intercepting those ^C events for dumping metrics. This may be an unpleasant user experience.
2. **Some details:** Go os/signal package treats both ^C and ^Break events on Windows as identical. These events are piped to os.Interrupt(syscall.SIGINT). If Notify is called for os.Interrupt, ^C or ^BREAK will cause os.Interrupt to be sent on the channel, and the program will not exit. https://pkg.go.dev/os/signal#hdr-Windows
3. I think the way go-metrics handles signaling for Windows should be rearchitected. Piggy backing on ^BREAK signal is not a clean approach and has unintended side effects as we see in our case. Should I open an issue against them @ryanfkeepers  @meain ? What do you think?
4. Also their code is broken for Windows. [DefaultSignal ](https://github.com/hashicorp/go-metrics/blob/master/const_windows.go)should be syscall.SIGINT for Windows.

ryanfkeepers: > Should I open an issue against them

@pandeyabs  Thanks for the overview.  Opening a PR to change/fix things in go-metrics seems reasonable.  I wonder if that package is intentionally using Break, on the belief that users want to dump stats on a signalled exit of their app. 

This also makes me curious if our implementation of signalling for log emissions is an abuse of their design.  I personally didn't find other ways to intermittently emit metrics with their API, but it's possible I missed something.  If you can get it working without an OS signal in the first place, we won't need any fixes in go-metrics.

pandeyabs: @ryanfkeepers thanks! I didn't consider that it's an artifact of our design. I'll explore another way to do this without using OS signals.

Issue #2486, opened 2023-02-13
Propagate location path for all data, migrate folder details to location path.


ryanfkeepers:
## Proposal

The introduction of the locationRef property in details entries allows Corso to track external data location in an user consumption-oriented format rather than a system storage-oriented format.  All data types, independent of storage path, should populate a locationRef during backup.

Once we have complete production of locationRef details, we can pivot the details folder entries from being based on storage path to being based on location path.  Since details folders are intended to represent the external item location anyway, this is the correct format for their structure.

## Effort, Value, Impact

Effort: moderate - lots of code to touch, but changes should be small and simple.
Value: moderate - ensures we always produce a folder display hierarchy that matches user expectations, not Corso design.
Impact: minimal - little or no changes to customer state or behavior.

PR #3054, opened 2023-04-06
Add basic prefix matcher type

ashmrtn:
Adds a basic generic prefix matcher type. The
first implementation just uses a map and iterates
through things

This is expected to be used for at least the
exclude list for backups (inner type of
`map[string]struct{}`) and logic for merging
backup details (inner type still TBD a bit,
but either `*path.Builder` or a custom struct)

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2486

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/common/prefix_matcher.go
+++ b/src/internal/common/prefix_matcher.go
@@ -1,0 +1,75 @@
+package common
+
+import (
+	"strings"
+
+	"github.com/alcionai/clues"
+)
+
+type PrefixMatcher[T any] interface {
+	Add(key string, value T) error
+	Get(key string) (T, bool)
+	LongestPrefix(key string) (string, T, bool)
+	Map() map[string]T
+	Empty() bool
+}
+
+type prefixMatcher[T any] struct {
+	data map[string]T
+}
+
+func (m *prefixMatcher[T]) Add(key string, value T) error {
+	if _, ok := m.data[key]; ok {
+		return clues.New("entry already exists").With("key", key)
+	}
+
+	m.data[key] = value
+
+	return nil
+}
+
+func (m *prefixMatcher[T]) Get(key string) (T, bool) {
+	if m == nil {
+		return *new(T), false
+	}
+
+	res, ok := m.data[key]
+	return res, ok
+}
+
+func (m *prefixMatcher[T]) LongestPrefix(key string) (string, T, bool) {
+	if m == nil {
+		return "", *new(T), false
+	}
+
+	var (
+		rk    string
+		rv    T
+		found bool
+		// Set to -1 so if there's "" as a prefix ("all match") we still select it.
+		longest = -1
+	)
+
+	for k, v := range m.data {
+		if strings.HasPrefix(key, k) && len(k) > longest {
+			found = true
+			longest = len(k)
+			rk = k
+			rv = v
+		}
+	}
+
+	return rk, rv, found
+}
+
+func (m prefixMatcher[T]) Map() map[string]T {
+	return m.data
+}
+
+func (m prefixMatcher[T]) Empty() bool {
+	return len(m.data) == 0
+}
+
+func NewPrefixMatcher[T any]() PrefixMatcher[T] {
+	return &prefixMatcher[T]{data: map[string]T{}}
+}
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -253,11 +253,11 @@ func runBackups(
 // genericDeleteCommand is a helper function that all services can use
 // for the removal of an entry from the repository
 func genericDeleteCommand(cmd *cobra.Command, bID, designation string, args []string) error {
-	ctx := clues.Add(cmd.Context(), "delete_backup_id", bID)
-
 	if utils.HasNoFlagsAndShownHelp(cmd) {
 		return nil
 	}
+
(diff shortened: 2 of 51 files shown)


src/internal/common/prefix_matcher.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type prefixMatcher[T any] struct {
+	data map[string]T
+}
+
+func (m *prefixMatcher[T]) Add(key string, value T) error {
+	if _, ok := m.data[key]; ok {
+		return clues.New("entry already exists").With("key", key)
+	}
```
ryanfkeepers, line 24: This feels overbearing.  It's not standard for a map implementations to error on collision.  Is an error necessary?  It seems acceptable to allow users to overwrite keys.

src/internal/common/prefix_matcher.go
```diff
@@ -0,0 +1,75 @@
...
+	return rk, rv, found
+}
+
+func (m prefixMatcher[T]) Map() map[string]T {
+	return m.data
+}
+
+func (m prefixMatcher[T]) Empty() bool {
+	return len(m.data) == 0
+}
```
ryanfkeepers, line 71: In interests of a smaller interface I think this could be removed, since it's also achieved by consumers calling:
```go
len(m.Map())
```

src/internal/common/prefix_matcher.go
```diff
@@ -0,0 +1,75 @@
...
+	Add(key string, value T) error
+	Get(key string) (T, bool)
+	LongestPrefix(key string) (string, T, bool)
+	Map() map[string]T
+	Empty() bool
+}
+
+type prefixMatcher[T any] struct {
+	data map[string]T
+}
```
ryanfkeepers, line 19: To challenge the design a bit: is it necessary to embed a map?  Seems to me the thing you really want here is a `LongestPrefixer`, and that the rest of this design is trying to recreate what a map already provides.
```suggestion
type LongestPrefixer[T any] interface {
	LongestPrefix(key string) (string, T, bool)
	// maybe get, as well
}

type prefixMatcher[T any] data map[string]T
```

Now, will we always require a map?  Of course not.  But I imagine that the usage pattern is for some producer to construct the underlying struct, and have all consumers pass around the above interface as a read-only interface.  Imo, that makes Add, Map, and Empty dubious requirements.

src/internal/common/prefix_matcher.go
```diff
@@ -0,0 +1,75 @@
+package common
```
ryanfkeepers, line 1: I feel nitpicky about "common" these days.  Thoughts on either of these?
1. adding this to Filters
2. putting it in a subpackage: common/prefix_matcher

ryanfkeepers (approved):
I'm picky about the design and size of the interface, but not enough to block anything, so you're approved and can refine as wanted.

src/internal/common/prefix_matcher.go
ryanfkeepers: Is empty a necessary call in the first place?  It strikes me as premature optimization.  How much performance do we save over letting the consumer call no-op Gets?

src/internal/common/prefix_matcher.go
ryanfkeepers: Ah, I think you got caught on the data structure part of my suggestion.  Let's ignore what data structure is used for backing and focus on the interface itself.

What I assume you're building is an object that facilitates a write-only (or read/write) producer, and a read-only consumer.  Ie: you don't want consumers calling `Add` in the first place (see your comment: "I don't want other code modifying the kv-pairs of the map").

Working within this pattern means your producer, most likely, has local access to the backing structure itself, and needs no interface to mutate it.  Could that backing structure still have an Add method?  Sure!  But it wouldn't make sense to expose that to consumers, much less expect that behavior to be necessary for a read-only environment.

Get() and LongestPrefix() seems like the core necessities for this object to be consumed, while the rest of the funcs are pushing local concerns up into the interface even though consumers won't need them.

ashmrtn: >Get() and LongestPrefix() seems like the core necessities for this object to be consumed, while the rest of the funcs are pushing local concerns up into the interface even though consumers won't need them.

Unfortunately there's no way to have common generic implementations of things that adhere to interfaces without exposing some sort of `Add` or `Put` function. Otherwise you'd need to reimplement the interface in every package that wants to use it so you could hide those functions. Or you could split the interface into one that has read/write functions and another that has only read functions but that seems like overkill

src/internal/common/prefix_matcher.go
ashmrtn: depends on what it costs to get the data to do the get. If getting the data requires reading backup details from kopia and then going through every item in it to check then it may be worth being able to ask if the matcher is empty. If it's just checking data that's already in-memory then probably doesn't matter as much

src/internal/common/prefix_matcher.go
ryanfkeepers: > Or you could split the interface into one that has read/write functions and another that has only read functions but that seems like overkill

Not overkill at all.  If production/consumption is write/read bound, then I'd say the split is _more_ idiomatic than a single, universal interface.

PR #3057, opened 2023-04-06
optionally look up and union fallback manifests

ryanfkeepers:
Fallback manifests allow us to find manifests and
metadata from snapshots that utilized earlier
versions of kopia Reasons, primarily across
resource owner deltas.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test


ryanfkeepers:
Manual testing involved:
1. On a prior commit: OneDrive backup with userPN.
2. On this branch: Incremental backup with userID matching PN.

Backup was successful.  Incremental history was carried over, no full backups occurred.  All details were updated to a repoRef containing the userID instead of the PN.

src/internal/operations/manifests.go
```diff
@@ -43,24 +43,38 @@ func produceManifestsAndMetadata(
...
-		ctx,
-		reasons,
-		map[string]string{kopia.TagBackupCategory: ""})
+	ms, err := mr.FetchPrevSnapshotManifests(ctx, reasons, tags)
 	if err != nil {
-		return nil, nil, false, err
+		return nil, nil, false, clues.Wrap(err, "looking up prior snapshots")
+	}
+
+	if len(fallbackReasons) > 0 {
```
ashmrtn, line 63: do we need a len check? Ideally the code to find previous manifests should return nothing if no reasons are given

src/internal/operations/manifests.go
```diff
@@ -137,6 +151,96 @@ func produceManifestsAndMetadata(
...
+		k := r.Service.String() + r.Category.String()
+		tups[k] = manTup{}
+	}
+
+	// track the manifests that were collected with the current lookup
+	for _, m := range mans {
+		for _, r := range m.Reasons {
+			k := r.Service.String() + r.Category.String()
+			t := tups[k]
+			// assume mans will have, at most, one complete and one incomplete per key
```
ashmrtn, line 194: this assumption should be added to the comment for the function as well

src/internal/operations/manifests.go
```diff
@@ -137,6 +151,96 @@ func produceManifestsAndMetadata(
...
+
+	// backfill from the fallback where necessary
+	for _, m := range fallback {
+		for _, r := range m.Reasons {
+			k := r.Service.String() + r.Category.String()
+			t := tups[k]
+
+			if t.complete != nil {
+				// assume fallbacks contains prior manifest versions.
+				// we dont want to stack a prior version incomplete onto
```
ashmrtn, line 213:
```suggestion
				// we don't want to stack a prior version incomplete snapshot onto
```

ashmrtn (approved):

src/internal/operations/manifests.go
ryanfkeepers: guess not!

PR #3064, opened 2023-04-07
Sanitize recurrenceTimezone prior to restoring calendar event

pandeyabs:
* Do not specify recurrenceTimeZone if it's not set in the item being restored. "" is not a valid value
---

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)
* COR-53

#### Test Plan

- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -117,7 +117,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreEvent() {
 	}()
 
 	info, err := RestoreExchangeEvent(ctx,
-		mockconnector.GetMockEventWithSubjectBytes(name),
+		mockconnector.GetMockEventWithAttendeesBytes(name),
 		suite.gs,
 		control.Copy,
 		calendarID,
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -125,6 +125,41 @@ func (suite *ExchangeRestoreSuite) TestRestoreEvent() {
 		fault.New(true))
 	assert.NoError(t, err, clues.ToCore(err))
 	assert.NotNil(t, info, "event item info")
+
+	// Recurrence testing
+	// 1. Event doesn't have recurrence timezone specified
+	info, err = RestoreExchangeEvent(ctx,
+		mockconnector.GetMockEventWithRecurrenceBytes(name, false, ""),
+		suite.gs,
+		control.Copy,
+		calendarID,
+		userID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
+
+	// 2. Event has recurrence timezone specified as ""
+	info, err = RestoreExchangeEvent(ctx,
+		mockconnector.GetMockEventWithRecurrenceBytes(name, true, ""),
+		suite.gs,
+		control.Copy,
+		calendarID,
+		userID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
+
+	// 3. Valid recurrence timezone field i.e. it is not the same timezone as
+	// start or end timezones
+	info, err = RestoreExchangeEvent(ctx,
+		mockconnector.GetMockEventWithRecurrenceBytes(name, true, "Pacific Standard Time"),
+		suite.gs,
+		control.Copy,
+		calendarID,
+		userID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
 }
 
 type containerDeleter interface {
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -117,7 +117,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreEvent() {
 	}()
 
 	info, err := RestoreExchangeEvent(ctx,
-		mockconnector.GetMockEventWithAttendeesBytes(name),
+		mockconnector.GetMockEventWithSubjectBytes(name),
 		suite.gs,
 		control.Copy,
 		calendarID,
--- a/src/internal/connector/mockconnector/mock_data_event.go
+++ b/src/internal/connector/mockconnector/mock_data_event.go
@@ -58,15 +58,15 @@ const (
 				"responseStatus": {
 					"response": "organizer",
 					"time": "0001-01-01T00:00:00Z"
-				},
+		},
 				"body": {
 					"contentType": "html",
 					"content": "<html><head><meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\">\r\n<meta name=\"Generator\" content=\"Microsoft Exchange Server\">\r\n<!-- converted from text -->\r\n<style><!-- .EmailQuote { margin-left: 1pt; padding-left: 4pt; border-left: #800000 2px solid; } --></style></head>\r\n<body>\r\n<font size=\"2\"><span style=\"font-size:11pt;\"><div class=\"PlainText\">&nbsp;</div></span></font>\r\n</body>\r\n</html>\r\n"
-				},
+		},
 				"start": {
 					"dateTime": "2000-01-01T00:00:00.0000000",
 					"timeZone": "UTC"
-				},
+		},
 				"end": {
 					"dateTime": "2000-01-02T00:00:00.0000000",
 					"timeZone": "UTC"
@@ -93,13 +93,13 @@ const (
 						"startDate": "2000-01-01",
 						"endDate": "0001-01-01",
 						"numberOfOccurrences": 0
-					}
-				},
+			}
+		},
 				"attendees": [{
 					"emailAddress": {
 						"address":"george.martinez@8qzvrj.onmicrosoft.com",
 						"name": "George Martinez"
-					},
+		},
 					"type": "required",
 					"status": {
 						"response": "none",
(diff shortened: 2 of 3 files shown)


src/internal/connector/support/m365Transform.go
```diff
@@ -88,6 +88,14 @@ func ToEventSimplified(orig models.Eventable) models.Eventable {
 	orig.SetICalUId(nil)
 	orig.SetId(nil)
 
+	// Sanitize recurrence timezone.
+	if orig.GetRecurrence() != nil {
+		recurrenceTimezone := orig.GetRecurrence().GetRange().GetRecurrenceTimeZone()
+		if recurrenceTimezone != nil && len(*recurrenceTimezone) == 0 {
+			orig.GetRecurrence().GetRange().SetRecurrenceTimeZone(nil)
+		}
```
ryanfkeepers, line 96: We now have helper funcs for pointer values like this:
```suggestion
		recurrenceTimezone := ptr.Val(orig.GetRecurrence().GetRange().GetRecurrenceTimeZone())
		if len(recurrenceTimezone) == 0 {
			orig.GetRecurrence().GetRange().SetRecurrenceTimeZone(nil)
		}
```

src/internal/connector/exchange/restore_test.go
```diff
@@ -125,6 +125,41 @@ func (suite *ExchangeRestoreSuite) TestRestoreEvent() {
 		fault.New(true))
 	assert.NoError(t, err, clues.ToCore(err))
 	assert.NotNil(t, info, "event item info")
+
```
ryanfkeepers, line 128: This test is a bit out of date.  While you're extending it, would you mind refactoring its implementation to a [table driven test](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)?  You should be able to copy boilerplate from much of the surrounding tests, if needed.

src/internal/connector/mockconnector/mock_data_event.go
```diff
@@ -266,3 +266,42 @@ func GetMockEventWithAttendeesBytes(subject string) []byte {
...
+func GetMockEventWithRecurrenceBytes(
+	subject string, hasRecurrenceTimeZone bool,
+	recurrenceTimeZone string,
+) []byte {
+	newTime := time.Now().AddDate(0, 0, 1)
+	conversion := common.FormatTime(newTime)
+	timeSlice := strings.Split(conversion, "T")
+
+	//nolint:lll
+	event := "{\"id\":\"AAMkAGZmNjNlYjI3LWJlZWYtNGI4Mi04YjMyLTIxYThkNGQ4NmY1MwBGAAAAAADCNgjhM9QmQYWNcI7hCpPrBwDSEBNbUIB9RL6ePDeF3FIYAAAAAAENAADSEBNbUIB9RL6ePDeF3FIYAABU_FdvAAA=\",\"@odata.etag\":\"W/\\\"0hATW1CAfUS+njw3hdxSGAAAVK7j9A==\\\"\"," +
```
ryanfkeepers, line 279: instead of adding another text blob, would it be possible to update the `eventTmpl` var to fit the needs of this test?  That'll keep up readability and reduce maintenance of these values in the long run.

src/internal/connector/exchange/restore_test.go
```diff
@@ -125,6 +125,41 @@ func (suite *ExchangeRestoreSuite) TestRestoreEvent() {
...
+	// 1. Event doesn't have recurrence timezone specified
+	info, err = RestoreExchangeEvent(ctx,
+		mockconnector.GetMockEventWithRecurrenceBytes(name, false, ""),
+		suite.gs,
+		control.Copy,
+		calendarID,
+		userID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
```
ryanfkeepers, line 139: Having a non-nil item isn't very pertinent to this test case.  Could you add an assertion for the recurrence timezone, to ensure the restored object matches your new conditional?

ryanfkeepers (approved):
Generally looks functional, so approving to unblock.  Comments are all nits about code quality/cleanliness.  Would prefer to see them fixed before merge, but it's nothing that would prevent moving forward.

src/internal/connector/exchange/restore_test.go
pandeyabs: Sounds good!

src/internal/connector/exchange/restore_test.go
pandeyabs: Good catch. Will do

src/internal/connector/mockconnector/mock_data_event.go
pandeyabs: My original thought was to update `eventTmpl` also. However, `recurrence` field is optional and may be nil. If we sprintf recurrence into `eventTmpl`, then we will lose the capability to test against nil `recurrence`. I think `Attendees` is along the same lines. 

Any thoughts on how to support nil references in the template? One option I was exploring was using [msgraph-sdk ](https://learn.microsoft.com/en-us/graph/api/calendar-post-events?view=graph-rest-1.0&tabs=go)to create test events instead of json payloads. Not sure if it defeats the purpose of having mock data.

ryanfkeepers: > Any thoughts on how to support nil references in the template? 

Heh, we're getting close to needing the text/template engine here.  But in the short term: a %s formatter that defaults to `"field": nil,` or even an empty string should give us what we need.  Most funcs can rely on the default, while the func you're implementing should populate both the key and value.  Does that sound reasonable?

> Not sure if it defeats the purpose of having mock data.

I don't think it would defeat the purpose entirely.  But a corner-case we rely on with a standard template is a additional awareness of when a change in microsoft libraries or server handling deviates from our expected behavior.  If we always create an event using the latest client state, rather than restoring a template of bytes (as if restoring the old data we previously persisted), it might handle those changes silently, which could turn into runtime gotchas after release.  It's an imperfect pattern, but one that's suited us this far.

pandeyabs: Thanks! 

> But in the short term: a %s formatter that defaults to "field": nil, or even an empty string should give us what we need

I tried with empty string. But it doesn't have the same effect as `"field":nil`. For e.g. instead of producing `"recurrence": null`, with an empty string, I get `"recurrence": ""`, which is not null. 

Am i doing it wrong? Could you tell me how to do this - "%s formatter that defaults to "field": nil"

ryanfkeepers: I'm thinking something like this: https://go.dev/play/p/3DlALI6PI1a

pandeyabs: Thank you so much!

src/internal/connector/exchange/restore_test.go
pandeyabs: We won't be able to check recurrence timezone details in the integration test ( we only have access to details.ExchangeInfo).
I'll add a UT for ToEventSimplified for this purpose.

src/internal/connector/mockconnector/mock_data_event.go
pandeyabs: @ryanfkeepers Thanks for all the suggestions! This PR is now ready for another review

src/cmd/factory/impl/exchange.go
```diff
@@ -115,7 +115,7 @@ func handleExchangeCalendarEventFactory(cmd *cobra.Command, args []string) error
 		func(id, now, subject, body string) []byte {
 			return mockconnector.GetMockEventWith(
 				User, subject, body, body,
-				now, now, false)
+				now, now, `null`, `[]`, false)
```
ryanfkeepers, line 118: not important here, but fyi, the grave-tics in go are primarily useful for constructing escaped and multiline strings.  Generally you should use `"` unless otherwise needed.  Example here: https://go.dev/play/p/U8VjEN2bSo7

src/internal/connector/support/m365Transform_test.go
```diff
@@ -37,7 +37,7 @@ func (suite *SupportTestSuite) TestToMessage() {
 	assert.NotEqual(t, message.GetId(), clone.GetId())
 }
 
-func (suite *SupportTestSuite) TestToEventSimplified() {
+func (suite *SupportTestSuite) TestToEventSimplifiedAttendees() {
```
ryanfkeepers, line 40: Fyi, the Golang test naming convention we want to follow is `Test<FuncOrStructName>_<MethodNameIfExists>_<conditionIfExists>()`.

So in this case, a more appropriate name would be:
```suggestion
func (suite *SupportTestSuite) TestToEventSimplified_attendees() {
```

src/internal/connector/mockconnector/mock_data_event.go
```diff
@@ -147,6 +150,48 @@ const (
 		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" +
 		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" +
 		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\"}],"
+
+	noRecurrence   = `null`
```
ryanfkeepers, line 154: You might want to export this and `noAttendees` so that callers can use the const vals instead of passing them in as plain strings.

ryanfkeepers, line 154: You might want to export this and `noAttendees` so that callers can use the const vals instead of passing them in as plain strings.

ryanfkeepers (commented):
some nits, but looks good overall; approving to unblock.

ryanfkeepers (approved):
some nits, but looks good overall; approving to unblock.

PR #3066, opened 2023-04-07
centralized sdk err is checker

ryanfkeepers:
Adds a package pkg/err which houses all internal-
to-external error mappings.  This helps centralize
error.Is checks for sdk consumers, without corso
distributing checkable error wrappers across
multiple pkg/** locations.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/err/err.go
+++ b/src/pkg/err/err.go
@@ -1,0 +1,45 @@
+package err
+
+import (
+	"errors"
+
+	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/pkg/repository"
+)
+
+// expose enums, rather than errors, for Is checks.  The enum should
+// map to a specific internal error that can be used for the actual
+// errors.Is comparison.
+type errEnum string
+
+const (
+	RepoAlreadyExists errEnum = "repository-already-exists"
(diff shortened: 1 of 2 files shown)


src/pkg/err/err.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	RepoAlreadyExists errEnum = "repository-already-exists"
+	BackupNotFound    errEnum = "backup-not-found"
+	ServiceNotEnabled errEnum = "service-not-enabled"
+)
+
+// map of enums to errors.  We might want to re-use an enum for multiple
+// internal errors (ex: "ServiceNotEnabled" may exist in both graph and
+// non-graph producers).
+var internalToExternal = map[errEnum][]error{
```
ashmrtn, line 24: we're going to need some process for corso devs to make sure this stays up to date

ashmrtn (approved):
LGTM, though may want to check with others if the API is good as changes to SDK APIs has been a recent pain point for folks

src/pkg/err/err.go
ryanfkeepers: Lets bring it up as a topic in the monday meeting.

PR #3068, opened 2023-04-07
clean up cli e2e tests

ryanfkeepers:
deduplicate some e2e boilerplate setup.
Also, transitions exchange e2e tests from an in-
test loop over categories, to a well-defined test
func per category.  The former design doesn't
play well with go test --run ./testName, and
causes all tests to run even if they don't match
the string.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E


ryanfkeepers:
@meain @neha-Gupta1 This is ready for review, please and thank you!

src/cli/backup/help_e2e_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/alcionai/corso/src/cli/config"
+	"github.com/alcionai/corso/src/internal/tester"
+	"github.com/alcionai/corso/src/pkg/account"
+	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/repository"
+	"github.com/alcionai/corso/src/pkg/storage"
+)
+
+func prepM365Test(
```
meain, line 20: Good call on this one :D

src/internal/tester/resource_owners.go
```diff
@@ -20,7 +20,7 @@ func M365TenantID(t *testing.T) string {
 	cfg, err := readTestConfig()
 	require.NoError(t, err, "retrieving m365 tenant ID from test configuration", clues.ToCore(err))
 
-	return cfg[TestCfgAzureTenantID]
+	return strings.ToLower(cfg[TestCfgAzureTenantID])
```
meain, line 23: Are we sure we wanna lowercase IDs? 🤔

src/cli/backup/exchange_e2e_test.go
```diff
@@ -140,110 +125,94 @@ func TestBackupExchangeE2ESuite(t *testing.T) {
...
 
-	for _, set := range backupDataSets {
-		recorder.Reset()
+func (suite *BackupExchangeE2ESuite) TestExchangeBackupCmd_userNotFound_events() {
+	runExchangeBackupCategoryTest(suite, "events")
+}
 
-		suite.Run(set.String(), func() {
-			t := suite.T()
+func runExchangeBackupUserNotFoundTest(suite *BackupExchangeE2ESuite, category string) {
```
meain, line 190: This seems to be unused.

src/internal/tester/resource_owners.go
ryanfkeepers: Do we have any reason not to?

src/cli/backup/exchange_e2e_test.go
ryanfkeepers: whoops, good catch, thanks!

src/internal/tester/resource_owners.go
meain: I was just wondering if we should lowercase in the first place and if anything would break just because of that. Any specific reason for lowercasing them?

meain: Since this is only in tests, let me approve it to unblock.

meain (approved):

Issue #3027, opened 2023-04-04
Expose (M365) item ID in backup details


ashmrtn:
SDK consumers have need of a semi-stable, unique item identifier to specify for restores. This will give them the ability to dedupe data across backups in some cases

After investigation, we've decided to expose M365 ID as it is unique and stable under most* conditions. Other things like display name are insufficient because some services, like Exchange events, can have multiple items with the same display name

PR #3070, opened 2023-04-07
add itemref to details entry

ryanfkeepers:
Adds an new details entry field: itemRef.
This holds a stable, semi-unique identifier
to the item represented by that entry.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3027

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/onedrive_test.go
+++ b/src/pkg/selectors/onedrive_test.go
@@ -262,8 +262,9 @@ func (suite *OneDriveSelectorSuite) TestOneDriveCategory_PathValues() {
 	t := suite.T()
 
 	fileName := "file"
+	fileID := fileName + "-id"
 	shortRef := "short"
-	elems := []string{"drive", "driveID", "root:", "dir1", "dir2", fileName + "-id"}
+	elems := []string{"drive", "driveID", "root:", "dir1", "dir2", fileID}
 
 	filePath, err := path.Build("tenant", "user", path.OneDriveService, path.FilesCategory, true, elems...)
 	require.NoError(t, err, clues.ToCore(err))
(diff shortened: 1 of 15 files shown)


src/pkg/selectors/onedrive_test.go
```diff
@@ -276,6 +277,7 @@ func (suite *OneDriveSelectorSuite) TestOneDriveCategory_PathValues() {
 	ent := details.DetailsEntry{
 		RepoRef:  filePath.String(),
 		ShortRef: shortRef,
+		ItemRef:  fileID,
```
ashmrtn, line 280: I think for upstream SDK consumers the behavior of `pathValues` for OneDrive/SharePoint will need to be adjusted. We'll need to re-enable matching on item IDs. We should also put out a "known issues" line that states we'll restore/select items with both the name and item ID if users happen to have a file named the same as an item ID. This should be a known issue because we never display OneDrive or SharePoint item IDs to CLI users

ryanfkeepers: That will be handled in a later PR.  This change is going to act as a feature branch for two further changes:
1. ensuring population of the item id.
2. selection on item id.

In the second one, I'll allow the selector producer to declare whether it wants to select on item names or item IDs.  The CLI will opt for item names.  Haven't decided if I want to default to including both name and id, or just the id.  The former is more lenient, the latter is reduces some collisions.  Probably going to do the latter.

Either way, once that configuration is in place, we won't have the aforementioned duplicate-id-name collision case.

PR #3076, opened 2023-04-10
change user flag

neha-Gupta1:
As part of handling share mailbox-
Introduced new flag - `mailbox` and deprecated existing `user` flag
This is to handle wider group of mailbox types.

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual


src/cli/utils/flags.go
```diff
@@ -121,7 +122,22 @@ func AddUserFlag(cmd *cobra.Command) {
 		&UserFV,
 		UserFN, nil,
 		"Backup a specific user's data; accepts '"+Wildcard+"' to select all users.")
-	cobra.CheckErr(cmd.MarkFlagRequired(UserFN))
+
+	// if we are marking the user flag as deprecated,
```
neha-Gupta1, line 126: @ryanfkeepers please let me know if there is a better way to handle this situation

src/cli/utils/flags.go
```diff
@@ -115,13 +116,27 @@ func AddRunModeFlag(cmd *cobra.Command, persistent bool) {
...
+	// exclusive check
+	cmd.MarkFlagsMutuallyExclusive(MailBoxFN, UserFN)
+	// cobra.CheckErr(cmd.MarkFlagRequired(UserFN))
+
+	cobra.CheckErr(cmd.Flags().MarkDeprecated(UserFN, fmt.Sprintf("user flag is deprecated. Use - %s", MailBoxFN)))
+
+	cmd.Flags().StringSliceVar(
+		&UserFV,
+		MailBoxFN, nil,
+		"Backup a specific user's data; accepts '"+Wildcard+"' to select all users.")
```
ryanfkeepers, line 139:
```suggestion
		"Backup a specific mailbox's data; accepts '"+Wildcard+"' to select all mailboxes.")
```

ryanfkeepers (commented):
Would like to review concerns and manual output.  But conceptually this change looks good.

neha-Gupta1:
Response of commands
- user flag provided- deprecated warning
userFlag

- mailbox flag provided - normal run of backup 
mailboxFlag

- both flag provided - error message specifying the same
bothFlag

- no flag provided- the cli request fails and shows help. But it does not gives a clear message that the mailbox flag is required. 
noFlag

src/cli/utils/flags.go
```diff
@@ -115,13 +116,28 @@ func AddRunModeFlag(cmd *cobra.Command, persistent bool) {
...
-	cmd.Flags().StringSliceVar(
+	flags := cmd.Flags()
+
+	flags.StringSliceVar(
 		&UserFV,
 		UserFN, nil,
 		"Backup a specific user's data; accepts '"+Wildcard+"' to select all users.")
-	cobra.CheckErr(cmd.MarkFlagRequired(UserFN))
+
+	cobra.CheckErr(flags.MarkDeprecated(UserFN, fmt.Sprintf("user flag is deprecated. Use - %s", MailBoxFN)))
```
meain, line 128:
```suggestion
	cobra.CheckErr(flags.MarkDeprecated(UserFN, fmt.Sprintf("use --%s instead", MailBoxFN)))
```

IMO, this would be better. The  will make the whole thing to be `Flag --user has been deprecated, use --mailbox instead`.

neha-Gupta1: updatedMsg

src/cli/backup/onedrive.go
```diff
@@ -26,34 +26,34 @@ import (
 
 const (
 	oneDriveServiceCommand                 = "onedrive"
-	oneDriveServiceCommandCreateUseSuffix  = "--user <email> | '" + utils.Wildcard + "'"
+	oneDriveServiceCommandCreateUseSuffix  = "--mailbox <email> | '" + utils.Wildcard + "'"
```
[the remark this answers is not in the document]
gmatev: @ryanfkeepers is correct. `mailbox` should only apply to `exchange`. 

For the scenario where a licensed user is converted to a shared mailbox but their license is kept in place, I'm ok if for `onedrive` we use `user`.

CHANGELOG.md
```diff
@@ -137,7 +137,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Issue where repository connect progress bar was clobbering backup/restore operation output.
 - Issue where a `backup create exchange` produced one backup record per data type.
-- Specifying multiple users in a onedrive backup (ex: `--user a,b,c`) now properly delimits the input along the commas.
+- Specifying multiple users in a onedrive backup (ex: `--mailbox a,b,c`) now properly delimits the input along the commas.
```
gmatev, line 140: Please keep `user` only for `onedrive`

gmatev:
> Error: if any flags in the group [mailbox user] are set none of the others can be; [mailbox user] were all set

> Pretty sure this is the default message provided by cobra, and it's terrible. Not sure how much leeway we have to change it, but something like one of these would be very much preferred:

@neha-Gupta1 If the message for the built-in Cobra exclusive flags is bad, can we please do the validation manually and output one of the more user friendly versions?

src/cli/utils/flags.go
```diff
@@ -115,13 +116,28 @@ func AddRunModeFlag(cmd *cobra.Command, persistent bool) {
...
-	cmd.Flags().StringSliceVar(
+	flags := cmd.Flags()
+
+	flags.StringSliceVar(
 		&UserFV,
 		UserFN, nil,
 		"Backup a specific user's data; accepts '"+Wildcard+"' to select all users.")
-	cobra.CheckErr(cmd.MarkFlagRequired(UserFN))
+
+	cobra.CheckErr(flags.MarkDeprecated(UserFN, fmt.Sprintf("use - %s instead", MailBoxFN)))
```
gmatev, line 128: What is the behavior here? Would it block the flag or let it through with an information message? We want the latter behavior for now.

src/cli/utils/flags.go
meain: I was thinking we could mimic mailbox as a flag to avoid confusion, ie `--mailbox.

neha-Gupta1:
Updates-
mailbox flag is only applicable for backup of exchange. And handled the user and mailbox flag manually. Sample command outputs-

- no flag- fail with  error message that atleast one required
noFlag

- both flag- fail with error message that only one aloud
bothflag

- user flag - move ahead with a deprecation message
userFlag

@meain have handled the `--mailbox` changes too.

meain:
Unless there was a specific requirement, in the no-flag scenario, we can just mention about the mailbox flag.

ryanfkeepers (approved):

PR #3079, opened 2023-04-10
rename package to avoid varname shadowing

ryanfkeepers:
#### Type of change


ashmrtn (commented):
aren't we going to have shadowing of `errs` too when we pass around the fault bus?

ashmrtn (approved):

PR #3085, opened 2023-04-11
Fix error with backup list --backup ID (backup  created with an older corso version)

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/repository/repository.go
```diff
@@ -515,7 +515,7 @@ func getBackupErrors(
 
 	ssid := b.StreamStoreID
 	if len(ssid) == 0 {
-		return nil, b, clues.New("no errors in backup").WithClues(ctx)
+		return nil, b, nil // Draft. Just to show the origin of the bug
```
ryanfkeepers, line 518: Hmm, a couple concerns to address here:
* I think this is a good change for UX, but only halfway there.  We need to report to the user that the error details we should have been able to access are not available.
* We could compare the version of the backup here to decide whether it's supposed to have a ssid or detailsID.  But rather than dragging out breaking change maintenance, this might be better handled by the first point.
* The ssid, for the appropriate backup version, should be guaranteed to exist.  It's a necessary part of a backup, and without it we're unable to run details or restore commands.  So if it is missing, and it's not a version issue, then some other thing caused the bug and this is only dealing with the symptoms.

pandeyabs: Thanks for the details @ryanfkeepers. This helps. So I think there are 2 problems here. 
1. **Missing fault error ssid for older backup versions**: Since this will cease to exist going forward, I think we can ignore it. What do you think? 
2. **Error: "<nil>" message**: We should have seen "no errors in backup" instead. I've fixed this issue with a newer commit in this PR. It was a small bug in our code.

pandeyabs: > I don't think "no backup" would be the correct message here. It seems like we do have a backup; just that it contains incomplete data.

I was quoting the current message we have in place for missing SSID. I'll improve the message.

pandeyabs: @ryanfkeepers this is ready for a review. thanks!

ryanfkeepers (approved):

PR #3086, opened 2023-04-11
Fetch individual attachments in case of full download failure

meain:
Failure can happen as the list of attachment and they data can get
pretty big.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/api/mail_test.go
```diff
@@ -145,3 +153,202 @@ func (suite *MailAPIUnitSuite) TestMailInfo() {
...
+		attachmentCount int
+		expect          assert.ErrorAssertionFunc
+	}{
+		{
+			name: "simple fetch",
+			setupf: func() {
+				mitem := models.NewMessage()
+				mitem.SetId(&mid)
+
+				gock.New("https://graph.microsoft.com").
```
meain, line 221: I was exploring how to test this and using something like `gock` seemed like a good choice. Let me know if you have any thoughts or other options.

I could think of 3 options:
- Option 1: Use [golang/mock](https://github.com/golang/mock) to create mocks for SDK  
	- ❌ Have to run codegen
	- ✅ Can run tests in parallel
- Option 2: Use [h2non/gock](https://github.com/h2non/gock) to intercept and substitute http requests  
	- ✅ Don't need a codegen  
	- ✅ We don't have to make any assumptions about SDK
	- ❌ Not thread safe and cannot run parallely  
- Inject a middleware to respond with custom http response  
	- ❌ Have to create implementation from scratch  
	- ✅ Don't have to make assumptions about SDK behaviour apart from middlewares  


/cc @ashmrtn @ryanfkeepers

src/internal/connector/exchange/api/mail_test.go
```diff
@@ -145,3 +153,202 @@ func (suite *MailAPIUnitSuite) TestMailInfo() {
...
+	aid := "fake-attachment-id"
+
+	tests := []struct {
+		name            string
+		setupf          func()
+		attachmentCount int
+		expect          assert.ErrorAssertionFunc
+	}{
+		{
+			name: "simple fetch",
```
ashmrtn, line 216: nit: just to be a little more descriptive
```suggestion
			name: "no attachments",
```

src/internal/connector/exchange/api/mail_test.go
ryanfkeepers: While I'm not usually a fan of external mocking packages (largely because they're normally overly heavy and come with their own corner cases), where a lightweight, local interface will do just as well.  But I don't think we're going to do any better in this case by writing anything locally.  Let's keep with Gock here, and we can revisit as needed.

Other notes:

> ❌ Have to run codegen

This is a big no for me.  Having to run codegen for mocking is absolutely something to avoid.

> Not thread safe and cannot run parallely

Only a problem if we register overlapping urls, right?  That can be avoided by standardizing around random uuids/strings as test vars instead of hardcoded ids.

meain: > Only a problem if we register overlapping urls, right?

Yup, AFAICT if we have separate urls, we should good.

> I love seeing tests like this as it gives us so much more control. ...

Yeah, I really like how gock fits it, especially the fact that we don't have to make assumptions about how the SDK works. This also has the advantage of making sure any SDK changes also gets tested. The setup is also relatively minimal for `gock`.

src/internal/connector/exchange/api/mail.go
```diff
@@ -152,10 +152,53 @@ func (c Mail) GetItem(
 			Attachments().
 			Get(ctx, options)
 		if err != nil {
-			return nil, nil, graph.Wrap(ctx, err, "mail attachment download")
-		}
+			// This can be caused by having a lot of attachments as we
+			// are trying to fetch the data within the attachments as well
+			// in the request. We instead fetch all the attachment ids and
+			// fetch each item individually.
```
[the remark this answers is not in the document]
meain: From what I can tell, this is an exception and in most cases(I'm assuming ~90%) there are relatively few attachments and that too small ones. This way we can avoid making multiple calls, reducing throttling any potential slow downs from pulling one by one.

src/internal/connector/exchange/api/mail.go
```diff
@@ -152,10 +152,53 @@ func (c Mail) GetItem(
...
+					QueryParameters: &users.ItemMessagesItemAttachmentsAttachmentItemRequestBuilderGetQueryParameters{
+						Expand: []string{"microsoft.graph.itemattachment/item"},
+					},
+				}
+
+				att, err := c.stable.
+					Client().
+					UsersById(user).
+					MessagesById(itemID).
+					AttachmentsById(*a.GetId()).
```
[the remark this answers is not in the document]
meain: Woops, missed that.

src/internal/connector/exchange/api/mock/mail.go
```diff
@@ -0,0 +1,67 @@
+package mock
```
ryanfkeepers, line 1: These mocks all read like they should belong to `graph/mock/service.go`.  They neither accept nor return exchange/api structures: it's all graph level.  I think it should be moved.

meain: `NewClient` and `NewService` are mockable version of their counterparts in `api` package under exchange. I've however moved the `CreateAdapter` to under `graph/service`.

src/internal/connector/exchange/api/mail.go
```diff
@@ -133,31 +133,81 @@ func (c Mail) GetItem(
...
+		return mail, MailInfo(mail), nil
+	}
+
+	// A failure can be caused by having a lot of attachments as
+	// we are trying to fetch the data within the attachments as
+	// well in the request. We instead fetch all the attachment
+	// ids and fetch each item individually.
+	// NOTE: Maybe filter for specific error:
+	// graph.IsErrTimeout(err) || graph.IsServiceUnavailable(err)
+	// TODO: Once MS Graph fixes pagination for this, we can
```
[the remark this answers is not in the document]
meain: I've opened a new ticket for this.

PR #3090, opened 2023-04-11
separate fault bus in fetch manifests

ryanfkeepers:
The global aggregation of recoverable errors
causes us to mark operations as failed even in
conditions where they should have succeeded,
because higher level callers don't have sufficient
control over ignorable error conditions.  This
is a short-term fix that allows us to log and dump
fault error accumulation in a specific, scoped instance.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/manifests.go
+++ b/src/internal/operations/manifests.go
@@ -46,7 +46,6 @@ func produceManifestsAndMetadata(
 	reasons, fallbackReasons []kopia.Reason,
 	tenantID string,
 	getMetadata bool,
-	errs *fault.Bus,
 ) ([]*kopia.ManifestEntry, []data.RestoreCollection, bool, error) {
 	var (
 		tags          = map[string]string{kopia.TagBackupCategory: ""}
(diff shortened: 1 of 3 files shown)


src/internal/operations/manifests.go
```diff
@@ -151,18 +150,33 @@ func produceManifestsAndMetadata(
...
+		// we avoid the global fault.Bus because all failures here are ignorable,
+		// and cascading errors up to the operation can cause a conflict that forces
+		// the operation into a failure state unnecessarily.
+		// TODO(keepers): this is not a pattern we want to
+		// spread around.  Need to find more idiomatic handling.
+		fb := fault.New(true)
+
+		colls, err := collectMetadata(mctx, mr, man, metadataFiles, tenantID, fb)
 		if err != nil && !errors.Is(err, data.ErrNotFound) {
+			LogFaultErrors(ctx, fb.Errors(), "collecting metadata")
```
ashmrtn, line 163: move this above the if-block and remove the one lower down?

ashmrtn (approved):

src/internal/operations/manifests.go
ryanfkeepers: Sure, why not

PR #3019, opened 2023-04-04
Cleanup birthday events for old contacts

gmatev:
Purge events birthdays corresponding to old contacts
---

#### Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Supportability/Tests
- [x] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Issue #2334, opened 2023-01-31
Refactor invalid metadata handling during incremental backups


ashmrtn:
There may be certain edge cases where we discard metadata from a previous backup in GraphConnector (i.e. incorrectly formatted data etc). In those cases, we don't mark collections as DoNotMerge which could theoretically lead to zombie items in kopia as the base directory will still be merged into the results of the collection.

Overall, we may need a more explicit and direct way to denote what kopia should and shouldn't merge. Whether something should be merged or not may also depend on the service (e.x. Exchange has a delta token per folder but OneDrive has a single delta token for all items)

ashmrtn:
a "big hammer" solution that could solve the merging issues is to make a collection for the category and mark it as deleted. Since kopia wrapper understands tree operations, this would be the equivalent of `rm -rf <root of tree for category>` and would stop merges for any folders under it

By "make a collection for the category", I mean something like a collection for the path `<tenant ID>/service/<user ID or PN>/<category>`. Collections like this (though not marked as deleted) are [already created](https://github.com/alcionai/corso/blob/896c05f62322bb2e4d9c31cb313f95a96236e4ed/src/internal/connector/exchange/data_collections.go#L214) by some other code to ensure that we at least have the base of the hierarchy so we don't see failures during merging if a user had no emails or something

It would be safe to have both a collection marked as deleted and a collection marked as new for the same path due to how kopia handles merging (one refers to the old version of the data while the other refers to the new)

ashmrtn:
can't quite close this yet. The logic for OneDrive/SharePoint metadata decoding should be updated slightly. It should be less aggressive about [removing incomplete metadata](https://github.com/alcionai/corso/blob/a9918d2f78b9419af553a6e0356bbf5f6078b9d8/src/internal/connector/onedrive/collections.go#L215) so we don't accidentally miss marking collections as do not merge or deleted. We should also probably just return an error and fail if we have trouble decoding metadata instead of [adding an error to the bus](https://github.com/alcionai/corso/blob/a9918d2f78b9419af553a6e0356bbf5f6078b9d8/src/internal/connector/onedrive/collections.go#L208)

PR #3092, opened 2023-04-11
always include delta metadata, additional logging

ryanfkeepers:
Ensure exchange always includes the delta metadata entry, even if it has zero deltaURLs to record.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* fixes #3019

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -48,6 +48,7 @@ func filterContainersAndFillCollections(
 		// copy of previousPaths.  any folder found in the resolver get
 		// deleted from this map, leaving only the deleted folders behind
 		tombstones = makeTombstones(dps)
+		category   = qp.Category
 	)
 
 	logger.Ctx(ctx).Infow("filling collections", "len_deltapaths", len(dps))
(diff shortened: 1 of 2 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -75,28 +76,34 @@ func filterContainersAndFillCollections(
...
 		var (
 			dp          = dps[cID]
 			prevDelta   = dp.delta
-			prevPathStr = dp.path
+			prevPathStr = dp.path // do not log: pii
 			prevPath    path.Path
+			ictx        = clues.Add(
+				ctx,
+				"container_id", cID,
+				"previous_delta", prevDelta)
```
ashmrtn, line 87: is the delta string going to have PII in it? Seems like it would have user display name or ID at least?

ashmrtn (approved):
I think we still need some sort of fix re: #2334 else we may end up with zombie folders because the path map was cleared due to missing delta token(s)

src/internal/connector/exchange/service_iterators.go
ryanfkeepers: Ah, check, thanks.

PR #3099, opened 2023-04-11
standardize around site names without host

ryanfkeepers:
Standardize around normalizing site names to
not include the host portion, and only include the
two following uuids, since those suffice as a unique
id for graph lookups and will remain stable.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
closing, can deal with later as a migration

PR #3100, opened 2023-04-11
Create abstraction to represent a unique location

ashmrtn:
UniqueLocation allows representing a folder path that has
both a location that will be stored in backup details and
a location that can be used as a key in maps. The key for
maps is guaranteed to be unique for all data types within
a service

Add a function to extract a unique location from a backup
details entry and tests for that

UniqueLocation will eventually be used in GraphConnector
and KopiaWrapper so it needs to be in a package that both
of them can import

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2486

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/version/backup.go
```diff
@@ -1,5 +1,7 @@
 package version
 
+// TODO(ashmrtn): If bumping this adjust OneDriveXLocRef. It's not used in prod
+// yet but does gate some tests.
```
ryanfkeepers, line 4: This makes me a uneasy.  I'd rather changes like this commit to a version and bump.  In the unlikely event that some other version 7 goes to prod while this is still in queue, we can change the PR.  In the similarly-unlikely event that two PRs declare a version bump, labeling them both 7 is benign.

Or- if you don't plan to commit to a version- set the const to `Backup+1`.

src/pkg/backup/details/details.go
```diff
@@ -13,9 +13,76 @@ import (
...
+	return ul.pb
+}
+
+func (ul uniqueLoc) DetailsLocation() *path.Builder {
+	return path.Builder{}.Append(ul.pb.Elements()[ul.prefixElems:]...)
+}
+
+// Having service-specific constructors can be kind of clunky, but in this case
+// I think they'd be useful to ensure the proper args are used since this
+// path.Builder is used as a key in some maps.
```
ryanfkeepers, line 45: This is going to be read by golang as the description for NewExchangeUniqueLocation, while the other two funcs don't have any description.  Seems unwanted.  Also, these constructors could use some description for callers about their behavior and expectations, for quick insight into their assumptions.  Ex:
```
// NewExchangeUniqueLocation builds a category-specific location identifier.
// it assumes that paths are maximally unique within a given data category.
func NewExchangeUniqueLocation(
```

src/pkg/path/onedrive_test.go
```diff
@@ -35,13 +35,13 @@ func (suite *OneDrivePathSuite) Test_ToOneDrivePath() {
 		{
 			name:         "Root path",
 			pathElements: []string{"drive", "driveID", "root:"},
-			expected:     &path.DrivePath{DriveID: "driveID", Folders: []string{}},
+			expected:     &path.DrivePath{DriveID: "driveID", Root: "root:", Folders: []string{}},
```
ryanfkeepers, line 38: `"root:"` might make a good const.

src/pkg/backup/details/details.go
```diff
@@ -13,9 +13,76 @@ import (
...
 )
 
+// UniqueLocationer provides access to location information but guarantees that
+// it can also generate a unique location (among items in the same service but
+// possibly across data types within the service) that can be used as a key in
+// maps and other structures. The unique location may be different than
+// DetailsLocation, the location used in backup details.
+type UniqueLocationer interface {
+	UniqueLocation() *path.Builder
+	DetailsLocation() *path.Builder
```
ryanfkeepers, line 27: These names are a bit of a mouthful.  Thoughts on this?
```suggestion
type LocationIDer interface {
	ID() *path.Builder
	InDetails() *path.Builder
```

ryanfkeepers (approved):
just nits

src/pkg/backup/details/details.go
ryanfkeepers: names are too difficult.  We should switch to emojis.

src/pkg/backup/details/details.go
ryanfkeepers: > The fact that it's category-specific should be hidden from the caller.

Hmm, have to disagree here.  The implementation can be hidden from the interface _consumer_, sure.  But this is a constructor.  The caller has the right to know the purpose of the parameters and their usage.  Not to mention that it's already: accessible via code reference without searching for an implementation (unlike an interface), partially inferrable from the parameters (so you're just clarifying what is already logical) and, most importantly, is a necessary consideration of the data that we work with (if other structs that comply with this interface want to do more to guarantee uniqueness, that's on them, but it's fair to declare that you're doing what's minimally necessary).

src/pkg/backup/details/details.go
```diff
@@ -13,9 +13,84 @@ import (
...
+// I think they'd be useful to ensure the proper args are used since this
+// path.Builder is used as a key in some maps.
+
+// NewExchangeLocationIDer builds a LocationIDer for the given category and
+// folder path. The path denoted by the folders should be unique within the
+// category.
+func NewExchangeLocationIDer(
+	category path.CategoryType,
+	escapedFolders ...string,
+) LocationIDer {
```
ryanfkeepers, line 53: Caught this on second pass.  I'm a big fan of the `accept interfaces, return structs` idiom, unless it's otherwise necessary.  I don't see any necessity here to return the core interface.  Would be more idiomatic to return the struct itself.

src/pkg/backup/details/details.go
ashmrtn: adding that it's category-specific invites people to rely on the fact that it _will_ have the category now and forever. Since this is something that is not directly persisted (i.e. we never persist the `ID()` result directly, only `InDetails()`) that's not a safe assumption to make about the design as we could change it without worrying about fixing up previously persisted data. Overall, the fact that the category is used for uniqueness is immaterial. If other things existed that could guarantee uniqueness those could be used instead.

In a similar vein, since OneDrive only has one category, it's not required to have category as part of the unique location. It's in there now, but we could just as easily remove it. In that case it would no longer comply with the "unofficial" interface we're discussing here

src/pkg/backup/details/details.go
ryanfkeepers: Imo, good to keep things in line with standards.  Plus, having the struct not be exported will both prevent ad-hoc struct generation and ensure any func which consumes it will prefer an interface over the struct itself.  Overall a safe move that matches golang expectations.

src/pkg/backup/details/details.go
ryanfkeepers: > In that case it would no longer comply with the "unofficial" interface we're discussing here

Ah, I think there's some misunderstanding here.  I'm not pushing category as an expected part of the interface.  Only as a descriptor for how this exchange _constructor_ behaves.  The onedrive constructor doesn't directly request a category, therefore the caller can already assume it's not utilized/required.

What I'm asking to expose is that these constructors describes their own behavior as, "I'm requesting you give me a <category|driveid|etc> because I'm going to use it like <implementation> in order to comply with <local assumptions>".   These are declarations that these constructors can make without implication about the interface design at large.

As a best practice around interfaces, I think it's important for implementations to explain their design choices and assumptions around compliance with the interface.  It's proper to not extend those assumptions onto the interface itself (ie: any one local implementation does not dictate the interface's behavior).  This additional description will both aid future contributors in understanding the design choices, and help us catch implementation-specific bugs, misbehavior, and misunderstanding.

ashmrtn: >"I'm requesting you give me a <category|driveid|etc> because I'm going to use it like in order to comply with ".

I still really don't see this as something users of an interface need to know in all the gory detail. Using abstractions is a way to get away from programmers needing to know all the gory details of every implementation. While having it there may give insight into design choices, it's not going to let anyone use the interface more quickly. To use it they just need to know what args to pass it, which is already encoded in the parameter names and types. Having a comment about why it's this way also opens us up to stale comments because stuff like that doesn't always get updated if the code changes. Design proposals/decisions should be outside of the codebase because having them in the codebase tends to add a lot of cruft and isn't necessarily useful to people using the code

>It's proper to not extend those assumptions onto the interface itself

This is what worries me. In practice people tend to blur the line between implementation and interface. If the data is already there, why not make use of it and extend the interface just a little bit more and all that 🙃

ryanfkeepers: > Using abstractions is a way to get away from programmers needing to know all the gory details of every implementation.

But there is no abstraction in the construction of a struct.  Programmers need to be able to ignore the gory details of implementation in any instance that consumes the interface behavior, certainly.  But extending that abstraction onto the construction of a variable which complies with an interface is an anti-pattern.

At the time of calling this constructor, the user either: 1/ has no other constructors to choose from, in which case the use of an interface is largely moot, since no abstraction is required, 2/ has multiple constructors to choose from, in which case clear knowledge about the implementation details of each is required to make the correct selection, or 3/ is in the same package as the constructor, and wouldn't bother to choose any other options, in which case the implementation details are both exposed and relevant within that context.

> it's not going to let anyone use the interface more quickly. To use it they just need to know what args to pass it, which is already encoded in the parameter names and types.

A distillation of the conversation at large, but: constructors are not interfaces.  Abstraction should be prioritized at the time of consumption.  Otherwise the duck-typing design of golang is irrelevant.  "to use [the constructor] they just need [the args]" and "to use [the interface] they just need [a compliant struct]" are separate concerns.

> If the data is already there, why not make use of it and extend the interface just a little bit more and all that

If the interface has multiple compliant structs throughout the repo, then it would be a smell to modify the interface just to satisfy one of those implementations, yeah?  That change isn't necessarily supported by other implementations.  If it is, great, maybe it does make sense for the interface to expect that.

ashmrtn: >But there is no abstraction in the construction of a struct.

These functions _are_ constructors for the abstraction essentially. The struct that it's constructing isn't exported, so people calling these functions aren't calling them to get a handle to instances of the struct, they're calling it to get a handle to the abstraction. The struct itself isn't even going to appear in generated godocs because it's not exported and it's confusing to discuss trivial things about a struct that isn't going to appear in docs

Whether the abstraction is backed by this struct or a different struct is immaterial. Likewise, as long as it's clear what the params should be it's immaterial in this case what precisely they're being used for because it's not related to performance tuning at all (i.e. for things like setting the number of buckets on a hash map it could matter)

The only reason these functions actually appear as opposed to having a more generic `NewLocationIDer(pb *path.Builder, prefixElems int)` is because at least with separate constructors the order of prefix elements (purely an internal concern) is well-controlled and things like drive ID/drive name are documented in the input args

>If the interface has multiple compliant structs throughout the repo, then it would be a smell to modify the interface just to satisfy one of those implementations, yeah?

At the moment this is the only compliant struct, and all constructed instances of it have the category. It's not always needed, but they have it at the moment. Requiring the category would be a code smell for an interface that is only meant to provide a unique key for an item and the item

ryanfkeepers: > they're calling it to get a handle to the abstraction

Per duck-typing, the struct _is_ a handle to the abstraction.  Broadly speaking, a constructor result may also provide a superset of the given abstraction, of which other parts may be valuable to other consumers.  

> is because at least with separate constructors the order of prefix elements (purely an internal concern) is well-controlled and things like drive ID/drive name are documented in the input args

Yes, I agree!  That's what I'm saying, too.  And, also, that it'd be nice for the constructor to have a comment saying, "assumes items are only unique within a drive", as explanation of its args.

> Requiring the category would be a code smell for an interface that is only meant to provide a unique key for an item and the item

Hmm, I still feel like I'm not making myself clear.  Total agreement that the inclusion of the category is an implementation detail, and does not need to be mentioned in all cases.  However, because the exchange constructor _does_ request the category as a parameter, a comment on that constructor like "assumes uniqueness can only be guaranteed within each exchange category" seems useful.

ashmrtn: >Per duck-typing, the struct is a handle to the abstraction. Broadly speaking, a constructor result may also provide a superset of the given abstraction, of which other parts may be valuable to other consumers.

If the struct is exported or has a broader API sure. Neither of those are the case for this

>a comment on that constructor like "assumes uniqueness can only be guaranteed within each exchange category" seems useful.

This is different than saying _for what (i.e. unique key generation)_ the value is used which was my interpretation of earlier comments

PR #3106, opened 2023-04-12
Shared mailbox

neha-Gupta1:
Handle user's API to provide user purpose of users

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -150,6 +150,22 @@ func (c Users) GetByID(ctx context.Context, userID string) (models.Userable, err
 	return resp, err
 }
 
+func (c Users) GetMailSetting(ctx context.Context, userID string) (models.MailboxSettingsable, error) {
+	var (
+		resp models.Userable
+		err  error
+	)
+
+	resp, err = c.stable.Client().UsersById(userID).Get(ctx, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "getting user")
+	}
+
+	userResponse := resp.GetMailboxSettings()
+
+	return userResponse, nil
+}
+
 func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
 	// Assume all services are enabled
 	// then filter down to only services the user has enabled
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -150,43 +150,22 @@ func (c Users) GetByID(ctx context.Context, userID string) (models.Userable, err
 	return resp, err
 }
 
-func (c Users) GetMailSetting(ctx context.Context, userID string) (models.MailboxSettingsable, error) {
+func (c Users) GetUserPurpose(ctx context.Context, userID string) (string, error) {
 	var (
-		resp models.Userable
-		err  error
+		err     error
+		rawURL  = "https://graph.microsoft.com/v1.0/users/" + userID + "/mailboxSettings"
+		apadtor = c.stable.Adapter()
+		builder = users.NewUserItemRequestBuilder(rawURL, apadtor)
 	)
 
-	resp, err = c.stable.Client().UsersById(userID).Get(ctx, nil)
+	newItem, err := builder.Get(ctx, nil)
 	if err != nil {
-		return nil, graph.Wrap(ctx, err, "getting user")
+		return "", graph.Wrap(ctx, err, "creating item")
 	}
 
-	userResponse := resp.GetMailboxSettings()
+	userPurpose := *(newItem.GetAdditionalData()["userPurpose"].(*string))
 
-	//TODO: work on this
-	// rawURL := "https://graph.microsoft.com/v1.0/users/" + userID + "/mailboxSettings" //fmt.Sprintf(itemChildrenRawURLFmt, driveID, parentFolderID)
-	// apadtor := c.stable.Adapter()
-	// // builder := users.NewDeltaRequestBuilder(rawURL, apadtor)
-	// // users.ItemSettingsRequestBuilder(rawURL, apadtor)
-	// builder := users.NewUserItemRequestBuilder(rawURL, apadtor)
-
-	// // builder := users.NewCollectionsRequestBuilder(rawURL, apadtor)
-	// newItem, err := builder.Get(ctx, nil)
-	// if err != nil {
-	// 	fmt.Println("Error: ", err)
-	// 	return nil, graph.Wrap(ctx, err, "creating item")
-	// }
-	// fmt.Println("NewItem: ", newItem.GetAdditionalData())
-	// data := newItem.GetMailboxSettings()
-	// // data := newItem.GetAdditionalData()
-	// // address := data["userPurpose"].(*string)
-	// // address1 := data["userPurpose"].(*models.MailboxSettingsable)
-	// fmt.Println("*******newItem: ", data) // ["userPurpose"], &address, address, address1
-
-	// // userable := newItem.GetValue()
-	// // fmt.Println("*******newItem userable: ", userable)
-
-	return userResponse, nil
+	return userPurpose, nil
 }
 
 func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
--- a/src/internal/connector/discovery/api/users.go
+++ b/src/internal/connector/discovery/api/users.go
@@ -163,6 +163,29 @@ func (c Users) GetMailSetting(ctx context.Context, userID string) (models.Mailbo
 
 	userResponse := resp.GetMailboxSettings()
 
+	//TODO: work on this
+	// rawURL := "https://graph.microsoft.com/v1.0/users/" + userID + "/mailboxSettings" //fmt.Sprintf(itemChildrenRawURLFmt, driveID, parentFolderID)
+	// apadtor := c.stable.Adapter()
+	// // builder := users.NewDeltaRequestBuilder(rawURL, apadtor)
+	// // users.ItemSettingsRequestBuilder(rawURL, apadtor)
+	// builder := users.NewUserItemRequestBuilder(rawURL, apadtor)
+
+	// // builder := users.NewCollectionsRequestBuilder(rawURL, apadtor)
+	// newItem, err := builder.Get(ctx, nil)
+	// if err != nil {
+	// 	fmt.Println("Error: ", err)
+	// 	return nil, graph.Wrap(ctx, err, "creating item")
+	// }
+	// fmt.Println("NewItem: ", newItem.GetAdditionalData())
+	// data := newItem.GetMailboxSettings()
+	// // data := newItem.GetAdditionalData()
+	// // address := data["userPurpose"].(*string)
+	// // address1 := data["userPurpose"].(*models.MailboxSettingsable)
+	// fmt.Println("*******newItem: ", data) // ["userPurpose"], &address, address, address1
+
+	// // userable := newItem.GetValue()
+	// // fmt.Println("*******newItem userable: ", userable)
+
 	return userResponse, nil
 }
 
--- a/src/internal/connector/discovery/discovery.go
+++ b/src/internal/connector/discovery/discovery.go
@@ -71,13 +71,13 @@ func UsersDetails(
 	acct account.Account,
 	userID string,
 	errs *fault.Bus,
-) (models.MailboxSettingsable, error) {
+) (string, error) {
 	client, err := apiClient(ctx, acct)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
-	return client.Users().GetMailSetting(ctx, userID)
+	return client.Users().GetUserPurpose(ctx, userID)
 }
 
 // User fetches a single user's data.
--- a/src/internal/connector/discovery/discovery.go
+++ b/src/internal/connector/discovery/discovery.go
@@ -65,6 +65,21 @@ func Users(
 	return client.Users().GetAll(ctx, errs)
 }
 
+// UsersDetails fetches detailed info like - userPurpose for all users in the tenant.
+func UsersDetails(
+	ctx context.Context,
+	acct account.Account,
+	userID string,
+	errs *fault.Bus,
+) (models.MailboxSettingsable, error) {
+	client, err := apiClient(ctx, acct)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.Users().GetMailSetting(ctx, userID)
+}
+
 // User fetches a single user's data.
 func User(ctx context.Context, gwi getWithInfoer, userID string) (models.Userable, *api.UserInfo, error) {
 	u, err := gwi.GetByID(ctx, userID)
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -36,6 +36,7 @@ const (
 	errCodeResourceNotFound            = "ResourceNotFound"
 	errCodeRequestResourceNotFound     = "Request_ResourceNotFound"
 	errCodeMailboxNotEnabledForRESTAPI = "MailboxNotEnabledForRESTAPI"
+	errCodeErrorAccessDenied           = "ErrorAccessDenied"
 )
 
 const (
@@ -96,6 +97,10 @@ func IsErrExchangeMailFolderNotFound(err error) bool {
 
 func IsErrUserNotFound(err error) bool {
 	return hasErrorCode(err, errCodeRequestResourceNotFound)
+}
+
+func IsErrAccessDenied(err error) bool {
+	return hasErrorCode(err, errCodeErrorAccessDenied)
 }
 
 func IsErrTimeout(err error) bool {
--- a/src/pkg/services/m365/m365.go
+++ b/src/pkg/services/m365/m365.go
@@ -18,6 +18,7 @@ type User struct {
 	PrincipalName string
 	ID            string
 	Name          string
+	userPurpose   string
 }
 
 // UsersCompat returns a list of users in the specified M365 tenant.
@@ -50,6 +51,13 @@ func Users(ctx context.Context, acct account.Account, errs *fault.Bus) ([]*User,
 			return nil, clues.Wrap(err, "parsing userable")
 		}
 
+		mailBoxSetting, err := discovery.UsersDetails(ctx, acct, pu.ID, errs)
+		if err != nil {
+			return nil, clues.Wrap(err, "parsing userable")
+		}
+
+		userPurpose := mailBoxSetting.GetUserPurpose()
+		pu.userPurpose = userPurpose.String()
 		ret = append(ret, pu)
 	}
 
--- a/src/pkg/services/m365/m365.go
+++ b/src/pkg/services/m365/m365.go
@@ -10,15 +10,17 @@ import (
 	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/discovery"
+	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 type User struct {
 	PrincipalName string
 	ID            string
 	Name          string
-	userPurpose   string
+	UserPurpose   string
 }
 
 // UsersCompat returns a list of users in the specified M365 tenant.
@@ -51,13 +53,16 @@ func Users(ctx context.Context, acct account.Account, errs *fault.Bus) ([]*User,
 			return nil, clues.Wrap(err, "parsing userable")
 		}
 
-		mailBoxSetting, err := discovery.UsersDetails(ctx, acct, pu.ID, errs)
+		pu.UserPurpose, err = discovery.UsersDetails(ctx, acct, pu.ID, errs)
 		if err != nil {
-			return nil, clues.Wrap(err, "parsing userable")
+			// TODO: handle the access denied scenario
+			if graph.IsErrAccessDenied(err) {
+				// // clues.Wrap(err, fmt.Sprintf("access denied for %s", pu.ID))
+				// errList = append(errList, fmt.Errorf("access denied for %s", pu.ID))
+				logger.Ctx(ctx).Infow("access denied for %s", pu.ID)
+			}
 		}
 
-		userPurpose := mailBoxSetting.GetUserPurpose()
-		pu.userPurpose = userPurpose.String()
 		ret = append(ret, pu)
 	}
 


gmatev:
@neha-Gupta1 after some more thinking, the logic whether something is designated as a shared mailbox should also consider if there is an active onedrive for the userId. Shared Mailboxes should not have a OneDrive.

src/internal/connector/discovery/api/users.go
```diff
@@ -165,6 +166,77 @@ func (c Users) GetByID(ctx context.Context, identifier string) (models.Userable,
...
+		userPurpose = ptr.Val(newItem.GetAdditionalData()["userPurpose"].(*string))
+	}
+
+	if err != nil {
+		if !(graph.IsErrAccessDenied(err) || graph.IsErrExchangeMailFolderNotFound(err)) {
+			return "", false, false, graph.Wrap(ctx, err, "getting purpose")
+		}
+
+		// not the best way to handle the errors. But since we don't won't to break
+		// this request and just want to convey a message that access are invalid have
```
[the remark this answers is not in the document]
neha-Gupta1: I had a discussion with @vkamra that since its a new change we don't won't to break things for existing users with lesser permissions. 
Can we add one more value here specifying err if any. And that goes as part of user's struct output?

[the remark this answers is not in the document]
meain: I don't have the full context of the issue, but is "user" a reasonable fallback in case of an error?

src/internal/connector/discovery/discovery.go
```diff
@@ -69,6 +69,22 @@ func Users(
 	return users, nil
 }
 
+// UsersDetails fetches detailed info like - userPurpose for all users in the tenant.
+func UsersDetails(
```
meain, line 73:
```suggestion
func UserDetails(
```

src/internal/connector/discovery/api/users.go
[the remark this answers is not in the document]
neha-Gupta1: No. its just that we can have different type of users like - user or shared.

src/pkg/services/m365/m365.go
```diff
@@ -29,9 +29,13 @@ type ServiceAccess struct {
...
 type User struct {
-	PrincipalName string
-	ID            string
-	Name          string
+	PrincipalName      string
+	ID                 string
+	Name               string
+	UserPurpose        string
+	HasMailBox         bool
+	HasOnedrive        bool
```
ryanfkeepers, line 37: This overlaps with UserInfo, which seems like the better owner for this sort of metadata.

src/internal/connector/discovery/discovery.go
```diff
@@ -69,6 +69,22 @@ func Users(
 	return users, nil
 }
 
+// UserDetails fetches detailed info like - userPurpose for all users in the tenant.
+func UserDetails(
```
ryanfkeepers, line 73: Rather than splitting it out, packing this info UserInfo would make for a better consolidated owner.

src/internal/connector/discovery/api/users.go
```diff
@@ -165,6 +166,78 @@ func (c Users) GetByID(ctx context.Context, identifier string) (models.Userable,
 	return resp, err
 }
 
+// GetUserInfo provide information about a user like- userPurpose/type,
+// if mailbox and onedrive is present for user.
+func (c Users) GetUserInfo(ctx context.Context, userID string) (string, bool, bool, string, error) {
+	var (
+		rawURL                         = "https://graph.microsoft.com/v1.0/users/" + userID + "/mailboxSettings"
```
ryanfkeepers, line 173: Please make a formattable template const for this url.

src/internal/connector/discovery/api/users.go
```diff
@@ -165,6 +166,78 @@ func (c Users) GetByID(ctx context.Context, identifier string) (models.Userable,
...
+		apadtor                        = c.stable.Adapter()
```
ryanfkeepers, line 174:
```suggestion
		adapter                        = c.stable.Adapter()
```

src/internal/connector/discovery/api/users.go
```diff
@@ -165,6 +166,78 @@ func (c Users) GetByID(ctx context.Context, identifier string) (models.Userable,
 	return resp, err
 }
 
+// GetUserInfo provide information about a user like- userPurpose/type,
+// if mailbox and onedrive is present for user.
+func (c Users) GetUserInfo(ctx context.Context, userID string) (string, bool, bool, string, error) {
```
ryanfkeepers, line 171: A couple issues here: 
* stuttering: `users.GetUserInfo` is redundant.  `users.GetInfo` is more sensible.
* collision: `users.GetInfo` already exists.  Having both funcs makes determining their distinct purposes difficult.
* Also, again (trying not to say this too many times, sorry), they could be combined.

src/internal/connector/discovery/api/users.go
```diff
@@ -165,6 +166,78 @@ func (c Users) GetByID(ctx context.Context, identifier string) (models.Userable,
...
+			return "", false, false, "", graph.Wrap(ctx, err, "getting onedrive")
+		}
+
+		logger.Ctx(ctx).Infof("resource owner does not have a drive")
+
+		hasOneDrive = false
+	}
+
+	// get userPurpose for user
+	newItem, err := builder.Get(ctx, nil)
```
ryanfkeepers, line 211: `builder` isn't getting re-used.  Rather than separating the builder creation from its usage, both should occur here in the same line.

Also, stepping back from the granular implementation a bit.  This function is a bit overloaded, and is going to be more difficult than necessary to write tests for.  Consider a layout like the following as an alternative, which breaks each call out to its own func and only uses this func to control aggregation of the results:

```go
func (c Users) GetInfo(ctx) (UserInfo, error) {
  userInfo := newUserInfo()
  
  userInfo.Exchange, err = c.allowsExchange(ctx)
  if err { return }
  
  userInfo.OneDrive, err = c.allowsOneDrive(ctx)
  if err { return }
  
  userInfo.purpose, err = c.getPurpose(ctx)
  if err { return }
  
  return userInfo, nil
}

func (c Users) allowsExchange(ctx) (bool, err) {...}
func (c Users) allowsOneDrive(ctx) (bool, err) {...}
func (c Users) getPurpose(ctx) (string, err) {...}
```

This does the following:
1. makes the GetInfo func easy to read and grok.
2. places all the granular handling into self-contained functions.  funcs like that are easy to mock and test and make clear assertions about without cross contamination from other calls.
3. provides a set of re-usable funcs in case we want to make the same individual calls elsewhere.

src/internal/connector/discovery/api/users.go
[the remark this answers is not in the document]
ryanfkeepers: Returning a value as an ignored-error indicator is a smell.  First and foremost, we need to ensure there's an error log in all of these cases.  That will do most of the work in communicating the occurrence.  If we need to pivot logic based on this state, then we'll want a better set of identifiers than a descriptive string, anyway (passing up an error and letting a higher layer handle it, for example).

ryanfkeepers (changes requested):
Many concerns about design.  Additionally, all of these new functions need testing.

gmatev:
@neha-Gupta1 for the things coming from /mailboxSettings can we just get all the data from https://learn.microsoft.com/en-us/graph/api/resources/mailboxsettings?view=graph-rest-1.0#json-representation. This will give flexibility to the caller to use some of this in the future as needed.

src/pkg/services/m365/m365.go
neha-Gupta1: To be clear, we have
- a struct named `UserInfo`
- a func named `UserInfo`

Do we want to accommodate the implementation in the `UserInfo` function and these value go as part of `UserInfo` struct?

neha-Gupta1: As of now I have merged the implementation and the struct too. The only issue I see with this is-
In cases where only `HasMailBox` info was required, we make too more calls. Which might be a overhead.

src/internal/connector/discovery/discovery.go
neha-Gupta1: Done that. Just a thought- 
In cases where only HasMailBox info was required, we make two more calls. Which might be a overhead.

ryanfkeepers: Good point. How about exposing a func that can make that call in isolation?

src/internal/connector/discovery/api/users.go
```diff
@@ -184,17 +195,112 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+	err = c.getAdditionalData(ctx, userID, userInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	return userInfo, nil
+}
+
+// verify mailbox enabled for user
+func (c Users) allowsExchange(ctx context.Context, userID string, userInfo *UserInfo) error {
```
ryanfkeepers, line 217: This function leans a bit too heavily towards the interpretation of its results.  What it's fundamentally doing is retrieving exchange mailfolders.  We want to eventually interpret the results as proof the user can access exchange.  But if we needed to make a different interpretation based on mailfolders, we'd have to change this func signature or create a new func, right?  Packing the interpretation into the func itself makes it less portable.

Instead of adding a set of `allowsFoo()` funcs, it'd be better to add thin wrappers on each call: getUsersMailFolders, getUsersDrives, getMailboxSettings, etc.  The caller can interpret the results however they want.

src/internal/connector/discovery/api/users.go
```diff
@@ -220,3 +326,31 @@ func validateUser(item any) (models.Userable, error) {
...
+		ok          bool
+		dataPointer *string
+		value       string
+	)
+
+	dataPointer, ok = data.(*string)
+	if !ok {
+		logger.Ctx(ctx).Infof("error getting mailboxSettings")
+
+		userInfo.ErrGetMailBoxSetting = "not found"
```
ryanfkeepers, line 341: Mutating the userInfo is a side effect that shouldn't be part of this function.  You could instead catch an empty string when handling the result for the same consequence, and no side effect.

Alternatively, pass back an error sentinel, like so:
```go
var ErrMailBoxSettingsNotFound = clues.New("mailbox settings not found")

s, err := convert(...)
if err != nil {
  userInfo.ErrGetMailboxSetting = err
}

func convert(...) (string, error) {
  if !ok {
    return "", ErrMailBoxSettingsNotFound
  }
} 
```

src/internal/connector/graph/errors.go
```diff
@@ -36,6 +36,7 @@ const (
 	errCodeResourceNotFound            = "ResourceNotFound"
 	errCodeRequestResourceNotFound     = "Request_ResourceNotFound"
 	errCodeMailboxNotEnabledForRESTAPI = "MailboxNotEnabledForRESTAPI"
+	errCodeErrorAccessDenied           = "ErrorAccessDenied"
```
ryanfkeepers, line 39: Is this different than a 401 status code?

src/internal/connector/discovery/api/users.go
```diff
@@ -220,3 +326,31 @@ func validateUser(item any) (models.Userable, error) {
 
 	return m, nil
 }
+
+func convertInterfaceToString(ctx context.Context, data interface{}, userInfo *UserInfo) string {
```
ryanfkeepers, line 330: nit on func name.  also, prefer `any` over `interface{}`.
```suggestion
func toString(ctx context.Context, data any, userInfo *UserInfo) string {
```

src/internal/connector/discovery/api/users.go
```diff
@@ -220,3 +326,31 @@ func validateUser(item any) (models.Userable, error) {
...
+	if !ok {
+		logger.Ctx(ctx).Infof("error getting mailboxSettings")
+
+		userInfo.ErrGetMailBoxSetting = "not found"
+
+		return ""
+	}
+
+	value, ok = ptr.ValOK(dataPointer)
+	if !ok {
```
ryanfkeepers, line 347: ok doesn't check for zero len strings.  Is this a better check?
```suggestion
	value := ptr.ValdataPointer)
	if len(value) == 0 {
```

src/internal/connector/discovery/api/users.go
```diff
@@ -220,3 +326,31 @@ func validateUser(item any) (models.Userable, error) {
 
 	return m, nil
 }
+
+func convertInterfaceToString(ctx context.Context, data interface{}, userInfo *UserInfo) string {
+	var (
+		ok          bool
+		dataPointer *string
+		value       string
+	)
```
ryanfkeepers, line 335: These pre-declarations are unnecesssary: the values aren't getting re-used, overwritten, referenced multiple times, nor are we conditionally using their zero value.  It's cleaner in this case to use normal declaration.  eg: datapointer, ok := ...

src/pkg/services/m365/m365.go
```diff
@@ -72,6 +81,28 @@ func Users(ctx context.Context, acct account.Account, errs *fault.Bus) ([]*User,
...
+
+		pu.HasMailBox = userInfo.HasMailBox
+		pu.HasOnedrive = userInfo.HasOneDrive
+
+		// mailbox and onedrive info is independent of mailboxSetting.
+		// so getting them and then checking if any error getting mailboxSettings
+		if userInfo.ErrGetMailBoxSetting != "" {
+			pu.ErrGettingUserInfo = userInfo.ErrGetMailBoxSetting
+			continue
+		}
```
ryanfkeepers, line 97: This conditional is more effort than necessary.  We have no nil-pointer issues to be concerned with, and golang zero-values are viable will suffice for all the other values.  Easier overall just to plug in the value and not worry about the rest.
```suggestion
		pu.ErrGettingUserInfo = userInfo.ErrGetMailBoxSetting
```

Although, replacing this with a userInfo struct is going to be the easiest way forward overall.

src/pkg/services/m365/m365.go
```diff
@@ -29,9 +29,18 @@ type ServiceAccess struct {
...
+	Name                          string
+	ArchiveFolder                 string
+	Timezone                      string
+	DelegateMeetingMsgDeliveryOpt string
+	DateFormat                    string
+	TimeFormat                    string
+	UserPurpose                   string
+	HasMailBox                    bool
+	HasOnedrive                   bool
+	ErrGettingUserInfo            string
```
ryanfkeepers, line 43: Why not embed a UserInfo struct here?
```suggestion
    Info UserInfo
```

src/internal/connector/discovery/api/users.go
```diff
@@ -34,7 +35,17 @@ type Users struct {
...
+	Purpose                    string
+	ArchiveFolder              string
+	Status                     string
+	DateFormat                 string
+	TimeFormat                 string
+	DelegateMeetMsgDeliveryOpt string
+	Timezone                   string
+	HasMailBox                 bool
+	HasOneDrive                bool
+	ErrGetMailBoxSetting       string
```
ryanfkeepers, line 48: Since these are all mailbox settings, it might be more maintainable for the future to construct a separate struct to hold them, like so:
```go
type mailboxInfo struct {

	Purpose                    string
	ArchiveFolder              string
	Status                     string
	DateFormat                 string
	TimeFormat                 string
	DelegateMeetMsgDeliveryOpt string
	Timezone                   string
	HasMailBox                 bool
	HasOneDrive                bool
	ErrGetMailBoxSetting       string
}
```

and then add that item to the userInfo.  That way you don't need to worry about future naming collisions, and more generic variables (Purpose, Status, Timezone) are already known to be part of the mailbox info, and not necessarily the user info at large (which may conflict across applications).
```suggestion
  mailbox mailboxInfo
```

ryanfkeepers (approved):

src/pkg/services/m365/m365.go
neha-Gupta1: Actually, it fails if the mailbox settings values are blanks.

neha-Gupta1: done

src/internal/connector/discovery/api/users.go
neha-Gupta1: there is a possibility that few values are blank and length will be zero in that case. We are fine with it. Only scenario we want to avoid is that - the code should not break with a nil pointer exception  if some field is unset.

src/internal/connector/graph/errors.go
neha-Gupta1: access denied is 403.  401 is unauthorised.

neha-Gupta1:
> @ryanfkeepers we will need to add - `MailboxSettings.Read` permission for getting mailbox setting and the testcases to pass.

@ryanfkeepers just a reminder.

src/internal/connector/discovery/api/users.go
neha-Gupta1: done

src/internal/connector/discovery/discovery.go
ryanfkeepers, line 73: Nit: since we're getting user info
```suggestion
func GetUserInfo(
```

src/internal/connector/discovery/api/users.go
ryanfkeepers: Ah, I don't think I explained my expectations clearly enough, since these functions are still of the `allowsFoo` pattern.

API functions should avoid interpreting a result when they could return that result instead.  That means, we don't want `allowsFoo`.  Rather, we want `getFoo`.  The caller can decide whether the result implies whether things are allowed or not (or, alternatively, we can have a second function which centralizes that evaluation).

So this shouldn't be `allowsExchange`.  It should be `GetMailboxes`, and it should return the mailboxes it fetched, and an error.

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+		adapter                    = c.stable.Adapter()
+		builder                    = users.NewUserItemRequestBuilder(rawURL, adapter)
+		ErrMailBoxSettingsNotFound = clues.New("mailbox settings not found")
+		mailBoundErr               clues.Err
+	)
+
+	newItem, err := builder.Get(ctx, nil)
+	if err != nil && !(graph.IsErrAccessDenied(err) || graph.IsErrExchangeMailFolderNotFound(err)) {
+		logger.Ctx(ctx).Errorf("err getting additional data: %s", err)
+
```
ryanfkeepers, line 319: nit: space not needed for two-line blocks
```suggestion
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+		return graph.Wrap(ctx, err, "getting additional data")
+	}
+
+	if graph.IsErrAccessDenied(err) {
+		logger.Ctx(ctx).Infof("err getting additional data: access denied")
+
```
ryanfkeepers, line 325: nit: space not needed between the logger and the setter (at least, I think the linter is okay with this?)
```suggestion
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+		rawURL                     = fmt.Sprintf("https://graph.microsoft.com/v1.0/users/%s/mailboxSettings", userID)
+		adapter                    = c.stable.Adapter()
+		builder                    = users.NewUserItemRequestBuilder(rawURL, adapter)
+		ErrMailBoxSettingsNotFound = clues.New("mailbox settings not found")
+		mailBoundErr               clues.Err
+	)
+
+	newItem, err := builder.Get(ctx, nil)
+	if err != nil && !(graph.IsErrAccessDenied(err) || graph.IsErrExchangeMailFolderNotFound(err)) {
+		logger.Ctx(ctx).Errorf("err getting additional data: %s", err)
```
ryanfkeepers, line 318: Please do not format values, error or otherwise, into the log message.  Errors can use CtxErr(...), other values should be added using `.With(...)`.  Also, per logging standards, avoid phrases like `"err doing foo"` in log messages.  Just say `"doing foo"`.
```suggestion
		logger.CtxErr(ctx, err).Error("getting mailbox settings")
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+
+	return nil
+}
+
+func (c Users) getAdditionalData(ctx context.Context, userID string, mailbox *MailboxInfo) error {
+	var (
+		rawURL                     = fmt.Sprintf("https://graph.microsoft.com/v1.0/users/%s/mailboxSettings", userID)
+		adapter                    = c.stable.Adapter()
+		builder                    = users.NewUserItemRequestBuilder(rawURL, adapter)
+		ErrMailBoxSettingsNotFound = clues.New("mailbox settings not found")
```
ryanfkeepers, line 312: I'm a bit confused by this variable.  It's capitalized, as if it's exported, but only declared within the func, so also inaccessible to other code.  You probably want it declared as a global variable at the top of the file, instead of here in the function.

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+	}
+
+	return nil
+}
+
+func (c Users) getAdditionalData(ctx context.Context, userID string, mailbox *MailboxInfo) error {
+	var (
+		rawURL                     = fmt.Sprintf("https://graph.microsoft.com/v1.0/users/%s/mailboxSettings", userID)
+		adapter                    = c.stable.Adapter()
+		builder                    = users.NewUserItemRequestBuilder(rawURL, adapter)
```
ryanfkeepers, line 311: This builder is only used on line 316.  Declaring it up here makes it look like we'll be re-using it, or overriding the value.  Since neither of those are the case, it would make more sense to not declare a variable and have line 317 look something like this:

```
settings, err := users.NewUserItemRequestBuilder(rawURL, adapter).Get(ctx, nil)
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+		ErrMailBoxSettingsNotFound = clues.New("mailbox settings not found")
+		mailBoundErr               clues.Err
+	)
+
+	newItem, err := builder.Get(ctx, nil)
```
ryanfkeepers, line 316: `newItem` is generic and could represent any type of item.  We know this contains mailbox settings, so variable name should reflect that.
```suggestion
	settings, err := builder.Get(ctx, nil)
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+	if err != nil && !(graph.IsErrAccessDenied(err) || graph.IsErrExchangeMailFolderNotFound(err)) {
+		logger.Ctx(ctx).Errorf("err getting additional data: %s", err)
+
+		return graph.Wrap(ctx, err, "getting additional data")
+	}
+
+	if graph.IsErrAccessDenied(err) {
+		logger.Ctx(ctx).Infof("err getting additional data: access denied")
```
ryanfkeepers, line 324: Infof implies formatting, but nothing is getting formatted here.  Also, err implies an error occurred, but no error is attached.  It's okay to have an error produce an info log, but we still want to attach the error value.
```suggestion
		logger.CtxErr(ctx, err).Info("mailbox settings access denied")
```

src/pkg/services/m365/api/users.go
```diff
@@ -35,6 +36,50 @@ type Users struct {
...
+	Purpose                    string
+	ArchiveFolder              string
+	DateFormat                 string
+	TimeFormat                 string
+	DelegateMeetMsgDeliveryOpt string
+	Timezone                   string
+	AutomaticRepliesSetting    AutomaticRepliesSettings
+	Language                   Language
+	WorkingHours               WorkingHours
+	ErrGetMailBoxSetting       clues.Err
```
ryanfkeepers, line 54: Can this get a TODO to be removed in favor of standard error handling?

Also, this should be an `error`, not a `clues.Err`.  We're not persisting the value (you'd want a clues.ErrCore for that, anyway), therefore no concrete struct is required.  If we need to extract the clues data later on, we can still do that with a generic `error`.
```suggestion
	ErrGetMailBoxSetting       error
```

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+	if err != nil && !(graph.IsErrAccessDenied(err) || graph.IsErrExchangeMailFolderNotFound(err)) {
+		logger.Ctx(ctx).Errorf("err getting additional data: %s", err)
+
+		return graph.Wrap(ctx, err, "getting additional data")
+	}
+
+	if graph.IsErrAccessDenied(err) {
+		logger.Ctx(ctx).Infof("err getting additional data: access denied")
+
+		mailbox.ErrGetMailBoxSetting = *clues.New("access denied")
```
ryanfkeepers, line 326: reminder: drop the dereference once this variable changes to an `error` type.

src/pkg/services/m365/api/users.go
```diff
@@ -193,19 +238,196 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
+
+		mailbox.ErrGetMailBoxSetting = *ErrMailBoxSettingsNotFound
+
+		return nil
+	}
+
+	additionalData := newItem.GetAdditionalData()
+
+	mailbox.ArchiveFolder = toString(ctx, additionalData["archiveFolder"], &mailBoundErr)
+
```
ryanfkeepers, line 342: Nit: The empty lines here feel a little arbitrary.  I think you could squish it all together.
```suggestion
```

src/pkg/services/m365/api/users.go
neha-Gupta1: No. The linter is throwing an error for it.

PR #3111, opened 2023-04-12
Populate and persist OneDrive LocationRef

### LINKED ISSUE #2486: Propagate location path for all data, migrate folder details to location path.

ashmrtn:
Leverage structs from previous PRs to
populate and persist the LocationRef
for OneDrive

Right now the source of the
LocationRef is still based off the
RepoRef path, but this is in the
OneDrive code only

Manually tested
* updating a subfolder path when parent folder was moved
* populating LocationRef when the base backup didn't have LocationRefs when nothing was changed in OneDrive
* populating LocationRef when the base backup didn't have LocationRefs when something was changed in OneDrive

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #2486

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph_connector_onedrive_test.go
```diff
@@ -213,7 +213,7 @@ func (c *onedriveCollection) withFile(name string, fileData []byte, perm permDat
 func (c *onedriveCollection) withFolder(name string, perm permData) *onedriveCollection {
 	switch c.backupVersion {
```
ryanfkeepers, line 214: This switch looks like an opportunity for gotchas.  What about this instead?  This way we only need to update it in case of actual exceptions to the rule.
```go
if c.backupVersion > version.Backup || c.backupVersion < 0 {
  assert.FailNowf(...)
}

switch c.backupVersion {
case version.OneDrive1DataAndMetaFiles, 2, version.OneDrive3IsMetaMarker:
  // ...
default:
  return c
}
```

src/pkg/backup/details/details.go
```diff
@@ -864,15 +856,8 @@ func (i OneDriveInfo) Values() []string {
...
-	newParent, err := path.GetDriveFolderPath(newPath)
-	if err != nil {
-		return clues.Wrap(err, "making oneDrive path").With("path", newPath)
-	}
-
-	i.ParentPath = newParent
-
-	return nil
+func (i *OneDriveInfo) UpdateParentPath(newLocPath *path.Builder) {
+	i.ParentPath = newLocPath.PopFront().String()
```
ryanfkeepers, line 860: This makes me want for a well defined `path.Location` instead of a `path.Builder`.  Something with built-in knowledge of concepts like "this location has prefix of len 1 (alt, this location has a drive prefix), which is separate from its folders".  The builder is a good underlying data structure, but it isn't very good at communicating contextual expectations like this.  And I worry that there's footguns abound for the next contributor who doesn't have that topology already in mind.

src/internal/operations/backup.go
```diff
@@ -540,12 +540,41 @@ func getNewPathRefs(
...
 
-	// OneDrive doesn't return prefixes yet.
-	if newPrefix == nil {
-		newPrefix = &path.Builder{}
+	// Exact match for this location, no need to pull in data from the entry.
+	if newPrefix != nil {
+		// This is kind of jank cause we're in a transitionary period, but even if
+		// we're consesrvative here about marking something as updated the RepoRef
+		// comparison in the caller should catch the change. Calendars is the only
+		// on that would matter, but we should already be populating the LocationRef
```
ryanfkeepers, line 548:
```suggestion
		// exception, since it uses ids for folders, but we should already be populating the LocationRef
```

ryanfkeepers (approved):

src/pkg/backup/details/details.go
ashmrtn: mmmm I think there's a broader question here about whether we need to continue providing `ParentPath` or not. Really this information is now rolled into `LocationRef`, but we've chosen different representations for `LocationRef` and `ParentPath` (hence this fun piece of code). Really this is a `path.Builder` that was obtained by calling `newLoc.InDetails()` :(

src/internal/connector/graph_connector_onedrive_test.go
ashmrtn: this is one of the things I'm on the fence on. It's not necessarily a bad thing to make sure that people check this when bumping the version as they may have changed things that require updates to this logic

PR #3121, opened 2023-04-13
use idnamer in queryParams

ryanfkeepers:
Preparation for following changes

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/data_collections_test.go
```diff
@@ -239,8 +239,11 @@ func (suite *DataCollectionsIntegrationSuite) TestMailFetch() {
 		userID    = tester.M365UserID(suite.T())
 		users     = []string{userID}
 		acct, err = tester.NewM365Account(suite.T()).M365Config()
+		ss        = selectors.Selector{}
 	)
 
+	ss.SetDiscreteOwnerIDName(userID, userID)
```
ashmrtn, line 245: can this use different values or will that tests stuff until we switch over?

ashmrtn (approved):

ashmrtn:
just double check this doesn't change the behavior we have now please

src/internal/connector/exchange/data_collections_test.go
ryanfkeepers: It could use different values.  Also, honestly I was exhausted while writing this and wanted the fasted way to a solution.  I already have a minor refactor planned for today that should add in mocks for testing.

PR #3122, opened 2023-04-13
migrate onedrive using prefix collection

ryanfkeepers:

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/onedrive/data_collections.go
```diff
@@ -110,3 +123,45 @@ func DataCollections(
 
 	return collections, allExcludes, el.Failure()
 }
+
+// adds data migrations to the collection set.
+func migrationCollections(
```
ashmrtn, line 128: this is (probably?) going to need to work for sharepoint too

hmmm though now I see why this becomes logically complex :/
I'm a little hesitant to have ungated (by backup version or something) folder moves as I'm not sure if we'll ever want to tighten things down re: assumptions about moves (i.e. actually check there was something in the base moved when we thought there should be)

Eventually we may end up passing the previous backup version down to this function since we consume metadata, but we don't have the wiring for it right now to help gate this. Ideally, we should just design some interface that allows passing this (and other) info down to stuff. We could maybe place it on the RestoreCollection, but it feels a tad out of place there since all collections should have the same backup version

It would make me uncomfortable, but we can go ahead with this. We definitely need a ticket to track using this only when we actually need to migrate something though

src/internal/kopia/upload.go
```diff
@@ -991,6 +990,11 @@ func inflateBaseTree(
 			return clues.Wrap(err, "subtree root is not directory").WithClues(ictx)
 		}
 
+		newSubtreePath := subtreePath
```
ashmrtn, line 993: nit: add a comment that this handles migration of the prefix directories. That way the comment below that explains that it assumes no prefix directory changes makes sense (idk how I missed that earlier 🙃)

src/pkg/path/path.go
```diff
@@ -280,6 +280,31 @@ func verifyInputValues(tenant, resourceOwner string) error {
 	return nil
 }
 
+func (pb Builder) ToServicePrefix(
+	tenant, resourceOwner string,
+	s ServiceType,
+	c CategoryType,
+) (Path, error) {
```
ashmrtn, line 287: why require a builder instance at all? Seems like you could just do `path.ToServicePrefix` and save some keystrokes

src/internal/kopia/path_encoder.go
```diff
@@ -20,6 +20,22 @@ func encodeElements(elements ...string) []string {
 	return encoded
 }
 
+func decodeElements(elements ...string) []string {
+	decoded := make([]string, 0, len(elements))
+
+	for _, e := range elements {
+		bs, err := encoder.DecodeString(e)
+		if err != nil {
+			decoded = append(decoded, "error decoding: "+e)
```
ashmrtn, line 29: this should be returned as a real error. Or this should be in a test file and it should be clear it's only for test functions

ashmrtn (commented):
looks good with a few nits. Some discussion needed around how to gate this if possible though

src/pkg/path/path.go
ryanfkeepers: Goooood question.

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: > this is (probably?) going to need to work for sharepoint too

As a general statement?  Or did you have a particular case in mind?

> Eventually we may end up passing the previous backup version down to this function since we consume metadata

I was thinking of doing that for this change already.  Pushing unwarranted migrations sounds like a fantastic footgun.  Wanted to hold off for this PR and re-evaluate today/later, since it's a two-way door.

> passing this (and other) info down to stuff

I waffled on that, myself.  There's an ownership tug-of-war between backup/repo versioning/design and collection sourcing.  If we're going to keep with the collection-based migration, the clearest owner would be the package that produces those collections, and what we'd want to pass down is an indicator of which migration to perform.  Operations is agnostic to the granular details of collection data, and passing in an interface from collections/data_collections is a bit moot.

> We definitely need a ticket to track using this only when we actually need to migrate something though

I'll either follow up today or make a ticket for it by eod.

src/internal/connector/graph/collections.go
```diff
@@ -48,22 +48,28 @@ func (c emptyCollection) DoNotMergeItems() bool {
...
-		lastErr error
+		res      = []data.BackupCollection{}
+		el       = errs.Local()
+		lastErr  error
+		collKeys = map[string]struct{}{}
 	)
 
+	for _, c := range colls {
+		collKeys[c.FullPath().String()] = struct{}{}
+	}
```
ashmrtn, line 67: hmmm looking at this with fresh eyes this needs a comment and a fixup

fixup: it's not going to work for collections denoting deleted folders because they have a nil `FullPath` so I think it will NPE

comment: this will not guarantee getting all collections, only those that aren't deleted (see above). It may be less confusing to only track collections related to the prefix directories?

src/internal/connector/onedrive/data_collections.go
ashmrtn: > >this is (probably?) going to need to work for sharepoint too
>
>As a general statement? Or did you have a particular case in mind?

There's probably SharePoint backups out there that will need migrating since I believe we've pointed at least a few folks to that feature so we likely need the ability to handle that

>Wanted to hold off for this PR and re-evaluate today/later, since it's a two-way door.

are you suggesting merge this and fixup later? If we do that I think we should have a solid timeline on when "later" is. As you said, it's a fantastic footgun

src/internal/kopia/upload_test.go
```diff
@@ -183,16 +183,22 @@ func expectDirs(
...
+
+	de, err := decodeElements(ents...)
+	require.NoError(t, err, clues.ToCore(err))
+
+	if exactly {
+		require.Lenf(t, entries, len(dirs), "expected %+v\ngot %+v", dd, de)
 	}
 
-	assert.Subset(t, names, dirs)
+	assert.Subsetf(t, dirs, ents, "expected %+v\ngot %+v", dd, de)
```
ashmrtn, line 201: nit: would be nice to say something about expecting "at least" since this is a subset check

src/pkg/path/path_test.go
```diff
@@ -749,3 +749,61 @@ func (suite *PathUnitSuite) TestPath_piiHandling() {
 		})
 	}
 }
+
```
ashmrtn, line 752: can you also make a test for calling `Folders()` on a `ServicePrefix` path? I'm not sure if that will throw an index out of bounds error right now since it's kind of escaping the original assumptions of Path

ashmrtn (commented):
LGTM except for a few minor things. Need to know more about the plan re: gating this with a backup version or something

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: > There's probably SharePoint backups out there that will need migrating since I believe we've pointed at least a few folks to that feature so we likely need the ability to handle that

I think the plan for sharepoint is not to include migrations, since it's barely been released at all.

> when "later" is.

already patched: https://github.com/alcionai/corso/pull/3137

src/pkg/path/path_test.go
ryanfkeepers: good idea

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: @ashmrtn 3137 has been squashed onto this PR.

ryanfkeepers:
Side note: the details.go file is feeling a little cluttered at this point.  Would be nice to see a follow-up that splits the various structures/handlers into their own files.

ashmrtn:
just to put this here as it was surfaced offline: this needs a little bit of coordination prior to merging to make sure everyone is alright with it going in

Removing the code that actually triggers the migration (i.e. version bump plus code in the repo package I think) would allow us to merge all the supporting logic while holding off on the actual migration

src/internal/connector/onedrive/data_collections.go
ashmrtn: >I think the plan for sharepoint is not to include migrations, since it's barely been released at all.

It may be better to just plan for SP migrations and handle it if needed. That way there's no surprises down the line

src/internal/version/backup.go
```diff
@@ -36,6 +39,10 @@ const (
 	// name.
 	OneDrive6NameInMeta = 6
 
+	// All7MigrateUserPNToID marks when we migrated repo refs from the user's
+	// PrincipalName to their ID for stability.
+	All7MigrateUserPNToID = 7
```
ashmrtn, line 44: this needs updated

ashmrtn (changes requested):
putting on a block on this since the version needs updated

ryanfkeepers:
> putting on a block on this since the version needs updated

@ashmrtn please see: https://github.com/alcionai/corso/pull/3187

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: > It may be better to just plan for SP migrations and handle it if needed.

Oh, I have it all planned out.  The problem with SP, what we don't have with OD, is that SP would migrate from a value that isn't passed to the caller onto one that is.  The migration itself would be an easy lift on top of this.  It's the additional plumbing specifically to surface data for that migration which I'd like to avoid.

ryanfkeepers:
@ashmrtn all PRs are in, this should be ready for a final OK.

src/internal/connector/onedrive/collection.go
```diff
@@ -173,12 +173,69 @@ func NewCollection(
...
+
+	c.locPath = locPath
+	c.prevLocPath = prevLocPath
+
+	return c, nil
+}
+
+// NewCollectionPrefix creates a Collection prefix, without parsing a location path.
+// Should not be used unless the paths are explicitly known to be prefixes.
+func NewCollectionPrefix(
```
ashmrtn, line 196: since we switched to using the graph package's prefix collection I don't think this function should be required anymore?

src/internal/connector/onedrive/data_collections.go
```diff
@@ -110,3 +126,49 @@ func DataCollections(
...
+	svc graph.Servicer,
+	lastBackupVersion int,
+	tenant string,
+	user idname.Provider,
+	su support.StatusUpdater,
+	ctrlOpts control.Options,
+) ([]data.BackupCollection, error) {
+	// assume a version < 0 implies no prior backup, thus nothing to migrate.
+	if lastBackupVersion < 0 {
```
ashmrtn, line 140: nit: could use `version.NoBackup`?

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+			mc, err := migrationCollections(nil, test.version, "t", u, nil, control.Options{})
+			require.NoError(t, err, clues.ToCore(err))
+
+			if test.expectLen == 0 {
+				assert.Nil(t, mc)
+				return
+			}
+
+			assert.LessOrEqual(t, test.expectLen, len(mc))
```
ashmrtn, line 81: this seems like it could be simplified. It seems like it should be either 1 or 0 collections so the `LessOrEqual` check should be removed and we can directly do a `Len` check instead of checking for nil

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+			migrs := make([]bool, len(test.expectMigration))
```
ashmrtn, line 83: The below could maybe be simplified to make `migr` instances in a slice `[]migr` and then do a `Contains` check at the bottom to ensure the migration collection is in there

src/internal/connector/sharepoint/data_collections.go
```diff
@@ -117,6 +117,7 @@ func DataCollections(
 	if len(collections) > 0 {
 		baseCols, err := graph.BaseCollections(
 			ctx,
+			nil,
```
ashmrtn, line 120: I would feel better if we passed in `collections` here as expected instead of nil
```suggestion
			collections,
```

src/internal/kopia/upload_test.go
```diff
@@ -183,16 +183,22 @@ func expectDirs(
...
-	for _, e := range entries {
-		names = append(names, e.Name())
+	dd, err := decodeElements(dirs...)
+	require.NoError(t, err, clues.ToCore(err))
+
+	de, err := decodeElements(ents...)
+	require.NoError(t, err, clues.ToCore(err))
+
+	if exactly {
+		require.Lenf(t, entries, len(dirs), "expected at least %+v\ngot %+v", dd, de)
```
ashmrtn, line 198:
```suggestion
		require.Lenf(t, entries, len(dirs), "expected exactly %+v\ngot %+v", dd, de)
```

src/internal/kopia/upload_test.go
```diff
@@ -2608,3 +2617,170 @@ func (suite *HierarchyBuilderUnitSuite) TestBuildDirectoryTreeSelectsCorrectSubt
...
+			},
+		)
+	}
+
+	// Check the following:
+	//   * contacts pulled from base1 unchanged even if no collections reference
+	//     it
+	//   * email pulled from base2
+	//   * new email added
+	//   * events not pulled from base2 as it's not listed as a Reason
```
ashmrtn, line 2708: these lines don't apply to this test
```suggestion
```

src/internal/kopia/upload_test.go
```diff
@@ -2608,3 +2617,170 @@ func (suite *HierarchyBuilderUnitSuite) TestBuildDirectoryTreeSelectsCorrectSubt
...
+	//
+	// Expected output:
+	// - a-tenant
+	//   - exchange
+	//     - user1new
+	//       - email
+	//         - Inbox
+	//           - file1 <- version of data from second base
+	//           - file2
```
ashmrtn, line 2717: these need updated

src/internal/operations/backup.go
```diff
@@ -33,11 +35,12 @@ import (
...
-	ResourceOwner common.IDNamer
+	ResourceOwner idname.Provider
 
-	Results   BackupResults      `json:"results"`
-	Selectors selectors.Selector `json:"selectors"`
-	Version   string             `json:"version"`
+	Results       BackupResults      `json:"results"`
+	Selectors     selectors.Selector `json:"selectors"`
+	Version       string             `json:"version"`
+	backupVersion int
```
ashmrtn, line 43: can you make a comment saying this is mostly expected to be used for tests?

src/internal/version/backup.go
```diff
@@ -39,4 +42,8 @@ const (
 	// OneDriveXLocationRef provides LocationRef information for Exchange,
 	// OneDrive, and SharePoint libraries.
 	OneDrive7LocationRef = 7
+
+	// AllXMigrateUserPNToID marks when we migrated repo refs from the user's
+	// PrincipalName to their ID for stability.
+	AllXMigrateUserPNToID = Backup + 1
```
ashmrtn, line 48: I'm a little on the fence about what the value of this should be. Right now, my understanding is setting it to `Backup + 1` will cause the OD migration code to always create a migration collection but the paths will be the same as selectors is wired to have the UPN for both. If we switch this to `NoBackup` though then we'll have no way to test the migration code :/

I'm not sure if there's a good path forward, do you have any thoughts on this?

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,109 @@
...
+	// usable  by whichever system has ownership of it.
+	ID() string
+	// the human-readable name of the thing.
+	Name() string
+}
+
+var _ Provider = &Is{}
+
+// Is provides an id-name tuple.
+type Is struct {
```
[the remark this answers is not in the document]
ryanfkeepers: huh?  Yeah!  what gives?  Where did that commit go?

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,109 @@
...
+	IDs() []string
+	Names() []string
+	ProviderForID(id string) Provider
+	ProviderForName(id string) Provider
+}
+
+var _ Cacher = &Cache{}
+
+// Cache holds a cache of id-name mappings.
+type Cache struct {
```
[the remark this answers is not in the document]
ryanfkeepers: I swear I did... now I'm worried.

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: made a helper func for it.

src/internal/connector/onedrive/data_collections_test.go
ryanfkeepers: I'm apparently too scarred by integration tests to remember I don't need to worry about it in unit testing. (:

src/internal/connector/sharepoint/data_collections.go
ryanfkeepers: oh good catch, thanks.

src/internal/version/backup.go
ryanfkeepers: @ashmrtn see: https://github.com/alcionai/corso/pull/3195

src/internal/common/idname/idname.go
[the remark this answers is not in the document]
ryanfkeepers: 😬 yeah.  here's the [merged, unexported version](https://github.com/alcionai/corso/pull/3158/commits/6ce3577c6edcc33c1083f99a1c77fa0efd9525c1).  And here's the [version merged to this branch](https://github.com/alcionai/corso/pull/3158/files#diff-8fd9848f6b38951a8e702aa0e07e5a4e0b3d532a9685f736b900e33edc5713dbR35).

/panic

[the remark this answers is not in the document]
ryanfkeepers: https://github.com/alcionai/corso/pull/3196

src/internal/version/backup.go
ryanfkeepers: 3195 is merged

src/internal/common/idname/idname.go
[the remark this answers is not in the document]
ryanfkeepers: 3196 is merged.

I am once again asking for a final review on this PR.

ashmrtn (approved):

Issue #3129, opened 2023-04-14
onedrive Item Downloads don't middleware


ryanfkeepers:
## Issue
We've built up a decent package of graph middleware for ensuring call throttling, metrics, logging, retries, etc.  However, the itemDownloader that uses a custom http client instead of the graph client avoids all that, and only includes a minimal retry handler.

## Acceptance Criteria
All http client requests against graph api should share in the middleware we've created.

PR #3131, opened 2023-04-14
fix up item download retry handling, add tests

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3129

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/api/mock/mail.go
```diff
@@ -1,36 +1,21 @@
...
 import (
-	"github.com/alcionai/clues"
-
 	"github.com/alcionai/corso/src/internal/connector/exchange/api"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/graph/mock"
 	"github.com/alcionai/corso/src/pkg/account"
 )
 
-func NewService(creds account.M365Config, opts ...graph.Option) (*graph.Service, error) {
```
meain, line 12: This was a new version of [`api.NewService`](https://github.com/alcionai/corso/blob/3e0f81dddbf66a7ef5d8901f21d2ed1676fe16dd/src/internal/connector/exchange/api/api.go#L91) and that is why I left it here instead of moving to `graph`

src/internal/connector/onedrive/collection_test.go
```diff
@@ -717,3 +718,103 @@ func (suite *GetDriveItemUnitTestSuite) TestGetDriveItemError() {
...
+	var (
+		svc     graph.Servicer
+		gr      graph.Requester
+		driveID string
+		iorc    = io.NopCloser(bytes.NewReader([]byte("fnords")))
+		item    = &models.DriveItem{}
+		itemWID = &models.DriveItem{}
+	)
+
+	itemWID.SetId(ptr.To("brainhooldy"))
```
meain, line 732: Interesting id 😅

src/internal/connector/onedrive/collection.go
```diff
@@ -390,6 +375,42 @@ func (oc *Collection) getDriveItemContent(
...
+	// Assume unauthorized requests are a sign of an expired jwt
+	// token, and that we've overrun the available window to
+	// download the actual file.  Re-downloading the item will
+	// refresh that download url.
+	di, err := igf(ctx, svc, driveID, ptr.Val(item.GetId()))
+	if err != nil {
+		return nil, clues.Wrap(err, "retrieving expired item")
+	}
+
+	_, content, err = irf(ctx, gr, di)
```
meain, line 406: Previously we used to try thrice to download expired content. Should we retain that behaviour instead of trying once.

meain (approved):
Approving to unblock, have a question on download retries.

src/internal/connector/exchange/api/mock/mail.go
ryanfkeepers: That may be the intent, but the code doesn't follow up.  What's implemented here is simply a graph service.  For that matter, now that I'm looking at it again, it isn't even a mock. It's just a regular graph.Service.

src/internal/connector/onedrive/collection.go
meain: The middleware does retry things, but it does not retry link expiring(and retrying using new link).

ryanfkeepers: I don't believe the link can expire more than once.  Since we regenerate and consume the link at the same time, the second one won't have any delay before its usage.

PR #3137, opened 2023-04-17
check backup version on onedrive migration

ryanfkeepers:
pre-fetches the prior backups from the manifests and passes down the oldest version into the backup producer as a way to check whether a migration is necessary or not.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -37,6 +37,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 	owner common.IDNamer,
 	sels selectors.Selector,
 	metadata []data.RestoreCollection,
+	priorVersion int,
 	ctrlOpts control.Options,
 	errs *fault.Bus,
 ) ([]data.BackupCollection, map[string]map[string]struct{}, error) {
@@ -99,6 +100,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 			sels,
 			sels,
 			metadata,
+			priorVersion,
 			gc.credentials.AzureTenantID,
 			gc.itemClient,
 			gc.Service,
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -37,7 +37,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 	owner common.IDNamer,
 	sels selectors.Selector,
 	metadata []data.RestoreCollection,
-	priorVersion int,
+	lastBackupVersion int,
 	ctrlOpts control.Options,
 	errs *fault.Bus,
 ) ([]data.BackupCollection, map[string]map[string]struct{}, error) {
@@ -100,7 +100,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 			sels,
 			sels,
 			metadata,
-			priorVersion,
+			lastBackupVersion,
 			gc.credentials.AzureTenantID,
 			gc.itemClient,
 			gc.Service,
--- a/src/internal/connector/onedrive/data_collections.go
+++ b/src/internal/connector/onedrive/data_collections.go
@@ -11,6 +11,7 @@ import (
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/internal/version"
 	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/logger"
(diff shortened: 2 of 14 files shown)


src/internal/connector/data_collections.go
```diff
@@ -37,6 +37,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 	owner common.IDNamer,
 	sels selectors.Selector,
 	metadata []data.RestoreCollection,
+	priorVersion int,
```
ashmrtn, line 40: nit: maybe either a comment on what this is or rename to `minMetadataVersion`? Longer term this does open up the question of if the version should be associated with the collections themselves as now that you've written the description it's possible we may treat different versions differently

src/internal/connector/onedrive/data_collections.go
```diff
@@ -127,11 +130,20 @@ func DataCollections(
 // adds data migrations to the collection set.
 func migrationCollections(
 	svc graph.Servicer,
+	priorVersion int,
 	tenant string,
 	user common.IDNamer,
 	su support.StatusUpdater,
 	ctrlOpts control.Options,
 ) ([]data.BackupCollection, error) {
+	if priorVersion < 1 {
```
ashmrtn, line 139: this seems a bit arbitrary. Can you add a comment for why it's v1? It seems like persisting metadata was enabled prior to versioning so in theory there's v0 backups that could be selected successfully as a base

Though on that note, I think there's a bug in this code. We should only allow reusing the delta token in the version is high enough to have files using IDs instead of names... going to make a ticket for that...

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -0,0 +1,113 @@
...
+				return
+			}
+
+			assert.LessOrEqual(t, test.expectLen, len(mc))
+
+			migrs := make([]bool, len(test.expectMigration))
+
+			for _, col := range mc {
+				oc, ok := col.(*Collection)
+				require.True(t, ok, "casting backupCollection iface to *onedrive.Collection")
```
ashmrtn, line 87: do you need to cast it? You should be able to access `FullPath` and `PrevPath` through the regular `BackupCollection` interface

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+				if oc.FullPath() != nil {
+					fp = oc.FullPath().String()
+				}
+
+				if oc.prevPath != nil {
+					pp = oc.prevPath.String()
+				}
+
+				t.Logf("Found migration collection:\n* full: %s\n* prev: %s\n", fp, pp)
```
ashmrtn, line 99: can be removed?

src/internal/version/backup.go
```diff
@@ -36,6 +36,10 @@ const (
 	// name.
 	OneDrive6NameInMeta = 6
 
+	// OneDrive7MigrateUserPNToID marks when we migrated repo refs from the user's
+	// PrincipalName to their ID for stability.
+	OneDrive7MigrateUserPNToID = 7
```
ashmrtn, line 41: nit: this change isn't just for OneDrive, it's for all services. It's just OneDrive is the only one that needs special handling. Not sure what prefix we should use besides OneDrive though
```suggestion
	All7MigrateUserPNToID = 7
```

ashmrtn (approved):

src/internal/connector/data_collections.go
ryanfkeepers: Going to rename this to `priorBackupVersion`, since it isn't specific for tracking metadata.

src/internal/connector/onedrive/data_collections.go
ryanfkeepers: I think your mind is on a different topic.  This isn't for metadata handling, specifically.  Per migrations, I'm using < 1 to interpret `no prior backup, nothing to migrate`.  (will add that as a comment)

src/internal/connector/onedrive/data_collections_test.go
ryanfkeepers: Oh!  I forgot that PrevPath was part of the interface.

src/internal/connector/onedrive/data_collections_test.go
ryanfkeepers: Nah, keep it.  Otherwise we don't list what collections we did find, which makes this difficult to debug.

src/internal/version/backup.go
ryanfkeepers: Yeah, I waffle between naming them for who is checking it, versus who it applies to.  Ah well, it's all fast and loose anyway.

src/internal/connector/onedrive/data_collections.go
ashmrtn: >This isn't for metadata handling, specifically. Per migrations, I'm using < 1 to interpret no prior backup, nothing to migrate. (will add that as a comment)

if it's tied to the backup version though it's possible to have a backup of v0 or v1 that will be selected as a base for incrementals. In those cases we still need to migrate the subtree. There's other reasons why we shouldn't be using those as a base, but it's still possible to select them

PR #3140, opened 2023-04-17
Redo folder generation in backup details

### LINKED ISSUE #2138: Optimize adding folders to backup details
Now that the helper function for getting folders to add to backup details is in the details package it can be merged with actually updating the folder items. This can help reduce the number of items allocated and skip repeated work

### LINKED ISSUE #3120: use LocationRef in folder details entries
Items now have a LocationRef and a RepoRef. However, the generated folder entries are still using RepoRef and that can break consumers if the underlying storage layout changes in kopia.

Going forward, folder entries should be generated using LocationRef, with some additional data to ensure uniqueness:
* the data category should be included in the folder entry (e.x. contacts, email, etc)
* drive ID and drive name should be included for OneDrive and SharePoint items
* future items that have elements similar to the drive name/ID should add those as fields to the folder struct

ashmrtn:
Augment the folder backup details entries with
LocationRef and some information about what
data type generated the entry. Also add top-level
container information like drive name/ID if
applicable

Refactor code to do folder generation to make it
more contained and require less parameters to add
entries to backup details

Important changes are in details.go. All other
changes are just to keep up with the slightly
modified function API or update tests

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3120
* closes #2138

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/upload_test.go
```diff
@@ -521,71 +536,6 @@ func (suite *CorsoProgressUnitSuite) TestFinishedFileCachedNoPrevPathErrors() {
 	assert.Error(t, cp.errs.Failure(), clues.ToCore(cp.errs.Failure()))
 }
 
-func (suite *CorsoProgressUnitSuite) TestFinishedFileBuildsHierarchyNewItem() {
```
ashmrtn, line 524: generating folders is now checked in details_test.go

src/pkg/backup/details/details.go
```diff
@@ -682,11 +705,30 @@ func (i ItemInfo) uniqueLocation(baseLoc *path.Builder) (LocationIDer, error) {
...
+}
+
 type FolderInfo struct {
 	ItemType    ItemType  `json:"itemType,omitempty"`
 	DisplayName string    `json:"displayName"`
 	Modified    time.Time `json:"modified,omitempty"`
 	Size        int64     `json:"size,omitempty"`
+	DataType    ItemType  `json:"dataType,omitempty"`
+	DriveName   string    `json:"driveName,omitempty"`
+	DriveID     string    `json:"driveID,omitempty"`
```
ryanfkeepers, line 731: This is a bit suspect, since not all folders are children of a drive.  I could see renaming it for something more generic, that expresses container of some kind between the prefix and folders/items, which would need to be interpreted on a case-by-case basis (eg: if sharepoint, the container is a drive).  Or, if more detail is preferred, adding a service subtyping so that folderInfo contains an `ItemInfo` which has the drive ID.

src/pkg/backup/details/details_test.go
```diff
@@ -173,6 +173,544 @@ func (suite *DetailsUnitSuite) TestDetailsEntry_HeadersValues() {
...
+					e.Updated = true
+
+					res = append(res, e)
+				}
+
+				return res
+			},
+		},
+		{
+			name: "Two None Updated Exchange Email And Contact",
```
ryanfkeepers, line 551:
```suggestion
			name: "Two Non-Updated Exchange Email And Contact",
```

src/pkg/backup/details/details_test.go
```diff
@@ -173,6 +173,544 @@ func (suite *DetailsUnitSuite) TestDetailsEntry_HeadersValues() {
...
+					e.Folder.Modified = exchange2.Exchange.Modified
+
+					res = append(res, e)
+				}
+
+				return res
+			},
+		},
+		{
+			name: "Two One Updated Exchange Emails",
```
ryanfkeepers, line 517: two one?

src/pkg/backup/details/details_test.go
```diff
@@ -173,6 +173,544 @@ func (suite *DetailsUnitSuite) TestDetailsEntry_HeadersValues() {
...
+func (suite *DetailsUnitSuite) TestDetailsAdd_LocationFolders() {
+	t := suite.T()
+
+	exchange1 := exchangeEntry(t, "foo1", 42, ExchangeMail)
+	oneDrive1 := oneDriveishEntry(t, "foo1", 42, OneDriveItem)
+	sharePoint1 := oneDriveishEntry(t, "foo1", 42, SharePointLibrary)
+	sharePointLegacy1 := oneDriveishEntry(t, "foo1", 42, SharePointLibrary)
+	sharePointLegacy1.SharePoint.ItemType = OneDriveItem
+
+	time.Sleep(1)
```
ryanfkeepers, line 341: What's going on here?

src/pkg/backup/details/details_test.go
```diff
@@ -173,6 +173,544 @@ func (suite *DetailsUnitSuite) TestDetailsEntry_HeadersValues() {
...
+
+	exchange2 := exchangeEntry(t, "foo2", 43, ExchangeMail)
+	exchange3 := exchangeEntry(t, "foo2", 43, ExchangeContact)
```
ryanfkeepers, line 344: For ease of reading and quick comparison, it would be nice for multi-typed services like exchange to contain type refs in the var names. 
```suggestion
	exchMail2 := exchangeEntry(t, "foo2", 43, ExchangeMail)
	exchContact1 := exchangeEntry(t, "foo2", 43, ExchangeContact)
```

ryanfkeepers (commented):
generally looks good. some concerns around details folder design.

src/internal/operations/backup_test.go
```diff
@@ -267,23 +267,25 @@ func makeMetadataPath(
...
+	dt details.ItemType,
 ) *details.DetailsEntry {
 	t.Helper()
 
 	return &details.DetailsEntry{
 		RepoRef:     pb.String(),
 		ShortRef:    pb.ShortRef(),
 		ParentRef:   pb.Dir().ShortRef(),
-		LocationRef: pb.PopFront().PopFront().PopFront().PopFront().Dir().String(),
+		LocationRef: loc.Dir().String(),
```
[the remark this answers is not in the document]
ashmrtn: no. I think we'll probably be able to remove parentRef at some point as it probably won't be needed in the future

src/pkg/repository/repository_unexported_test.go
```diff
@@ -411,15 +417,19 @@ func (suite *RepositoryModelIntgSuite) TestGetBackupErrors() {
...
-			Folder: &details.FolderInfo{
-				DisplayName: "test",
+			Exchange: &details.ExchangeInfo{
+				ItemType: details.ExchangeMail,
 			},
 		}
 	)
 
-	builder := &details.Builder{}
+	repoPath, err2 := path.FromDataLayerPath(tenantID+"/exchange/user-id/email/test/foo", true)
```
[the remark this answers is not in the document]
ashmrtn: `err`'s type is `*clues.Err` so I can't reuse it

src/pkg/backup/details/details.go
```diff
@@ -340,29 +374,29 @@ type Details struct {
 }
 
 func (d *Details) add(
-	repoRef, shortRef, parentRef, locationRef string,
+	repoRef path.Path,
+	locationRef *path.Builder,
```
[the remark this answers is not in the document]
ashmrtn: we are not using the location to generate shortrefs at the moment and I don't think we can until SDK consumers change their behavior. Otherwise we'll cause breakages by changing what shortref represents when the item's path didn't change

There's also not really much of an interface right now. We can debate about what it should have in the long run, but this PR is already large enough that I don't want to add more to it

src/pkg/backup/details/details.go
ashmrtn: >I could see renaming it for something more generic, that expresses container of some kind between the prefix and folders/items

I debated about this with someone else and I'm not real happy with it either. However, it is the most flexible going forward as not populating fields won't cause JSON bloat thanks to `omitempty`. Reusing fields between different data types also doesn't guarantee that we won't need more fields in the future to cover similarly positioned (in the hierarchy) containers. Reusing fields also makes it difficult to determine what they actually mean since the name may not match what it's representing

>adding a service subtyping

service subtyping can be done by examining `DataType`

> folderInfo contains an ItemInfo which has the drive ID

folderInfo is already itself contained within ItemInfo. Additional nesting seems like it would just complicate stuff further

src/pkg/backup/details/details_test.go
ashmrtn: wanted later mod times. Sure they'll always compare later just by virtue of getting the `1`s first, but nice to have more than a few milliseconds (hopefully) difference

src/pkg/backup/details/details_test.go
```diff
@@ -173,6 +173,544 @@ func (suite *DetailsUnitSuite) TestDetailsEntry_HeadersValues() {
...
+
+	exchange1 := exchangeEntry(t, "foo1", 42, ExchangeMail)
+	oneDrive1 := oneDriveishEntry(t, "foo1", 42, OneDriveItem)
+	sharePoint1 := oneDriveishEntry(t, "foo1", 42, SharePointLibrary)
+	sharePointLegacy1 := oneDriveishEntry(t, "foo1", 42, SharePointLibrary)
+	sharePointLegacy1.SharePoint.ItemType = OneDriveItem
+
+	time.Sleep(1)
+
+	exchange2 := exchangeEntry(t, "foo2", 43, ExchangeMail)
```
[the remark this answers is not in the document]
ashmrtn: resource owner is not modified. It does create a separate resource ID for the object which yes, is intentional so we can test setting the mod time to the most recent value when generating folders

src/pkg/backup/details/details_test.go
ryanfkeepers: could use a comment with that explanation.  This looks unwanted, or possibly vestigial, otherwise.

src/pkg/backup/details/details_test.go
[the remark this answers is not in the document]
ryanfkeepers: mmkay might also want a comment here, so that expectation is maintained.

src/pkg/backup/details/details.go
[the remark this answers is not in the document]
ryanfkeepers: > we are not using the location to generate shortrefs

Doh, my bad.  I misread the line when I glanced at it.  s/shortReffer/stringer/

> There's also not really much of an interface right now. We can debate about what it should have in the long run, but this PR is already large enough that I don't want to add more to it

For sure, this isn't a current-PR concern.  Per "not much in the interface", that's precisely what I'd hope to make clear.  Give consumers clear understanding that there is not a prefix, item, etc.

src/pkg/backup/details/details.go
ryanfkeepers: > folderInfo is already itself contained within ItemInfo

Ah, my brain had that swapped around, and I was still reading this from the perspective of folder entries, not info.

ryanfkeepers (approved):

PR #3151, opened 2023-04-17
Restore with OneDrive/SharePoint ID

### LINKED ISSUE #3142: Allow restoring OneDrive files by M365 ID
Some upstream consumers want to restore items by M365 ID. OneDrive currently doesn't allow this, but we should re-enable it and add a note to known issues that it could cause unexpected file restores on the CLI

ashmrtn:
Allow restoring OneDrive and SharePoint items by
M365 ID as well as file name and ShortRef

Does create an edgecase where attempting to
restore an item named the same as another item's
M365 ID causes both items to be restored

Moves the consts for OneDrive file suffixes to a
different package so that the selectors package
can also import them without a cycle

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3142

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/common/consts.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "strings"
+
+const (
+	MetaFileSuffix    = ".meta"
+	DirMetaFileSuffix = ".dirmeta"
+	DataFileSuffix    = ".data"
+)
+
+func IsMetaFile(name string) bool {
```
ryanfkeepers, line 11: This feels more like a `hasMetaSuffix` than an `isMetaFile` check.   To avoid gotchas, I think we should move the graph/metadata func into this file and have a singular owner, rather than split ownership.

src/internal/connector/onedrive/common/consts.go
```diff
@@ -0,0 +1,13 @@
+package common
```
ryanfkeepers, line 1: I'd like us to minimize the notion of a "common" package: the naming isn't helpful, the packages tend to bloat, and that all results in collisions (see: `odcommon` aliased imports).  Thoughts on moving this into `internal/common/metadata` instead?  The functionality will overextend onedrive soon (sharepoint), and potentially again in the future.  Plus, that move would jive well with the other suggestion for moving the logic out of graph, too.

ryanfkeepers (approved):
behavior looks good.  Just nits about package location/naming.

src/internal/connector/onedrive/common/consts.go
ashmrtn: I'm not really sure it belongs outside of the OneDrive tree though. No other service outside of OneDrive stuff really has the notion of metadata. And really nothing besides OneDrive and tests need this. We have it in details and selectors as a hack to work around how things are now or some old ways of doing things, but I'm not really sure yet if we have a stable idea of what "metadata" is or isn't

ryanfkeepers: That's all fine, I'm not terribly concerned with the location, so much as being nitpicky about naming and not spreading logic between two packages.  `/connector/onedrive/metadata`, maybe?

src/internal/connector/onedrive/common/consts.go
ashmrtn: the one in `graph/metadata` is used by kopia because we wanted to filter out metadata files from details when merging for incrementals. We could probably scrap that code (in another PR) and remove that function completely if we wanted to. Unfortunately though, we'll probably always need something that filters out metadata entries from backup details because folks can still download old releases that had that in it

ryanfkeepers: So, contrary to the other comment, that does sound like a service outside of onedrive has the notion of metadata.

ryanfkeepers: > Again, there's also nothing stopping folks from downloading an old corso binary that generates metadata details entries and running that

We should keep these things in mind when we look toward a v1 release.  That will allow us some leeway in dictating what is and isn't portable across versions.

PR #3155, opened 2023-04-18
send notification on action failure

neha-Gupta1:
send a Slack notification if Github Actions fails

- [ ] :no_entry: No

#### Type of change

#### Test Plan

- [ ] :muscle: Manual


meain:
I've triggered a manual run on this branch at https://github.com/alcionai/corso/actions/runs/4729929001 just to see how this works.

.github/workflows/sanity-test.yaml
```diff
@@ -364,4 +364,25 @@ jobs:
           path: src/testlog/*
           if-no-files-found: error
           retention-days: 14
-      
\ No newline at end of file
+      
+      - name: Send Github Action failure to Slack
```
meain, line 368: Can we try a manual run for this to validate the integration?

neha-Gupta1: We need to add slack-webhook in the secrets and then can test it. I will let you know once its added.

ryanfkeepers: Could we add a similar action to the cleanup scripts?  Would be nice to know when those fail, too.  Thanks!

vkamra: @neha-Gupta1 - created SLACK_WEBHOOK_URL secret

neha-Gupta1: As Vaibhav mentioned the secret is added and I have done some testing too. Its ready for review now.

ryanfkeepers (approved):

PR #3157, opened 2023-04-18
Add immutable ID header to Get requests

### LINKED ISSUE #2861: switch to using stable IDs for Exchange items
Using stable IDs from M365 for Exchange items will help ensure items don't disappear during a backup. With the current implementation and the proposal to use a set of item IDs as a sort of "set of things to fetch," if an item is moved between folders between the time the item's ID is recorded and the time Corso goes to fetch the item then it will be excluded from the backup (marked as not found by Graph as the ID changed during the move)

Switching to stable IDs from M365 shields us from that by ensuring we'll only get item not found errors if the item was deleted. Although the data for the item we pull may not be consistent with the location of the item (i.e. parent ID reference may be off), at least we will ensure the item was actually backed up rather than not appearing in the backup at all

ashmrtn:
This allows using immutable IDs for Exchange objects. It does not add this header to put requests as we don't record the IDs of restored objects anywhere.

May slightly increase performance of events backups as
it now sets the delta query page size to 200 like with
other requests

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2861

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cmd/getM365/exchange/get_item.go
+++ b/src/cmd/getM365/exchange/get_item.go
@@ -98,11 +98,11 @@ func runDisplayM365JSON(
 
 	switch cat {
 	case path.EmailCategory:
-		bs, err = getItem(ctx, ac.Mail(), user, itemID, errs)
+		bs, err = getItem(ctx, ac.Mail(), user, itemID, true, errs)
 	case path.EventsCategory:
-		bs, err = getItem(ctx, ac.Events(), user, itemID, errs)
+		bs, err = getItem(ctx, ac.Events(), user, itemID, true, errs)
 	case path.ContactsCategory:
-		bs, err = getItem(ctx, ac.Contacts(), user, itemID, errs)
+		bs, err = getItem(ctx, ac.Contacts(), user, itemID, true, errs)
 	default:
 		return fmt.Errorf("unable to process category: %s", cat)
 	}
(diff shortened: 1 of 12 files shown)


ashmrtn:
still need a way to force a full backup with this change (see linked issue comments for why we need a full backup or extra work)

vkamra:
> still need a way to force a full backup with this change (see linked issue comments for why we need a full backup or extra work)

Can we enable this only for newer repositories? Is there a mechanism for us to detect a repository created after this change was introduced? If not - we should add that.

vkamra:
> > still need a way to force a full backup with this change (see linked issue comments for why we need a full backup or extra work)
> 
> Can we enable this only for newer repositories? Is there a mechanism for us to detect a repository created after this change was introduced? If not - we should add that.

@ashmrtn @ryanfkeepers - here's what I was thinking - https://github.com/alcionai/corso/pull/3161

ashmrtn:
@vkamra @ryanfkeepers this should be ready for review. 

I'v added a non-user-facing feature flag that can be used to fetch immutable IDs. Note that right now it's only safe to enable this flag if either of the following are true
* this will be a full backup
* this will be an incremental backup and the base backup also used immutable IDs

Manual testing showed we got mutable IDs for a backup and an incremental on top of it with the flag set to false (no changes beyond this PR). Forcing the flag to true (by setting in the Default() options c'tor) caused immutable IDs to be used for a backup and an incremental on top of it

ryanfkeepers (approved):

PR #3158, opened 2023-04-18
cleanup idnamer

ryanfkeepers:
Tech debt refactor

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"strings"
+
+	"golang.org/x/exp/maps"
+)
+
+type Provider interface {
+	// the canonical id of the thing, generated and usable
+	//  by whichever system has ownership of it.
```
ashmrtn, line 11: nit
```suggestion
	// ID returns the canonical id of the thing, generated and usable
	// by whichever system has ownership of it.
```

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,89 @@
+package idname
+
+import (
+	"strings"
+
+	"golang.org/x/exp/maps"
+)
+
+type Provider interface {
```
ashmrtn, line 9: since some of the functions below return `&Is{}` this should state what the behavior of the nil/empty `Provider` is

src/pkg/repository/repository.go
```diff
@@ -329,7 +330,7 @@ func (r repository) NewBackupWithLookup(
 		gc,
 		r.Account,
 		sel,
-		sel,
+		inMock.NewProvider(ownerID, ownerName),
```
ashmrtn, line 333: this shouldn't be using the mock since it's prod code

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,89 @@
...
+	Name() string
+}
+
+var _ Provider = &Is{}
+
+// Is provides an id-name tuple.
+type Is struct {
+	IDV   string
+	NameV string
+}
```
ashmrtn, line 23: this doesn't seem to be used outside this package so it and all its fields should be unexported

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,89 @@
...
+	ProviderForName(id string) Provider
+}
+
+var _ Cacher = &Cache{}
+
+// Cache holds a cache of id-name mappings.
+type Cache struct {
+	IDToName map[string]string
+	NameToID map[string]string
+}
```
ashmrtn, line 43: I'd really feel more comfortable if we didn't export this or any of its fields

ashmrtn (approved):

src/internal/common/idname/idname.go
ryanfkeepers: ah, had a moment where it did need an export.  Good catch.

src/pkg/repository/repository.go
ryanfkeepers: oh, wow, how did I miss that?

src/internal/common/idname/idname.go
ryanfkeepers: sure, can do.  Forgot I'd meant to give that a constructor.

src/internal/common/idname/idname.go
```diff
@@ -0,0 +1,109 @@
...
+	// usable  by whichever system has ownership of it.
+	ID() string
+	// the human-readable name of the thing.
+	Name() string
+}
+
+var _ Provider = &Is{}
+
+// Is provides an id-name tuple.
+type Is struct {
```
ashmrtn, line 27: I think this struct is not referenced outside this package and could be unexported as well?

ashmrtn (approved):

PR #3160, opened 2023-04-18
add itemref to details entry

ryanfkeepers:
Adds an new details entry field: itemRef.
This holds a stable, semi-unique identifier to the item represented by that entry.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3027

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -903,8 +903,7 @@ func traverseBaseDir(
 			oldDirPath,
 			currentPath,
 			dEntry,
-			roots,
-		)
+			roots)
 	})
 	if err != nil {
 		return clues.Wrap(err, "traversing base directory")
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -388,6 +388,7 @@ func (d *Details) add(
 		ShortRef:    repoRef.ShortRef(),
 		ParentRef:   repoRef.ToBuilder().Dir().ShortRef(),
 		LocationRef: locationRef.String(),
+		ItemRef:     repoRef.Item(),
 		Updated:     updated,
 		ItemInfo:    info,
 	}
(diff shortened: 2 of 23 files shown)


src/internal/kopia/upload.go
```diff
@@ -216,6 +216,7 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 		d.repoPath.String(),
 		d.repoPath.ShortRef(),
 		parent.ShortRef(),
+		d.repoPath.Item(),
```
ashmrtn, line 219: I don't think this will do quite what we want for all types. For OneDrive and SharePoint I think it will end up with the `.data` suffix on them because that's part of the item's name. It seems like at least for now we should try to shuffle this into service-specific code. Either the Stream should have another function to get this or details.go should be extracting it in a service-specific manner. The former is probably a bit more flexible (or at least closer to what's actually "generating" the ID?)

src/pkg/backup/details/details.go
```diff
@@ -388,6 +388,7 @@ func (d *Details) add(
 		ShortRef:    repoRef.ShortRef(),
 		ParentRef:   repoRef.ToBuilder().Dir().ShortRef(),
 		LocationRef: locationRef.String(),
+		ItemRef:     repoRef.Item(),
```
ashmrtn, line 391: I think this is still going to generate ItemRefs with suffixes for OD/SP (e.x. `<item ID>.data`). I thought we were aiming for just the item ID without any suffixes?

src/internal/kopia/upload.go
ryanfkeepers: fixed in https://github.com/alcionai/corso/pull/3174

ashmrtn (approved):

PR #3164, opened 2023-04-18
configure selector to prefer item name for reduce

ryanfkeepers:
add selector configuration, beginning with a toggle that tells pathValues to prefer item names (where available) instead of the item ID.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3027

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/onedrive_test.go
+++ b/src/pkg/selectors/onedrive_test.go
@@ -173,6 +173,7 @@ func (suite *OneDriveSelectorSuite) TestOneDriveRestore_Reduce() {
 			Entries: []details.DetailsEntry{
 				{
 					RepoRef: file,
+					ItemRef: "file",
 					ItemInfo: details.ItemInfo{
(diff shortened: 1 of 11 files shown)


src/pkg/selectors/onedrive_test.go
```diff
@@ -211,36 +211,69 @@ func (suite *OneDriveSelectorSuite) TestOneDriveRestore_Reduce() {
 		deets        *details.Details
```
ashmrtn, line 211: should deets (above that I can't comment on directly) be populating ItemRef since that's a fallback in the code?

ashmrtn (approved):

src/pkg/selectors/onedrive_test.go
ryanfkeepers: yeah, probably should.

PR #3174, opened 2023-04-19
clean metadata suffix from itemRef

ryanfkeepers:
Ensures that all added details entries have the metadata suffix removed from the itemRef, for any service that applies a metadata suffix to the items.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3027

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/backup/details/details_test.go
+++ b/src/pkg/backup/details/details_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"strings"
 	"testing"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/stretchr/testify/suite"
 
 	"github.com/alcionai/corso/src/internal/common"
+	"github.com/alcionai/corso/src/internal/connector/onedrive/metadata"
 	"github.com/alcionai/corso/src/internal/tester"
 	"github.com/alcionai/corso/src/internal/version"
 	"github.com/alcionai/corso/src/pkg/path"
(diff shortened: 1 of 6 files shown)


src/pkg/backup/details/details_test.go
```diff
@@ -987,6 +993,45 @@ func (suite *DetailsUnitSuite) TestDetails_Add_ShortRefs_Unique_From_Folder() {
...
+	err = b.Add(dirMetaSfx, loc, false, info)
+	require.NoError(t, err, clues.ToCore(err))
+
+	err = b.Add(metaSfx, loc, false, info)
+	require.NoError(t, err, clues.ToCore(err))
+
+	for _, ent := range b.Details().Items() {
+		assert.False(t, strings.HasSuffix(ent.ItemRef, metadata.DirMetaFileSuffix))
+		assert.False(t, strings.HasSuffix(ent.ItemRef, metadata.MetaFileSuffix))
+		assert.False(t, strings.HasSuffix(ent.ItemRef, metadata.DataFileSuffix))
```
ashmrtn, line 1031: nit: in theory this could use the metadata package's `HasMetaSuffix` iirc

ashmrtn (approved):

src/pkg/backup/details/details_test.go
ryanfkeepers: could, but I believe we're safe with no-ops here.

PR #3182, opened 2023-04-20
Add concurrency limiter middleware package

pandeyabs:
Graph service only allows [4 concurrent requests per exchange mailbox.](https://learn.microsoft.com/en-us/graph/throttling-limits#outlook-service-limits) We are currently not honoring this limit in corso. This causes 429 errors with “Application is over its MailboxConcurrency limit”.  This PR introduces a concurrency limiter middleware. This middleware is selectively added for exchange backups only. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/exchange_data_collection.go
pandeyabs, line 1: Please ignore changes to this file. I'll put the semaphore logic back on here. Reason is that i am going to control the concurrency limiter with a kill flag, and exposing the kill flag to this file would be an unnecessary short term change.

pandeyabs: Having both concurrency limiter and the semaphore logic in streamItems isn't going to impact performance.

src/cli/backup/exchange.go
```diff
@@ -147,6 +148,9 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	// TODO: Add hidden flag to disable this feature
+	graph.InitializeConcurrencyLimiter(options.Control().ItemFetchParallelism)
```
ryanfkeepers, line 152: suggestion: move this into https://github.com/alcionai/corso/blob/main/src/internal/connector/exchange/data_collections.go.  I don't think we'll need it established at a higher level than that.

src/internal/connector/graph/service.go
```diff
@@ -253,6 +253,14 @@ func GetMiddlewares(maxRetry int, delay time.Duration) []khttp.Middleware {
 		&ThrottleControlMiddleware{},
 		&MetricsMiddleware{},
 	}
+
+	// Optionally add concurrency limiter middleware if it has been initialized
+	cl := GetConcurrencyLimiter()
```
ryanfkeepers, line 258: nit: this should be moved to the top.  It feels a bit out of order to save a prepend for the last step.  Something like:
```go
func GetMiddlewares(...) {
  mw := []khttp.Middleware{}

  if cl != nil {
    mw = append(mw, cl)
  }

  mw = append(mw, theRest...)
  return mw
}
```

src/internal/connector/exchange/exchange_data_collection.go
ashmrtn: (I've vaguely been following this every now and then just to keep up with the state of the codebase)

As an aside, it may actually be good to keep some sort of semaphore in this file as it can help us from creating too many goroutines at once. It won't limit the total number of goroutines across all folders being backed up (not something we need to tackle right now imo), but it will help ensure a folder with 10k emails doesn't spawn 10k goroutines right off the bat

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,52 @@
...
+	"net/http"
+	"sync"
+
+	khttp "github.com/microsoft/kiota-http-go"
+)
+
+// ConcurrencyLimiter middleware limits the number of concurrent requests to graph API.
+// Currently we only enforce concurrency limit for exchange
+type ConcurrencyLimiter struct {
+	semaphoreExch chan struct{}
```
ryanfkeepers, line 13: Corso can only run a single service at a time.  I don't think you need to namespace exchange for this case.  Lets name it `semaphore`, and we can namespace further as refinements arrive.

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+var (
+	once               sync.Once
+	concurrencyLimiter *ConcurrencyLimiter
+	// Outlooks expects max 4 concurrent requests
+	// https://learn.microsoft.com/en-us/graph/throttling-limits#outlook-service-limits
+	maxConcurrentRequestsExchange = 4
```
ryanfkeepers, line 21: Not much point in calling out exchange here unless you're going to take in a service identifier as part of the `initialize` func.  Otherwise the correlation between the two is purely in cognitive load, and not in code.

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"net/http"
+	"sync"
+
+	khttp "github.com/microsoft/kiota-http-go"
+)
+
+// ConcurrencyLimiter middleware limits the number of concurrent requests to graph API.
+// Currently we only enforce concurrency limit for exchange
+type ConcurrencyLimiter struct {
```
ryanfkeepers, line 12: This doesn't need to be exported, unless we plan for external packages to compose their own ConcurrencyLimiter struct, which sounds like something we'd rather avoid.

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,52 @@
...
+func GetConcurrencyLimiter() *ConcurrencyLimiter {
+	return concurrencyLimiter
+}
+
+func (cl *ConcurrencyLimiter) Intercept(
+	pipeline khttp.Pipeline,
+	middlewareIndex int,
+	req *http.Request,
+) (*http.Response, error) {
+	cl.semaphoreExch <- struct{}{}
```
ryanfkeepers, line 45: In the unlikely case that this gets called without cl getting initialized, this is going to panic.  You have two options: 1, if cl or cl.semaphore is nil, call `Initialize` here with a default concurrency value;  2, if cl or cl.semaphore is nil, return (nil, err).  I'd suggest 2, personally.

src/internal/connector/graph/concurrency_limiter_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	khttp "github.com/microsoft/kiota-http-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/alcionai/corso/src/internal/tester"
+)
+
+type ConcurrencyLimiterTestSuite struct {
```
ryanfkeepers, line 19: Nit: for quick reference when reviewing test outputs in ci
```suggestion
type ConcurrencyLimiterUnitTestSuite struct {
```

src/internal/connector/graph/concurrency_limiter_test.go
```diff
@@ -0,0 +1,70 @@
...
+	tester.Suite
+}
+
+func TestConcurrencyLimiterSuite(t *testing.T) {
+	suite.Run(t, &ConcurrencyLimiterTestSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+func (suite *ConcurrencyLimiterTestSuite) TestConcurrencyLimiterMiddleware() {
```
ryanfkeepers, line 27:
```suggestion
func (suite *ConcurrencyLimiterTestSuite) TestConcurrencyLimiter() {
```

src/internal/connector/graph/concurrency_limiter_test.go
```diff
@@ -0,0 +1,70 @@
...
+	t := suite.T()
+	maxConcurrentRequests := 4
+
+	InitializeConcurrencyLimiter(maxConcurrentRequests)
```
ryanfkeepers, line 31: Since Initialize is a singleton, a pattern you might want to employ here instead is to move the internal func of `once.Do()` to its own func.  That way Initialize can call `once.Do(func() { cl = generateConcurrencyLimiter() })`, and this test can call `generateConcurrencyLimiter()` directly, without populating the global singleton.

src/internal/connector/exchange/exchange_data_collection.go
pandeyabs: Thanks! That's a great point. I'll keep the logic as is.

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,52 @@
...
+			capacity = maxConcurrentRequestsExchange
+		}
+
+		concurrencyLimiter = &ConcurrencyLimiter{
+			semaphoreExch: make(chan struct{}, capacity),
+		}
+	})
+}
+
+func GetConcurrencyLimiter() *ConcurrencyLimiter {
```
ryanfkeepers, line 36: Seems like an unnecessary func, unless we expect external packages to call it?  Again, exporting these values feels like a smell.

src/internal/connector/graph/concurrency_limiter_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	cl := GetConcurrencyLimiter()
+	client := khttp.GetDefaultClient(cl)
+
+	sem := make(chan struct{}, maxConcurrentRequests)
+	reqHandler := func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case sem <- struct{}{}:
```
ryanfkeepers, line 39: Wow, it's been a while.  I actually forgot about non-blocking writes and thought this was a bug at first (:

ryanfkeepers (commented):
Generally looks good!

src/cli/backup/exchange.go
pandeyabs: Would createCollections be a good place?

src/internal/connector/graph/concurrency_limiter.go
pandeyabs: Will do!

src/internal/connector/graph/concurrency_limiter.go
pandeyabs: fair point. will make the change

src/internal/connector/graph/concurrency_limiter.go
pandeyabs: Good catch! Thank you!

src/internal/connector/graph/concurrency_limiter.go
pandeyabs: True. I had ignored this because we only attach `ConcurrencyLimiter` if it's not nil. But I'll make this change.

src/internal/connector/graph/concurrency_limiter_test.go
pandeyabs: Ah! I see. I thought each test would be self contained and would get torn down on completion, thereby eliminating the singleton. 
But this is a great idea! I'll make this change

src/internal/connector/graph/service.go
pandeyabs: Good catch!

src/internal/connector/exchange/data_collections.go
```diff
@@ -182,6 +182,9 @@ func DataCollections(
 		categories  = map[path.CategoryType]struct{}{}
 	)
 
+	// TODO: Add hidden flag to disable this feature
```
ryanfkeepers, line 185:
```suggestion
	// TODO: Add hidden cli flag to disable this feature
```

src/internal/connector/graph/concurrency_limiter.go
```diff
@@ -0,0 +1,53 @@
...
+		concurrencyLim = generateConcurrencyLimiter(capacity)
+	})
+}
+
+func (cl *concurrencyLimiter) Intercept(
+	pipeline khttp.Pipeline,
+	middlewareIndex int,
+	req *http.Request,
+) (*http.Response, error) {
+	if cl == nil {
```
ryanfkeepers, line 43: Just for total safety, might want to nil check the channel as well.  It's unlikely that we'll have a non-nil cl and a nil sempahore, but better to have the panic protection in place.
```suggestion
	if cl == nil || cl.semaphore == nil {
```

src/internal/connector/graph/concurrency_limiter_test.go
```diff
@@ -0,0 +1,117 @@
...
+			defer func() {
+				<-sem
+			}()
+
+			time.Sleep(time.Duration(rand.Intn(50)+50) * time.Millisecond)
+			w.WriteHeader(http.StatusOK)
+
+			return
+		default:
+			w.WriteHeader(http.StatusTooManyRequests)
```
ryanfkeepers, line 48: purely as a sanity check to ensure the test is running as expected, how about adding an incremental counter to this default block, and asserting that it's non-zero after the wg.Wait()?

ryanfkeepers (approved):
Some minor nits, looks good overall.

src/internal/connector/graph/concurrency_limiter_test.go
pandeyabs: If the concurrency limiter semaphore capacity <= reqHandler semaphore capacity, we would never hit the default block. 

I've manually tested that the reqHandler default block is working as expected. I could add another test which hits the default block, but then that'd be testing the test handler. Not sure if there is value in doing that.

PR #3184, opened 2023-04-20
Rework itemPager interfaces to handle delta and non-delta

meain:
When the user has a full mailbox, delta queries does not work, we need to use non delta endpoints. This changes all of the exchange backup options to revert to non delta queries when delta query fails with quota exceeded error.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -207,7 +207,6 @@ type contactPager struct {
 func NewContactPager(
 	gs graph.Servicer,
 	user, directoryID, deltaURL string,
-	nonDelta bool,
 	immutableIDs bool,
 ) (contactPager, error) {
 	options, err := optionsForContactFoldersItem([]string{"parentFolderId"}, immutableIDs)
@@ -229,7 +228,7 @@ func NewContactPager(
 		gs:          gs,
 		user:        user,
 		directoryID: directoryID,
-		nonDelta:    nonDelta,
+		nonDelta:    false,
 
 		// for non-delta based pagination
 		builder: gs.Client().UsersById(user).ContactFoldersById(directoryID).Contacts(),
@@ -319,7 +318,7 @@ func (c Contacts) GetAddedAndRemovedItemIDs(
 		"category", selectors.ExchangeContact,
 		"container_id", directoryID)
 
-	pgr, err := NewContactPager(service, user, directoryID, oldDelta, false, immutableIDs)
+	pgr, err := NewContactPager(service, user, directoryID, oldDelta, immutableIDs)
 	if err != nil {
 		return nil, nil, DeltaUpdate{}, err
 	}
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -295,10 +295,6 @@ func (p *contactPager) getNextPage(ctx context.Context) ([]getIDAndAddtler, bool
 	return p.getNextPageDelta(ctx)
 }
 
-func (p *contactPager) valuesIn(pl api.DeltaPageLinker) ([]getIDAndAddtler, error) {
-	return toValues[models.Messageable](pl)
-}
-
 func (p *contactPager) reset(nonDelta bool) {
 	if nonDelta {
 		p.nonDelta = true
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -190,26 +190,122 @@ func (c Contacts) EnumerateContainers(
 var _ itemPager = &contactPager{}
 
 type contactPager struct {
-	gs      graph.Servicer
-	builder *users.ItemContactFoldersItemContactsDeltaRequestBuilder
-	options *users.ItemContactFoldersItemContactsDeltaRequestBuilderGetRequestConfiguration
+	gs          graph.Servicer
+	user        string
+	directoryID string
+
+	// switch between using delta vs non-delta fetch
+	nonDelta bool
+
+	builder *users.ItemContactFoldersItemContactsRequestBuilder
+	options *users.ItemContactFoldersItemContactsRequestBuilderGetRequestConfiguration
+
+	deltaBuilder *users.ItemContactFoldersItemContactsDeltaRequestBuilder
+	deltaOptions *users.ItemContactFoldersItemContactsDeltaRequestBuilderGetRequestConfiguration
 }
 
-func (p *contactPager) getPage(ctx context.Context) (api.DeltaPageLinker, error) {
-	resp, err := p.builder.Get(ctx, p.options)
+func NewContactPager(
+	gs graph.Servicer,
+	user, directoryID, deltaURL string,
+	nonDelta bool,
+	immutableIDs bool,
+) (contactPager, error) {
+	options, err := optionsForContactFoldersItem([]string{"parentFolderId"}, immutableIDs)
 	if err != nil {
-		return nil, graph.Stack(ctx, err)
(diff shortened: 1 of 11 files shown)


src/internal/connector/exchange/api/contacts.go
```diff
@@ -190,95 +190,140 @@ func (c Contacts) EnumerateContainers(
...
+	deltaBuilder := gs.Client().UsersById(user).ContactFoldersById(directoryID).Contacts().Delta()
+	if deltaURL != "" {
+		deltaBuilder = users.NewItemContactFoldersItemContactsDeltaRequestBuilder(deltaURL, gs.Adapter())
+	}
+
+	return contactPager{
+		gs:          gs,
+		user:        user,
+		directoryID: directoryID,
+		nonDelta:    false,
```
meain, line 231: We could add an option  to start with non delta, but we already know the behaviour of delta when mailbox is full and it will anyways take us one extra request either way.

src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +59,43 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
...
+		return added, removed, DeltaUpdate{deltaURL, false}, nil
+	}
+
+	// if we fail with quota exceeded error, retry with non delta
+	if graph.IsErrQuotaExceeded(err) {
+		pgr.reset(true)
+
+		added, removed, deltaURL, err = getItemsAddedAndRemovedFromContainer(ctx, pgr)
+		if err == nil {
+			// TODO(meain): Should we return resetDelta here? Will
```
meain, line 78: @ashmrtn Wouldn't just returning `resetDelta` here take care of "non incremental" backup?

src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +59,47 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
...
+	added, removed, deltaURL, err := getItemsAddedAndRemovedFromContainer(ctx, pgr)
+	if err == nil {
+		return added, removed, DeltaUpdate{deltaURL, false}, nil
+	}
+
+	// if we fail with quota exceeded error, retry with non delta
+	if graph.IsErrQuotaExceeded(err) {
+		pgr.reset(true)
+
+		logger.CtxErr(ctx, err).Info("switching to non-delta pagination")
```
ashmrtn, line 76: nit: may want to say the mailbox is full, just for some context

src/internal/connector/exchange/api/shared.go
ashmrtn: with the way the code is arranged, yes that should work

ryanfkeepers:
@meain I remain critical of this approach.  There's a lot of complexity and state management being introduced to solve only one of the many places where a quota issue could cause failures.  This feels like a rush to solve a pointed issue without tackling the larger concern.  At large, the goal should be to identify and address how Corso behaves if the user's account is in a state of being unable to use delta queries; a state that we currently know we can hit from the mailbox being over quota.

In my opinion the aforementioned account state should produce a systematic, high level change in behavior for the entire operation.  IE: if the user cannot make delta queries (in this case, if the user is at or over quota), then any process involving delta queries must use an alternate code path.  

This PR produces a singular reactionary fallback case, not a systematic change in behavior.  Reactionary fallbacks have limited portability, increase code and cognitive complexity, and are more difficult to maintain overall.  These issues all concern me.  A targeted fix might produce more long-term issues than tackling the whole issue appropriately from the start.

meain:
@ryanfkeepers Makes sense, let me rethink how we can approach it. Also, as ashmrtn suggested, there are a few other places where we use delta queries which might not support simply switching over to non-delta as the response format/content changes.

meain:
@ashmrtn Thanks for pointing out about `mailFolders`. Also, you are right in that the non delta variant (`https://graph.microsoft.com/beta/users/:user/mailFolders`) does not return any subfolders, but the `beta` version seems to be returning all the subfolders. We can consider switching to it.

meain:
@ryanfkeepers I have another proposal in https://github.com/alcionai/corso/compare/separate-pagers which implements two separate pagers `mailPager` and `mailDeltaPager` and we choose what to use based on a check to see if an exchange box is full. This retains the previous format for the most part. *Only code for mail is hashed out, rest of them(contacts and events) have just enough changes to make them compile.* Also, just to clean up building [`options`](https://github.com/alcionai/corso/blob/dbfb902f4759285ff96b60add4ea31dd5aa1238f/src/internal/connector/exchange/api/mail.go#L302), I've created a `NewMailPager` function which just returns a `mailPager` after taking values as args. We could store these values in `mailPager` and implement reset for reselting to delta but without a previous delta link. This is what [OneDrive does](https://github.com/alcionai/corso/blob/dbfb902f4759285ff96b60add4ea31dd5aa1238f/src/internal/connector/onedrive/api/drive.go#L34).

ryanfkeepers:
@meain Thanks for the exploration.  Per the given proposal, I think you're overcomplicating the handling here.  We already have a well-established path for non-delta lookups in itemPager.  All you'd need to do here is extend the func params to include a `skipDeltaQuery bool`, and change https://github.com/alcionai/corso/blob/main/src/internal/connector/exchange/api/mail.go#L341 to something like `if len(oldDelta) > 0 && !skipDeltaQuery {`.  As a result, the focus of the change is largely about identifying the cases where `skipDeltaQuery` is true (imo, seems like a good userInfo addition).

> we might have to convert them to pagers as well

I'm dubious about the need to extend the pager pattern where it isn't already implemented.  It exists in the exchange api purely as a duplicate-code reduction across similarly behaving types.  Do all pageable resources need to use it?  Maybe not.  I say this from the perspective of knowing how many hoops I made things jump through just to get it to this point.  We can keep evaluating the pattern as these changes progress, but let's save it as an option rather than the goal.  Prioritize simplicity and directness where possible here.  If we find out we're writing the same pattern over and over again, we can look at consolidating it.

meain:
@ryanfkeepers

>  We already have a well-established path for non-delta lookups in itemPager. All you'd need to do here is extend the func params to include a skipDeltaQuery bool, and change ...

I think there is a misunderstanding here. What we currently have is a way to query using delta api, with or without the delta token. Querying the delta API without a delta token can be used when we don't have a delta token(first backup) or when the token that we have has expired. What we need here is non delta query(which is a different endpoint), as delta queries do not work when the mailbox is full.

For example in the case of messages, we were using [`GET /me/mailFolders/{id}/messages/delta`](https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-1.0&tabs=http) both with and without `$deltatoken` depending on whether we had the previous delta token or not. When the mailbox is full, we have to get the same using [`GET /users/{id | userPrincipalName}/messages`] which is a different endpoint(which translates to different data types in the api) and is not current something that we have support for.

> imo, the quota investigation seems like a good userInfo addition

Yup, that was the plan.

> I'm dubious about the need to extend the pager pattern where it isn't already implemented. It exists in the exchange api purely as a duplicate-code reduction across similarly behaving types. Do all pageable resources need to use it?

Makes sense. As long as we don't have to add this in, we can stick to directly using them.

ryanfkeepers:
@meain 

> I think there is a misunderstanding here.

Whoops!  Good catch, I indeed did not remember that correctly.

meain:
Closing in favor of https://github.com/alcionai/corso/pull/3212

PR #3189, opened 2023-04-20
fixup mailbox info

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E


src/internal/connector/discovery/discovery_test.go
```diff
@@ -197,25 +198,17 @@ func (suite *DiscoveryIntegrationSuite) TestUserInfo() {
...
-				DiscoveredServices: map[path.ServiceType]struct{}{},
-				HasMailBox:         false,
-				HasOneDrive:        false,
-				Mailbox: api.MailboxInfo{
-					ErrGetMailBoxSetting: api.ErrMailBoxSettingsNotFound,
-				},
-			},
+			name:      "user does not exist",
+			user:      uuid.NewString(),
+			expectErr: require.Error,
```
ashmrtn, line 211: are we sure the results of this test are stable? Since it got merged to start with it seems like it could have indeterminate results in some way

src/internal/connector/discovery/discovery_test.go
```diff
@@ -197,25 +198,17 @@ func (suite *DiscoveryIntegrationSuite) TestUserInfo() {
 					path.ExchangeService: {},
 					path.OneDriveService: {},
 				},
-				HasMailBox:  true,
-				HasOneDrive: true,
 				Mailbox: api.MailboxInfo{
 					Purpose:              "user",
 					ErrGetMailBoxSetting: nil,
 				},
```
ashmrtn, line 204: seems like this can be removed since only `DiscoveredServices` is checked

src/pkg/services/m365/api/users.go
```diff
@@ -229,42 +227,33 @@ func (c Users) GetIDAndName(ctx context.Context, userID string) (string, string,
...
-		options = users.ItemMailFoldersRequestBuilderGetRequestConfiguration{
-			QueryParameters: requestParameters,
-		}
-	)
+	options := users.ItemMailFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &requestParameters,
+	}
 
-	userInfo.HasMailBox = true
+	if _, err := c.GetMailFolders(ctx, userID, options); err != nil {
```
ashmrtn, line 241: this is a bit out of the scope of this PR, but will we ever need to check all of the data types for Exchange? I seem to recall some reports of seeing users with no contacts folders but I'm unsure if they had email folders

ashmrtn (approved):

src/internal/connector/discovery/discovery_test.go
ryanfkeepers: For the life of me I don't see how they'd be unstable.

src/pkg/services/m365/api/users.go
ryanfkeepers: Good question. To the best of my memory, any issue that involved missing "contact folders" was, at base, a failure to get MailFolders.  So I think we're safe?  I believe, since contacts is a well known folder, that microsoft will guarantee its creation at the time of creating mailfolders.

src/pkg/services/m365/api/users.go
```diff
@@ -453,50 +458,52 @@ func validateUser(item any) (models.Userable, error) {
...
-	dataPointer, ok := data.(*string)
-	if !ok {
-		logger.Ctx(ctx).Info("error getting data from mailboxSettings")
-
-		*mailBoxErr = *ErrMailBoxSettingsNotFound
+func toString(ctx context.Context, key string, data map[string]any) (string, error) {
+	ctx = clues.Add(ctx, "setting_name", key)
 
-		return ""
+	if len(data) == 0 {
```
neha-Gupta1, line 464: if we just want to check the size of map we can check it at a central place. Taking example of `additionalData` - in current scenario the size of same map is getting checked 6 7 times.

src/pkg/services/m365/api/users.go
```diff
@@ -41,8 +41,6 @@ type Users struct {
 
 type UserInfo struct {
 	DiscoveredServices map[path.ServiceType]struct{}
-	HasMailBox         bool
-	HasOneDrive        bool
```
neha-Gupta1, line 45: Since DiscoveredServices is already having services info, it makes a lot of sense to not have these flags. 
But aren't we adding a extra burden on functions like `checkServiceEnabled` which used to call the `UserInfo` function for just getting `DiscoveredServices` info and are not fetching the MailBoxSettings too?

Infact `checkServiceEnabled` will call the `UserInfo` function to know about one service at a time(this is my understanding of the code and could be wrong) why burden it with other calls?

src/internal/connector/discovery/discovery_test.go
ryanfkeepers: and the first test run does:
```
  2023-04-20T23:58:53.016Z	INFO	mailfolders not found
      discovery_test.go:222: 
          	Error Trace:	/home/runner/work/corso/corso/src/internal/connector/discovery/discovery_test.go:222
          	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.2/suite/suite.go:112
          	Error:      	An error is expected but got nil.
          	Test:       	TestDiscoveryIntegrationSuite/TestUserInfo/user_does_not_exist
          	Messages:   	<nil>
```
:stumped:

ryanfkeepers: and then, on the first test:
```
      discovery_test.go:222: 
          	Error Trace:	/home/runner/work/corso/corso/src/internal/connector/discovery/discovery_test.go:222
          	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.2/suite/suite.go:112
          	Error:      	An error is expected but got nil.
          	Test:       	TestDiscoveryIntegrationSuite/TestUserInfo/user_does_not_exist
          	Messages:   	<nil>
```
:stumped:

PR #3195, opened 2023-04-21
2825 skip migration

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/data_collections_test.go
+++ b/src/internal/connector/onedrive/data_collections_test.go
@@ -39,24 +39,28 @@ func (suite *DataCollectionsUnitSuite) TestMigrationCollections() {
 	table := []struct {
 		name            string
 		version         int
+		forceSkip       bool
 		expectLen       int
 		expectMigration []migr
 	}{
 		{
 			name:            "no backup version",
 			version:         version.NoBackup,
+			forceSkip:       false,
 			expectLen:       0,
 			expectMigration: []migr{},
 		},
 		{
 			name:            "above current version",
 			version:         version.Backup + 5,
+			forceSkip:       false,
 			expectLen:       0,
 			expectMigration: []migr{},
 		},
 		{
 			name:      "user pn to id",
 			version:   version.AllXMigrateUserPNToID - 1,
+			forceSkip: false,
 			expectLen: 1,
 			expectMigration: []migr{
 				{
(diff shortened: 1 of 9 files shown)


src/pkg/control/options.go
```diff
@@ -13,6 +13,7 @@ type Options struct {
 	RestorePermissions   bool            `json:"restorePermissions"`
 	SkipReduce           bool            `json:"skipReduce"`
 	ToggleFeatures       Toggles         `json:"ToggleFeatures"`
+	SkipMigrations       bool            `json:"skipMigrations"`
```
ashmrtn, line 16: nit: as Options is exported I would prefer inverting this flag so that even if someone uses just `control.Options{}` we don't accidentally enable code paths we don't want to

ashmrtn: also this may be more suited as a toggle feature? I assume it's temporary for right now

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -97,13 +108,13 @@ func (suite *DataCollectionsUnitSuite) TestMigrationCollections() {
 
 				for i, cm := range test.expectMigration {
 					if cm.full == fp && cm.prev == pp {
-						migrs[i] = true
+						migrs[i] = cm
 					}
 				}
```
ashmrtn, line 113: we already did a length check above, we don't need to be picky here. We can just do something like
```go
migrs[i] = migr{full: fp, prev: pp}
```

The contains check below will just ensure we have at least what we expect

src/internal/connector/onedrive/data_collections_test.go
```diff
@@ -97,13 +108,13 @@ func (suite *DataCollectionsUnitSuite) TestMigrationCollections() {
...
 			}
 
 			for i, m := range migrs {
-				assert.Truef(t, m, "expected to find migration: %+v", test.expectMigration[i])
+				assert.Contains(t, migrs, m, "expected to find migration: %+v", test.expectMigration[i])
 			}
```
ashmrtn, line 118: actually since I missed that the expected value is a slice this whole loop could be further simplified into just
```go
assert.Subset(t, migrs, test.expectMigration)
```

Or we could skip the len check above and just use `ElementsMatch` I think

ashmrtn: second though, Len check above is needed but can exit after it if the expected len is 0

ashmrtn (approved):

src/pkg/control/options.go
ryanfkeepers: sounds good

PR #3196, opened 2023-04-21
2825 unexport

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2825

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/common/idname/idname.go
+++ b/src/internal/common/idname/idname.go
@@ -21,16 +21,15 @@ type Provider interface {
 	Name() string
 }
 
-var _ Provider = &Is{}
+var _ Provider = &is{}
 
-// Is provides an id-name tuple.
-type Is struct {
+type is struct {
 	id   string
 	name string
 }
 
-func (is Is) ID() string   { return is.id }
-func (is Is) Name() string { return is.name }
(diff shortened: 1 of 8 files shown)


src/internal/common/idname/idname.go
```diff
@@ -41,68 +40,72 @@ type Cacher interface {
...
 	idToName map[string]string
 	nameToID map[string]string
 }
 
-func NewCache(idToName map[string]string) Cache {
+func NewCache(idToName map[string]string) cache {
+	if len(idToName) == 0 {
+		// in case of nil
+		idToName = map[string]string{}
+	}
```
ashmrtn, line 54: do we need to ensure the map is non-nil? The golang [playground](https://play.golang.com/p/h00HGR73sz0) indicates that both `[]` operator and the maps package can handle a nil map

ashmrtn (approved):

src/internal/common/idname/idname.go
ryanfkeepers: yeah, I guess not.  Thanks, golang.

PR #3203, opened 2023-04-24
Add --disable-concurrency-limiter cli flag

pandeyabs:
The concurrency limiter middleware is enabled by default for exchange backups. Adding a hidden --disable-concurrency-limiter flag to disable this middleware.

Reasons for adding this flag:
We have done limited perf testing with the concurrency limiter. In addition, our understanding of exchange concurrency limits is also a bit fuzzy. This flag acts as a killswitch in case we start to see perf regressions in prod.

I see this as a temporary flag. Ideally we would never have to use it. Also, once we have a better way to configure concurrency limiter, we can eliminate this flag.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :zap: Unit test

### DIFF
--- a/src/cli/options/options_test.go
+++ b/src/cli/options/options_test.go
@@ -32,6 +32,7 @@ func (suite *OptionsUnitSuite) TestAddExchangeCommands() {
 			assert.True(t, restorePermissionsFV, RestorePermissionsFN)
 			assert.True(t, skipReduceFV, SkipReduceFN)
 			assert.Equal(t, 2, fetchParallelismFV, FetchParallelismFN)
+			assert.True(t, disableConcurrencyLimiterFV, DisableConcurrencyLimiterFN)
 		},
 	}
 
@@ -42,8 +43,8 @@ func (suite *OptionsUnitSuite) TestAddExchangeCommands() {
 	AddDisableIncrementalsFlag(cmd)
 	AddRestorePermissionsFlag(cmd)
 	AddSkipReduceFlag(cmd)
-
 	AddFetchParallelismFlag(cmd)
+	AddDisableConcurrencyLimiterFlag(cmd)
 
 	// Test arg parsing for few args
 	cmd.SetArgs([]string{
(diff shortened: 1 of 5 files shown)


src/pkg/control/options.go
```diff
@@ -105,4 +105,7 @@ type Toggles struct {
 	ExchangeImmutableIDs bool `json:"exchangeImmutableIDs,omitempty"`
 
 	RunMigrations bool `json:"runMigrations"`
+
+	// DisableConcurrencyLimiter disables the concurrency limiter package
```
ryanfkeepers, line 109: Nit, since you're not technically turning off a package.
```suggestion
	// DisableConcurrencyLimiter removes concurrency restraints when communicating with graph API
```

src/internal/connector/exchange/data_collections.go
```diff
@@ -182,8 +182,11 @@ func DataCollections(
 		categories  = map[path.CategoryType]struct{}{}
 	)
 
-	// TODO: Add hidden cli flag to disable this feature
-	graph.InitializeConcurrencyLimiter(ctrlOpts.Parallelism.ItemFetch)
+	// Turn on concurrency limiter middleware for exchange backups
+	// unless explicitly disabled through the --disable-concurrency-limiter cli flag
```
ryanfkeepers, line 186: Out of curiosity: under what conditions to we expect to want to run without the concurrency limiter in place?

src/cli/options/options_test.go
```diff
@@ -60,3 +60,50 @@ func (suite *OptionsUnitSuite) TestAddExchangeCommands() {
...
+			name: "--disable-concurrency-limiter not set",
+			args: []string{"test"},
+			assertFn: func(t *testing.T, f bool, s string) {
+				assert.False(t, f, s)
+			},
+			addFlag: false,
+		},
+		{
+			name: "--disable-concurrency-limiter set",
+			args: []string{"test", "--disable-concurrency-limiter"},
```
ryanfkeepers, line 81: Flag names should always reference their const values.
```suggestion
			args: []string{"test", "--"+DisableConcurrencyLimiterFN},
```

src/cli/options/options_test.go
```diff
@@ -60,3 +60,50 @@ func (suite *OptionsUnitSuite) TestAddExchangeCommands() {
...
+		{
+			name: "--disable-concurrency-limiter not set",
+			args: []string{"test"},
+			assertFn: func(t *testing.T, f bool, s string) {
+				assert.False(t, f, s)
+			},
+			addFlag: false,
+		},
+		{
+			name: "--disable-concurrency-limiter set",
```
ryanfkeepers, line 80: I think this case is better covered by adding the flag to the `TestAddExchangeCommands` above.  After that, I'm not sure we need a test case for not adding the flag, since the addition of the flag itself is guaranteed, not conditional.

src/pkg/control/options.go
pandeyabs: Thanks!

ryanfkeepers (approved):
Nothing blocking here, just nits.  But also, do we actually need this flag?  When do we expect to utilize it?

src/cli/options/options_test.go
pandeyabs: Fair point. I wanted to add a test case to assert on default value of the flag, which is false. If it's too extra, I can remove  this test and add to `TestAddExchangeCommands`

src/internal/connector/exchange/data_collections.go
pandeyabs: We have done limited perf testing with the concurrency limiter. In addition, our understanding of exchange concurrency limits is also a bit fuzzy. This flag acts as a killswitch in case we start to see perf regressions in prod. 

I see this as a temporary flag. Ideally we would never have to use it. Also, once we have a better way to configure concurrency limiter, we can eliminate this flag.

src/cli/options/options_test.go
pandeyabs: Update: I think there is value to checking for default flag behavior. So i've kept this as is.

pandeyabs: Good idea. Let me do this. Although I prefer dedicated unit tests per flag, but I think your suggestion is more reasonable.

pandeyabs: Oh wait. My understanding was wrong. The flag defaults only get set when the Add*(cmd) functions are called. 
If the Add* funcs are not called, these flags would be initialized to false(or 0). Hence testing for flag == false is moot. 

I'm gonna add to `TestAddExchangeCommands` and call it done. There is no value add testing for defaults.

Issue #3135, opened 2023-04-17
Enable SharePoint permissions backup


ashmrtn:

ashmrtn:
related to #2528 though not sure how easy it would be to just continue with that PR

meain:
@ryanfkeepers FYI, for [permissions](https://learn.microsoft.com/en-us/graph/api/resources/sharepointidentityset?view=graph-rest-1.0), we do not currently track `siteUser` and `siteGroup`. We should probably include that as we are now starting to backup permissions for sharepoint. Also, we currently skip `application` and `device` permissions, as we were previously using email and not id. Now that we have switched to ids I think we can enable them as well as we just have to send back the id.

ryanfkeepers:
@meain Thanks for the heads up.  Added those in https://github.com/alcionai/corso/pull/3311.

PR #3204, opened 2023-04-24
add sharepoint to sanity tests

ryanfkeepers:
Add sharepoint sanity testing.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3135

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


.github/workflows/sanity-test.yaml
```diff
@@ -353,7 +354,129 @@ jobs:
           SANITY_RESTORE_SERVICE: "onedrive"
         run: |
           set -euo pipefail
-          ./sanityCheck
+          ./sanityTest
+
+##########################################################################################################################################
+
+ # Sharepoint test
```
meain, line 361: Just wondering if we can move the common stuff into an `.github/action` and reuse that for each of them.

meain (commented):
If I am not mistaken, this currently shows few changes that has already been merged(probably since there is merge conflict). It would be better to take a look at this after a rebase.

.github/workflows/sanity-test.yaml
ryanfkeepers: Possibly, I can look at that as a later refactor, since there's ongoing churn with the sanity tests atm.

ryanfkeepers:
@ashmrtn Could I get a review on these sanity test additions, please?

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -363,27 +416,55 @@ func checkOnedriveRestoration(
...
-			logger.Ctx(ctx).Info("skipped empty folder: ", itemName)
-			fmt.Println("skipped empty folder: ", itemName)
-
+			logAndPrint(ctx, "skipped empty folder: %s", itemName)
 			continue
 		}
 
-		folderPermission[itemName] = permissionIn(ctx, client, driveID, itemID)
-		getOneDriveChildFolder(ctx, client, driveID, itemID, itemName, fileSizes, folderPermission, startTime)
+		folderPermissions[itemName] = permissionIn(ctx, client, driveID, itemID)
```
ashmrtn, line 423: I know it's not directly related to this PR, but the fact that we're using file names here means we're making the assumption that file names are unique. We should have a note about that somewhere probably

ashmrtn (approved):

src/cmd/sanity_test/sanity_tests.go
```diff
@@ -292,36 +295,88 @@ func checkAllSubFolder(
...
+// ---------------------------------------------------------------------------
+
+func checkDriveRestoration(
+	ctx context.Context,
+	client *msgraphsdk.GraphServiceClient,
+	resourceOwner,
+	folderName,
+	driveID,
+	driveName,
+	dataFolder string,
```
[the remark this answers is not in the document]
ryanfkeepers: I've been waffling on that.  I think this format is okay, so long as the variables are all listed either in a single line, or each to their own line.  Thoughts?

src/cmd/sanity_test/sanity_tests.go
ryanfkeepers: Could you kindly add that note/issue/whatever, please?

src/cmd/sanity_test/sanity_tests.go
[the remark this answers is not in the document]
ashmrtn: my nit here is that type information is only given once but params are split across lines so it's confusing to read. I tend to prefer having each param on a separate line with the type information on each line as well and the closing paren on a different line so the diff is smaller when parameters are added/removed. I know you have different preferences for some of those things

Overall, I'd prefer either

```go
folderName string,
driveID string,
driveName string,
dataFolder string,
```

or

```go
folderName, driveID, driveName, dataFolder string,
```

I personally find the former much easier to skim. The pattern currently in the PR makes skimming difficult because any given line does not contain all the information about the param name and the param type

PR #3205, opened 2023-04-24
use nanosecond folders for testing

ryanfkeepers:
we've been failing on folder name collision, particularly in this package.  Probably because it's using second-granularity folder names.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -68,7 +68,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreContact() {
 		t          = suite.T()
 		userID     = tester.M365UserID(t)
 		now        = time.Now()
-		folderName = "TestRestoreContact: " + common.FormatSimpleDateTime(now)
+		folderName = "TestRestoreContact: " + common.FormatTimeWith(now, common.SimpleTimeTesting)
 	)
 
 	aFolder, err := suite.ac.Contacts().CreateContactFolder(ctx, userID, folderName)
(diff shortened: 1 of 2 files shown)


src/internal/connector/exchange/restore_test.go
```diff
@@ -68,7 +68,7 @@ func (suite *ExchangeRestoreSuite) TestRestoreContact() {
 		t          = suite.T()
 		userID     = tester.M365UserID(t)
 		now        = time.Now()
-		folderName = "TestRestoreContact: " + common.FormatSimpleDateTime(now)
+		folderName = "TestRestoreContact: " + common.FormatTimeWith(now, common.SimpleTimeTesting)
```
pandeyabs, line 71: nit: Isn't `SimpleTimeTesting` in microseconds? That being said, microseconds should do just fine. 

It could use a better name imo e.g. `SimpleTimeUsec` . Does the word "Simple" have a significance btw?

pandeyabs (approved):
LGTM otherwise

src/internal/connector/exchange/restore_test.go
ryanfkeepers: > nit: Isn't SimpleTimeTesting in microseconds?

heh, yes, and it'll be fine.  I really should have named it "use more decimal spots for testing".  That would've been sufficient.  Realistically we'd only really need a tenth or hundredth decimal point to get enough granularity to avoid test folder collision.

> It could use a better name imo e.g. SimpleTimeUsec.

This is one of the times where I prefer the name to follow the intent, rather than the underlying structure.  This is a format we should only use for testing.  There's no real world case where we need Usec granularity, and lots of test cases where we do, thus the name.

> Does the word "Simple" have a significance btw?
Yeah!  It's an arbitrary title, but basically means "the human readable timestamp format" (specifically: `"02-Jan-2006_15:04:05"`).  IE: what we'd want to use for generating names for folders and such.  You can [expect](https://github.com/alcionai/corso/blob/main/design/time_standards.md) that any other time format follow RFC3339 standards. (I need to migrate that doc to co-exist with the time package).

pandeyabs: Thanks for the context!

PR #3206, opened 2023-04-24
ditch userInfo middle struct

ryanfkeepers:
we don't need a m365.UserInfo when we already have an api.UserInfo

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/m365.go
+++ b/src/pkg/services/m365/m365.go
@@ -12,7 +12,6 @@ import (
 	"github.com/alcionai/corso/src/internal/connector/discovery"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/fault"
-	"github.com/alcionai/corso/src/pkg/path"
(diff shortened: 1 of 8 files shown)


src/pkg/services/m365/m365.go
```diff
@@ -135,7 +130,7 @@ func GetUserInfo(
 	ctx context.Context,
 	acct account.Account,
 	userID string,
-) (*UserInfo, error) {
```
neha-Gupta1, line 138: Out of curiosity, what is the significance of `GetUserInfo` func. And where all it is getting used?

ryanfkeepers: SDK users can call it before running a backup for a user to ensure they have up-to-date info about that user.  Supposing they aren't retrieving all users every run, and instead use their own library of user ids that they iterate through to kick off backups, GetUserInfo allows them to check whether the user has services enabled, among other checks, before initializing the backup.

juleslasarte (approved):

PR #3207, opened 2023-04-24
get users for cli backup without getting info

ryanfkeepers:
When the cli aggregates all users, it doesn't need to look up user info for anyone, it only needs their id and PN.  The rest of the info (such as service support) will get looked up on a one-by-one basis in connector during backup run.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange.go
+++ b/src/cli/backup/exchange.go
@@ -17,7 +17,6 @@ import (
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
 	"github.com/alcionai/corso/src/pkg/selectors"
-	"github.com/alcionai/corso/src/pkg/services/m365"
 )
 
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/cli/backup/exchange.go
```diff
@@ -161,10 +160,7 @@ func createExchangeCmd(cmd *cobra.Command, args []string) error {
 
 	sel := exchangeBackupCreateSelectors(utils.UserFV, utils.CategoryDataFV)
 
-	// TODO: log/print recoverable errors
-	errs := fault.New(false)
-
-	ins, err := m365.UsersMap(ctx, *acct, errs)
```
neha-Gupta1, line 167: https://github.com/alcionai/corso/blob/c38a43107962b509d07b79b72b9c4f3773315559/src/cmd/factory/impl/common.go#L125 also uses `m365.UsersMap`

neha-Gupta1 (approved):
The code will save us some computation. Only a change is required in common.go.

src/cli/backup/exchange.go
ryanfkeepers: whoops, thanks!

PR #3212, opened 2023-04-25
Add non delta pagers to exchange

meain:
When the user's mailbox is full, we cannot make use of delta apis. This adds initial changes needed to create separate delta and non delta pagers for all of exchange. There will be follow up PRs to this to do the following. Will merge everything into this and merge together at the end.

- [ ] Thread through `deltaAvailable` after getting it at the start for `userInfo`
- [ ] Tests and any other docs stuff

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +56,62 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
...
+	}
+
+	added, removed, deltaURL, err := getItemsAddedAndRemovedFromContainer(ctx, pgr)
+	// note: happy path, not the error condition
+	if err == nil {
+		return added, removed, DeltaUpdate{deltaURL, len(oldDelta) != 0}, err
+	}
+
+	// return error if invalid delta error or if we did a non-delta fetch
+	if !graph.IsErrInvalidDelta(err) && !graph.IsErrQuotaExceeded(err) {
```
meain, line 85: @ryanfkeepers I'm switching based on the error instead of trying to see if we fetch via delta at the start as that requires an extra permissions, plus we have validated this behaviour.

ryanfkeepers: I have to disagree with that choice.  The core issue is that this is still _reacting_ to an over-quota issue instead of behaving consistently throughout the process runtime.

The overwhelming majority of over-quota cases are issues of user account state, not of folder state.  This approach treats it as folder state.  That means every single folder we enumerate, account-wide, will have to fail, catch the failure, change behavior, and then correct course.  And since the most likely case for over-quota issues is that the user's mailbox will be entirely over quota for the duration of the backup, or not, as a binary condition, this means we'll end up in this fallback for every folder, or none of them.

The purpose of discovering quota state early on is to avoid the per-folder fallback.  How we do the check is flexible; catching the error instead of asking for another permission does solve this case (though I can imagine good reasons to do the latter).  The method isn't as important as the location.  The location we want is at the start of the backup, before any enumeration.  I'm thinking [here where we checkServiceEnabled](https://github.com/alcionai/corso/blob/main/src/internal/connector/data_collections.go#L59), expanding that logic a bit to transform or pass down userInfo.

By generating a high-level flag and passing it to lower layers, we can make clear assumptions about system behavior: _all_ enumeration will avoid deltas, _all_ of the metadata will get dropped in favor of a full backup.  It minimizes the cognitive load about how parts of the system should behave, and when, ensure we have the flag available for all cases that rely on the identification, and increases our capacity to easily test and maintain the code.  The more conditional fallbacks we have, especially those that are dependent on broad account state, the more we encounter questions like "folder A did a fallback, but folder B didn't; is that expected or a bug?" and "what corner cases arise for a backup if we arbitrarily fallback on N of M containers?"

For the most part, the changes you've made here are great and provide stepping stones to handle the behavior as outlined above.  It's just the matter of pinning down how we establish the pivot between delta and non-delta behaviors.  From here, to continue with iteration, I think additional small branches that can be folded back into this one are the best way to iterate and review.  If needed, I'm happy to jump on a call, even if its late for me (morning for you).

ryanfkeepers (changes requested):

src/internal/connector/exchange/api/mail.go
```diff
@@ -221,6 +225,89 @@ func (c Mail) GetItem(
...
+func (p *mailFolderPager) setNext(nextLink string) {
+	p.builder = users.NewItemMailFoldersRequestBuilder(nextLink, p.service.Adapter())
+}
+
+func (p *mailFolderPager) valuesIn(pl api.PageLinker) ([]models.MailFolderable, error) {
+	// Ideally this should be `users.ItemMailFoldersResponseable`, but
+	// that is not a thing as stable returns different result
+	page, ok := pl.(models.MailFolderCollectionResponseable)
+	if !ok {
+		return nil, clues.New("unable to convert to ItemMailFoldersResponseable")
```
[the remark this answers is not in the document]
pandeyabs: also `ItemMailFoldersResponsible`

src/internal/connector/exchange/api/contacts.go
```diff
@@ -190,12 +190,92 @@ func (c Contacts) EnumerateContainers(
...
+	options, err := optionsForContactFoldersItem([]string{"parentFolderId"}, immutableIDs)
+	if err != nil {
+		return &contactPager{}, err
+	}
+
+	builder := gs.Client().UsersById(user).ContactFoldersById(directoryID).Contacts()
+
+	if len(os.Getenv("CORSO_URL_LOGGING")) > 0 {
+		gri, err := builder.ToGetRequestInformation(ctx, options)
+		if err != nil {
```
ashmrtn, line 213: should we be returning this error somehow? How will callers know that they can't use the pager? Is the pager going to check later if it was setup properly?

this would apply to other places this pattern is used as well

src/internal/connector/exchange/api/mail.go
```diff
@@ -221,6 +225,89 @@ func (c Mail) GetItem(
 	return mail, MailInfo(mail), nil
 }
 
+type mailFolderDeltaPagerer interface {
```
ashmrtn, line 228: We don't need to maintain delta and non-delta versions of these. We only need something that supports enumerating all folders at all levels of the hierarchy efficiently. If the beta API can handle that then it's fine to just switch to that implementation. The delta tokens we get from enumerating mail folders are discarded anyway

src/internal/connector/exchange/api/mail.go
```diff
@@ -221,6 +225,89 @@ func (c Mail) GetItem(
...
+func NewMailFolderPager(service graph.Servicer, user string) mailFolderDeltaPagerer {
+	// Sable /mailFolders endpoint does not return any of the nested folders
+	rawURL := fmt.Sprintf(mailFoldersBetaURLTemplate, user)
+	builder := users.NewItemMailFoldersRequestBuilder(rawURL, service.Adapter())
+
+	return &mailFolderPager{service, builder}
+}
+
+func (p *mailFolderPager) getPage(ctx context.Context) (api.PageLinker, error) {
+	page, err := p.builder.Get(ctx, nil)
```
ashmrtn, line 288: should we be passing options like a page size here?

src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +56,56 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
 	return r, nil
 }
 
+func getAddedAndRemovedItemIDs(
+	ctx context.Context,
+	service graph.Servicer,
+	user, directoryID, oldDelta string,
+	pagerGetter func(context.Context, graph.Servicer, string, string, bool) (itemPager, error),
+	deltaPagerGetter func(context.Context, graph.Servicer, string, string, string, bool) (itemPager, error),
```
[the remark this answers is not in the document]
ashmrtn: +1 to passing in instances instead of constructors. Creating these pagers is a fairly lightweight operation anyway

src/internal/connector/exchange/api/shared.go
```diff
@@ -104,19 +154,19 @@ func getItemsAddedAndRemovedFromContainer(
 			}
 		}
 
-		nextLink, delta := api.NextAndDeltaLink(resp)
+		nextLink, deltaLink := api.NextAndDeltaLink(resp)
 		if len(os.Getenv("CORSO_URL_LOGGING")) > 0 {
-			if !api.IsNextLinkValid(nextLink) || api.IsNextLinkValid(delta) {
-				logger.Ctx(ctx).Infof("Received invalid link from M365:\nNext Link: %s\nDelta Link: %s\n", nextLink, delta)
+			if !api.IsNextLinkValid(nextLink) || api.IsNextLinkValid(deltaLink) {
+				logger.Ctx(ctx).Infof("Received invalid link from M365:\nNext Link: %s\nDelta Link: %s\n", nextLink, deltaLink)
```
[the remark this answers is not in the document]
ashmrtn: also: do we need to conceal these values at all? Or will concealing be done automatcially?

src/internal/connector/graph/api/api.go
```diff
@@ -24,6 +24,12 @@ func NextLink(pl PageLinker) string {
 	return ptr.Val(pl.GetOdataNextLink())
 }
 
-func NextAndDeltaLink(pl DeltaPageLinker) (string, string) {
-	return NextLink(pl), ptr.Val(pl.GetOdataDeltaLink())
+func NextAndDeltaLink(pl PageLinker) (string, string) {
+	dpl, ok := pl.(DeltaPageLinker)
```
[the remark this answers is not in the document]
ashmrtn: I'm not sure we can completely comply with that in the graph package without jumping through a lot of hoops. In some places we probably could, but in others, due to the way graph is written with lots of different types, it may not be possible while retaining the common code we have already

ashmrtn (commented):
I think there may still be something that needs addressed. If we make a backup after falling back to the non-delta pager then we won't persist any delta tokens. This may cause the next backup to incorrectly calculate the set of deleted folders as the previous path map may not be passed in properly

The way to get to the above would be
1. create a backup with delta tokens
2. fill the mailbox so delta requests return quota exceeded
3. make another backup that will fall back to non-delta enumeration
4. delete a folder
5. make a backup, may still have the old folder due to strict metadata handling we have right now

The code I'm worried about is the stuff that deserializes metadata from the previous backup. I think the block in question would be [this one](https://github.com/alcionai/corso/blob/c95a07660ae6c3b350bd7bd67faa2cc79206413e/src/internal/connector/exchange/data_collections.go#L148)

src/internal/connector/exchange/api/contacts.go
```diff
@@ -190,12 +190,92 @@ func (c Contacts) EnumerateContainers(
...
+	options, err := optionsForContactFoldersItemDelta([]string{"parentFolderId"}, immutableIDs)
+	if err != nil {
+		return &contactDeltaPager{}, err
+	}
+
+	builder := gs.Client().UsersById(user).ContactFoldersById(directoryID).Contacts().Delta()
+	if deltaURL != "" {
+		builder = users.NewItemContactFoldersItemContactsDeltaRequestBuilder(deltaURL, gs.Adapter())
+
+		if len(os.Getenv("CORSO_URL_LOGGING")) > 0 {
```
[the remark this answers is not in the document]
meain: I kept it that way as we were doing that only when we did not have previous delta currently.
https://github.com/alcionai/corso/blob/edef23bfc4eb57acc7ced0a36830a58d76acbf15/src/internal/connector/exchange/api/mail.go#L364

src/internal/connector/exchange/api/mail.go
```diff
@@ -221,6 +225,89 @@ func (c Mail) GetItem(
...
+func (p *mailFolderDeltaPager) valuesIn(pl api.PageLinker) ([]models.MailFolderable, error) {
+	page, ok := pl.(users.ItemMailFoldersDeltaResponseable)
+	if !ok {
+		return nil, clues.New("unable to convert to ItemMailFoldersDeltaResponseable")
+	}
+
+	return page.GetValue(), nil
+}
+
+var _ mailFolderDeltaPagerer = &mailFolderPager{}
```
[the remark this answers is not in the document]
meain: This was supposed to be `mailFolderDeltaPagerer`, changed.

src/internal/connector/exchange/api/shared.go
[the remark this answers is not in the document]
meain: I've wrapped the URLs in `graph.LoggableURL`.

src/internal/connector/exchange/api/mail.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
+const (
+	mailFoldersBetaURLTemplate = "https://graph.microsoft.com/beta/users/%s/mailFolders"
```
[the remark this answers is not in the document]
meain: v1.0 version of non delta pager for `mailFolders` does not return nested folders.
https://github.com/alcionai/corso/pull/3184#issuecomment-1517526350

src/internal/connector/exchange/api/mail.go
```diff
@@ -221,6 +225,89 @@ func (c Mail) GetItem(
...
+
+var _ mailFolderDeltaPagerer = &mailFolderPager{}
+
+type mailFolderPager struct {
+	service graph.Servicer
+	builder *users.ItemMailFoldersRequestBuilder
+}
+
+func NewMailFolderPager(service graph.Servicer, user string) mailFolderDeltaPagerer {
+	// Sable /mailFolders endpoint does not return any of the nested folders
```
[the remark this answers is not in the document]
meain: v1.0 non delta version

src/internal/connector/exchange/api/mail.go
meain: I kept both since the the new one is a **beta** endpoint and thought it would be better to use the stable version for by default.

src/internal/connector/exchange/api/mail.go
meain: The current version does not pass in any options for now. I could add it in a future PR if we should add it.

src/internal/connector/exchange/api/contacts.go
meain: The current version just logs them as well. I believe we should be fine as this data is only used for logging and if the endpoint actually fails, we will fail.
https://github.com/alcionai/corso/blob/edef23bfc4eb57acc7ced0a36830a58d76acbf15/src/internal/connector/exchange/api/mail.go#L364

src/internal/connector/graph/api/api.go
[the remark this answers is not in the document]
meain: I'm assuming https://github.com/alcionai/corso/pull/3212/commits/0b945758572967ee4414028710edcaa301ca6d17 is what you meant.

src/internal/connector/exchange/api/shared.go
meain: Thanks for the detailed response. I think you make very good points. For some reason I forgot the fact that we have multiple deltas in exchange. Let me see how I can rework this.

src/internal/connector/exchange/api/mail.go
meain: Removing it.

src/internal/connector/exchange/api/shared.go
meain: Created https://github.com/alcionai/corso/pull/3232 for this.

src/internal/connector/graph/api/api.go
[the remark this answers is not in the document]
ryanfkeepers: yes, just the PageLinker/DeltaPageLinker in this case.

src/internal/connector/exchange/api/shared.go
ryanfkeepers: Looking good so far, thank you for the changes!

src/internal/connector/exchange/api/shared.go
```diff
@@ -18,9 +18,14 @@ import (
 // ---------------------------------------------------------------------------
 
 type itemPager interface {
-	getPage(context.Context) (api.DeltaPageLinker, error)
+	// getPage get a page with the specified options from graph
+	getPage(context.Context) (api.PageLinker, error)
+	// setNext is used to pass in the next url got from graph
 	setNext(string)
-	valuesIn(api.DeltaPageLinker) ([]getIDAndAddtler, error)
+	// reset is used to reset delta url in delta pagers
```
ashmrtn, line 25: nit: it's that we're not using the delta at all so maybe a different word would be more meaningful. "Reset" could make it sound like it's going to start at the beginning again
```suggestion
	// reset is used to clear delta url in delta pagers
```

src/internal/connector/exchange/api/mail.go
```diff
@@ -238,22 +325,31 @@ func (c Mail) EnumerateContainers(
 	}
 
 	el := errs.Local()
-	builder := service.Client().
-		UsersById(userID).
-		MailFolders().
-		Delta()
+
+	var pgr mailFolderPagerer
+	pgr = NewMailFolderPager(service, userID)
```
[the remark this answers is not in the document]
ashmrtn: ah, I see, this could have been added to this PR by accident as it's also present in the other PR on top of this one

ashmrtn:
>quickest solution that I can think of is to either add another flag in the config, or have the delta be populated with a fixed value (like *) which can be used as an identifier that the run was successful, but we did not have any delta

Adding another flag to the config requires us to know that the backup sourcing the metadata didn't use delta. As of now I don't think we have the infrastructure for that. It's essentially the same as checking which features were enabled on a given backup

I'm on the fence about using a dummy value. On the one hand, it allows us to move forward quickly as well as ensuring we can tell error cases apart from non-error cases. On the other hand, whatever we pick we're going to be stuck supporting for a good while and we're may have to address the problem of incomplete metadata at some point

The checks we have right now are fragile because they assume complete metadata. However, without prior knowledge of that assumption it's easy to break the system by passing in no metadata or some similar thing (i.e. in response to a condition discovered higher up the stack like a full mailbox)

I guess as a general question, what if we did the following:
* set the delta token for items that had no path entry to an empty string
* kept path entries even if the corresponding delta token was an empty string

Would the remainder of the code function? Would it cause collections corresponding to those entries to be created as `DoNoMerge`?

The above hinges on the idea that passing "" as a delta would result in `deltaReset` being returned by the getAddedRemoved logic and that would subsequently flow into the created collections. The lift to get the change in would also be fairly small as it's not a huge departure from the existing logic assuming getAddedRemoved behaves according to the assumption. It does still assume that individually the paths entries and the delta entries are complete (i.e. all paths/deltas are contained in the file respectively) but I think that's probably a safe assumption to make given that kopia shouldn't expose partial writes (or at least not ones we can't detect)

src/internal/connector/exchange/api/mail.go
meain: Created https://github.com/alcionai/corso/issues/3243

src/internal/connector/exchange/api/events.go
```diff
@@ -241,54 +265,58 @@ func (p *eventPager) getPage(ctx context.Context) (api.DeltaPageLinker, error) {
...
-	ctx context.Context,
-	user, calendarID, oldDelta string,
-	immutableIDs bool,
-) ([]string, []string, DeltaUpdate, error) {
-	service, err := c.service()
-	if err != nil {
-		return nil, nil, DeltaUpdate{}, err
-	}
+// ---------------------------------------------------------------------------
+// non-delta item pager
```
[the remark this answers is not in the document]
meain: Woops!

src/internal/connector/exchange/api/contacts.go
```diff
@@ -205,80 +244,129 @@ func (p *contactPager) getPage(ctx context.Context) (api.DeltaPageLinker, error)
...
-	builder := service.Client().UsersById(user).ContactFoldersById(directoryID).Contacts().Delta()
-	pgr := &contactPager{service, builder, options}
+	return resp, nil
+}
+
+func (p *contactDeltaPager) setNext(nextLink string) {
+	p.builder = users.NewItemContactFoldersItemContactsDeltaRequestBuilder(nextLink, p.gs.Adapter())
+}
 
+func (p *contactDeltaPager) reset(ctx context.Context) {
```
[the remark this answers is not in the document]
meain: I wanted to convert the logging into a util function, but don't think we will be able to do that easily. Let me make the delta builder creation all together into a separate function like in events.

meain:
> The above hinges on the idea that passing "" as a delta would result in deltaReset

This is true for both [exchange](https://github.com/alcionai/corso/blob/f70f44debf08ec2547979870db19cc5d9efd2d6a/src/internal/connector/exchange/api/shared.go#L79) and [onedrive](https://github.com/alcionai/corso/blob/867df943082d090fa4cf4ac8671f6e227bb945b1/src/internal/connector/onedrive/drive.go#L122).


> I guess as a general question, what if we did the following:
> - set the delta token for items that had no path entry to an empty string
> - kept path entries even if the corresponding delta token was an empty string

IIUC we just don't filter based on empty delta, but just based on empty path. I guess even in the error scenario where corso somehow wrote the path but not the delta we will still require the same behaviour of doing a full backup.

This is a small change, added in 867df9430.

---

On a related note, should we have a check to make sure we have seen both files? *FYI, with the non-delta pagers combined with `omitempty`, we end up with `{}` as the data within `delta` metadata file.*

```diff
diff --git a/src/internal/connector/exchange/data_collections.go b/src/internal/connector/exchange/data_collections.go
index cbb67211d..0d148189d 100644
--- a/src/internal/connector/exchange/data_collections.go
+++ b/src/internal/connector/exchange/data_collections.go
@@ -145,9 +145,18 @@ func parseMetadataCollections(
 	// Remove any entries that contain a path or a delta, but not both.
 	// That metadata is considered incomplete, and needs to incur a
 	// complete backup on the next run.
-	for _, dps := range cdp {
+	for cat, dps := range cdp {
+		_, hasPath := found[cat]["path"]
+		_, hasDelta := found[cat]["delta"]
+		if !hasPath && !hasDelta {
+			// if both files are not available, we might have
+			// incomplete backups
+			delete(cdp, cat)
+			continue
+		}
+
 		for k, dp := range dps {
-			if len(dp.delta) == 0 || len(dp.path) == 0 {
+			if len(dp.path) == 0 {
 				delete(dps, k)
 			}
 		}
```

src/internal/connector/exchange/api/contacts.go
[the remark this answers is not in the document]
meain: Added in https://github.com/alcionai/corso/pull/3212/commits/f47e9b91751e6775d6acd1c798419b0ab4174634

ashmrtn:
Sorry, this got forgotten in the other stuff that I ended up doing today.

>IIUC we just don't filter based on empty delta, but just based on empty path. I guess even in the error scenario where corso somehow wrote the path but not the delta we will still require the same behaviour of doing a full backup.

yup, it's safe (correctness wise) for us to have a superset of paths compared to delta tokens. The tokens we can always fall back to something else with. The paths however are used to to calculate deleted folders, which means we don't really want to remove items from the set if the delta is weird (the original code made fairly strong assumptions which is why it had the check for both)

In a different PR we should probably do better at checking/enforcing stuff like this

The diff you posted seems like it should work as it will cause the system to fallback to full enumeration and flag a collection with `DoNotMerge` when going from `no delta backup` -> `delta backup` if we just persist empty strings for each delta (i.e. delta map would look like `map{folder1ID: ""}`)

meain:
> The diff you posted seems like it should work as it will cause the system to fallback..

On second thought I don't think we need the whole thing, I think just https://github.com/alcionai/corso/commit/867df943082d090fa4cf4ac8671f6e227bb945b1 should be good(I have this added already in the PR).

meain:
@ashmrtn @ryanfkeepers  Other PRs in the chain:

- [ ] https://github.com/alcionai/corso/pull/3289
- [ ] https://github.com/alcionai/corso/pull/3319

meain:
This should now be good to. I've merged everything into this PR. If there is nothing else, we can merge it into main.
@ashmrtn @ryanfkeepers

src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +61,54 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
...
+		return added, removed, DeltaUpdate{deltaURL, resetDelta}, err
+	}
+
+	// If we already tried with a non-delta url, we can return
+	if !canMakeDeltaQueries {
+		return nil, nil, DeltaUpdate{}, err
+	}
+
+	// return error if invalid delta error or oldDelta was empty
+	if !graph.IsErrInvalidDelta(err) || len(oldDelta) == 0 {
```
ryanfkeepers, line 97: This condition doesn't match the comment.  It returns if the error is *not* an invalid delta error.

src/internal/connector/exchange/api/shared.go
```diff
@@ -56,6 +61,54 @@ func toValues[T any](a any) ([]getIDAndAddtler, error) {
...
+		return nil, nil, DeltaUpdate{}, err
+	}
+
+	// reset deltaPager
+	pgr.reset(ctx)
```
ryanfkeepers, line 102: The reset behavior isn't well explained, both here and in the interface.  I'm not sure what to expect of the state before and after this call, or what the act of "resetting" the pager is intended to solve.  This not only makes it difficult to grok code flow at this point, but will make future contribution around these expectations difficult to maintain.

src/internal/connector/exchange/api/shared.go
```diff
@@ -104,19 +159,29 @@ func getItemsAddedAndRemovedFromContainer(
...
 		}
 
-		nextLink, delta := api.NextAndDeltaLink(resp)
+		dresp, ok := resp.(api.DeltaPageLinker)
+		if ok {
+			nextLink, deltaLink = api.NextAndDeltaLink(dresp)
+		} else {
+			nextLink = api.NextLink(resp)
+			deltaLink = "" // to make sure we don't use an old value
+		}
```
ryanfkeepers, line 168: Still concerned about side-loading interfaces.  Seems like this could be changed so that all responses are DeltaPageLinkers, and in the case of non-delta queries the deltaLink is always an empty string.

src/internal/connector/exchange/data_collections.go
```diff
@@ -55,6 +55,16 @@ func (dps DeltaPaths) AddPath(k, p string) {
 	dps[k] = dp
 }
 
+// RemoveDelta removes the delta from the DeltaPaths.
+// It is *ONLY* for use in tests.
+func (dps DeltaPaths) RemoveDelta(k string) {
```
ryanfkeepers, line 60: This should be moved to a test file.  There's no reason for it to exist within the primary package if its only used for testing.
```suggestion
func RemoveDelta(t *testing.T, dps DeltaPaths, k string) {
```

src/internal/connector/exchange/service_iterators.go
```diff
@@ -170,7 +171,9 @@ func filterContainersAndFillCollections(
 			qp.ResourceOwner.ID(),
 			cID,
 			prevDelta,
-			ctrlOpts.ToggleFeatures.ExchangeImmutableIDs)
+			ctrlOpts.ToggleFeatures.ExchangeImmutableIDs,
+			!ctrlOpts.ToggleFeatures.DisableDelta,
+		)
```
ryanfkeepers, line 176:
```suggestion
			!ctrlOpts.ToggleFeatures.DisableDelta)
```

src/internal/connector/exchange/data_collections.go
```diff
@@ -303,7 +313,8 @@ func createCollections(
 		scope,
 		dps,
 		ctrlOpts,
-		errs)
+		errs,
+	)
```
ryanfkeepers, line 317:
```suggestion
		errs)
```

src/internal/connector/graph/errors.go
```diff
@@ -111,6 +112,10 @@ func IsErrInvalidDelta(err error) bool {
 		errors.Is(err, ErrInvalidDelta)
 }
 
+func IsErrQuotaExceeded(err error) bool {
```
ryanfkeepers, line 115: Can you please add the standard unit tests for this func?

src/pkg/services/m365/api/users.go
```diff
@@ -295,6 +303,27 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
 	userInfo.Mailbox = mbxInfo
 
+	if mfs != nil {
+		mf := mfs.GetValue()[0] // we will always have one
+		options := &users.ItemMailFoldersItemMessagesDeltaRequestBuilderGetRequestConfiguration{
+			QueryParameters: &users.ItemMailFoldersItemMessagesDeltaRequestBuilderGetQueryParameters{
+				Top: ptr.To[int32](1), // just one item is enough
+			},
+		}
+		_, err = c.stable.Client().
```
ryanfkeepers, line 313: Can you please add a todo that this is a quick hack to interpret over-quota errors, and that future changes could instead look up the mailbox quota status itself to evaluate delta usage?

ashmrtn (commented):
looks pretty good to me, just a few minor things. Will leave final approval to Keepers

src/internal/connector/exchange/api/shared.go
```diff
@@ -104,19 +159,29 @@ func getItemsAddedAndRemovedFromContainer(
...
+
 		if len(os.Getenv("CORSO_URL_LOGGING")) > 0 {
-			if !api.IsNextLinkValid(nextLink) || api.IsNextLinkValid(delta) {
-				logger.Ctx(ctx).Infof("Received invalid link from M365:\nNext Link: %s\nDelta Link: %s\n", nextLink, delta)
+			if !api.IsNextLinkValid(nextLink) || api.IsNextLinkValid(deltaLink) {
```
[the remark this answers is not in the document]
ryanfkeepers: I think it's time to remove the CORSO_URL_LOGGING handlers, anyway.  They were meant to track a specific test case and we haven't gotten around to cleanup.  @meain Feel free to remove that code (in another pr).

src/internal/connector/exchange/api/shared.go
meain: Add an explanation to the reset interface.

src/internal/connector/exchange/api/shared.go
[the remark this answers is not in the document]
meain: Ahh, I never looked into the code. That a copy paste job. Good catch. Keepers, I'll create a follow up to remove it.

src/internal/connector/exchange/data_collections.go
meain: This function is used in another package and the values in DeltaPaths were private. I'm converting them to public and moving this out to the test package.

src/internal/connector/exchange/api/shared.go
meain: Continuation of https://github.com/alcionai/corso/pull/3212#discussion_r1176697616

> Seems like this could be changed so that all responses are DeltaPageLinkers, and in the case of non-delta queries the deltaLink is always an empty string.

Unfortunately that is not possible as the object that is returned from non-delta calls does not have `GetOdataDeltaLink` and so cannot satisfy [`DeltaPageLinker`](https://github.com/alcionai/corso/blob/1b417af5bb1e6f250adb7e21216a731e9decb909/src/internal/connector/graph/api/api.go#L13).

> ... we should probably move NextAndDeltaLink to be a function on the pageLinker instead of a function that takes pageLinker

That would be better, but as we can't satisfy the interface, it will still be moving the check to within the function as done [here](https://github.com/alcionai/corso/pull/3212#discussion_r1176697616) unless I am mistaken.

ryanfkeepers (approved):

src/internal/connector/exchange/api/shared.go
ryanfkeepers: > Unfortunately that is not possible as the object that is returned from non-delta calls does not have GetOdataDeltaLink and so cannot satisfy [DeltaPageLinker](https://github.com/alcionai/corso/blob/1b417af5bb1e6f250adb7e21216a731e9decb909/src/internal/connector/graph/api/api.go#L13).

You can wrap the object to normalize the interface.  We do [similar things with calendar containers in graph](https://github.com/alcionai/corso/blob/main/src/internal/connector/graph/cache_container.go#L156).

ryanfkeepers:
Generally looks good (thanks for all the effort through the churn), approved to unblock.  Would still like to see the interface side-loading solved with the use of a wrapping struct.

PR #3220, opened 2023-04-25
cache created folders when restoring drive items

ryanfkeepers:
drive items aren't currently getting cached after creation.  Instead, we do a lookup on the folder name every time we walk the folder hierarchy toward the next child to restore.  Normally this would work fine, but apparently if the folder is named `folder`, then graph api returns nothing, causing the process to fail when repeatedly creating the folder.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/folder_cache.go
+++ b/src/internal/connector/onedrive/folder_cache.go
@@ -18,11 +18,11 @@ func NewFolderCache() *folderCache {
 	}
 }
 
-func (c *folderCache) Get(loc *path.Builder) (models.DriveItemable, bool) {
+func (c *folderCache) get(loc *path.Builder) (models.DriveItemable, bool) {
 	mdi, ok := c.cache[loc.String()]
 	return mdi, ok
 }
 
-func (c *folderCache) Set(loc *path.Builder, mdi models.DriveItemable) {
+func (c *folderCache) set(loc *path.Builder, mdi models.DriveItemable) {
 	c.cache[loc.String()] = mdi
 }
--- a/src/internal/connector/onedrive/folder_cache.go
+++ b/src/internal/connector/onedrive/folder_cache.go
@@ -1,0 +1,28 @@
+package onedrive
+
+import (
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+// TODO: refactor to comply with graph/cache_container
+
+type folderCache struct {
+	cache map[string]models.DriveItemable
+}
+
+func NewFolderCache() *folderCache {
+	return &folderCache{
+		cache: map[string]models.DriveItemable{},
+	}
+}
+
+func (c *folderCache) Get(loc *path.Builder) (models.DriveItemable, bool) {
+	mdi, ok := c.cache[loc.String()]
+	return mdi, ok
+}
+
+func (c *folderCache) Set(loc *path.Builder, mdi models.DriveItemable) {
+	c.cache[loc.String()] = mdi
+}
--- a/src/internal/connector/onedrive/restore.go
+++ b/src/internal/connector/onedrive/restore.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/internal/connector/onedrive/api"
 	"github.com/alcionai/corso/src/internal/connector/onedrive/metadata"
 	"github.com/alcionai/corso/src/internal/connector/support"
 	"github.com/alcionai/corso/src/internal/data"
@@ -52,6 +53,8 @@ func RestoreCollections(
 		// permissionIDMappings is used to map between old and new id
 		// of permissions as we restore them
 		permissionIDMappings = map[string]string{}
+		fc                   = NewFolderCache()
+		rootIDCache          = map[string]string{}
 	)
 
 	ctx = clues.Add(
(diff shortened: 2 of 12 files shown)


ryanfkeepers:
manual run success:
```
corso restore onedrive \
	--backup c7d64dff-9e18-4067-8215-670a7f8aa052 \
	--folder /folder
Logging to file: stderr
Restoring to folder Corso_Restore_25-Apr-2023_23-33-09
  ID            ItemName       ParentPath  Size    Owner  Created               Modified
  48b7f942bb60  wat.jpg                    6.1 kB         2023-04-25T23:33:15Z  2023-04-25T23:33:15Z
  b476a9c76ac6  woohoo.png                 184 kB         2023-04-25T23:33:17Z  2023-04-25T23:33:17Z
  7b7c4df912b3  huehuehue.GIF              89 kB          2023-04-25T23:33:19Z  2023-04-25T23:33:19Z
```

src/internal/connector/onedrive/folder_cache.go
```diff
@@ -0,0 +1,28 @@
+package onedrive
+
+import (
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+// TODO: refactor to comply with graph/cache_container
```
vkamra, line 9: Do we need this TODO? It seems like `cache_container` does a lot more

src/internal/connector/onedrive/restore.go
```diff
@@ -541,43 +562,112 @@ func restoreV6File(
...
+
 // CreateRestoreFolders creates the restore folder hierarchy in the specified
 // drive and returns the folder ID of the last folder entry in the hierarchy.
+// folderCache is mutated, as a side effect of populating the items.
 func CreateRestoreFolders(
 	ctx context.Context,
 	service graph.Servicer,
-	driveID string,
-	restoreFolders []string,
+	driveID, driveRootID string,
```
vkamra, line 624: Is the driveRootID addition here required or is that an optimization? It looks like the latter but want to confirm I'm reading the PR right?

vkamra:
@ryanfkeepers - changes LGTM. Can you add a test case with "folder" in the path to `TestCreateGetDeleteFolder` or just another test in that suite that calls `CreateRestoreFolders` with a path that includes `folder`

src/internal/connector/onedrive/restore.go
vkamra: Prefer we move that to a separate PR if indeed it's an optimization - there's a lot of lines of code changes for that.

vkamra (approved):

src/internal/connector/onedrive/folder_cache.go
ryanfkeepers: maybe, maybe not.  It's not a high priority concern.

src/internal/connector/onedrive/restore.go
ryanfkeepers: It's required in order to look up/create the top folder.

ryanfkeepers:
> Can you add a test case with "folder" in the path to TestCreateGetDeleteFolder or just another test in that suite that calls CreateRestoreFolders with a path that includes folder

Yup, will do.  Was thinking about adding this to the expected set of folders we require for every CI user account.  I at least want to extend that expectation to have some nested folders.  Not sure it needs the folder folder, but maybe two birds with one stone that way?

src/internal/connector/onedrive/restore.go
vkamra: Hmm - I'm missing something but why can't we continue to call `service.Client().DrivesById(driveID).Root().Get(ctx, nil)` in this method?

src/internal/connector/onedrive/restore.go
```diff
@@ -541,43 +562,112 @@ func restoreV6File(
...
+		location = location.Append(folder)
+		ictx := clues.Add(
+			ctx,
+			"creating_restore_folder", folder,
+			"restore_folder_location", location,
+			"parent_of_restore_folder", parentFolderID)
 
-	parentFolderID := ptr.Val(driveRoot.GetId())
-	ctx = clues.Add(ctx, "drive_root_id", parentFolderID)
+		if fc, ok := fc.Get(location); ok {
```
ashmrtn, line 642: nit: this should use a different name than `fc` for the first variable assigned

src/internal/connector/onedrive/folder_cache.go
```diff
@@ -0,0 +1,28 @@
...
+	cache map[string]models.DriveItemable
+}
+
+func NewFolderCache() *folderCache {
+	return &folderCache{
+		cache: map[string]models.DriveItemable{},
+	}
+}
+
+func (c *folderCache) Get(loc *path.Builder) (models.DriveItemable, bool) {
```
ashmrtn, line 21: it doesn't seem like this and `Set` need to be exported?

ashmrtn (commented):
mostly just skimmed. A few nits

src/internal/connector/onedrive/folder_cache.go
ryanfkeepers: Of note here, from Georgi's point, lookups by folder name are a little vulnerable.  It'd be nice to run this as a two step enumerate-then-check cache of folders, much like exchange does with CacheContainer.

src/internal/connector/onedrive/restore.go
ryanfkeepers: We could, I moved it to avoid redundancy.  We're calling it N times here (where N is equal to folder count), when we only need to call it once.  Not so much a performance concern as it is recognizing the behavior of the world.

PR #3222, opened 2023-04-26
Add nil check for ServicesEnabled

pandeyabs:
Prevent potential panic if ui.ServicesEnabled field is nil. 

---

- [x ] :no_entry: No

#### Type of change

- [ x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x ] :muscle: Manual

### DIFF
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -100,7 +100,7 @@ func newUserInfo() *UserInfo {
 // ServiceEnabled returns true if the UserInfo has an entry for the
 // service.  If no entry exists, the service is assumed to not be enabled.
 func (ui *UserInfo) ServiceEnabled(service path.ServiceType) bool {
-	if ui == nil || len(ui.ServicesEnabled) == 0 {
+	if ui == nil || ui.ServicesEnabled == nil || len(ui.ServicesEnabled) == 0 {
 		return false
 	}
 


src/pkg/services/m365/api/users.go
```diff
@@ -100,7 +100,7 @@ func newUserInfo() *UserInfo {
 // ServiceEnabled returns true if the UserInfo has an entry for the
 // service.  If no entry exists, the service is assumed to not be enabled.
 func (ui *UserInfo) ServiceEnabled(service path.ServiceType) bool {
-	if ui == nil || len(ui.ServicesEnabled) == 0 {
+	if ui == nil || ui.ServicesEnabled == nil || len(ui.ServicesEnabled) == 0 {
```
ryanfkeepers, line 103: This is unnecessary.  maps are nil-safe in golang.  See: https://go.dev/play/p/f31OX8MbUx5

pandeyabs: Hmm, if someone defines ui := &UserInfo{}, then ui.ServicesEnabled would be nil. 
https://go.dev/play/p/cOF4pR4T47U

ryanfkeepers: nil, yes. But still safe for usage.  see this amendment to your code, that uses the map in the case of `ui := &UserInfo{}`.

https://go.dev/play/p/Li1_fqfncaA

pandeyabs: hmm, interesting. I was seeing a segmentation fault with existing code if I call `ServiceEnabled` with `UserInfo{}`. My guess was that it's due to len(ui.ServicesEnabled), but it doesn't repro it in go playground. 

Thanks for the discussion. I'll close this one, as you said it's safe for usage.

Issue #3077, opened 2023-04-10
Add hidden CLI command to run kopia maintenance


ashmrtn:
Some kopia maintenance operations needs to run every now and then. This helps keep the repo clean and removes data that is no longer referenced by snapshots

Corso should have a command to hook into kopia's snapshot maintenance. Corso should be able to control whether Kopia does full maintenace or quick maintenance

PR #3223, opened 2023-04-26
Allow overriding the default username/hostname in kopia

ashmrtn:
This allows setting the username/hostname
kopia will use for maintenance. This
information is recorded in the kopia config
file during connect and reused during open

If no values are given, kopia pulls the
values from the OS

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3077

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/conn.go
+++ b/src/internal/kopia/conn.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 	"github.com/pkg/errors"
 
+	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/storage"
 )
 
@@ -69,7 +70,7 @@ func NewConn(s storage.Storage) *conn {
 	}
 }
 
-func (w *conn) Initialize(ctx context.Context) error {
+func (w *conn) Initialize(ctx context.Context, opts control.Options) error {
 	bst, err := blobStoreByProvider(ctx, w.storage)
 	if err != nil {
 		return clues.Wrap(err, "initializing storage")
@@ -92,6 +93,7 @@ func (w *conn) Initialize(ctx context.Context) error {
 
 	err = w.commonConnect(
 		ctx,
+		opts,
 		cfg.KopiaCfgDir,
 		bst,
 		cfg.CorsoPassphrase,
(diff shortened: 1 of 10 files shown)


src/internal/kopia/conn.go
```diff
@@ -131,16 +134,21 @@ func (w *conn) Connect(ctx context.Context) error {
...
+	opts control.Options,
 	configDir string,
 	bst blob.Storage,
 	password, compressor string,
 ) error {
-	var opts *repo.ConnectOptions
+	kopiaOpts := &repo.ConnectOptions{
+		ClientOptions: repo.ClientOptions{
+			Username: opts.User,
+			Hostname: opts.Host,
```
ryanfkeepers, line 145: is it okay for these to be zero valued?  Do we need a sane default established in `control.Default()`?  (wrote this before seeing the tests below; seems safe, but still want to sanity check.)

Likewise, is there anything we need to do for common integration testing cases here?

ryanfkeepers (approved):
Behavior looks good; would prefer renaming before merger.

src/internal/kopia/conn.go
ashmrtn: fine for them to have a zero value, that's essentially what's happening right now

If we do integration testing I'd much rather have something mock KW and use the mock to ensure the passed in values are what are expected. The tests in this PR already show that KW will behave as expected when given different values and there's no multi-component complexity here that would require a non-mocked integration test (i.e. it's not like details merging where several different components touch the data and we're only really interested in the end result)

As of now, we don't have any infrastructure for mocking KW or other things in corso's kopia package

src/pkg/control/options.go
```diff
@@ -13,6 +13,8 @@ type Options struct {
 	SkipReduce         bool            `json:"skipReduce"`
 	ToggleFeatures     Toggles         `json:"toggleFeatures"`
 	Parallelism        Parallelism     `json:"parallelism"`
+	User               string          `json:"user"`
+	Host               string          `json:"host"`
```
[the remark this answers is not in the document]
ashmrtn: I'm fine with adding `Repo` to it. I'd prefer to stay away from mentioning kopia directly in places and/or exporting anything kopia related to the SDK level because really the fact that corso is built on kopia is an implementation detail

PR #3224, opened 2023-04-26
Create wrapper around kopia maintenance

ashmrtn:
Create a wrapper function around the kopia
maintenance operation. Will allow users to
run maintenance, though they still need to
be careful not to run maintenance
concurrently if they override the username
and hostname that kopia uses

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3077

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/wrapper.go
```diff
@@ -515,3 +518,118 @@ func isErrEntryNotFound(err error) bool {
...
+	p *maintenance.Params,
+	userAtHost string,
+) error {
+	// This will source user/host from the kopia config file or fallback to
+	// fetching the values from the OS.
+	p.Owner = userAtHost
+	// Disable automatic maintenance for now since it can start matching on the
+	// user/host of at least one machine now.
+	p.QuickCycle.Enabled = false
+	p.FullCycle.Enabled = false
```
ashmrtn, line 623: @vkamra should we disable these by default or leave them enabled?

src/pkg/control/options.go
```diff
@@ -112,3 +112,19 @@ type Toggles struct {
...
+
+// ---------------------------------------------------------------------------
+// Repo Maintenance flags
+// ---------------------------------------------------------------------------
+
+type Safety string
+
+const (
+	FullSafety Safety = "full"
+	NoSafety   Safety = "none"
```
ryanfkeepers, line 124: These are understandable comparison to their kopia usage, but don't have much context when read on their own.  Further, I'm not sure we want all services to observe a common notion of `safety`, as those concepts could be critically different between services.

It's a tough call, because I don't want to expose kopia configuration at higher levels, but kopia _does_ feel like the correct owner here.  I think it would be reasonable to have it export these two enums.  Understandably, if we don't want to do that, I think some other form of namespacing like `control.Maintenance{ Safety: enum }` or `control.Repository{ MaintenanceSafety: enum }` would help identify ownership, and prevent this from accidentally spilling over into unintended spaces.

src/internal/kopia/wrapper.go
```diff
@@ -515,3 +518,118 @@ func isErrEntryNotFound(err error) bool {
...
 }
+
+func (w Wrapper) Maintenance(
+	ctx context.Context,
+	safety control.Safety,
+	quickMaintenance, force bool,
+) error {
+	kopiaSafety, err := translateSafety(safety)
+	if err != nil {
+		return clues.Wrap(err, "getting safety level")
```
ryanfkeepers, line 529: Nit
```suggestion
		return clues.Wrap(err, "identifying safety level")
```

src/internal/kopia/wrapper.go
```diff
@@ -515,3 +518,118 @@ func isErrEntryNotFound(err error) bool {
 	return strings.Contains(err.Error(), "entry not found") &&
 		!strings.Contains(err.Error(), "parent is not a directory")
 }
+
+func (w Wrapper) Maintenance(
+	ctx context.Context,
+	safety control.Safety,
+	quickMaintenance, force bool,
```
ryanfkeepers, line 525: Feels like all three of these could get rolled into some kind of `control.RepoMaintenance`

src/internal/kopia/wrapper.go
```diff
@@ -515,3 +518,118 @@ func isErrEntryNotFound(err error) bool {
...
+
+	// Below write session options pulled from kopia's CLI code that runs
+	// maintenance.
+	err = repo.DirectWriteSession(
+		ctx,
+		dr,
+		repo.WriteSessionOptions{
+			Purpose: "Corso maintenance",
+		},
+		func(ctx context.Context, dw repo.DirectRepositoryWriter) error {
```
ryanfkeepers, line 555: nit, could this get moved into a var declaration above this?  Imo, declaring funcs within a func call isn't as readable as the following.  (plus, reduces indent depth, and makes it easier to declare if we ever decide to not make it an ad-hoc func for testing purposes)
```go
ctx = ...
dr, ok := ...
if !ok {...}

theFunc := func(...)

err = repo.DirectWriteSession(ctx, dr, {}, theFunc)
```

ryanfkeepers (commented):
nits on naming/enum ownership.  behavior generally looks good.  Will let Vaibhav take approval duties, since there are outstanding questions for him.

src/internal/kopia/wrapper.go
vkamra: Let's keep these disabled till we have more experience with this.

src/internal/kopia/wrapper.go
```diff
@@ -515,3 +518,118 @@ func isErrEntryNotFound(err error) bool {
...
+			params, err := maintenance.GetParams(ctx, w.c)
+			if err != nil {
+				return clues.Wrap(err, "getting maintenance user/host").WithClues(ctx)
+			}
+
+			// Need to do some fixup here as the user/host may not have been set.
+			if len(params.Owner) == 0 || (params.Owner != currentOwner && force) {
+				// TODO(ashmrtn): Print some message if the recorded owner doesn't match
```
vkamra, line 563: Address TODO before merging

vkamra:
Changes here LGTM

src/pkg/control/repository/repo.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Maintenance flags
+// ---------------------------------------------------------------------------
+
+type MaintenanceType int
+
+// Can't be reordered as we rely on iota for numbering.
+//
```
ryanfkeepers, line 22: Any reason in particular to use iota over string enums here?

src/internal/observe/observe.go
```diff
@@ -151,7 +151,17 @@ func Message(ctx context.Context, msgs ...any) {
...
-	logger.Ctx(ctx).Info(loggable)
+	MessageWithStructure(ctx, plain, loggable)
+}
+
+// MessageWithStructure is used to display a progress message as well as a
+// structured log.
+func MessageWithStructure(
+	ctx context.Context,
+	msg, logMsg string,
+	logArgs ...any,
```
ryanfkeepers, line 162: I don't believe this func is necessary.  `observe.Message` should already be set up to handle hiding arbitrary strings values.  Ie:

```
// instead of this
observe.MessageWithStructure(
  ctx,
  "updating maintenance user/host to "+currentOwner,
  "updating maintenance user/host",
  "new_maintenance_owner",
  clues.Hide(currentOwner))

// do this
observe.MessageWithStructure(
  ctx,
  "updating maintenance user/host to",
  clues.Hide(currentOwner))
  
// or, if you want to add structure explicitly, in addition to the message
observe.MessageWithStructure(
  clues.Add(ctx, "new_maintenance_owner", clues.Hide(currentOwner)),
  "updating maintenance user/host to",
  clues.Hide(currentOwner))

ryanfkeepers (approved):

src/pkg/control/repository/repo.go
ashmrtn: I kind of like how iotas give us the ability to have meaningful default (zero) values while also allowing string output if we use generated code. I feel that having values in a struct (for namespacing) makes it easier to have unitialized values than passing parameters to a function explicitly

src/internal/observe/observe.go
ashmrtn: mmm part of that was so we could have structured logging of stuff as well. It's kind of sixes though as long as data is hidden

PR #3225, opened 2023-04-26
Create basic maintenance operation

ashmrtn:
Add a maintenance operation to run
kopia maintenance

Using this instead of calling kopia
directly will allow us to hook into
metrics reporting in a more
consistent manner if we want metrics

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3077

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/maintenance.go
+++ b/src/internal/operations/maintenance.go
@@ -11,15 +11,14 @@ import (
 	"github.com/alcionai/corso/src/internal/kopia"
 	"github.com/alcionai/corso/src/internal/stats"
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/control/repository"
 )
 
 // MaintenanceOperation wraps an operation with restore-specific props.
 type MaintenanceOperation struct {
 	operation
-	Results          MaintenanceResults
-	safety           control.Safety
-	force            bool
-	quickMaintenance bool
+	Results MaintenanceResults
+	mOpts   repository.Maintenance
 }
 
 // MaintenanceResults aggregate the details of the results of the operation.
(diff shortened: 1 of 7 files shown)


src/internal/operations/maintenance.go
```diff
@@ -0,0 +1,76 @@
...
+	stats.StartAndEndTime
+}
+
+// NewMaintenanceOperation constructs and validates a restore operation.
+func NewMaintenanceOperation(
+	ctx context.Context,
+	opts control.Options,
+	kw *kopia.Wrapper,
+	safety control.Safety,
+	quickMaintenance, force bool,
```
ashmrtn, line 36: is having `quickMaintenace` be a bool sufficiently generic? Should we change it to an enum instead so that we can add more values in the future if we need to?

This is going to be the SDK-level API so it would be best to try to make it fit our future needs if we can

ryanfkeepers: Enums over flags are always a good idea, imo.

src/internal/operations/maintenance.go
```diff
@@ -0,0 +1,76 @@
...
+	err = op.operation.kopia.Maintenance(
+		ctx,
+		op.safety,
+		op.quickMaintenance,
+		op.force)
+	if err != nil {
+		return clues.Wrap(err, "running maintenance operation")
+	}
+
+	return nil
```
ryanfkeepers, line 75: No plans for storing/retrieving the results?  I assume not, since we only currently do that for backups, but wanted to check.

src/internal/operations/maintenance.go
```diff
@@ -0,0 +1,76 @@
...
+
+	op.Results.StartedAt = time.Now()
+
+	// TODO(ashmrtn): Send usage statistics?
+
+	err = op.operation.kopia.Maintenance(
+		ctx,
+		op.safety,
+		op.quickMaintenance,
+		op.force)
```
ryanfkeepers, line 70: Can this get added as an interface in `operations/inject.go`, please?  Would follow the pattern we have with backup/restore, and allow you to write an easily mocked unit test at this layer and above.

src/internal/operations/maintenance.go
```diff
@@ -0,0 +1,76 @@
...
+		// TODO(ashmrtn): Send success/failure usage stat?
+
+		op.Results.CompletedAt = time.Now()
+	}()
+
+	op.Results.StartedAt = time.Now()
+
+	// TODO(ashmrtn): Send usage statistics?
+
+	err = op.operation.kopia.Maintenance(
```
ryanfkeepers, line 66: Nit on naming: `RepoMaintenance`, maybe?  Operations interfaces with a lot of dependencies.  Who knows when a backup producer or other package might have something that overlaps with the notion of maintenance.

ryanfkeepers (approved):
lgtm.  `Maintenance` interface is a want, and I wouldn't mind seeing a sanity integration/unit test.

src/internal/operations/maintenance.go
```diff
@@ -0,0 +1,76 @@
...
+	force            bool
+	quickMaintenance bool
+}
+
+// MaintenanceResults aggregate the details of the results of the operation.
+type MaintenanceResults struct {
+	stats.StartAndEndTime
+}
+
+// NewMaintenanceOperation constructs and validates a restore operation.
```
pandeyabs, line 30: restore operation? is that a typo?

pandeyabs (commented):
Changes look good to me!

PR #3226, opened 2023-04-26
Create CLI command for maintenance

ashmrtn:
Creates hidden CLI command to run
maintenance.

Flag names and descriptions can be
updated if something else would fit
better

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3077

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/repo/repo.go
```diff
@@ -84,3 +103,65 @@ func connectCmd() *cobra.Command {
 func handleConnectCmd(cmd *cobra.Command, args []string) error {
 	return cmd.Help()
 }
+
+func maintenanceCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   maintenanceCommand,
+		Short: "Run maintenance on an existing repository",
+		Long:  `Run maintenance on an existing repository to keep it running smoothly`,
```
ryanfkeepers, line 111: Not sure I have a good suggestion for replacement, but `to keep it running smoothly` doesn't strike me as useful; it's sort of a generic extension to the notion of "maintenance".  Also, Corso always runs smoothly.  Not a bump or pothole in sight, amirite? 😉 

Spitballing: `to keep memory usage low`?  `to keep processing time down`?  `to compress backup history`?

src/cli/utils/flags.go
```diff
@@ -167,6 +177,36 @@ func AddSiteFlag(cmd *cobra.Command) {
...
+	cobra.CheckErr(fs.MarkHidden(QuickMaintenanceFN))
+}
+
+func AddForceMaintenanceFlag(cmd *cobra.Command) {
+	fs := cmd.Flags()
+	fs.BoolVar(
+		&ForceMaintenanceFV,
+		ForceMaintenanceFN,
+		false,
+		"Always run maintenance even if it may be unsafe")
```
ryanfkeepers, line 206: nit: `always run` sounds like a persistent, cron-like configuration.
```suggestion
		"Force maintenance even if it may be unsafe")
```

src/cli/utils/flags.go
```diff
@@ -167,6 +177,36 @@ func AddSiteFlag(cmd *cobra.Command) {
...
+	cobra.CheckErr(fs.MarkHidden(MaintenanceSafetyFN))
+}
+
+func AddQuickMaintenanceFlag(cmd *cobra.Command) {
+	fs := cmd.Flags()
+	fs.BoolVar(
+		&QuickMaintenanceFV,
+		QuickMaintenanceFN,
+		false,
+		"Run quick maintenance that won't clean up and optimize the repo as much")
```
ryanfkeepers, line 196:
```suggestion
		"Run a faster maintenance that does minimal clean-up and optimization")
```

src/cli/utils/flags.go
```diff
@@ -167,6 +177,36 @@ func AddSiteFlag(cmd *cobra.Command) {
 		"Backup data by site URL; accepts '"+Wildcard+"' to select all sites.")
 }
 
+func AddMaintenanceSafetyFlag(cmd *cobra.Command) {
+	fs := cmd.Flags()
+	fs.StringVar(
+		&MaintenanceSafetyFV,
+		MaintenanceSafetyFN,
+		string(control.FullSafety),
+		"Disable safety measures to run maintenance faster. Improper usage can break a repository")
```
ryanfkeepers, line 186: admittedly, there's a part of me that wants it to say `but can bork the repository`
```suggestion
		"Set the level of safety measures; lower safety makes maintenance faster, but can break the repository")
```

src/cli/repo/repo.go
```diff
@@ -84,3 +103,65 @@ func connectCmd() *cobra.Command {
...
+		return print.Only(ctx, err)
+	}
+
+	return nil
+}
+
+func validateMaintenanceFlags(s control.Safety) error {
+	_, ok := control.SafetyValues[s]
+	if !ok {
+		values := maps.Keys(control.SafetyValues)
```
ryanfkeepers, line 152: `values = keys` hits the ocd part of me hard.
```suggestion
		safetys := maps.Keys(control.SafetyValues)
```

src/cli/utils/flags.go
```diff
@@ -58,6 +63,11 @@ const (
 	FileCreatedBeforeFN  = "file-created-before"
 	FileModifiedAfterFN  = "file-modified-after"
 	FileModifiedBeforeFN = "file-modified-before"
+
+	// Maintenance stuff.
+	MaintenanceSafetyFN = "safety"
+	QuickMaintenanceFN  = "quick"
```
ryanfkeepers, line 69: Since we're entertaining an enum instead of a flag for this elsewhere, I'd prefer `Mode` flag that defaults to `slow` (or whatever not-quick will be).  If we never add a third value, no big deal.  If we do, then it's not a breaking change to the cli.
```suggestion
	MaintenanceModeFN  = "mode"
```

ryanfkeepers (approved):
code generally looks good, just nits.

ryanfkeepers:
adding @gmatev for final review on language.

src/cli/repo/repo.go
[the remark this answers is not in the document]
gmatev: For the more technical Corso audience something like this is ok.

src/cli/utils/flags.go
gmatev: Some of these are Kopia terms and it may make sense to keep them.

src/cli/utils/flags.go
gmatev: We should look at some of the Kopia language here and make this more specific. For Kopia, safety means that you need to make sure no concurrent operations are happening 

https://kopia.io/docs/advanced/maintenance/

src/cli/utils/flags.go
gmatev: We can benefit from some of the Kopia docs here https://kopia.io/docs/advanced/maintenance/

src/cli/utils/flags.go
```diff
@@ -36,6 +37,10 @@ var (
 
 	// for selection of data by category.  eg: `--data email,contacts`
 	CategoryDataFV []string
+
+	MaintenanceSafetyFV string
```
pandeyabs, line 41: Should we add some comments explaining the various safety levels?

gmatev:
Several points:
* For the descriptions of the operations, we can use some of the Kopia language which is pretty good. 
* I'm not seeing the corresponding docs changes:
   * Changes that may be needed to publish the auto-gen CLI docs
   * Any other hand written docs that may be helpful to better educate on the topic of maintenance. 

We can probably split these in separate PRs but they should be a fast follow

src/cli/utils/flags.go
pandeyabs: `Force maintenance. Caution: This may be unsafe` ?

gmatev:
Also the maintenance action should be logged in telemetry as well.

src/cli/repo/repo.go
vkamra: Run maintenance on an existing repository to optimize performance and storage use.

src/cli/utils/flags.go
vkamra: @ashmrtn - let's revisit this. We have safety options in the SDK but let's not expose in CLI yet.

ashmrtn:
Can I get a final review on language in this PR? Also what follow ups are required? I think we were going to punt on doing doc updates since this is marked as a hidden command at the moment. Telemetry does need to be added from the sounds of it

gmatev:
> Can I get a final review on language in this PR? Also what follow ups are required? I think we were going to punt on doing doc updates since this is marked as a hidden command at the moment. Telemetry does need to be added from the sounds of it

What are the flag values that the `mode` flag takes? Probably should be included in the inline help. 

And yes, we would ideally log a telemetry event when this occurs. It's ok to defer as a separate issue and PR if more convenient.

ashmrtn:
I've merged some telemetry changes into this PR. Going to queue this on main since it seems like the loose ends have been tied up

PR #3227, opened 2023-04-26
create onedrive custom drive

neha-Gupta1:
Run Sanity test with custom onedrive data

- [x] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix
#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual


.github/workflows/sanity-test.yaml
```diff
@@ -130,7 +132,7 @@ jobs:
 
           resultjson=$(sed -e '1,/Completed Backups/d' $TEST_RESULT/backup_exchange.txt ) 
 
-          if  [[ $( echo $resultjson | jq -r '.[0] | .errorCount') -ne 0 ]]; then
+          if  [[ $( echo $resultjson | jq -r '.[0] | .stats.errorCount') -ne 0 ]]; then
```
vkamra, line 135: @neha-Gupta1 - is this the fix for the unbounded variable error?

neha-Gupta1: Yes

vkamra: 👍 - got it. Suggest we just push a separate PR for this ASAP just to get the tests moving. Also - maybe disable the failing test in that PR for now. That way we have a green run with less tests vs a red run with more tests.

The failing test can be re-enabled once the rest of this current PR merges.

neha-Gupta1: @vkamra, this is the PR for it - https://github.com/alcionai/corso/pull/3228

neha-Gupta1:
Passed run- https://github.com/alcionai/corso/actions/runs/4807405458/jobs/8556131744?pr=3227

.github/workflows/sanity-test.yaml
meain: I think GitHub is showing this as it is diffing this with an old version due to the fact that you have a merge conflict. If you could fix and rebase, I think we will have a better diff to review.

ryanfkeepers: Taken care of.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+) {
+	ctx, flush := tester.NewContext()
+	defer flush()
+
+	// TODO: fit the destination to the containers
+	dest := control.DefaultRestoreDestination(common.SimpleTimeTesting)
+	dest.ContainerName = destFldr
+	print.Infof(ctx, "Restoring to folder %s", dest.ContainerName)
+
+	d, _ := gc.Service.Client().UsersById(resourceOwner).Drive().Get(ctx, nil)
```
ryanfkeepers, line 278: This should call the drive api for re-use, instead of building the call here in place.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+	count int,
+	errs *fault.Bus,
+) (
+	*details.Details,
+	error,
+) {
+	ctx, flush := tester.NewContext()
+	defer flush()
+
+	// TODO: fit the destination to the containers
```
ryanfkeepers, line 273: not sure what this todo means.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+	driveID := ptr.Val(d.GetId())
+
+	var (
+		cols []onedriveColInfo
+
+		rootPath = []string{
+			"drives",
+			driveID,
+			"root:",
+		}
```
ryanfkeepers, line 288: nit, this isn't a very long line
```suggestion
		rootPath = []string{ "drives", driveID, "root:"}
```

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+		var col onedriveColInfo
+
+		// basic folder and file creation
+		if i == 0 {
+			col = onedriveColInfo{
+				pathElements: rootPath,
+				files: []itemData{
+					{
+						// Test restoring a file that doesn't inherit permissions.
+						name: fmt.Sprintf("testFile-%s-1-%d", currentTime, i),
```
ryanfkeepers, line 323: this name isn't very human friendly.  I think longer, but more readable, is acceptable here.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+
+		now         = time.Now()
+		currentTime = fmt.Sprintf("%d-%d-%d", now.Hour(), now.Minute(), now.Second())
+	)
+
+	for i := 0; i < count; i++ {
+		var col onedriveColInfo
+
+		// basic folder and file creation
+		if i == 0 {
```
ryanfkeepers, line 317: This pattern (including the modulus below) is more clever than helpful.  A simpler pattern would be to say ("for each count in --count, the factory will construct a set of files that adequately express permission").  IE: if `--count 1`, then you create 4 files (I think that's the standard, right?).  if `--count 2`, you repeat the set to make 8 files, and so on.  Additional fiddling with complexity beyond that isn't of any benefit.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
 
 	return collections, nil
 }
+
+type permData struct {
```
ryanfkeepers, line 203: These are all copied from `graph_connector_onedrive_test.go`, right?  Instead of duplication, I'd rather see those moved into something like `onedrive/stub/collections.go` (or testdata, if not stub) and exported to both the test and to this file.  Normally we wouldn't want to export testdata packages to non-test code, but since this is an isolated helper, not part of the corso binary, it'll be safe to do.

By sharing this code we have much less duplication to track and maintain, and won't have this factory getting out of sync across version changes.  I bet that consolidation alone would bring the changes in this file from ~700 lines down to one or two hundred.

For the sake of time, we can move forward with the duplication as it exists here, but the refactor should come quickly after that.

src/cmd/factory/impl/common.go
```diff
@@ -179,3 +199,547 @@ func buildCollections(
...
+	for i := 0; i < count; i++ {
+		var col onedriveColInfo
+
+		// basic folder and file creation
+		if i == 0 {
+			col = onedriveColInfo{
+				pathElements: rootPath,
+				files: []itemData{
+					{
+						// Test restoring a file that doesn't inherit permissions.
```
ryanfkeepers, line 322: These are not tests, and should not suggest that they involve testing.

ryanfkeepers (approved):
Lots of refactoring needed here.  But since this is a non-critical code path, I'm going to approve and let that refactoring happen as a follow-up to this PR.  Please take care of the smaller nits for now, and focus on the larger refactors after this goes in.

src/cmd/factory/impl/common.go
neha-Gupta1: I have updated the name to something like -`file-count-1-at-2023-April-30-13-27-18`. If its still not that readable we can work on it.

src/cmd/factory/impl/common.go
neha-Gupta1: I will get the PR with this refactoring ready soon.

PR #3232, opened 2023-04-26
Check and pass state of delta request availability

meain:
To be merged into https://github.com/alcionai/corso/pull/3212

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -57,7 +58,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 		return nil, nil, clues.Stack(err).WithClues(ctx)
 	}
 
-	serviceEnabled, err := checkServiceEnabled(
+	serviceEnabled, deltaAvailable, err := checkServiceEnabled(
```
ryanfkeepers, line 61: naming nit: `deltaAvailable` sounds like you're flagging "we have a prior delta available to use".  A small fix would be to name it something like `mustNotDeltaQuery` or `canMakeDeltaQueries`.

A somewhat larger lift, butt one that might be more flexible, would be to retrieve the entire `api.UserInfo` here and pass it down (no strong feelings on how far to pass it down).  Adding a method like `userInfo.CanMakeDeltaQueries()` will allow us to establish an interface (if wanted), and will generally provide easy portability/extensibility by centralizing the ownership of delta-able identification within the api.  It also give us a userInfo struct in case we need to examine it for any other downstream/future case (mailbox settings, etc).  Finally, it exposes the check for SDK users to easily consume.

Either way works, I leave it up to you.

src/pkg/services/m365/api/users.go
```diff
@@ -58,6 +58,7 @@ type MailboxInfo struct {
 	Language                   Language
 	WorkingHours               WorkingHours
 	ErrGetMailBoxSetting       []error
+	DeltaAvailable             bool
```
ryanfkeepers, line 61: `DeltaAvailble` is an inference on the quota issue.  I think this will be easier to support if we expose the case as its own boolean (ie: `DeltaAvailble` -> `QuotaExceeded`) and provide a func to evaluate the mailbox state, eg:
```go
func (ui UserInfo) CanMakeDeltaQueries() bool { return QuotaExceeded }
```
If a second condition comes along which also nullifies delta query usage, we would get both a consolidated check (the func) and clear identification of which cases caused the evaluation (the flags).

src/internal/connector/exchange/service_iterators.go
```diff
@@ -114,7 +116,9 @@ func filterContainersAndFillCollections(
 			qp.ResourceOwner.ID(),
 			cID,
 			prevDelta,
-			ctrlOpts.ToggleFeatures.ExchangeImmutableIDs)
+			ctrlOpts.ToggleFeatures.ExchangeImmutableIDs,
+			deltaAvailable,
+		)
```
ryanfkeepers, line 121: nit:
```suggestion
			deltaAvailable)
```

src/internal/connector/exchange/data_collections.go
```diff
@@ -306,7 +309,9 @@ func createCollections(
 		scope,
 		dps,
 		ctrlOpts,
-		errs)
+		deltaAvailable,
+		errs,
+	)
```
ryanfkeepers, line 314: Nit
```suggestion
		errs)
```

src/internal/connector/exchange/data_collections_test.go
```diff
@@ -604,6 +609,7 @@ func (suite *DataCollectionsIntegrationSuite) TestEventsSerializationRegression(
 				test.scope,
 				DeltaPaths{},
 				control.Defaults(),
+				true,
```
ryanfkeepers, line 612: I think we need some `false` case tests in regression to be fully comfortable with this change.

ryanfkeepers (approved):
Some nits worth fixing up, and additional testing needed around non-delta-able regression.  But generally lgtm, so approved to unblock.

src/internal/connector/data_collections.go
ashmrtn: I would say start with the low lift for now. I'm not sure adding more information to the backup path is what we want in the long run. There's already quite a bit going on

src/pkg/services/m365/api/users.go
```diff
@@ -295,6 +297,27 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
 	userInfo.Mailbox = mbxInfo
 
+	if mfs != nil {
+		mf := mfs.GetValue()[0] // we will always have one
+		options := &users.ItemMailFoldersItemMessagesDeltaRequestBuilderGetRequestConfiguration{
+			QueryParameters: &users.ItemMailFoldersItemMessagesDeltaRequestBuilderGetQueryParameters{
+				Top: ptr.To[int32](1), // just one item is enough
+			},
+		}
+		_, err = c.stable.Client().
```
ashmrtn, line 307: hm, looking at this with fresh eyes I'm wondering if we're missing something. This is built on the assumption that the user won't accrue more data in their mailbox during the backup process. Since we have no number to denote how much data is in the mailbox or how full the mailbox is now we're unable to change behavior if we think the mailbox may get full during the backup

It's also unclear if Graph will start returning errors if the mailbox gets full during the backup, even if that happens midway through enumerating items in a folder

Overall, it's hard to say if this will solve _all_ mailbox full failures or just the subset that occur when the mailbox is full prior to launching a backup

To be clear, it does sort of seem like checking up front for fullness simplifies the code (and what's happening on a conceptual level) a bit if we can keep it. However I'm unsure both on the failure modes we get with this approach and what the acceptance criteria for this task is

src/internal/connector/data_collections.go
meain: I don't think we should pass down entire `userInfo` just yet. I was thinking we could do that once we have the requirement where we have to pass down more items which could also help us make more sense of how far to pass it down. I've made the other changes.

src/internal/connector/exchange/data_collections_test.go
meain: Augmented existing tests to test both in e9b5de8c3

meain:
I'm merging this in as this looks good. If it requires any more changes, I can add it in the other PR.

PR #3236, opened 2023-04-26
Don't use RepoRef in selector reduction

### LINKED ISSUE #3194: Use LocationRef/ParentPath in selectors
LocationRef is now populated for all services during backup but selectors still relies on extracting information from the RepoRef in many cases (OneDrive, SharePoint, at least some of Exchange)

Selectors should be updated to use the LocationRef so the RepoRef can change without breaking things

Since we have multiple backup versions with different amounts of information, whether the RepoRef or LocationRef is used will depend on which is populated for a given entry

ashmrtn:
A few high-level things of note:
* things will no longer match on folder ID. Folder IDs weren't displayed
  to the user via CLI and SDK consumers have no insight into folder IDs
  so this shouldn't be an issue
* OneDrive and SharePoint match on ParentPath (derived from
  LocationRef). ParentPath *does not* include root: in the path

Not matching on folder ID should be the only user-visible change in
this PR

First commit contains the required logic changes. All other changes
are test updates

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3194

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/backup/details/testdata/testdata.go
+++ b/src/pkg/backup/details/testdata/testdata.go
@@ -1,7 +1,7 @@
 package testdata
 
 import (
-	stdpath "path"
+	"strings"
 	"time"
 
 	"github.com/alcionai/corso/src/pkg/backup/details"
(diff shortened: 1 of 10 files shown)


src/pkg/selectors/sharepoint_test.go
```diff
@@ -204,67 +205,123 @@ func (suite *SharePointSelectorSuite) TestToSharePointRestore() {
...
+		itemElems3 = []string{"folderD", "folderE"}
+		pairAC     = "folderA/folderC"
+		pairGH     = "folderG/folderH"
+		item       = toRR(
+			path.LibrariesCategory,
+			"sid",
+			append(
+				append(
+					[]string{},
+					prefixElems...),
```
ryanfkeepers, line 240: Nit: for brevity
```suggestion
				slices.Clone(prefixElems),
```

src/pkg/backup/details/testdata/testdata.go
```diff
@@ -33,7 +33,97 @@ func mustAppendPath(p path.Path, newElement string, isItem bool) path.Path {
...
+	return p.loc.Append(strings.TrimSuffix(lastElem, folderSuffix)).String()
+}
+
+func mustPathRep(ref string, isItem bool) repoRefAndLocRef {
+	res := repoRefAndLocRef{}
+	tmp := mustParsePath(ref, isItem)
+
+	// Now append stuff to the RepoRef elements so we have distinct LocationRef
+	// and RepoRef elements to simulate using IDs in the path instead of display
+	// names.
```
ryanfkeepers, line 97: I'm on board with the future this sets up, but do you have any concerns about it allowing current cases that depend on identical naming to slip past undetected?

src/pkg/backup/details/testdata/testdata.go
```diff
@@ -33,7 +33,97 @@ func mustAppendPath(p path.Path, newElement string, isItem bool) path.Path {
...
+
+	res.RR = rr
+	res.loc = locFromRepo(rr, isItem)
+
+	return res
+}
+
 const (
+	folderSuffix = ".d"
+	fileSuffix   = ".f"
```
ryanfkeepers, line 125: seeing this in both details and in selectors kinda makes me want a shared owner.  tester?  That's probably too heavy handed of me.

ryanfkeepers (approved):
Looks pretty good, just a couple nits/questions.  Thanks for all the effort here.

src/pkg/backup/details/testdata/testdata.go
ashmrtn: I think getting tests in selectors and the CLI to use this is probably "good enough" for the moment as those are the things that most heavily exercise different paths. There will likely be more changes in the future for details merging, but we'll get there when we actually change the RepoRef to something else. Doing those changes now would disconnect them from the reason it's being changed

src/pkg/backup/details/testdata/testdata.go
ashmrtn: I think long-term we should [re]move a good chunk of tests for selectors. TBH I'm not interested in testing every function. I'm interested in the system working for the given inputs and outputs (black box testing). Testing every function in the selectors package provides minimal benefits as it ends up testing large chunks of the code multiple times while generating lots of work to make even minor changes (as seen in this PR)

Moving tests that check edge cases to the test data in the details package would go a long way to reducing the overhead for changes while also reducing duplicate work in tests

My general rule of thumb is not to test unexported functions if
* tests on exported functions allow good coverage of unexported code paths
* not having tests for the unexported functions will not greatly hinder debugging of other test failures (i.e. unexported functions aren't too complex)

If other packages start wanting to use data like this we should centralize the data definitions, not the bits that people may use to build their own version of the data. That will help cut down on mismatches in the expectations of different modules when stuff like the separation between RepoRef and LocationRef goes in

ryanfkeepers: > TBH I'm not interested in testing every function. I'm interested in the system working for the given inputs and outputs (black box testing). 

I'm not on board with 100% test coverage, either.  But the more fiddly-bits there are to mess up, the more need there is to ensure implementation gets all the fiddly bits right.  And, as we both know, selectors has many fiddly bits.  So the complaint is really about the core design, and the symptom involves the test requirements surfaced by that design.  I'm sure you agree about that.  Still, per `we should [re]move a good chunk of tests for selectors`, before we attack the symptom, I think we should first investigate a design that removes the underlying _need_ to have this granularity of testing.

PR #3239, opened 2023-04-26
permission basics

ryanfkeepers:
this doesn't solve all permission handling in sharepoint, but it sets a lot of groundwork that needs to be done anyway.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3135

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/onedrive/item.go
```diff
@@ -250,11 +249,12 @@ func filterUserPermissions(ctx context.Context, perms []models.Permissionable) [
...
 			logm := logger.Ctx(ctx)
+
 			if gv2.GetApplication() != nil {
-				logm.With("application_id", ptr.Val(gv2.GetApplication().GetId()))
+				entityID = ptr.Val(gv2.GetApplication().GetId())
+			} else if gv2.GetDevice() != nil {
+				entityID = ptr.Val(gv2.GetDevice().GetId())
 			}
 			if gv2.GetDevice() != nil {
 				logm.With("device_id", ptr.Val(gv2.GetDevice().GetId()))
```
ashmrtn, line 260: this is now duplicated

src/internal/connector/onedrive/permission.go
```diff
@@ -89,8 +89,9 @@ func getCollectionMetadata(
 // permissions. folderMetas is expected to have all the parent
 // directory metas for this to work.
 func computeParentPermissions(
-	itemPath path.Path,
-	folderMetas map[string]metadata.Metadata,
+	originDir path.Path,
```
ashmrtn, line 92: I don't usually mind renaming things, but please make separate PRs for it. Including name changes with other changes makes it much harder to tell what's an actual change and what's just cleanup

src/internal/connector/onedrive/restore.go
```diff
@@ -157,29 +159,29 @@ func RestoreCollection(
 		return metrics, clues.Wrap(err, "creating drive path").WithClues(ctx)
 	}
 
-	if rootIDCache == nil {
-		rootIDCache = map[string]string{}
+	if rootFolderIDCache == nil {
+		rootFolderIDCache = map[string]string{}
 	}
```
ashmrtn, line 164: this seems like a code smell. Either `rootFolderIDCache` doesn't need to be an in-out param, and thus doesn't need to be passed in as a parameter at all (it's not returned), or it _is_ an in-out param and stuff is could break in weird ways if nil is passed in

src/internal/connector/sharepoint/data_collections.go
ashmrtn, line 1: These changes aren't related to restore. They should have been in a different PR. Also these changes should not be merged to main as they'll cause the `excluded` map to be clobbered (thus losing items) if a backup happens to have multiple scopes

ashmrtn (commented):
What are the actual logical changes happening in this PR? There's a lot of variable renaming and other things going on so I'm having a difficult time figuring out what's changed and I should pay attention to, what's a variable rename, and what's everything else (which itself doesn't always seem related to permissions)

this should probably be split into multiple PRs corresponding to what the PR does. For example there could be a PR for CLI-level changes, a PR for code cleanup/renaming, and a PR for changes to stuff below the CLI

Also, what is the ordering of this PR relative to the PRs stacked on it? What is each PR addressing? What outcome will we get if they merge in this order rather than a different one (i.e. does this enable backup up permissions but not get all that data? That could throw off the restore code or have other unintended consequences if we're not careful)

src/internal/connector/onedrive/item.go
```diff
@@ -250,11 +249,12 @@ func filterUserPermissions(ctx context.Context, perms []models.Permissionable) [
 		} else if gv2.GetGroup() != nil {
 			entityID = ptr.Val(gv2.GetGroup().GetId())
 		} else {
-			// TODO Add application permissions when adding permissions for SharePoint
-			// https://devblogs.microsoft.com/microsoft365dev/controlling-app-access-on-specific-sharepoint-site-collections/
```
meain, line 254: I would like to keep this link. This was the only useful documentation that I could find around application permissions.

src/internal/connector/onedrive/item.go
```diff
@@ -250,11 +249,12 @@ func filterUserPermissions(ctx context.Context, perms []models.Permissionable) [
...
 			logm := logger.Ctx(ctx)
+
 			if gv2.GetApplication() != nil {
-				logm.With("application_id", ptr.Val(gv2.GetApplication().GetId()))
+				entityID = ptr.Val(gv2.GetApplication().GetId())
```
meain, line 255: This is not required for restore as of now, but I was just wondering if we should store what kind of permissions we are storing(user or group or device etc) just in case we need that information later on for filtering permissions to restore.

ryanfkeepers: @meain Can you please make a tech-debt ticket to add that?  We can follow up from there as needed.

src/internal/connector/onedrive/restore.go
ryanfkeepers: This gets resolved in the next PR.

ryanfkeepers:
@ashmrtn renaming and other cleanup moved to https://github.com/alcionai/corso/pull/3330.  Will continue changes from there, and will rebase this branch on to that chain at the appropriate stage.

ryanfkeepers:
closing this PR, replaced by https://github.com/alcionai/corso/pull/3333

Issue #3139, opened 2023-04-17
Don't allow OneDrive incremental backup if the previous version had file names


ashmrtn:
Incremental backups shouldn't be enabled if the base backup that's selected is a version before we switched to using M365 IDs for kopia file names. If we select a backup that uses file display names as the base then we'll end up with zombie files

ashmrtn:
Closing as this is not something that affects only older backups, does not cause loss of data, and can be cleared up by running a full backup of OneDrive or SharePoint if needed

PR #3241, opened 2023-04-27
Force full item enumeration if a backup that used file names is the base

ashmrtn:
At some point we switched from using file names
to file IDs in kopia as IDs were required for
incremental backups. However, before that change
went through we started persisting metadata for
incrementals. This means it was possible to get
delta changes when the base used file names which
would make it impossible to remove moved/deleted
files from the base as we wouldn't know the old
name

This PR forces full enumeration if the base
backup uses file names. It stops hierarchy merging
in kopia by deleting the old prefix and everything
under it and then passing in no metadata to the
get collections operation

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #3139

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/data_collections.go
+++ b/src/internal/connector/data_collections.go
@@ -114,6 +114,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 			sels,
 			owner,
 			metadata,
+			lastBackupVersion,
 			gc.credentials,
 			gc.Service,
 			gc,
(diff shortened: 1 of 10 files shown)


ashmrtn:
Closing as this is not something that affects only older backups, does not cause loss of data, and can be cleared up by running a full backup of OneDrive or SharePoint if needed

PR #3245, opened 2023-04-27
Add workflow to manually publish image via CI

meain:
Option to manually publish a binary from a specific branch. This is useful when  we want to share test builds without having to merge to main. Once merged, this will let us release binaries for any branch just like publish website.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/actions/publish-binary/action.yml
+++ b/.github/actions/publish-binary/action.yml
@@ -1,0 +1,75 @@
+name: Publish Binary
+
+inputs:
+  version:
+    description: Corso version to use for publishing
+    required: true
+  github_token:
+    description: GitHub token for publishing
+    required: true
+  rudderstack_write_key:
+    description: Write key for RudderStack
+    required: true
+  rudderstack_data_plane_url:
+    description: Data plane URL for RudderStack
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # needed to pull changelog
+
+    - name: Setup Golang with cache
+      uses: magnetikonline/action-golang-cache@v4
+      with:
+        go-version-file: src/go.mod
+
+    - name: Mark snapshot release
+      shell: bash
+      if: ${{ !startsWith(github.ref , 'refs/tags/') }}
+      run: |
+        echo "grflags=--snapshot" >> $GITHUB_ENV
+
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v4
+      with:
+        version: latest
+        args: release --rm-dist --timeout 500m --parallelism 1 ${{ env.grflags }}
+        workdir: src
+      env:
+        GITHUB_TOKEN: ${{ inputs.github_token }}
+        RUDDERSTACK_CORSO_WRITE_KEY: ${{ inputs.rudderstack_write_key }}
+        RUDDERSTACK_CORSO_DATA_PLANE_URL: ${{ inputs.rudderstack_data_plane_url }}
+        CORSO_VERSION: ${{ inputs.version }}
+
+    - name: Upload darwin arm64
+      uses: actions/upload-artifact@v3
+      with:
+        name: corso_Darwin_arm64
+        path: src/dist/corso_darwin_arm64/corso
+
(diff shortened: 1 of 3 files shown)


neha-Gupta1 (approved):

.github/workflows/binary-publish.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get version string
+        id: version
+        run: |
+          if ${{ startsWith(github.ref, 'refs/tags/') }}; then
+            echo "version=$(git describe --exact-match --tags $(git rev-parse HEAD))" | tee -a $GITHUB_OUTPUT
+          else
+            echo "version=$(echo unreleased-$(git rev-parse --short HEAD))" | tee -a $GITHUB_OUTPUT
```
pandeyabs, line 20: Few comments:

1. For unreleased versions, should we also attach branch name `--abbrev-ref` along with commit id? Probably for easier bookkeeping. 
2. Not related to this change, but just curious - When we share unreleased binaries with customers, do we share any terms & conditions with them?

.github/workflows/binary-publish.yml
```diff
@@ -0,0 +1,37 @@
...
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get version string
+        id: version
+        run: |
+          if ${{ startsWith(github.ref, 'refs/tags/') }}; then
+            echo "version=$(git describe --exact-match --tags $(git rev-parse HEAD))" | tee -a $GITHUB_OUTPUT
```
pandeyabs, line 18: YAML question - Do we typically not do any error handling? What's the behavior if say `git describe` fails?

pandeyabs (approved):
Looks good to me in general! Left some questions/suggestions.

.github/actions/publish-binary/action.yml
```diff
@@ -0,0 +1,75 @@
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: corso_Linux_amd64
+        path: src/dist/corso_linux_amd64_v1/corso
+
+    - name: Upload windows amd64
+      uses: actions/upload-artifact@v3
+      with:
+        name: corso_Windows_amd64
+        path: src/dist/corso_windows_amd64_v1/corso.exe
```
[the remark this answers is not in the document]
meain: I don't think adding it would be hard, but if I recall correctly we had decided that it was not really necessary at this point. That said, we could add it in a follow up PR if necessary.

.github/workflows/binary-publish.yml
meain: > For unreleased versions, should we also attach branch name --abbrev-ref along with commit id? Probably for easier bookkeeping.

- In most cases the unknown version will be released from `main` and so won't really have a useful branch name.
- We have been using this version format for quite some time and it would be better to stick to it
- The users have no real use for the branch name, and we can always just lookup the branch name from commit

> Not related to this change, but just curious - When we share unreleased binaries with customers, do we share any terms & conditions with them?

Apart from the fact that we highlight that it is a development build and not a release, nothing else.

.github/workflows/binary-publish.yml
meain: We don't really have much in terms of error handling in the CI. I might be useful to improve them, but don't think they are high up in the priority list.

.github/actions/publish-binary/action.yml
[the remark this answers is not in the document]
pandeyabs: sounds good. Given we are supporting windows amd64, i think we should also support windows arm64 similarly. 
but i do agree its low priority.

.github/workflows/binary-publish.yml
pandeyabs: Ah okay, didn't know that `main` would be the source. Thanks for answering my questions!

Issue #3050, opened 2023-04-06
[Bug]: cannot delete empty Sharepoint backup


serverless-mom:
### What happened?

I'm experiencing an error with corso backup delete sharepoint , after running a `--site '*'` backup, I have a number of backups listed as 'No Data' which, all well and good but I cannot delete those backups. Trying to gives me `Error: Deleting backup a6c6473e-10ed-4fb5-8c36-76b0523eca8d: snapshot ID required for deletion`

any backup that didn't have 'No Data' is deleting just fine

### Corso Version?

Corso v0.6.1

### Where are you running Corso?

OSx 13.3 (22E252)

ashmrtn:
hmmm could be a change between corso versions. We should now save metadata for incremental backups even if there's no data (items) to backup

It seems like there's a bug in the [delete code](https://github.com/alcionai/corso/blob/f5a4c3c0ba40ae40c3796d9d2afa37d9320d62d7/src/pkg/repository/repository.go#L558) that could be causing this. The linked block is duplicated in that statement and the if-block below. We should probably just keep the if-block version

neha-Gupta1:
PR - https://github.com/alcionai/corso/pull/3250

PR #3250, opened 2023-04-27
Don't go for snapshot deletion in case of empty backup

neha-Gupta1:
Fix: https://github.com/alcionai/corso/issues/3050 
In case of a blank backup, don't go for snapshot deletion. 

- [ ] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3050

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test


meain (approved):
LGTM

PR #3251, opened 2023-04-27
Fixup archive upload for sanity test

ashmrtn:
* now always upload logs
* don't use relative paths for things (archive doesn't like it)
* build variables off each other
* upload whole log file output directory
* fix upload action to use the right directory (it doesn't pay attention to working-directory)

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

tested locally with nektos act

### DIFF
--- a/.github/workflows/sanity-test.yaml
+++ b/.github/workflows/sanity-test.yaml
@@ -29,14 +29,18 @@ jobs:
       AZURE_CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
       AZURE_TENANT_ID: ${{ secrets.TENANT_ID }}
       CORSO_BUCKET: ${{ secrets.CI_TESTS_S3_BUCKET }}
-      CORSO_LOG_FILE: ./src/testlog/testlogging.log
+      CORSO_LOG_DIR: testlog
+      CORSO_LOG_FILE: ${{ env.CORSO_LOG_DIR }}/testlogging.log
       CORSO_M365_TEST_USER_ID: ${{ github.event.inputs.user != '' && github.event.inputs.user || vars.CORSO_M365_TEST_USER_ID }}
       CORSO_PASSPHRASE: ${{ secrets.INTEGRATION_TEST_CORSO_PASSPHRASE }}
-      TEST_RESULT: "test_results"
+      TEST_RESULT: test_results
+      # The default working directory doesn't seem to apply to things without
+      # the 'run' directive. https://stackoverflow.com/a/67845456
+      WORKING_DIR: src
 
     defaults:
       run:
-        working-directory: src
+        working-directory: ${{ env.WORKING_DIR }}
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity-test.yaml
```diff
@@ -70,7 +74,7 @@ jobs:
         run: |
           set -euo pipefail 
           prefix=`date +"%Y-%m-%d-%T"`
```
pandeyabs, line 76: Just curious, where does this `prefix` get used?

ashmrtn: looks like it's used as the corso repo prefix (read s3 prefix) for later operations

.github/workflows/sanity-test.yaml
```diff
@@ -29,14 +29,18 @@ jobs:
       AZURE_CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
       AZURE_TENANT_ID: ${{ secrets.TENANT_ID }}
       CORSO_BUCKET: ${{ secrets.CI_TESTS_S3_BUCKET }}
-      CORSO_LOG_FILE: ./src/testlog/testlogging.log
+      CORSO_LOG_DIR: testlog
+      CORSO_LOG_FILE: ${{ env.CORSO_LOG_DIR }}/testlogging.log
```
pandeyabs, line 33: Does the log file get cleaned up post run? I've noticed an interesting behavior with corso - If you supply an existing log file to corso, it concatenates to it rather than overwriting it. 
Is this by design? If the testlogging.log file doesn't get cleaned up, it might get really big across runs. Also, this we might not want this.

.github/workflows/sanity-test.yaml
pandeyabs: thanks! i should have noticed that

.github/workflows/sanity-test.yaml
ashmrtn: the log file won't be cleaned up between runs

I didn't think about log file size, mostly just how to make it more readable by adding some extra output in there for the different tests. I was looking to see if there was a way to extract the name of the current step that's running, but unfortunately that doesn't appear possible. If it was that would make for an easy way to set the log name

I can split the log into different files if that would be useful, otherwise we can test this out and adjust as needed

ashmrtn: hmmm actually may be able to dig more for how to extract the step name 🤔

ashmrtn: ah, no it has no good way to do that. Only accessing by ID can you get the name, but there's no way to get the ID of the current step it seems

pandeyabs: Yeah, your changes are fine. I was just commenting on the evergrowing corso log file. 
We should be able to verify the _evergrowingness_ from archives. Also, we could switch to running jobs in a container, so that state gets cleaned up after runs. But I don't know if this has any other repercussions. 

https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

pandeyabs (approved):

PR #3254, opened 2023-04-28
Type out path instead of using env

ashmrtn:
Env inside of an env declaration isn't allowed it seems.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix


meain (approved):
I think the problem is that the env is only defined within the job and not here.

PR #3257, opened 2023-04-28
Introduce a UserCompatNoInfo call

vkamra:
Instead of returning `Info` in `User`, callers should call the `GetUserInfo` API to get extended user information.

This allows the caller to decide whether info is needed *and* control parallelism for these queries.

This change introduces a `UserCompatNoInfo` call that allows us to make the switch without breaking existing `UserCompat` references. We will remove this API after callers have switched.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/m365.go
+++ b/src/pkg/services/m365/m365.go
@@ -90,6 +90,7 @@ func usersNoInfo(ctx context.Context, acct account.Account, errs *fault.Bus) ([]
 		if err != nil {
 			return nil, clues.Wrap(err, "formatting user data")
 		}
+
 		puNoInfo := &UserNoInfo{
 			PrincipalName: pu.PrincipalName,
 			ID:            pu.ID,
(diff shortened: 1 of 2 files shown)


src/pkg/services/m365/m365.go
```diff
@@ -47,6 +47,48 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+	if err != nil {
+		return nil, clues.Wrap(err, "getting users").WithClues(ctx)
+	}
+
+	users, err := discovery.Users(ctx, uapi, errs)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := make([]*User, 0, len(users))
```
neha-Gupta1, line 78: Info values will be empty in the `User` struct. But that's understood from function name itself.

vkamra: To be safe - I'm going to clone the `User` struct also

vkamra: Fixed

PR #3263, opened 2023-04-28
Check for new and older content URLs

### LINKED ISSUE #3262: [Bug]: Graph API returning OneDrive links under different property name
### What happened?

Graph API appears to be returning download links for some instances under the `@content.downloadUrl` property instead of `@microsoft.graph.downloadUrl` property. This is causing backup failures. [SO](https://stackoverflow.com/questions/45502275/microsoftgraph-querying-for-microsoft-graph-downloadurl-returns-character-i) seems to hint that the latter should be what we should expect under ideal conditions

We should check both just in case so backups don't fail

### Corso Version?

Corso dev-v0.6.1-222-g6e3403f71

N/A

### Relevant log output

```shell
"recovered_error": "{"onedrive reader: extracting file url", ...}
```

ashmrtn:
Some requests appear to be returning an older version of the content URL key. This was causing corso backup failures. This commit expands the logic to check for both the new and old keys for the URL.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* closes #3262 

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -20,11 +20,11 @@ import (
 	"github.com/alcionai/corso/src/pkg/logger"
 )
 
-const (
-	// downloadUrlKey is used to find the download URL in a
-	// DriveItem response
-	downloadURLKey = "@microsoft.graph.downloadUrl"
-)
+// downloadUrlKeys is used to find the download URL in a DriveItem response.
+var downloadURLKeys = []string{
+	"@microsoft.graph.downloadUrl",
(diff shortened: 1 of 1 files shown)


ashmrtn:
successfully backed up a user's OneDrive locally even though it was failing prior to the changes. Temporary log statement showed the data was returned under the newly added property

ryanfkeepers (approved):

Issue #3240, opened 2023-04-26
[Bug]: SharePoint doesn't propagate deleted items during incremental backup


ashmrtn:
### What happened?

SharePoint libraries backup isn't passing up a set of excluded items. This will result in zombie items in incremental backups if an item is moved between folders or is deleted

### Corso Version?

Corso dev-v0.6.1-203-gedef23bfc

### Where are you running Corso?

N/A

PR #3264, opened 2023-04-28
swap map for a prefix matcher

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3240 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -41,7 +42,7 @@ func (gc *GraphConnector) ProduceBackupCollections(
 	lastBackupVersion int,
 	ctrlOpts control.Options,
 	errs *fault.Bus,
-) ([]data.BackupCollection, map[string]map[string]struct{}, error) {
+) ([]data.BackupCollection, *excludes.ParentsItems, error) {
```
ashmrtn, line 45: all functions except maybe onedrive.Get should be returning `MapReader` instead of a concrete instance of the interface. Once the collections are created we should only do lookups for excluded items

src/internal/connector/onedrive/excludes/excludes.go
```diff
@@ -1,22 +1,80 @@
...
+	"github.com/alcionai/corso/src/internal/common/prefixmatcher"
+)
+
+var _ prefixmatcher.MapBuilder = &ParentsItems{}
 
 // Items that should be excluded when sourcing data from the base backup.
 // Parent Path -> item ID -> {}
-type ParentsItems map[string]map[string]struct{}
+type ParentsItems struct {
+	m map[string]map[string]struct{}
```
ashmrtn, line 16: I believe it should be possible to just use the `Builder` interface from the prefixmatcher package. iirc, maps are passed by reference so a `Get` for a key and then copying into the returned map should allow equivalent behavior. If the `Get` returns nothing then a new entry should be added

src/internal/common/prefixmatcher/prefix_matcher.go
```diff
@@ -2,24 +2,40 @@ package prefixmatcher
 
 import (
 	"strings"
+
+	"golang.org/x/exp/maps"
 )
 
-type View[T any] interface {
+type MapReader interface {
```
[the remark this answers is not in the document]
ryanfkeepers: Done, and also renamed to better identify the type/purpose.

src/internal/common/prefixmatcher/prefix_matcher.go
```diff
@@ -63,6 +79,10 @@ func (m prefixMatcher[T]) Empty() bool {
 	return len(m.data) == 0
 }
 
-func NewMatcher[T any]() Matcher[T] {
+func (m prefixMatcher[T]) Keys() []string {
+	return maps.Keys(m.data)
+}
+
+func NewMatcher[T any]() Builder[T] {
```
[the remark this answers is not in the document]
ryanfkeepers: Nope!

src/internal/connector/data_collections.go
ryanfkeepers: Returning interfaces where a concrete struct is available is not an idiomatic golang pattern.  I'd like to keep with established standards wherever possible.

src/internal/connector/onedrive/excludes/excludes.go
ryanfkeepers: sure, why not

src/internal/connector/data_collections.go
ashmrtn: I'm not really saying "never return concrete structs" I'm just saying to move the boundary of where they are returned closer to the service-specific code rather than pushing them into the generic (mostly glue) code

Returning in this fashion causes other packages like exchange to import the onedrive package just to add the proper return type which seems like a code smell. I think perhaps a more helpful way to view things is that the code in the connector package, or any non-service-specific package should be viewed as "outside" the package where the idiomatic patterns apply. Really we don't want the connector package to be modifying anything in the exclude map, nor does it really care what's in there. It's mostly glue code that shuttles data between the service-specific packages and the operations package.

If we really wanted to we could take the code that's in the connector package and move it to a different package and call into the separate service-specific code that was in an unrelated package subtree. At that point it would maybe be more clear that although the connector package interacts with service-specific packages like exchange and onedrive it's not particularly related to them in a way that requires getting non-interface return values like this

I believe we had this same discussion over whether to return `BackupCollection` or service-specific collection types as well

This pattern would also preclude the use of a standardized interface for service-specific calls in the future if we wanted to move to that if other service-specific packages also needed to define their own exclude map type. At that point it wouldn't be possible for the interface to allow return types other than the onedrive package's

ryanfkeepers: mmm, I think you've bundled together three different concerns in all of that.
1. should onedrive own this package?  Probably not.  There's nothing service-specific about the behavior, outside of that being the place of its primary usage.  It's reasonable to move this to a disconnected location.
2. should an interface be returned here?  I don't see any particular benefit there.  Unlike backupCollections, we're not aggregating different structs in the same data structure.  If it becomes necessary in the future, we can pivot easily to that value.
3. would an interface safeguard the read-only behavior?  Superficially, yes, but realistically no.  The only real barrier between a read-only interface and downstream mutation is to guarantee no maintainer ever approves a side-load typecast to the Builder interface.  The functionality is still potentially present in the underlying struct.  But I can meet you in the middle here, and produce a read-only variant of the struct to return, for stronger guarantees.

ryanfkeepers: Updated.

ashmrtn: language idioms and general rules are often more guidelines than anything. At some point it's necessary to think a little outside the box to make things readable and usable. Just like writing grammatically correct and formal English (or any other language) is both tedious to do an extremely dull to read following every little rule and idiom in all places through a codebase tends to make things more difficult

Taking a more abstract approach to things tends to make stuff easier to expand on in the future and leaves room for mocks etc. I've found it's often easier to use concrete types in the places where the values are generated/modified, but after that it's often better to use abstract concepts. Then readers need only know the abstract type and don't have to worry about the particulars of any one implementation. Not to mention that mocking becomes possible only when interfaces are used

>should an interface be returned here? I don't see any particular benefit there. Unlike backupCollections, we're not aggregating different structs in the same data structure

this means we'll have to do a lot of wiring changes for something small in the future instead of leveraging tools the language gives us. It generates more work in the long run while not giving any benefits. If connector package isn't expected to modify anything having to do with the prefix collection there's no reason _not_ to use an interface in code that is completely unrelated to generating the values backing the interface and not concerned with the values at all.

Not returning the interface means it's also impossible to mock the map being used. If people wish to test any of the interface functions defined in the inject package they need to import the struct and use the functions they have access to. There's no other options or flexibility there because the inject interface definition _requires_ returning a specific type

>would an interface safeguard the read-only behavior? Superficially, yes, but realistically no. The only real barrier between a read-only interface and downstream mutation is to guarantee no maintainer ever approves a side-load typecast to the Builder interface.

This seems unlikely to be allowed anyway. There's a good chance that casting something from a read-only interface to a read-write struct would be flagged as a code smell by reviewers. That's also ignoring that neither of us are particularly fond of side-casting. I'm alright with it in limited places, but I by no means see it as a solution to things. It's more the way things just happen to play out sometimes due to the overall language design. Golang itself is also full of instances of side-casting where it makes sense (and some where it doesn't really make sense but happens anyway). The [ast package](https://pkg.go.dev/go/ast) is a great example of this

ashmrtn (approved):

PR #3271, opened 2023-05-01
retry on gateway timeout

ryanfkeepers:
We don't currently retry on graph API gateway timeout/bad gateway responses.  This change adds those statuses to our retryable codes.

Also adds some qol clues tracking and error formatting.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/graph/middleware.go
```diff
@@ -181,6 +182,40 @@ type RetryHandler struct {
 	Delay time.Duration
 }
 
+// Intercept implements the interface and evaluates whether to retry a failed request.
```
pandeyabs, line 185: Should we move retry handler middleware to its own file? Also suggest adding some tests

src/internal/connector/graph/middleware.go
```diff
@@ -181,6 +182,40 @@ type RetryHandler struct {
...
+) (*http.Response, error) {
+	ctx := req.Context()
+
+	response, err := pipeline.Next(req, middlewareIndex)
+	if err != nil && !IsErrTimeout(err) {
+		return response, stackReq(ctx, req, response, err)
+	}
+
+	exponentialBackOff := backoff.NewExponentialBackOff()
+	exponentialBackOff.InitialInterval = middleware.Delay
```
pandeyabs, line 199: Curious, what is the maximum possible exponential backoff period in our setup?

src/internal/connector/graph/errors.go
```diff
@@ -241,6 +241,26 @@ func Stack(ctx context.Context, e error) *clues.Err {
 	return setLabels(clues.Stack(e).WithClues(ctx).With(data...), innerMsg)
 }
 
+// stackReq is a helper function that extracts ODataError metadata from
+// the error, plus http req/resp data.  If the error is not an ODataError
+// type, returns the error with only the req/resp values.
+func stackReq(
```
pandeyabs, line 247: This is great!

src/internal/connector/graph/middleware.go
pandeyabs: Could we use chaos handler middleware to simulate graph API errors?

src/internal/connector/graph/middleware.go
```diff
@@ -232,15 +270,22 @@ func (middleware RetryHandler) retryRequest(
...
 	}
 
 	return resp, nil
 }
 
-func (middleware RetryHandler) isRetriableErrorCode(req *http.Request, code int) bool {
-	return code == http.StatusInternalServerError || code == http.StatusServiceUnavailable
+var retryableRespCodes = []int{
+	http.StatusInternalServerError,
+	http.StatusServiceUnavailable,
```
pandeyabs, line 282: Reading MSFT's guidance on how to handle service unavailable errors. 
`This is likely because the services is busy. You should employ a back-off strategy similar to 429. Additionally, you should always make new retry requests over a new HTTP connection.`

Just to confirm, will retries be done over a new HTTP connection?

src/internal/connector/graph/middleware.go
```diff
@@ -181,6 +182,40 @@ type RetryHandler struct {
...
+// Intercept implements the interface and evaluates whether to retry a failed request.
+func (middleware RetryHandler) Intercept(
+	pipeline khttp.Pipeline,
+	middlewareIndex int,
+	req *http.Request,
+) (*http.Response, error) {
+	ctx := req.Context()
+
+	response, err := pipeline.Next(req, middlewareIndex)
+	if err != nil && !IsErrTimeout(err) {
```
pandeyabs, line 194: Do we also need `!IsErrConnectionReset(err)` check here, similar to that in `retryRequest` code?

src/internal/connector/graph/middleware.go
```diff
@@ -190,9 +225,14 @@ func (middleware RetryHandler) retryRequest(
...
-	respErr error,
+	initialErr error,
 ) (*http.Response, error) {
-	if (respErr != nil || middleware.isRetriableErrorCode(req, resp.StatusCode)) &&
+	ctx = clues.Add(
+		ctx,
+		"retry_count", executionCount,
+		"prev_resp_status", resp.Status)
+
+	if (initialErr != nil || middleware.isRetriableRespCode(req, resp.StatusCode)) &&
```
pandeyabs, line 235: This set of conditions has some cognitive load. Would be great to have a comment describing it. 
Also, dumb question: Wouldn't the errors listed in `isRetriableRespCode` cause an err too? Wondering if `initialErr` is sufficient?

pandeyabs (changes requested):
I've left some comments. Overall the changes look good to me. But I'd love to see tests, with chaos handler middleware or otherwise.

src/internal/connector/graph/middleware.go
pandeyabs: ^ Taken from https://learn.microsoft.com/en-us/graph/best-practices-concept#handling-expected-errors?WT.mc_id=M365-MVP-5002900

src/internal/connector/graph/middleware.go
ryanfkeepers: > Should we move retry handler middleware to its own file?
No need, imo.  It's plenty small to fit in alongside the others.

> Also suggest adding some tests
Not entirely certain there's much gain in testing here, but I added a quick suite for coverage nonetheless.

src/internal/connector/graph/middleware.go
ryanfkeepers: Whatever we configure it to.  Default is `3`.

src/internal/connector/graph/middleware.go
ryanfkeepers: I think so?  iirc, default behavior for the graph client is to construct new connections on each call (barring more extreme circumstances that easily allow re-use, such as long dormancy with all prior connections being closed).

src/internal/connector/graph/middleware.go
ryanfkeepers: good eye, yes.

src/internal/connector/graph/middleware.go
ryanfkeepers: I don't know what errors you're referring to here.

src/internal/connector/graph/middleware.go
pandeyabs: Sounds good, we can do chaos handler based testing later. Please create an issue. I think it would help to have this test capability longer term.

src/internal/connector/graph/middleware.go
pandeyabs: sorry, i meant errors listed in `retryableRespCodes`

src/internal/connector/graph/middleware_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+			defer gock.Off()
+			gock.New(uri).Get(path).Reply(test.status)
+
+			_, err := suite.srv.Client().UsersById("user").MessagesById("foo").Get(ctx, nil)
+			assert.Error(t, err)
+
+			// hacky, but not much better way to catch the retry count
+			ce := clues.ToCore(err)
+			assert.Equal(t, test.retryCount, ce.Values["retry_count"])
```
pandeyabs, line 97: pretty neat way to test this!

src/internal/connector/graph/middleware_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/alcionai/corso/src/internal/tester"
+	"github.com/alcionai/corso/src/pkg/account"
+)
+
+type RetryMWIntgSuite struct {
+	tester.Suite
+	credentials account.M365Config
+	srv         Servicer
+	user        string
```
pandeyabs, line 22: Linter warning

src/internal/connector/graph/middleware.go
```diff
@@ -232,18 +274,33 @@ func (middleware RetryHandler) retryRequest(
...
+func (mw RetryMiddleware) isRetriableRespCode(ctx context.Context, resp *http.Response, code int) bool {
+	if slices.Contains(retryableRespCodes, code) {
+		return true
+	}
+
+	// not a status code, but the message itself might indicate a connectivity issue that
+	// can be retried independent of the status code.
+	return strings.Contains(
+		strings.ToLower(getRespDump(ctx, resp, true)),
+		strings.ToLower(string(IOErrDuringRead)))
```
pandeyabs, line 300: Probably an overkill, but we could save on `ToLower` calls by making `IOErrDuringRead` lower case.

pandeyabs (approved):

src/internal/connector/graph/middleware.go
ryanfkeepers: Ah, so that's a common misconception.  The standard golang http client does not return errors for non-2xx status codes.  A 400 resp code, for example, will return a valid http.Response with statusCode=400, and a nil error.  Http client errors in golang are reserved for an inability to make the request in at all (malformed config, missing necessary data, etc).

Now, in this case, we primarily use the graph api client wrapping the golang http client.  And the graph client _does_ transform non-2xx responses into errors (it also does not return the http.Response, so this transformation is somewhat necessary).  But since the middleware intercepts the response before that transformation exists, we're still working with response codes, not error interfaces.

src/internal/connector/graph/middleware_test.go
ryanfkeepers: I swear I ran that

src/internal/connector/graph/middleware.go
ryanfkeepers: Per company culture, minor performance improvements such as this are kept outside of our scope of concern.  O(n) growth performance or unbounded conditions, sure.  But we try not to sweat the little stuff.  What's more important in cases like this is whether the behavior causes unwanted outcomes (ie: unintended matches).

src/internal/connector/graph/middleware.go
pandeyabs: Thanks, this is useful!

PR #3275, opened 2023-05-01
retry on read IO error message

ryanfkeepers:
A corner case in graph api responses may return a 400 status code with an io-read error.  The cause is undetermined, but known cases of this have been manually retriable.  It's possible a flipped bit, encoding issue, or some internal fault caused a misread of the body.  Thus, we'll retry the request in cases with this message.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (approved):
This looks good to me.

PR #3278, opened 2023-05-02
Handle ErrorFolderExists for contact restore folders

pandeyabs:
Handle 409, `ErrorFolderExists` error in restore path while creating destination contact folder. We need to fix this for all m365 services. Starting with contacts to get a buy in on the fix. Will add integration tests + similar fixes for rest of exchange, onedrive, etc. in later PRs. 

Context:
`CreateContactFolder()` may fail with `ErrorFolderExists` under certain error conditions. For e.g. consider below scenario. 
1. `CreateContactFolder()` does a `POST` to graph to create restore destination folder but this fails with 5xx. 
2. It's possible that step 1 may have left some dirty state in graph. For e.g. it's possible that the destination folder in step 1 was actually created, but 5xx was returned due to other reasons. 
3. So when we reattempt `POST` in such a scenario, we sometimes observe `ErrorFolderExists` error .
4. Corso should be resilient to such errors. To fix this, when we encounter such an error, we will do a `GET` to fetch the restore destination folder and add it to folder cache.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -55,6 +55,22 @@ func (c Contacts) CreateContactFolder(
 	return mdl, nil
 }
 
+// CreateContactFolder makes a contact folder with the displayName of folderName.
+// If successful, returns the created folder object.
+func (c Contacts) GetContactFolders(
+	ctx context.Context,
+	user string,
+) (models.ContactFolderCollectionResponseable, error) {
+	// TODO: Add pagination or just reuse EnumerateContainer logic
+	// with base id at root
+	mdl, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "creating contact folder")
+	}
+
+	return mdl, nil
+}
+
 // DeleteContainer deletes the ContactFolder associated with the M365 ID if permissions are valid.
 func (c Contacts) DeleteContainer(
 	ctx context.Context,
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -55,14 +55,12 @@ func (c Contacts) CreateContactFolder(
 	return mdl, nil
 }
 
-// CreateContactFolder makes a contact folder with the displayName of folderName.
-// If successful, returns the created folder object.
+// TODO: Add pagination or just reuse EnumerateContainer logic
+// with base id at root
 func (c Contacts) GetContactFolders(
 	ctx context.Context,
 	user string,
 ) (models.ContactFolderCollectionResponseable, error) {
-	// TODO: Add pagination or just reuse EnumerateContainer logic
-	// with base id at root
 	mdl, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, nil)
 	if err != nil {
 		return nil, graph.Wrap(ctx, err, "creating contact folder")
--- a/src/internal/connector/exchange/api/contacts.go
+++ b/src/internal/connector/exchange/api/contacts.go
@@ -55,18 +55,39 @@ func (c Contacts) CreateContactFolder(
 	return mdl, nil
 }
 
-// TODO: Add pagination or just reuse EnumerateContainer logic
-// with base id at root
-func (c Contacts) GetContactFolders(
+// GetContactFolderByDisplayName fetches contact folder which has the displayName of
+// folderName. This is only done at root level.
+func (c Contacts) GetContactFolderByDisplayName(
 	ctx context.Context,
-	user string,
-) (models.ContactFolderCollectionResponseable, error) {
-	mdl, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, nil)
-	if err != nil {
-		return nil, graph.Wrap(ctx, err, "creating contact folder")
+	user, folderName string,
+) (models.ContactFolderable, error) {
+	filter := fmt.Sprintf("displayName eq '%s'", folderName)
+	options := &users.ItemContactFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &users.ItemContactFoldersRequestBuilderGetQueryParameters{
+			Filter: &filter,
+		},
 	}
 
-	return mdl, nil
+	resp, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, options)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "getting contact folder").
+			With("folder name", folderName)
+	}
+
+	// We only expect one folder with provided folderName. Return an error if
+	// multiple folders exist(unlikely) or if no folder was found
+	if len(resp.GetValue()) != 1 {
+		return nil, graph.Wrap(ctx, err, "unexpected number of folders").
+			With("folder count", len(resp.GetValue())).
+			With("folder name", folderName)
+	}
+
+	fold := resp.GetValue()[0]
+	if err := checkIDAndName(fold); err != nil {
+		return nil, err
+	}
+
+	return fold, nil
 }
 
 // DeleteContainer deletes the ContactFolder associated with the M365 ID if permissions are valid.
--- a/src/internal/connector/exchange/service_restore.go
+++ b/src/internal/connector/exchange/service_restore.go
@@ -649,28 +649,29 @@ func establishContactsRestoreLocation(
 		return cached, nil
 	}
 
-	folders[0] = "Corso_Restore_02-May-2023_00:06:19"
 	ctx = clues.Add(ctx, "is_new_cache", isNewCache)
 
-	// This is where destination folder gets created by corso in graph
-	// This is the POST operation which may fail if the folder exists
-	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
-	if err != nil {
-		// TODO:: Add status code check too
-		if graph.IsErrFolderExists(err) {
-			result, _ := ac.Contacts().GetContactFolders(ctx, user)
+	fold, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
+	// 409 handling: Fetch folder if it exists.
+	// This is rare, but it may happen if an earlier CreateContactFolder POST failed
+	// with 5xx, potentially leaving dirty state in graph.
+	if graph.IsErrFolderExists(err) {
+		fold, err = ac.Contacts().GetContactFolderByDisplayName(ctx, user, folders[0])
+		if err != nil {
+			return "", err
 		}
+	} else if err != nil {
 		return "", err
 	}
 
-	folderID := ptr.Val(temp.GetId())
+	folderID := ptr.Val(fold.GetId())
 
 	if isNewCache {
 		if err := cfc.Populate(ctx, errs, folderID, folders[0]); err != nil {
 			return "", clues.Wrap(err, "populating contact cache")
 		}
 
-		if err = cfc.AddToCache(ctx, temp); err != nil {
+		if err = cfc.AddToCache(ctx, fold); err != nil {
 			return "", clues.Wrap(err, "adding contact folder to cache")
 		}
 	}
--- a/src/internal/connector/exchange/service_restore.go
+++ b/src/internal/connector/exchange/service_restore.go
@@ -649,10 +649,18 @@ func establishContactsRestoreLocation(
 		return cached, nil
 	}
 
+	folders[0] = "Corso_Restore_02-May-2023_00:06:19"
 	ctx = clues.Add(ctx, "is_new_cache", isNewCache)
 
+	// This is where destination folder gets created by corso in graph
+	// This is the POST operation which may fail if the folder exists
 	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
 	if err != nil {
+		// TODO:: Add status code check too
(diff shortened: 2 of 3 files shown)


src/internal/connector/exchange/service_restore.go
```diff
@@ -649,10 +649,17 @@ func establishContactsRestoreLocation(
...
 
+	// This is where destination folder gets created by corso in graph
+	// This is the POST operation which may fail if the folder exists
 	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
 	if err != nil {
+		// TODO:: Add status code check too
+		if graph.IsErrFolderExists(err) {
+			result, _ := ac.Contacts().GetContactFolders(ctx, user)
+		}
 		return "", err
```
ryanfkeepers, line 663:
```suggestion
		}
		
		return "", err
```

src/internal/connector/exchange/service_restore.go
```diff
@@ -649,10 +649,17 @@ func establishContactsRestoreLocation(
...
 	}
 
+	folders[0] = "Corso_Restore_02-May-2023_00:06:19"
 	ctx = clues.Add(ctx, "is_new_cache", isNewCache)
 
+	// This is where destination folder gets created by corso in graph
+	// This is the POST operation which may fail if the folder exists
 	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
 	if err != nil {
+		// TODO:: Add status code check too
```
ryanfkeepers, line 659: Status code checks shouldn't be necessary, since the graph client transforms status codes into error interfaces.  It's only in explicit corner-case conditions (eg: using a non-graph http client, or catching an error graph doesn't handle) that we compare status codes.

src/internal/connector/exchange/service_restore.go
```diff
@@ -649,10 +649,17 @@ func establishContactsRestoreLocation(
...
+		if graph.IsErrFolderExists(err) {
+			result, _ := ac.Contacts().GetContactFolders(ctx, user)
```
ryanfkeepers, line 661: Swallowing errors is always an anti-pattern.  We should never see `foo, _ := ` in place of `foo, err := ` in the code.

Likewise, there should be a linter that watches for this case.  If you aren't seeing it complain within this file, it might still be excluded (for some reason), and we'll want to fix that.

src/internal/connector/exchange/service_restore.go
```diff
@@ -649,10 +649,17 @@ func establishContactsRestoreLocation(
 		return cached, nil
 	}
 
+	folders[0] = "Corso_Restore_02-May-2023_00:06:19"
 	ctx = clues.Add(ctx, "is_new_cache", isNewCache)
 
+	// This is where destination folder gets created by corso in graph
+	// This is the POST operation which may fail if the folder exists
```
ryanfkeepers, line 656: technically not true, since you're handling this case, yeah?

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,20 @@ func (c Contacts) CreateContactFolder(
 	return mdl, nil
 }
 
+// TODO: Add pagination or just reuse EnumerateContainer logic
+// with base id at root
+func (c Contacts) GetContactFolders(
```
ryanfkeepers, line 60: This function already exists, see: `EnumerateContainers`.

src/internal/connector/exchange/service_restore.go
pandeyabs: Lol, these were just some pointers I added for myself.

src/internal/connector/graph/errors.go
```diff
@@ -172,6 +173,10 @@ func IsMalwareResp(ctx context.Context, resp *http.Response) bool {
 	return false
 }
 
+func IsErrFolderExists(err error) bool {
```
ryanfkeepers, line 176: Please add some unit tests for this.  There should be some easy boilerplate to steal from.

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
...
+	options := &users.ItemContactFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &users.ItemContactFoldersRequestBuilderGetQueryParameters{
+			Filter: &filter,
+		},
+	}
+
+	resp, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, options)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "getting contact folder").
+			With("folder name", folderName)
```
ryanfkeepers, line 74: no spaces in structured keys, please.  Use snake_case by default.  
```suggestion
		return nil, graph.Wrap(ctx, err, "searching contact folder by name").
			With("folder_name", folderName)
```

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
...
+	}
+
+	// We only expect one folder with provided folderName. Return an error if
+	// multiple folders exist(unlikely) or if no folder was found
+	if len(resp.GetValue()) != 1 {
+		return nil, graph.Wrap(ctx, err, "unexpected number of folders").
+			With("folder count", len(resp.GetValue())).
+			With("folder name", folderName)
```
ryanfkeepers, line 82: more descriptive keys tend to be more helpful.
```suggestion
			With("returned_folder_count", len(resp.GetValue()))
```

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
...
+	ctx context.Context,
+	user, folderName string,
+) (models.ContactFolderable, error) {
+	filter := fmt.Sprintf("displayName eq '%s'", folderName)
+	options := &users.ItemContactFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &users.ItemContactFoldersRequestBuilderGetQueryParameters{
+			Filter: &filter,
+		},
+	}
+
```
ryanfkeepers, line 70: Since you're adding `folder_name` to multiple error returns, I'd suggest adding it to the context here instead, to make sure that all subsequent errors incorporate it:
```suggestion
ctx = clues.Add(ctx, "folder_name", folderName)
```

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
...
+	resp, err := c.Stable.Client().UsersById(user).ContactFolders().Get(ctx, options)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "getting contact folder").
+			With("folder name", folderName)
+	}
+
+	// We only expect one folder with provided folderName. Return an error if
```
ryanfkeepers, line 77: I think we could match within the result set, instead of discarding on n>1 responses.  It shouldn't require us to jump through that many hoops, we have some good aids for identification (such as the parent folder ID), and I think the resilience will pay off with our users. 

```suggestion
	// We only allow the api to match one folder with provided folderName. Return an error if
```

src/internal/connector/exchange/service_restore.go
```diff
@@ -651,19 +651,27 @@ func establishContactsRestoreLocation(
 
 	ctx = clues.Add(ctx, "is_new_cache", isNewCache)
 
-	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
-	if err != nil {
+	fold, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
```
ryanfkeepers, line 654: email and events coming in a follow-up PR?

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
 	return mdl, nil
 }
 
+// GetContactFolderByDisplayName fetches contact folder which has the displayName of
```
ryanfkeepers, line 58: I'd love to see a gock-mocked integration test for this.  The mocks should help to avoid external data dependencies, making it fairly easy to inject.

src/internal/connector/exchange/api/contacts.go
```diff
@@ -55,6 +55,41 @@ func (c Contacts) CreateContactFolder(
...
+// folderName. This is only done at root level.
+func (c Contacts) GetContactFolderByDisplayName(
```
ryanfkeepers, line 60: A design we're moving towards, but haven't reached yet, is the ability to use the api iterations (mail, contacts, events) as common interfaces throughout connector/exchange.  In order to do that, we need to keep common func naming in mind.  I'd suggest the following here (just swapping "contactFolder" for "container").  This will make it practical and easy to, later on, establish an interface like `ContainerByNamer { GetContainerByDisplayName(...) }`

```suggestion
func (c Contacts) GetContainerByDisplayName(
```

src/internal/connector/exchange/service_restore.go
```diff
@@ -651,19 +651,27 @@ func establishContactsRestoreLocation(
...
+	// 409 handling: Fetch folder if it exists.
+	// This is rare, but it may happen if an earlier CreateContactFolder POST failed
+	// with 5xx, potentially leaving dirty state in graph.
+	if graph.IsErrFolderExists(err) {
+		fold, err = ac.Contacts().GetContactFolderByDisplayName(ctx, user, folders[0])
+		if err != nil {
+			return "", err
+		}
+	} else if err != nil {
```
ryanfkeepers, line 663: Very minor: I prefer the following code ordering.  It establishes the worst case/immediate exits at the top, and gradually moves towards happy path handling. 
```suggestion
	if err != nil && !graph.IsErrFolderExists(err) {
	  return "", err
	}
	
	if graph.IsErrFolderExists(err)
		fold, err = ac.Contacts().GetContactFolderByDisplayName(ctx, user, folders[0])
		if err != nil {
			return "", err
		}
	}
```

src/internal/connector/exchange/service_restore.go
```diff
@@ -651,19 +651,27 @@ func establishContactsRestoreLocation(
...
-	temp, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
-	if err != nil {
+	fold, err := ac.Contacts().CreateContactFolder(ctx, user, folders[0])
+	// 409 handling: Fetch folder if it exists.
+	// This is rare, but it may happen if an earlier CreateContactFolder POST failed
+	// with 5xx, potentially leaving dirty state in graph.
+	if graph.IsErrFolderExists(err) {
+		fold, err = ac.Contacts().GetContactFolderByDisplayName(ctx, user, folders[0])
+		if err != nil {
+			return "", err
```
ryanfkeepers, line 661: The inner error here kinda slices the outer one.  A little extra context will go a long way when trying to discover what processing occurred.
```suggestion
			return "", clues.Wrap(err, "getting folder after collision on create")
```

ryanfkeepers (commented):
Generally looks good.  Mostly just nits and wanting for more tests.

src/internal/connector/exchange/service_restore.go
pandeyabs: yeah! just wanted to keep this PR small and focused.

src/internal/connector/exchange/api/contacts.go
pandeyabs: Good point! I'll make the change.

src/internal/connector/exchange/service_restore.go
pandeyabs: 100%

src/internal/connector/exchange/service_restore.go
pandeyabs: Kinda subjective.  I didn't want to call `IsErrFolderExists` twice. But your pattern seems better( assuming we use it throughout the codebase). I'll make this change.

src/internal/connector/exchange/api/contacts.go
pandeyabs: ✍️

src/internal/connector/exchange/api/contacts.go
pandeyabs: Fair. I could make this function accept a parent folder id, and match against that in the result set. 
@ryanfkeepers  is there a convenient way to get the default contact folder id ( root)? I don't want to search by displayName.

src/internal/connector/exchange/api/contacts.go
pandeyabs: sgtm

src/internal/connector/exchange/api/contacts.go
pandeyabs: just found a solution for this - https://learn.microsoft.com/en-us/graph/known-issues#contacts

ryanfkeepers: oof, yeah, the default contacts folder is weird.  You might be able to find some other hacks we've used in the code as well (can't remember where they are, off the top of my head).

Or maybe this is something the container resolver could track for you?  Since it's aggregating all the containers, it should already know the root folder id.

pandeyabs:
not needed. closing.

PR #3280, opened 2023-05-02
reset body reader on corso retry handling

ryanfkeepers:
The kiota compressor middleware will attempt to compress the request body.  In the event that we have a corso-middleware- retriable response (eg: status 500), we need to reset the seek position of the req.Body, similar to how graph api does in their retry handler, or else the re-run of the compressor will already have read the full req.Body and the retried call will have a zero len body.

---

- [y] :white_check_mark: Yes, it's included

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/graph/middleware.go
+++ b/src/internal/connector/graph/middleware.go
@@ -3,6 +3,7 @@ package graph
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httputil"
 	"os"
@@ -250,7 +251,6 @@ func (mw RetryMiddleware) retryRequest(
 		executionCount++
 
 		delay := mw.getRetryDelay(req, resp, exponentialBackoff)
-
 		cumulativeDelay += delay
 
 		req.Header.Set(retryAttemptHeader, strconv.Itoa(executionCount))
@@ -264,6 +264,18 @@ func (mw RetryMiddleware) retryRequest(
 			return resp, clues.Stack(ctx.Err()).WithClues(ctx)
 
 		case <-timer.C:
+		}
+
+		// we have to reset the original body reader for each retry, or else the graph
+		// compressor will produce a 0 length body following an error response such
+		// as a 500.
+		if req.Body != nil {
+			if s, ok := req.Body.(io.Seeker); ok {
+				_, err := s.Seek(0, io.SeekStart)
+				if err != nil {
+					return nil, Wrap(ctx, err, "resetting request body reader")
+				}
+			}
 		}
 
 		nextResp, err := pipeline.Next(req, middlewareIndex)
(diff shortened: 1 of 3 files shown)


src/internal/connector/graph/middleware.go
```diff
@@ -259,6 +259,18 @@ func (mw RetryMiddleware) retryRequest(
 		case <-timer.C:
 		}
 
+		// we have to reset the original body reader for each retry, or else the graph
+		// compressor will produce a 0 length body following an error response such
```
pandeyabs, line 263: msgraph documentation seems to suggest that the compression handler should be added before retry handler to avoid this exact problem. Any reason we are not doing it this way in our code?

`This handler MUST be inserted BEFORE the [Retry handler](https://microsoftgraph.github.io/msgraph-sdk-design/middleware/RetryHanlder.md) to avoid compressing the request payload multiple times over.` 

https://microsoftgraph.github.io/msgraph-sdk-design/middleware/CompressionHandler.html

ryanfkeepers: Well that's _very interesting_, because we took this ordering from [the kiota http client factory code](https://github.com/microsoft/kiota-http-go/blob/main/kiota_client_factory.go#L78).

Considering that this mimics the graph client itself, I think I'm comfortable with the ordering.

pandeyabs: Good point. Might be worthwhile raising an issue with msgraph folks here https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/middleware/CompressionHandler.md

src/internal/connector/graph/middleware_test.go
```diff
@@ -150,3 +190,43 @@ func (suite *RetryMWIntgSuite) TestRetryMiddleware_Intercept_byStatusCode() {
...
+
+	var (
+		t     = suite.T()
+		body  = models.NewMailFolder()
+		check = func(req *http.Request) {
+			bs, err := io.ReadAll(req.Body)
+			require.NoError(t, err, clues.ToCore(err))
+
+			// an expired body, after graph compression, will
+			// normally contain 25 bytes.  So we should see more
```
pandeyabs, line 206: Might be worth explaining what constitutes 25 bytes, or how it was calculated.

src/internal/connector/graph/middleware_test.go
```diff
@@ -150,3 +190,43 @@ func (suite *RetryMWIntgSuite) TestRetryMiddleware_Intercept_byStatusCode() {
...
 }
+
+func (suite *RetryMWIntgSuite) TestRetryMiddleware_RetryRequest_resetBodyAfter500() {
+	ctx, flush := tester.NewContext()
+	defer flush()
+
+	var (
+		t     = suite.T()
+		body  = models.NewMailFolder()
+		check = func(req *http.Request) {
```
pandeyabs, line 201: nit: Maybe rename this to `intercept` or `onIntercept` for easier association?

pandeyabs (approved):
LGTM. Added minor nits.

PR #3282, opened 2023-05-02
Add periodic log message for loading items during restore

ashmrtn:

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

*

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -466,6 +466,8 @@ func (w Wrapper) ProduceRestoreCollections(
 		res = append(res, c)
 	}
 
+	logger.Ctx(ctx).Infow("done loading items from kopia", "loaded_count", loadCount)
+
 	return res, el.Failure()
 }
 
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -413,6 +413,7 @@ func (w Wrapper) ProduceRestoreCollections(
 	}
 
 	var (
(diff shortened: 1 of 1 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -451,6 +452,11 @@ func (w Wrapper) ProduceRestoreCollections(
 		}
 
 		c.streams = append(c.streams, ds)
+
+		loadCount++
+		if loadCount%1000 == 0 {
+			logger.Ctx(ctx).Infow("loading items from kopia", "loaded_count", loadCount)
```
pandeyabs, line 458: This looks good to me. Did you also consider sampling loadCount every x seconds or so?

pandeyabs: Should we also add another log which indicates that we finished loading all items from kopia?

ashmrtn: >Did you also consider sampling loadCount every x seconds or so?

we probably don't need to do sampling yet. Mostly we're trying to make sure we keep making some progress as a debugging aid. The timestamps in the log can also be used to calculate rate if we need it

>Should we also add another log which indicates that we finished loading all items from kopia?

good idea, I've added a log for that

pandeyabs (approved):

Issue #2106, opened 2023-01-11
Add support for specifying a AWS credential provider


vkamra:
This will allow API consumers to plug in custom credential providers

vkamra:
Follow-ups before this is closed -> Once we verify this works as intended, we need to upstream this PR

vkamra:
Upstream: https://github.com/kopia/kopia/pull/2689

PR #3284, opened 2023-05-03
Allow specifying a role to assume when initializing a storage provider

vkamra:
Allows caller to specify a IAM role to assume in the Kopia storage provider

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2106 

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/storage/storage.go
```diff
@@ -36,6 +36,11 @@ const (
 type Storage struct {
 	Provider storageProvider
 	Config   map[string]string
+	// TODO: These are AWS S3 specific -> move these out
```
ryanfkeepers, line 39: per this todo: reasons not to move them to `/storage/s3.go` right away?  Extending the `S3Config` struct (or adding a S3ConfigWithRole struct) would also also save the effort of wiring up the `NewStorageUsingRole` func below, since NewStorage would do everything you'd need.

ryanfkeepers (approved):
Going to approve to unblock (not clear on urgency here), but I really think we're better off jumping straight to integrating the new fields into the storage/s3 handling instead of the current workaround that uses the parent config struct.

vkamra:
> Going to approve to unblock (not clear on urgency here), but I really think we're better off jumping straight to integrating the new fields into the storage/s3 handling instead of the current workaround that uses the parent config struct.

Discussed offline - if this approach does work out - will follow up and make this part of the S3 configurator.

PR #3289, opened 2023-05-03
Add flag to disable delta queries

meain:
Adds an internal flag `--disable-delta` which can be used to disable delta based queries.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/options/options.go
```diff
@@ -106,6 +111,18 @@ func AddDisableIncrementalsFlag(cmd *cobra.Command) {
 	cobra.CheckErr(fs.MarkHidden(DisableIncrementalsFN))
 }
 
+// Adds the hidden '--disable-delta' cli flag which, when set, disables
+// delta based backups.
+func AddDisableDeltaFlag(cmd *cobra.Command) {
```
ryanfkeepers, line 116: pitch, instead of adding this (and in the future, more --disable-foo) flags, how about a single `--disable` flag with a string slice arg?  Ex:
```go
	fs.BoolVar(
		&disableFV,
		DisableFN,
		false,
		"Disable specific functionality; accepts any of [incrementals, delta]")
```

This way we have one easily extensible flag instead of a handful of bools.

ryanfkeepers (approved):
a couple nits/proposals, generally looks good.

src/internal/connector/data_collections.go
```diff
@@ -80,6 +80,8 @@ func (gc *GraphConnector) ProduceBackupCollections(
 
 	if !canMakeDeltaQueries {
 		logger.Ctx(ctx).Info("delta requests not available")
+
+		ctrlOpts.ToggleFeatures.DisableDelta = true
```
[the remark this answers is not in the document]
meain: This is something that I debated myself. The reason why I removed the `canMakeDeltaQueries` is because it was getting passed down quite a bit along with this which mostly carries the same(depends on how you look at it) data.

> Turning on the "I, the user, want to disable deltas" flag based on the state of data mixes concerns in ways that might be difficult to puzzle out if we're investigating state down the line.

The idea was that the log line just above this is what we would use to distinguish what caused us to not use the delta queries. I'm merging it for now, but let me know if the reasoning is not valid and I can switch it.

src/cli/options/options.go
meain: Hmm, interesting idea. I am wondering if we should be reserving `--disable` and `--enable` for non internal flags though.

Issue #3293, opened 2023-05-03
Optimize restore item loading


ashmrtn:
Restore isn't super optimized right now. This tracks some improvements we're making on that front

ashmrtn:
#3294 improved the logic to only load the directory that items are in once. This saves traversing the kopia hierarchy multiple times when loading items in the same kopia directory. However, this change keeps the logic that opens `io.ReaderCloser`s for every item being restored prior to returning from `ProduceRestoreCollections`

There's another set of changes that we could make where items are opened as the RestoreCollection returns them in the `Items` call. This results in a minor semantics change as errors related to finding/opening/reading those items would be returned at the end of `Items` instead of at the end of `ProduceRestoreCollections`. While I don't believe this would have a major impact on how Corso functions it may require some rewiring

ashmrtn:
As we appear to have solved the major issue we were hitting I'm going to close this for now. We can revisit if we want further improvements to restore time

PR #3294, opened 2023-05-03
Smarter item load ordering during ProduceRestoreCollections

ashmrtn:
Optimize loading items for restore a little bit by first grouping items
by directory, loading the directory once, and then loading all items
from the loaded directory. This brings item loading on my local machine
(communicating with remote S3) down to ~1.5min/1k items

Future improvements could lazily load items as they're returned in the
Items() call of each collection but doing so would change the semantics
of ProduceRestoreCollections() (specifically item not found errors would
be returned during Items() instead of during
ProduceRestoreCollections())

The kopia data collection has also been updated to hold onto a
reference to the folder it corresponds to. This folder reference is
used to service Fetch() calls

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3293

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/data_collection.go
+++ b/src/internal/kopia/data_collection.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/alcionai/corso/src/internal/data"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
@@ -60,6 +61,9 @@ func (kdc kopiaDataCollection) FullPath() path.Path {
 	return kdc.path
 }
 
+// Fetch returns the file with the given name from the collection as a
+// data.Stream. Returns a data.ErrNotFound error if the file isn't in the
+// collection.
 func (kdc kopiaDataCollection) Fetch(
 	ctx context.Context,
 	name string,
@@ -71,7 +75,7 @@ func (kdc kopiaDataCollection) Fetch(
 	}
 
 	if len(name) == 0 {
-		return nil, clues.Wrap(errNoRestorePath, "getting item stream").WithClues(ctx)
+		return nil, clues.Wrap(errNoRestorePath, "unknown item").WithClues(ctx)
 	}
 
 	e, err := kdc.dir.Child(ctx, encodeAsPath(name))
@@ -89,6 +93,11 @@ func (kdc kopiaDataCollection) Fetch(
 	}
 
 	size := f.Size() - int64(versionSize)
+	if size < 0 {
+		logger.Ctx(ctx).Infow("negative file size; resetting to 0", "file_size", size)
+
+		size = 0
+	}
 
 	if kdc.counter != nil {
 		kdc.counter.Count(size)
--- a/src/internal/kopia/data_collection.go
+++ b/src/internal/kopia/data_collection.go
@@ -18,10 +18,25 @@ var (
 )
 
 type kopiaDataCollection struct {
-	path         path.Path
-	streams      []data.Stream
-	snapshotRoot fs.Entry
-	counter      ByteCounter
+	path            path.Path
+	streams         []data.Stream
+	dir             fs.Directory
+	counter         ByteCounter
+	expectedVersion uint32
+}
+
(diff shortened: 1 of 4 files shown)


src/internal/kopia/data_collection.go
```diff
@@ -49,19 +64,49 @@ func (kdc kopiaDataCollection) Fetch(
...
-	if kdc.snapshotRoot == nil {
-		return nil, clues.New("no snapshot root")
+	ctx = clues.Add(ctx, "item_name", clues.Hide(name))
+
+	if kdc.dir == nil {
+		return nil, clues.New("no snapshot directory")
+	}
+
+	if len(name) == 0 {
+		return nil, clues.Wrap(errNoRestorePath, "getting item stream").WithClues(ctx)
```
ryanfkeepers, line 74:
```suggestion
		return nil, clues.Wrap(errNoRestorePath, "unknown item").WithClues(ctx)
```

ryanfkeepers (commented):
Generally LGTM.  I can approve if wanted, but I also think it's a good idea for @pandeyabs to review as well before merging.  For his opportunity to dig into/keep up with kopia changes, if nothing else.

pandeyabs (commented):
@ashmrtn I'll take a look at this PR later today. Since I am new to kopia, could you please tell me how were we loading items before this PR? I can go through the code but this would be easier. Trying to understand how this PR is improving item load time!

ashmrtn:
> I'll take a look at this PR later today.

sounds good, thanks!

>Since I am new to kopia, could you please tell me how were we loading items before this PR? I can go through the code but this would be easier. Trying to understand how this PR is improving item load time!

Previously the restore logic was pretty naive. The `ProduceRestoreCollections` function was passed a set of `path.Path` entries, each of which denotes the precise location in a kopia snapshot of the item to load (this info is sourced from RepoRef in backup details). The old logic would take each of those paths and [ask](https://github.com/alcionai/corso/blob/d067a7948326a42ade0b5d1d6f39d062f9dc5b78/src/internal/kopia/wrapper.go#L344) kopia to [fetch the nested entry](https://pkg.go.dev/github.com/kopia/kopia/snapshot/snapshotfs#GetNestedEntry), starting at the snapshot root. This would cause lots of traversals of the snapshot's hierarchy and apparently kopia doesn't cache those intermediate results (well or at all, not sure which at the moment)

This new PR adds a different call to [Directory.Child](https://pkg.go.dev/github.com/kopia/kopia/fs#Directory) to get entries directly from a loaded directory (the directory is still loaded by asking for a nested entry from the snapshot root). That alone cuts down on hierarchy traversals, thus helping load times

src/internal/kopia/wrapper.go
```diff
@@ -353,37 +352,89 @@ func getItemStream(
...
+type dirAndItems struct {
+	dir   path.Path
+	items []string
+}
 
-	return &kopiaDataStream{
-		uuid: decodedName,
-		reader: &restoreStreamReader{
-			ReadCloser:      r,
+func loadDirsAndItems(
```
pandeyabs, line 372: Could you add a short description for this function as well?

src/internal/kopia/data_collection.go
```diff
@@ -49,19 +64,49 @@ func (kdc kopiaDataCollection) Fetch(
...
+
+		return nil, clues.Wrap(err, "getting item").WithClues(ctx)
+	}
+
+	f, ok := e.(fs.File)
+	if !ok {
+		return nil, clues.New("object is not a file").WithClues(ctx)
+	}
+
+	size := f.Size() - int64(versionSize)
```
pandeyabs, line 91: Does this need a `size` < 0 error check?

pandeyabs (approved):
Added some minor follow ups, but overall looks good to me!

src/internal/kopia/data_collection.go
ashmrtn: hmmm probably a good check to have. I'll add a log message so we know it happened. Ideally it shouldn't though because this should be the version that's injected by the backup logic

PR #3295, opened 2023-05-03
add indeets test helper, implement in exchange op

ryanfkeepers:
Adds a helper for building expected details entries and checking
them after a backup.  Implements the helper in the exchange backup
tests in operations/backup integration.  Will follow with a onedrive
implementation.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3240

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,372 @@
...
+				_, ok := result.Sets[set].Locations[lr][ir]
+				assert.Falsef(t, ok, "deleted item in %s found in result: %s", set, lr)
+			}
+		}
+	}
+}
+
+func GetDeetsInBackup(
+	t *testing.T,
+	ctx context.Context, //revive:disable:context-as-argument
```
ashmrtn, line 344:
```suggestion
	ctx context.Context, //revive:disable-line:context-as-argument
```

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,372 @@
...
+	ss := ls.Subset(fromLocation)
+
+	for lr, items := range ss.Locations {
+		newLoc := toLocation
+		trimLoc := strings.TrimPrefix(lr, fromLocation)
+		trimLoc = strings.TrimPrefix(trimLoc, "/")
+
+		if len(trimLoc) > 0 {
+			newLoc += "/" + trimLoc
+		}
```
ashmrtn, line 135: I feel like a lot of this could be simplified by converting to a path.Builder, calling UpdateParent, and converting back to a string

ashmrtn (commented):
I've looked at most of the code to see how it works but haven't left comments for a full review

I think what concerns me looking at this code is that the details (and location information) that is used to compare correctness for later backups is itself sourced from a backup. If part of the reason for these tests is to ensure things like "LocationRef doesn't have container IDs" then sourcing the "golden image" of details that we then operate on from a backup cannot ensure that. It *will* ensure that things are moved around as expected and details contains the things we say it should which is useful in its own right. However, there's another set of important things related to actual values exposed to users and SDK consumers in backup details that this sort of test cannot check

ryanfkeepers:
> then sourcing the "golden image" of details that we then operate on from a backup cannot ensure that.

Good call.  I'll patch this up in a PR tailing on the drive addition of these tests.

src/pkg/backup/details/testdata/in_deets.go
ryanfkeepers: ooh, fancy.

ryanfkeepers:
@ashmrtn golden-style test expectations have been added in a follow-up PR: https://github.com/alcionai/corso/pull/3314

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,366 @@
...
+	return &locSet{
+		Locations: map[string]map[string]struct{}{},
+		Deleted:   map[string]map[string]struct{}{},
+	}
+}
+
+func (ls *locSet) AddItem(locationRef, itemRef string) {
+	if _, ok := ls.Locations[locationRef]; !ok {
+		ls.Locations[locationRef] = map[string]struct{}{}
+	}
```
ashmrtn, line 46: I think this should actually just call `AddLocation` as otherwise there may be an edge case with the following:
1. RemoveLocation(x)
2. AddItem(x, a)
3. // Location x still marked as removed

I'm unsure if that's intentional or not

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+
+		ls.RemoveLocation(lr)
+		ls.AddLocation(newLoc)
+	}
+
+	return toBuilder.String()
+}
+
+func (ls *locSet) RenameLocation(fromLocation, toLocation string) {
```
ashmrtn, line 120: This can call MoveLocation. Or you could remove the function to have a smaller API surface

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,366 @@
...
+	return odp.DriveID, nil
+}
+
+// ---------------------------------------------------------------------------
+// helpers and comparators
+// ---------------------------------------------------------------------------
+
+func CheckBackupDetails(
+	t *testing.T,
+	ctx context.Context, //revive:disable:context-as-argument
```
ashmrtn, line 293:
```suggestion
	ctx context.Context, //revive:disable-line:context-as-argument
```

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,366 @@
...
+
+		assert.Truef(
+			t,
+			strings.HasPrefix(ent.RepoRef, expect.RRPrefix),
+			"all details should begin with the expected prefix\nwant: %s\ngot:  %s",
+			expect.RRPrefix, ent.RepoRef)
+	}
+
+	for set := range expect.Sets {
+		assert.Subsetf(
```
ashmrtn, line 318: For our tests do we want a subset or a something like `ElementsMatch`? With a subset it seems like it won't error out on folders that should have been deleted (removed from the backup) but weren't? Would we need the deleted set if we could do an exact match?

If I remember right, we're backing up a specific directory so we shouldn't have to worry about other tests messing with state we'll see right? Do we intend to use this for stuff where we can't carefully control the data set?

src/internal/operations/backup_integration_test.go
```diff
@@ -815,43 +824,84 @@ func (suite *BackupOpIntegrationSuite) TestBackup_Run_exchangeIncrementals() {
...
+		for destName, cd := range gen.dests {
+			var longestLR string
+
+			for _, ent := range bupDeets.Entries {
+				// generated destinations should always contain items
+				if ent.Folder != nil {
+					continue
+				}
+
+				p, err := path.FromDataLayerPath(ent.RepoRef, false)
```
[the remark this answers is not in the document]
ryanfkeepers: Reporef isn't used to populate the inDeets checker.  It's just used as a filtering mechanism to ensure we're using the correct category in this loop.

src/pkg/backup/details/testdata/in_deets.go
ryanfkeepers: Yeah, waffled on this a lot.  I stuck with Subset because I want this to be generically usable by any test that needs to verify details, and not just these tests, without concern over external cross-pollination.  I think I'll add a flag that lets the caller opt for exact-match, since test cases like this should provide some concrete expectations about scope.

src/pkg/backup/details/testdata/in_deets.go
ashmrtn: I'm a little confused here. The comment you made makes it sound like it could just call `MoveLocation` but the code shows no changes. I also think it would be advantageous to have one of them call the other (or both call into some more generic code) since they're both doing almost exactly the same thing if you want to keep both functions exported. The only difference in the written code between the two is how `toBuilder` is generated

src/pkg/backup/details/testdata/in_deets.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// Subset produces a new locSet containing only Items and Locations
+// whose location matches the locationPfx
+func (id *InDeets) Subset(set, locationPfx string) *locSet {
+	return id.getSet(set).Subset(locationPfx)
+}
+
+// ---------------------------------------------------------------------------
+// whatSet helpers for extracing a set identifier from an arbitrary repoRef
```
ashmrtn, line 256: I found your t
```suggestion
// whatSet helpers for extracting a set identifier from an arbitrary repoRef
```

src/pkg/backup/details/testdata/in_deets.go
ashmrtn: this still needs to be fixed

src/pkg/backup/details/testdata/in_deets.go
ashmrtn: this still needs fixed

ashmrtn (approved):
couple of minor fixes that need addressed regarding revive linting. Would still strongly recommend combining code/logic for MoveLocation and RenameLocation

src/pkg/backup/details/testdata/in_deets.go
ryanfkeepers: Ah, I meant to say that I don't plan on changing it.  But after revisiting, you're right, and I've updated it.

src/pkg/backup/details/testdata/in_deets.go
ryanfkeepers: bargle!

Issue #3197, opened 2023-04-21
Properly handle cases of duplicate folder names 


gmatev:
For context on this see Corso Discord discussion - https://discord.com/channels/1022200980487557130/1022200981376745474/1098987178232651859

seems to be related to clients misbehaving and creating the duplicates. In this case this seemed to be "Detected Items" which seems like an known issue - https://forum.eset.com/topic/22459-outlook-detected-items-folder-behaving-strangely/

Should have a way of handling this case more gracefully. Given that this generally does not play well with Exchange and Outlook (See reports on crashes), I don't think it needs to be fully supported in a comprehensive way but some treatment that does not result in failed backups is desirable. 

Basic options:
* Pick the first occurrence only
* Pick the occurrence with most items only
* Handle explicitly "Detected items" - obviously not reliable

More comprehensive:
* Treat duplicate folders at the same level as a single folder for backup (merge). In place restores will go to the first occurrence

ashmrtn:
The more I think about this, the more wary I become of it from an implementation standpoint. It feels like all of the options given above make assumptions about how the data will evolve over time. Given that we have no reproducer of this, we're kind of guessing as to what behavior we'll see as things more forward

Below are some notes about potential issues we could run into as state evolves across backups

>Pick the first occurrence only

We'd need a clear definition of what "first" means. Would it be the one with the lexicographically lesser ID? We need some stable way to determine across backups. Also what happens if the "first" one we chose is deleted? Does the remaining one replace it entirely?


>Pick the occurrence with most items only

This assumes that the one we pick will always have the most items. If at some point it doesn't have the most items then we'll pick the other one, but may also fail to remove the items from the other one in the previous backup because we won't know we picked the other one on the previous backup

>Handle explicitly "Detected items" - obviously not reliable

yea, this seems tough to do and there's possibly other cases we'd run into this besides just something called "Detected items"

>Treat duplicate folders at the same level as a single folder for backup (merge). In place restores will go to the first occurrence

[...]

PR #3300, opened 2023-05-03
Temporary solution to duplicate Exchange folders

ashmrtn:
For Exchange email or contacts backup, pick on the folder with
the highest ID in the hopes that it will be the most recently
used one

A better solution for this situation should be coming soon

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #3197

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/service_iterators.go
+++ b/src/internal/connector/exchange/service_iterators.go
@@ -115,7 +115,7 @@ func filterContainersAndFillCollections(
 					"skipping duplicate folder with lesser ID",
 					"previous_folder_id", clues.Hide(oldCID),
 					"current_folder_id", clues.Hide(cID),
-				)
+					"duplicate_path", locPath)
 
 				// Readd this entry to the tombstone map because we remove it first off.
 				if oldDP, ok := dps[cID]; ok {
@@ -130,7 +130,8 @@ func filterContainersAndFillCollections(
 			logger.Ctx(ictx).Infow(
 				"switching duplicate folders as newer folder found",
 				"previous_folder_id", clues.Hide(oldCID),
-				"current_folder_id", clues.Hide(cID))
+				"current_folder_id", clues.Hide(cID),
+				"duplicate_path", locPath)
 
 			// Remove the previous collection from the maps. This will make us think
 			// it's a new item and properly populate it if it ever:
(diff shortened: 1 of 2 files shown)


src/internal/connector/exchange/service_iterators.go
```diff
@@ -99,6 +103,52 @@ func filterContainersAndFillCollections(
...
+		// should be skipped or the collection we previously added should be
+		// skipped.
+		//
+		// Calendars is already using folder IDs so we don't need to pick the
+		// "newest" folder for that.
+		if oldCID := dupPaths[locPath.String()]; category != path.EventsCategory && len(oldCID) > 0 {
+			if cID < oldCID {
+				logger.Ctx(ictx).Infow(
+					"skipping duplicate folder with lesser ID",
+					"previous_folder_id", clues.Hide(oldCID),
```
vkamra, line 116: Let's log the path as well (here and message below)

src/internal/connector/exchange/service_iterators.go
```diff
@@ -51,6 +51,10 @@ func filterContainersAndFillCollections(
 		// deleted from this map, leaving only the deleted folders behind
 		tombstones = makeTombstones(dps)
 		category   = qp.Category
+
+		// Stop-gap: Track folders by LocationPath and if there's duplicates pick
+		// the one with the lexicographically larger ID.
```
pandeyabs, line 56: Curious, is this documented behavior that larger ID == more recently created?

vkamra (approved):
Reviewed test cases offline with @ashmrtn . LGTM.

@ashmrtn - will address logging comment prior to merge.

src/internal/connector/exchange/service_iterators.go
ashmrtn: This isn't documented anywhere, just something we've noticed as we've gone along. We're not hoping to rely on it in the long run, this is more a temporary thing to unblock some folks until we have a better solution ready

pandeyabs: Gotcha! Better solution is moving to folder ids?

pandeyabs (commented):
The logic looks good to me.

PR #3317, opened 2023-05-05
Parse weburl to construct drive name for OneDrive

vkamra:
OneDrive drive names are always set to `OneDrive`. For users that have multiple drives, this makes it hard to distinguish
which drive items belong to.

This commit introduces logic where we parse the webURL to extract the underlying sharepoint document library name
and use that as the drive name for OneDrive.

For SP - this behavior is already implemented by the API - so we do not do this.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/drive.go
+++ b/src/internal/connector/onedrive/drive.go
@@ -3,6 +3,7 @@ package onedrive
 import (
 	"context"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/alcionai/clues"
@@ -26,6 +27,12 @@ const (
 	nextLinkKey           = "@odata.nextLink"
 	itemChildrenRawURLFmt = "https://graph.microsoft.com/v1.0/drives/%s/items/%s/children"
 	itemNotFoundErrorCode = "itemNotFound"
+
+	// https://learn.microsoft.com/en-us/graph/api/resources/drive?view=graph-rest-1.0#properties
+	// OneDrive personal drives will return personal. OneDrive for Business will return business.
+	// SharePoint document libraries will return documentLibrary.
+	oneDriveDriveType = "business"
+	spDriveDriveType  = "documentLibrary"
 )
 
 // DeltaUpdate holds the results of a current delta token.  It normally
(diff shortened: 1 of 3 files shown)


vkamra:
I'm working on a sharepoint integration test as well - currently only unit tests + onedrive integration coverage exists in this PR. Let's not block behind that though.

src/internal/connector/onedrive/drive.go
```diff
@@ -343,3 +350,21 @@ func DeleteItem(
...
+// driveName returns the drive name for a drive
+// For OneDrive - Graph always sets d.Name to `OneDrive` so
+// instead we parse the webURL and use the last element
+// e.g. if it is "https://<site name>/personal/<user>_onmicrosoft_com/Documents",
+// we return `Documents`
+func driveName(d models.Driveable) string {
+	dt := ptr.Val(d.GetDriveType())
+	switch dt {
+	case oneDriveDriveType:
+		if webURL := ptr.Val(d.GetWebUrl()); webURL != "" {
```
ashmrtn, line 363: nit: I think the standard is to use len checks for empty string?
```suggestion
		if webURL := ptr.Val(d.GetWebUrl()); len(webURL) > 0 {
```

src/internal/connector/onedrive/drive.go
```diff
@@ -343,3 +350,21 @@ func DeleteItem(
...
+			return path.Base(webURL)
```
ashmrtn, line 364:
```suggestion
			//nolint:forbidigo
			return path.Base(webURL)
```

src/internal/connector/onedrive/drive_test.go
```diff
@@ -277,6 +277,46 @@ func (suite *OneDriveUnitSuite) TestDrives() {
 	}
 }
 
+func (suite *OneDriveUnitSuite) Test_driveName() {
+	_, flush := tester.NewContext()
+	defer flush()
+
+	webURL := "https://<site name>/personal/<user>_onmicrosoft_com/Documents"
```
ashmrtn, line 284: have you happened to come across any documentation that shows the drive name is always the last path element? I know we've seen that some other parts of the URL can be changed by the admin so I thought I'd ask

ashmrtn (approved):

vkamra:
We're going to try something different

PR #3318, opened 2023-05-05
Fetch calendar by name on 409 ErrorFolderExist

pandeyabs:
Handle 409, `ErrorFolderExists` error in restore path while creating destination calendar. We need to fix this for all m365 services. This PR is focused on calendar only.

Context:
`CreateCalendar()` may fail with ErrorFolderExists under certain error conditions. For e.g. consider below scenario.

1. `CreateCalendar()` does a POST to graph to create restore destination calendar but this fails with 5xx.
2. It's possible that step 1 may have left some dirty state in graph. For e.g. it's possible that the destination folder in step 1 was actually created, but 5xx was returned due to other reasons.
3. So when we reattempt POST in such a scenario, we sometimes observe ErrorFolderExists error .
4. Corso should be resilient to such errors. To fix this, when we encounter such an error, we will do a GET to fetch the restore destination folder and add it to folder cache.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

Integration tests and mock tests will be added in a follow up PR. 
- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/api/events.go
+++ b/src/internal/connector/exchange/api/events.go
@@ -117,15 +117,16 @@ func (c Events) GetContainerByName(
 
 	resp, err := c.Stable.Client().UsersById(userID).Calendars().Get(ctx, options)
 	if err != nil {
-		return nil, graph.Wrap(ctx, err, "searching calendar by name")
+		return nil, graph.Stack(ctx, err).WithClues(ctx)
 	}
 
 	// We only allow the api to match one calendar with provided name.
 	// Return an error if multiple calendars exist(unlikely) or if no calendar
 	// is found.
 	if len(resp.GetValue()) != 1 {
-		return nil, graph.Wrap(ctx, err, "unexpected number of calendars").
+		err = clues.New("unexpected number of calendars returned").
 			With("returned_calendar_count", len(resp.GetValue()))
+		return nil, err
 	}
 
 	// Sanity check ID and name
--- a/src/internal/connector/exchange/api/events.go
+++ b/src/internal/connector/exchange/api/events.go
@@ -106,20 +106,16 @@ func (c Events) GetContainerByName(
 	ctx context.Context,
 	userID, name string,
 ) (models.Calendarable, error) {
-	service, err := c.service()
-	if err != nil {
-		return nil, graph.Stack(ctx, err)
-	}
-
 	filter := fmt.Sprintf("name eq '%s'", name)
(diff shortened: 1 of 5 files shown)


pandeyabs (commented):
Please don't review this yet. Integration test is failing

src/internal/connector/exchange/api/events.go
```diff
@@ -99,6 +99,47 @@ func (c Events) GetContainerByID(
 	return graph.CalendarDisplayable{Calendarable: cal}, nil
 }
 
+// GetContainerByName fetches calendar by name.
```
ryanfkeepers, line 102:
```suggestion
// GetContainerByName fetches a calendar by name.
```

src/internal/connector/exchange/api/events.go
```diff
@@ -99,6 +99,47 @@ func (c Events) GetContainerByID(
...
+
+	ctx = clues.Add(ctx, "calendar_name", name)
+
+	resp, err := c.Stable.Client().UsersById(userID).Calendars().Get(ctx, options)
+	if err != nil {
+		return nil, graph.Stack(ctx, err).WithClues(ctx)
+	}
+
+	// We only allow the api to match one calendar with provided name.
+	// Return an error if multiple calendars exist(unlikely) or if no calendar
```
ryanfkeepers, line 124:
```suggestion
	// Return an error if multiple calendars exist (unlikely) or if no calendar
```

src/internal/connector/exchange/api/events.go
```diff
@@ -99,6 +99,47 @@ func (c Events) GetContainerByID(
 	return graph.CalendarDisplayable{Calendarable: cal}, nil
 }
 
+// GetContainerByName fetches calendar by name.
+// Note: At the moment, we only support lookups for calendars which belong to
+// the default calendar group.
```
ryanfkeepers, line 104: This is a bit misleading, since we don't investigate calendar groups at all.  (and by default, all calendars are part of the default calendar group).  I think you could remove it, as that might produce less confusion overall.
```suggestion
```

ryanfkeepers (approved):
lgtm, just comment nits

src/internal/connector/exchange/api/events.go
pandeyabs: good point

PR #3319, opened 2023-05-05
More tests for non-delta queries

meain:
This add a sanity tests with delta queries disabled as well as more unit and integration tests to check this behaviour.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


.github/workflows/sanity-test.yaml
```diff
@@ -248,6 +248,53 @@ jobs:
           set -euo pipefail
           ./sanityTest
 
+      # non-delta backup
+      - name: Backup exchange incremental
```
ryanfkeepers, line 252:
```suggestion
      - name: Backup exchange incremental w/out delta
```

.github/workflows/sanity-test.yaml
```diff
@@ -248,6 +248,53 @@ jobs:
...
+        id: exchange-incremental-test
+        run: |
+          set -euo pipefail
+          echo -e "\nBackup Exchange incremental test\n" >> ${CORSO_LOG_FILE}
```
ryanfkeepers, line 256:
```suggestion
          echo -e "\nBackup Exchange incremental test w/out delta\n" >> ${CORSO_LOG_FILE}
```

src/internal/connector/data_collections.go
```diff
@@ -198,7 +200,7 @@ func checkServiceEnabled(
 		canMakeDeltaQueries = info.CanMakeDeltaQueries()
 	}
 
-	return true, canMakeDeltaQueries, nil
```
ryanfkeepers, line 201: very minor: imo this intersection should occur when handling the result of checkServiceEnabled, instead of passing down the opts into this call.  It's not much difference, but it does keep concerns separate.  The concern of this func is to report on the results of looking up the resource owner info.  The concern of the caller is to evaluate the combination of passed in options and lookup results.

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -1144,11 +1144,12 @@ func (suite *ServiceIteratorsSuite) TestFilterContainersAndFillCollections_incre
...
-		getter   mockGetter
-		resolver graph.ContainerResolver
-		dps      DeltaPaths
-		expect   map[string]endState
+		name            string
+		getter          mockGetter
+		resolver        graph.ContainerResolver
+		dps             DeltaPaths
+		expect          map[string]endState
+		skipForNonDelta bool
```
ryanfkeepers, line 1152: nit: the name here is slightly confusing
```suggestion
		skipWhenForcedNoDeltas bool
```

src/internal/connector/exchange/service_iterators_test.go
```diff
@@ -1403,54 +1405,71 @@ func (suite *ServiceIteratorsSuite) TestFilterContainersAndFillCollections_incre
...
 
-			ctx, flush := tester.NewContext()
-			defer flush()
+			if canMakeDeltaQueries {
+				name += "-delta"
+			} else {
+				if test.skipForNonDelta {
+					continue
+				}
+				name += "-non-delta"
```
ryanfkeepers, line 1421: Instead of silently skipping, lets make this loud:
```suggestion
				name += "-non-delta"
				
				// move this just inside suite.Run, line 1426 or so.
				if test.skipForNonDelta {
					t.Skip("intentionally skipped non-delta case")
				}
```

ryanfkeepers (approved):
lgtm, just nits

src/internal/connector/data_collections.go
meain: This was moved out. Looks like I missed rebased with that commit in this PR.

src/internal/connector/exchange/service_iterators_test.go
ashmrtn, line 1: these tests help ensure coverage, but I think one more test (or test case) may still need to be added to cover an edge case. I think we need a test with
* DeltaPaths (previous backup metadata) equal to what non-deltas would produce, needs at least one folder
* container resolver returning all but one folder in the DeltaPath set
* ensuring the folder not returned was marked as deleted

From what I've gathered by skimming the test cases, these tests ensure mostly that running a non-delta backup works assuming the previous backup was a delta backup (i.e. input metadata is consistent with what would be generated by a delta backup). We'll need the other way around too since we want to rely on deltas where possible

meain: I've modified `TestFilterContainersAndFillCollections_incrementals_nondelta` to make sure that it test all combinations of before/after * delta/non-delta.

PR #3330, opened 2023-05-05
renaming/cleanup for sharepoint perms pt.1

ryanfkeepers:
No logic changes, just renaming and minor cleanup.

PRs to follow:
1. collect various maps in onedrive collections into a single cache.
2. logic changes and tests to produce sharepoint permissions backup/restore e2e.
3. extend permission identity-type retention.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3135

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/connector/onedrive/permission.go
+++ b/src/internal/connector/onedrive/permission.go
@@ -4,11 +4,12 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
-	msdrive "github.com/microsoftgraph/msgraph-sdk-go/drive"
+	"github.com/microsoftgraph/msgraph-sdk-go/drive"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/graph"
+	"github.com/alcionai/corso/src/internal/connector/onedrive/api"
 	"github.com/alcionai/corso/src/internal/connector/onedrive/metadata"
 	"github.com/alcionai/corso/src/internal/data"
 	"github.com/alcionai/corso/src/internal/version"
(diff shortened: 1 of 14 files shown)


src/internal/connector/onedrive/permission.go
```diff
@@ -207,12 +209,12 @@ func UpdatePermissions(
 
 		pbody.SetRecipients([]models.DriveRecipientable{rec})
 
-		np, err := service.Client().DrivesById(driveID).ItemsById(itemID).Invite().Post(ctx, pbody, nil)
+		newPerm, err := api.PostItemPermissionUpdate(ctx, service, driveID, itemID, pbody)
 		if err != nil {
-			return graph.Wrap(ctx, err, "setting permissions")
+			return err
```
ashmrtn, line 214: will we get nice stack trace info that shows this is where `PostItemPermissionUpdate` was called if we don't wrap the error here? I haven't really looked through all of the stack trace generation code to know

src/internal/connector/onedrive/restore.go
```diff
@@ -157,17 +157,17 @@ func RestoreCollection(
 		return metrics, clues.Wrap(err, "creating drive path").WithClues(ctx)
 	}
 
-	if rootIDCache == nil {
-		rootIDCache = map[string]string{}
+	if rootFolderIDCache == nil {
+		rootFolderIDCache = map[string]string{}
```
ashmrtn, line 161: I still feel like this is a code smell. It seems like either we should be passing it in all the time and error out if it's not or we should never pass it in and create it in this function

ashmrtn (approved):

src/internal/connector/onedrive/restore.go
ryanfkeepers: It is a smell, but it gets cleaned up in the next PR.

src/internal/connector/onedrive/permission.go
ryanfkeepers: Stack trace generation is attached to any clues wrap or stack call.  So we can stack here, too, if wanted.

PR #3334, opened 2023-05-05
add sanity testing for sharepoint permissions

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3135

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/cmd/factory/impl/sharepoint.go
+++ b/src/cmd/factory/impl/sharepoint.go
@@ -14,21 +14,21 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
-var odFilesCmd = &cobra.Command{
+var spFilesCmd = &cobra.Command{
 	Use:   "files",
 	Short: "Generate OneDrive files",
-	RunE:  handleOneDriveFileFactory,
+	RunE:  handleSharePointLibraryFileFactory,
 }
 
-func AddOneDriveCommands(cmd *cobra.Command) {
-	cmd.AddCommand(odFilesCmd)
+func AddSharePointCommands(cmd *cobra.Command) {
+	cmd.AddCommand(spFilesCmd)
 }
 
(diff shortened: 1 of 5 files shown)


src/cmd/factory/impl/onedrive.go
```diff
@@ -35,20 +36,23 @@ func handleOneDriveFileFactory(cmd *cobra.Command, args []string) error {
...
 	}
 
-	gc, acct, inp, err := getGCAndVerifyUser(ctx, User)
+	gc, acct, inp, err := getGCAndVerifyResourceOwner(ctx, connector.Users, User)
 	if err != nil {
 		return Only(ctx, err)
 	}
 
-	deets, err := generateAndRestoreOnedriveItems(
+	sel := selectors.NewOneDriveBackup([]string{User}).Selector
```
ashmrtn, line 44: Should this be a restore selector?

```suggestion
	sel := selectors.NewOneDriveRestore([]string{User}).Selector
```

src/cmd/factory/impl/sharepoint.go
```diff
@@ -0,0 +1,71 @@
...
+	if utils.HasNoFlagsAndShownHelp(cmd) {
+		return nil
+	}
+
+	gc, acct, inp, err := getGCAndVerifyResourceOwner(ctx, connector.Sites, Site)
+	if err != nil {
+		return Only(ctx, err)
+	}
+
+	sel := selectors.NewSharePointBackup([]string{Site}).Selector
```
ashmrtn, line 44: Should this also be a restore selector?

ashmrtn (approved):

src/cmd/factory/impl/onedrive.go
ryanfkeepers: technically doesn't matter, since we're only using it as an IDNameProvider.  I think I wrote this before the provider mock existed.  I'll switch it to use that instead of the selector.

Issue #3335, opened 2023-05-05
Only backup primary drive for a user 


vkamra:
OneDrive users have only "one drive"/document library. There are some cases where additional libraries can get associated with a user (e.g. [1])

These libraries should not be backed up. 

[1]
https://learn.microsoft.com/en-us/microsoft-365/compliance/retention-policies-sharepoint?view=o365-worldwide#how-retention-works-for-sharepoint-and-onedrive

PR #3336, opened 2023-05-05
lookup user default drive for onedrive

ryanfkeepers:
instead of enumerating all drives during backup, if we're processing a onedrive source, opt to look up only the default drive instead of enumerating all drives.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3335

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
closing: the approach here is better for long term handling, but getting tests to work is a non-trivial feat.  Replaced with https://github.com/alcionai/corso/pull/3339

PR #3338, opened 2023-05-05
Create a collection to merge other collection data

ashmrtn:
Now that Corso changed how it loads data from
kopia during restore create some code to merge
results from several underlying collections.
This will come in handy when we have multiple
external service folders with distinct IDs but
which have the same display name and path
(LocationRef)

This code also defines the behavior when
multiple collections in the merge set have the
same item and `Fetch` is called

This should be merged into #3337 prior to it
merging to main

This is not expected to change system
functionality

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3197

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/merge_collection.go
+++ b/src/internal/kopia/merge_collection.go
@@ -1,0 +1,110 @@
+package kopia
+
+import (
+	"context"
+	"errors"
+
+	"github.com/alcionai/clues"
+	"golang.org/x/exp/slices"
+
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+type col struct {
+	storagePath string
+	data.RestoreCollection
+}
+
+type mergeCollection struct {
+	cols []col
+	// Technically don't need to track this but it can help detect errors.
+	fullPath path.Path
+}
+
+func (mc *mergeCollection) addCollection(
+	storagePath string,
+	c data.RestoreCollection,
+) error {
+	if c == nil {
+		return clues.New("adding nil collection").
+			With("current_path", mc.FullPath())
+	} else if mc.FullPath().String() != c.FullPath().String() {
+		return clues.New("attempting to merge collection with different path").
+			With("current_path", mc.FullPath(), "new_path", c.FullPath())
+	}
+
+	mc.cols = append(mc.cols, col{storagePath: storagePath, RestoreCollection: c})
+
+	// Keep a stable sorting of this merged collection set so we can say there's
+	// some deterministic behavior when Fetch is called. We don't expect to have
+	// to merge many collections.
+	slices.SortStableFunc(mc.cols, func(a, b col) bool {
+		return a.storagePath < b.storagePath
+	})
+
+	return nil
+}
+
+func (mc mergeCollection) FullPath() path.Path {
+	return mc.fullPath
+}
+
+func (mc *mergeCollection) Items(
+	ctx context.Context,
+	errs *fault.Bus,
+) <-chan data.Stream {
+	res := make(chan data.Stream)
+
+	go func() {
+		defer close(res)
+
+		logger.Ctx(ctx).Infow(
+			"getting items for merged collection",
+			"merged_collection_count", len(mc.cols))
+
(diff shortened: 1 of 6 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -399,14 +399,23 @@ func loadDirsAndItems(
 			return nil, el.Failure()
 		}
 
-		ictx := clues.Add(ctx, "restore_path", col.restorePath)
+		var (
+			mergeDC *mergeCollection
+			ictx    = clues.Add(ctx, "restore_path", col.restorePath)
+			merge   = len(col.storageDirs) > 0
```
ashmrtn, line 405: this should be `> 1` instead of `> 0`. I'll push a change for it later

src/internal/kopia/wrapper.go
```diff
@@ -399,14 +399,23 @@ func loadDirsAndItems(
 			return nil, el.Failure()
 		}
 
-		ictx := clues.Add(ctx, "restore_path", col.restorePath)
+		var (
+			mergeDC *mergeCollection
```
ryanfkeepers, line 403: nit
```suggestion
			mergeCol *mergeCollection
```

src/internal/kopia/merge_collection.go
```diff
@@ -0,0 +1,110 @@
...
+
+		logger.Ctx(ctx).Infow(
+			"getting items for merged collection",
+			"merged_collection_count", len(mc.cols))
+
+		for _, c := range mc.cols {
+			// Unfortunately doesn't seem to be a way right now to see if the
+			// iteration failed and we should be exiting early.
+			ictx := clues.Add(ctx, "merged_collection_storage_path", c.storagePath)
+			logger.Ctx(ictx).Infow("sending items from merged collection")
```
ryanfkeepers, line 72: This feels like it'll be a bit chatty.  How about debug?
```suggestion
			logger.Ctx(ictx).Debug("sending items from merged collection")
```

src/internal/kopia/merge_collection.go
```diff
@@ -0,0 +1,110 @@
...
+	name string,
+) (data.Stream, error) {
+	logger.Ctx(ctx).Infow(
+		"fetching item in merged collection",
+		"merged_collection_count", len(mc.cols))
+
+	for _, c := range mc.cols {
+		ictx := clues.Add(ctx, "merged_collection_storage_path", c.storagePath)
+
+		logger.Ctx(ictx).Infow("looking for item in merged collection")
```
ryanfkeepers, line 98: Similar concern about chattiness:
```suggestion
		logger.Ctx(ictx).Debug("looking for item in merged collection")
```

src/internal/kopia/wrapper.go
```diff
@@ -399,14 +399,23 @@ func loadDirsAndItems(
...
+			ictx    = clues.Add(ctx, "restore_path", col.restorePath)
+			merge   = len(col.storageDirs) > 0
+		)
+
+		if merge {
+			mergeDC = &mergeCollection{fullPath: col.restorePath}
+			res = append(res, mergeDC)
+		}
```
ryanfkeepers, line 411: What if, instead of the conditional, you made everything a mergeCollection?  That would simplify the code a bit and remove some conditional paths (plus extra mental load about state).  And any mergeCollection with a single item should still behave as expected.

ryanfkeepers (approved):
lgtm, just nits

src/internal/kopia/merge_collection.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+type col struct {
+	storagePath string
+	data.RestoreCollection
+}
+
+type mergeCollection struct {
```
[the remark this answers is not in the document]
ashmrtn: that's a good idea, thanks

src/internal/kopia/wrapper.go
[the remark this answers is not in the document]
ashmrtn: sure, we can start with that. We may revisit down the road though as it will cost us an extra goroutine for every call to `Items()`

PR #3339, opened 2023-05-05
disguise user's default drive as enum

ryanfkeepers:
When calling drive enumeration, secretly return only the user's default drive if running in onedrive, to prevent us from loading multiple drives unintentionally.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3335

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - --user and --mailbox flags have been removed from CLI examples for details and restore commands (they were already not supported, this only updates the docs).
 - Improve restore time on large restores by optimizing how items are loaded from the remote repository.
 - Remove exchange item filtering based on m365 item ID via the CLI.
+- OneDrive backups no longer include a user's non-default drives.
```
gmatev, line 19: This was never the case to begin with actually

ryanfkeepers: Kind of a regression, really.  It wasn't the case to begin with.  Then for sharepoint compliance we put in a generic drive enumerator that, we assumed, would only pick up the "one" oneDrive, opening us up to the issue.  So it's been this way for some time now, even in versions before we officially released sharepoint.

vkamra (approved):

Issue #3344, opened 2023-05-08
[Bug]: kopia error with corso version 0.7.0 


abaldoni:
### What happened?

During backup the following error is generated (no errors with version 0.6.1, the destination repository was created with corso 0.6.0 and it already contains other backups):

Error: running backup: persisting collection backups: building kopia snapshot

### Corso Version?

Corso v0.7.0

### Where are you running Corso?

VMware Photon OS 4 x86_64 
Storage provided by local minio version RELEASE.2023-05-04T21-44-30Z

### Relevant log output

```shell
corso backup create onedrive --user foo@bar.baz --log-level backup
Logging to file: /home/corso/.cache/corso/logs/2023-05-08T10-05-37Z.log
Connecting to repository:      0s done
Connecting to M365:         done
Backing Up ∙ foo@bar.baz
Discovering items to backup:      2s done
∙ files:      1s done
Discovered 66 items to backup
Backing up data:      5s done
Error: running backup: persisting collection backups: building kopia snapshot

Completed Backups:
No backups available
Error: 1 of 1 backups failed:
∙ foo@bar.baz: running backup: persisting collection backups: building kopia snapshot
```

ryanfkeepers:
Hi @abaldoni, thanks for the bug report!  We're going to need a couple more details in order to dig into the cause here, primarily the log file itself `/home/corso/.cache/corso/logs/2023-05-08T10-05-37Z.log` (or relevant snippets of it).

If you'd like to join us on discord (https://discord.gg/ewuWryaj), we can discuss this with you live, plus get a secure transfer of that info.

heimdal-31:
I have seen the same.  Sharepoint backups only.  Repo created with 0.6.1.  7.0 client gives the same error.  Error details seem to indicate it is not handling a 302 correctly:

 "error": "sharepoint reader: non-2xx http response: 302 Found", "errorVerbose": "302 Found\n\t/home/runner/work/corso/corso/src/internal/connector/onedrive/item.go:167\nnon-2xx http response\n\t/home/runner/work/corso/corso/src/internal/connector/onedrive/item.go:167\nsharepoint reader\n\t/home/runner/work/corso/corso/src/internal/connector/onedrive/item.go:40", "error_labels": {"status_code_302":{}}}
2023-05-08T20:22:55.903Z        INFO    graph/middleware.go:158 graph api error: 302 Found

[...]

PR #3348, opened 2023-05-08
some quick logging and error naming updates

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3344

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/graph/middleware.go
```diff
@@ -122,42 +124,61 @@ func (mw *LoggingMiddleware) Intercept(
 		return resp, err
 	}
 
-	ctx = clues.Add(ctx, "status", resp.Status, "statusCode", resp.StatusCode)
-	log := logger.Ctx(ctx)
+	ctx = clues.Add(
+		ctx,
+		"status", resp.Status,
+		"statusCode", resp.StatusCode,
```
pandeyabs, line 130: nit: Should this be `status_code` ?

src/internal/connector/graph/middleware.go
```diff
@@ -100,6 +99,9 @@ func LoggableURL(url string) pii.SafeURL {
 	}
 }
 
+// 1 MB
+const logMBLimit = 1 * 1048576
```
pandeyabs, line 103: minor nit: 1 * 1024 * 1024

src/internal/connector/graph/middleware.go
```diff
@@ -122,42 +124,61 @@ func (mw *LoggingMiddleware) Intercept(
...
 
-	// Return immediately if the response is good (2xx).
-	// If api logging is toggled, log a body-less dump of the request/resp.
-	if (resp.StatusCode / 100) == 2 {
-		if logger.DebugAPIFV || os.Getenv(log2xxGraphRequestsEnvKey) != "" {
-			log.Debugw("2xx graph api resp", "response", getRespDump(ctx, resp, os.Getenv(log2xxGraphResponseEnvKey) != ""))
-		}
+	var (
+		log       = logger.Ctx(ctx)
+		respClass = resp.StatusCode / 100
```
pandeyabs, line 135: nit: I've seen this statusCode / 100 technique being used in several places in the code to find resp class. It's clever! 
Can we move this to a small helper? Doing raw divisions in the code can be prone to bugs.

src/internal/connector/graph/middleware.go
```diff
@@ -122,42 +124,61 @@ func (mw *LoggingMiddleware) Intercept(
...
+		log.With("response", getRespDump(ctx, resp, true)).
+			Error("graph api error: " + resp.Status)
+
 		return resp, err
 	}
 
-	msg := fmt.Sprintf("graph api error: %s", resp.Status)
+	// Log api calls according to api debugging configurations.
+	switch respClass {
+	case 2:
```
pandeyabs, line 161: nit: Enums for response classes would be better

src/internal/connector/graph/middleware.go
```diff
@@ -122,42 +124,61 @@ func (mw *LoggingMiddleware) Intercept(
...
+		if logExtra {
+			// only dump the body if it's under a size limit.  We don't want to copy gigs into memory for a log.
+			dump := getRespDump(ctx, resp, os.Getenv(log2xxGraphResponseEnvKey) != "" && resp.ContentLength < logMBLimit)
```
pandeyabs, line 164: 1Mb per 2xx response still seems like a lot? With 10k such responses, we would reach 10GB ( over a ~10 min period).

pandeyabs: I guess the average response would be much smaller.

src/internal/connector/graph/middleware.go
ryanfkeepers: whoops, yes, thanks

src/internal/connector/graph/middleware.go
ryanfkeepers: I think a helper is overbearing in this case, since it's pretty well localized code with clear inputs and outputs.  But if we start to see it pop up in multiple packages (say, if we generate clients outside of graph), that'll be a good time to centralize.

src/internal/connector/graph/middleware.go
ryanfkeepers: I'm honestly not sure there's any cases where dumping a body on a 2xx helps us at all.  This is one step towards memory management before removing it altogether.  I think we'd be smarter to remove it, but I'm going to keep it in for now just to be sure.

pandeyabs: Fair. I agree this is the best approach right now

pandeyabs (approved):

PR #3350, opened 2023-05-08
handle large item client redirects

ryanfkeepers:
adds the khttp redirectMiddleware to the http wrapper used for large item downloads, to ensure that proxy servers and other 3xx class responses appropriately follow their redirection.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3344

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/http_wrapper.go
+++ b/src/internal/connector/graph/http_wrapper.go
@@ -140,13 +140,20 @@ func defaultTransport() http.RoundTripper {
 }
 
 func internalMiddleware(cc *clientConfig) []khttp.Middleware {
-	return []khttp.Middleware{
+	mw := []khttp.Middleware{
 		&RetryMiddleware{
 			MaxRetries: cc.maxRetries,
 			Delay:      cc.minDelay,
 		},
+		khttp.NewRedirectHandler(),
 		&LoggingMiddleware{},
 		&ThrottleControlMiddleware{},
 		&MetricsMiddleware{},
 	}
+
+	if len(cc.appendMiddleware) > 0 {
+		mw = append(mw, cc.appendMiddleware...)
+	}
+
+	return mw
 }
--- a/src/internal/connector/graph/http_wrapper_test.go
+++ b/src/internal/connector/graph/http_wrapper_test.go
@@ -85,7 +85,7 @@ func (suite *HTTPWrapperUnitSuite) TestNewHTTPWrapper_redirectMiddleware() {
 	)
 
 	// can't use gock for this, or else it'll short-circuit the transport,
-	// and thus skip all the middelware
+	// and thus skip all the middleware
 	hdr := http.Header{}
 	hdr.Set("Location", "localhost:99999999/smarfs")
 	toResp := &http.Response{
--- a/src/internal/connector/graph/http_wrapper_test.go
+++ b/src/internal/connector/graph/http_wrapper_test.go
@@ -2,9 +2,11 @@ package graph
 
 import (
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/alcionai/clues"
+	khttp "github.com/microsoft/kiota-http-go"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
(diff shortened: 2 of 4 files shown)


src/internal/connector/graph/http_wrapper.go
```diff
@@ -140,13 +140,20 @@ func defaultTransport() http.RoundTripper {
...
 			MaxRetries: cc.maxRetries,
 			Delay:      cc.minDelay,
 		},
+		khttp.NewRedirectHandler(),
 		&LoggingMiddleware{},
 		&ThrottleControlMiddleware{},
 		&MetricsMiddleware{},
 	}
+
+	if len(cc.appendMiddleware) > 0 {
```
pandeyabs, line 154: One limitation of `appendMiddleware` slice is that it can only be appended to the tail of middlewares. 
Is the longer term thought process here to supply the entire middleware chain through `appendMiddleware`?

src/internal/connector/graph/service.go
```diff
@@ -225,6 +227,14 @@ func MinimumBackoff(dur time.Duration) Option {
 	}
 }
 
+func appendMiddleware(mw ...khttp.Middleware) Option {
+	return func(c *clientConfig) {
+		if len(mw) > 0 {
+			c.appendMiddleware = mw
```
pandeyabs, line 233: Should this be made thread safe?

src/internal/connector/graph/service.go
```diff
@@ -173,6 +173,8 @@ type clientConfig struct {
 	// The minimum delay in seconds between retries
 	minDelay           time.Duration
 	overrideRetryCount bool
+
+	appendMiddleware []khttp.Middleware
```
pandeyabs, line 177: very minor nit: `additionalMiddleware` feels like a better fit.

src/internal/connector/graph/http_wrapper_test.go
```diff
@@ -43,3 +45,70 @@ func (suite *HTTPWrapperIntgSuite) TestNewHTTPWrapper() {
...
+
+type HTTPWrapperUnitSuite struct {
+	tester.Suite
+}
+
+func TestHTTPWrapperUnitSuite(t *testing.T) {
+	suite.Run(t, &HTTPWrapperUnitSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+func (suite *HTTPWrapperUnitSuite) TestNewHTTPWrapper_redirectMiddleware() {
```
pandeyabs, line 76: Could you add a small description for this test?

src/internal/connector/graph/http_wrapper_test.go
```diff
@@ -43,3 +45,70 @@ func (suite *HTTPWrapperIntgSuite) TestNewHTTPWrapper() {
...
+
+	var (
+		t    = suite.T()
+		uri  = "https://graph.microsoft.com"
+		path = "/fnords/beaux/regard"
+		url  = uri + path
+	)
+
+	// can't use gock for this, or else it'll short-circuit the transport,
+	// and thus skip all the middelware
```
pandeyabs, line 88: middleware*

pandeyabs (approved):
Left a few minor nits. Other than that, LGTM!

src/internal/connector/graph/http_wrapper.go
ryanfkeepers: This alone feels pretty hacky and test oriented, and we have no need for additional middleware injection.  We definitely do not want to supply the entire middleware chain through this option, as that opens us up to all sorts of unexpected behaviors.  I'd strongly hesitate before extending this behavior at all.

src/internal/connector/graph/service.go
ryanfkeepers: I like `append` because it can be very important to know whether the middleware gets called at the beginning of the stack or the end of it.

src/internal/connector/graph/http_wrapper.go
pandeyabs: It would be good to add a small note next to `appendMiddleware` field description, saying that it's for test use only.

PR #3353, opened 2023-05-09
Fix json tag for modified time in sharepoint info

meain:
Small typo fix for metadata store.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/backup/details/details.go
+++ b/src/pkg/backup/details/details.go
@@ -872,7 +872,7 @@ type SharePointInfo struct {
 	DriveID    string    `json:"driveID,omitempty"`
 	ItemName   string    `json:"itemName,omitempty"`
 	ItemType   ItemType  `json:"itemType,omitempty"`
-	Modified   time.Time `josn:"modified,omitempty"`
+	Modified   time.Time `json:"modified,omitempty"`
 	Owner      string    `json:"owner,omitempty"`
 	ParentPath string    `json:"parentPath,omitempty"`
 	Size       int64     `json:"size,omitempty"`


meain:
Converting this to draft as this could cause issues when restoring. Delaying merging this until that is discussed.

src/pkg/backup/details/details.go
```diff
@@ -872,7 +872,7 @@ type SharePointInfo struct {
 	DriveID    string    `json:"driveID,omitempty"`
 	ItemName   string    `json:"itemName,omitempty"`
 	ItemType   ItemType  `json:"itemType,omitempty"`
-	Modified   time.Time `josn:"modified,omitempty"`
+	Modified   time.Time `json:"modified,omitempty"`
```
ryanfkeepers, line 875: we should add a linter for this typo, so we can protect against it in the future.

We could add an `OldModified time.Time `json:"Modified,omitempty"` field, and run a post-read transformer.  but that feels a little heavy.

meain: > we should add a linter for this typo, so we can protect against it in the future.

Ahh, good call. I'll add that.


> We could add an OldModified time.Time json:"Modified,omitempty"` field, and run a post-read transformer. but that feels a little heavy.

Hmm, that actually doesn't sound that bad.

meain: > we should add a linter for this typo, so we can protect against it in the future.

I looked a bit into this. Does not look like we can't really target this easily. I was thinking about using `forbidigo` or `misspell` within golangci-lint, but does not look like we can use either.

ashmrtn: hmmm wonder how hard it would be to write a linter that looks for this... 🤔 😂

meain: I don't expect it to be that hard  🤔

ashmrtn: true, Golang has some good tools for building a linter. Mostly it's about figuring out what you need. I wrote one a while back in my spare time for comment wording and it didn't take all that much effort. Not sure how well directives like `json` are supported in the AST package though

meain: Looks like I paniced for nothing. Golang does not care about case when unmarshalling, TIL.

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match. By default, object keys which don't have a corresponding struct field are ignored (see Decoder.DisallowUnknownFields for an alternative). 

https://pkg.go.dev/encoding/json#Unmarshal

meain: That said, I'll probably look into the linter at a later stage. This PR should now be good to go.

ashmrtn: deserializing things is always such a struggle 😞 at least we've got some more info on this case

ashmrtn (approved):

Issue #2300, opened 2023-01-27
Use LargeFile http client for restore operations


vkamra:
We're using a Resty client in `upload_session.go`. We should replace this with the LargeFile http client and add the same retry handling we have in the download file logic

vkamra:
Closed with #3355

PR #3355, opened 2023-05-09
use graph client for large data set

neha-Gupta1:
- Use graph client instead of Resty client. This brings uniformity and will help in utilising other features like Middlewares from graph client wrapper
- Moved the upload session to graph package to use Graph http-client wrapper

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* https://github.com/alcionai/corso/issues/2300

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/internal/connector/graph/uploadsession.go
+++ b/src/internal/connector/graph/uploadsession.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"net/http"
 
 	"github.com/alcionai/clues"
 
@@ -20,7 +19,7 @@ const (
 
 // Writer implements an io.Writer for a M365
 // UploadSession URL
-type writer struct {
+type largeItemWriter struct {
 	// Identifier
 	id string
 	// Upload URL for this item
@@ -32,15 +31,17 @@ type writer struct {
 	client            httpWrapper
 }
 
-func NewWriter(id, url string, size int64) *writer {
-	return &writer{id: id, url: url, contentLength: size, client: *NewNoTimeoutHTTPWrapper()}
+func NewLargeItemWriter(id, url string, size int64) *largeItemWriter {
+	return &largeItemWriter{id: id, url: url, contentLength: size, client: *NewNoTimeoutHTTPWrapper()}
 }
 
 // Write will upload the provided data to M365. It sets the `Content-Length` and `Content-Range` headers based on
 // https://docs.microsoft.com/en-us/graph/api/driveitem-createuploadsession
-func (iw *writer) Write(p []byte) (int, error) {
+func (iw *largeItemWriter) Write(p []byte) (int, error) {
 	rangeLength := len(p)
-	logger.Ctx(context.Background()).
+	ctx := context.Background()
+
+	logger.Ctx(ctx).
 		Debugf("WRITE for %s. Size:%d, Offset: %d, TotalSize: %d",
 			iw.id, rangeLength, iw.lastWrittenOffset, iw.contentLength)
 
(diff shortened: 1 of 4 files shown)


src/internal/connector/graph/uploadsession.go
```diff
@@ -48,17 +48,25 @@ func (iw *writer) Write(p []byte) (int, error) {
...
-			contentRangeHeaderKey: fmt.Sprintf(
-				contentRangeHeaderValueFmt,
-				iw.lastWrittenOffset,
-				endOffset-1,
-				iw.contentLength),
-			contentLengthHeaderKey: fmt.Sprintf("%d", rangeLength),
-		}).
-		SetBody(bytes.NewReader(p)).
-		Put(iw.url)
+	req, err := http.NewRequest("PUT", iw.url, bytes.NewReader(p))
```
ryanfkeepers, line 51: It worries me that this is based off of the HTTPWrapper client, but then avoids calling its `Request()` func, which is the primary point of centralizing that struct.  Seems like the headers are the only standout conflict between the two uses?  What if you updated Request() to swap the `map[string]string` for a `http.Headers` value?  That should help you to work within the architecture.

src/internal/connector/graph/uploadsession.go
```diff
@@ -48,17 +48,25 @@ func (iw *writer) Write(p []byte) (int, error) {
...
+	if err != nil {
+		return 0, clues.Wrap(err, "uploading item").With(
```
ryanfkeepers, line 53: These `with` values seem useful for the whole call chain.  Can you pull them out to a `clues.Add(ctx...)` above the request instead, please?

src/internal/connector/graph/uploadsession.go
```diff
@@ -29,11 +29,11 @@ type writer struct {
 	contentLength int64
 	// Last item offset that was written to
 	lastWrittenOffset int64
-	client            *resty.Client
+	client            httpWrapper
 }
 
 func NewWriter(id, url string, size int64) *writer {
```
ryanfkeepers, line 35: The very generic names here concern me more after moving this file to the graph package.  When it was in `uploadsession`, the scope of the writer was self evident by the package name.  But a `graph.writer` could mean a lot of things, whereas this has a very specific purpose.  How about a name like `LargeItemWriter`?

src/internal/connector/graph/uploadsession.go
neha-Gupta1: Make sense.

src/internal/connector/graph/uploadsession.go
neha-Gupta1: Agreed. I have done the changes.

src/internal/connector/graph/uploadsession.go
neha-Gupta1: With refactoring this is not required now.

src/internal/connector/graph/uploadsession.go
```diff
@@ -29,36 +28,36 @@ type writer struct {
...
-		Put(iw.url)
+	headers := make(map[string]string)
+	headers[contentRangeHeaderKey] = fmt.Sprintf(
+		contentRangeHeaderValueFmt,
+		iw.lastWrittenOffset,
+		endOffset-1,
+		iw.contentLength)
+	headers[contentLengthHeaderKey] = fmt.Sprintf("%d", rangeLength)
+
+	_, err := iw.client.Request(ctx, "PUT", iw.url, bytes.NewReader(p), headers)
```
ryanfkeepers, line 60: Always prefer http library consts for http values, please.
```suggestion
	_, err := iw.client.Request(ctx, http.MethodPut, iw.url, bytes.NewReader(p), headers)
```

ryanfkeepers (changes requested):
Looks good, one remaining nit.

Per requested changes tag: the code is actually fine, but at this time because we want to delay these changes until after the next release.  After that time, this can be approved by anyone and merged.

ryanfkeepers (approved):

ryanfkeepers:
> Per requested changes tag: the code is actually fine, but at this time because we want to delay these changes until after the next release. After that time, this can be approved by anyone and merged.

v0.8.0 is released, the block on this PR is lifted.

Issue #2951, opened 2023-03-27
Refine graph api global limiter


ryanfkeepers:
## Issue
The current implementation assumes all calls are limited to to exchange's 10k/10min threshold.  However, different services, methods, and even data types may all engage with different, or altogether separate, limits.  The limiter needs additional refinement so that we don't unnecessarily restrict throughput.

## Improvements
- [x] Separate limits for onedrive.
- [x] Additional token consumption when querying permissions.

PR #3357, opened 2023-05-09
introduce per-service rate limiter configurations

ryanfkeepers:
Adds a context passdown that allows GC to define the service being queried at a high level, and the rate limiter to utilize different rate limiters based on that info.  Malformed or missing limiter config uses the default limiter.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2951

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/data_collections.go
```diff
@@ -213,15 +217,16 @@ func (gc *GraphConnector) ConsumeRestoreCollections(
 		return nil, clues.Wrap(err, "malformed azure credentials")
 	}
 
-	switch selector.Service {
+	switch sels.Service {
 	case selectors.ServiceExchange:
+		ctx = graph.BindRateLimiterConfig(ctx, graph.LimiterCfg{Service: path.ExchangeService})
```
pandeyabs, line 222: Is this a duplicate call? Line 208 above should be sufficient for all services, right?

src/internal/connector/graph/middleware.go
```diff
@@ -369,18 +370,60 @@ func (mw RetryMiddleware) getRetryDelay(
...
+	defaultLimiter = rate.NewLimiter(defaultPerSecond, defaultMaxCap)
+)
+
+type LimiterCfg struct {
+	Service path.ServiceType
+}
+
+type limiterCfgKey string
+
+const limiterCfgCtxKey = "corsoGaphRateLimiterCfg"
```
pandeyabs, line 391: Graph*

pandeyabs (approved):
Left a few comments. Please also fix linter errors. 
Overall, looks good!

src/internal/connector/data_collections.go
ryanfkeepers: whoops, yep!

src/internal/connector/graph/middleware.go
ryanfkeepers: what a gaff

PR #3358, opened 2023-05-09
refine rate limiter: per-query token consumption

ryanfkeepers:
Now that the rate limiter is split by service, we can further split by per-query token consumption.  Two primary service cases exist: all exchange queries assume to cost a single token (for now).  Drive-service queries are split between permissions (5), default cost (2), and single-item or delta gets (1).

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #2951

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (commented):
Hey @ryanfkeepers  could you rebase this on main? Currently this PR is also showing changes from #3357

ryanfkeepers:
@pandeyabs Done!  Thanks for the ping.

src/internal/connector/graph/middleware.go
```diff
@@ -419,13 +424,55 @@ func extractRateLimiterConfig(ctx context.Context) (LimiterCfg, bool) {
 	return lc, ok
 }
 
+type limiterConsumptionKey string
```
pandeyabs, line 427: nit: Very subjective, but to me this typedef isn't adding much value.

src/internal/connector/graph/middleware.go
```diff
@@ -419,13 +424,55 @@ func extractRateLimiterConfig(ctx context.Context) (LimiterCfg, bool) {
...
+
+const (
+	limiterConsumptionCtxKey       limiterConsumptionKey = "corsoGraphRateLimiterConsumption"
+	defaultLimiterConsumption                            = 1
+	driveDefaultLimiterConsumption                       = 2
```
pandeyabs, line 432: nit: driveLC*? Just to be consistent with other constants defined below.

src/internal/connector/graph/middleware.go
```diff
@@ -419,13 +424,55 @@ func extractRateLimiterConfig(ctx context.Context) (LimiterCfg, bool) {
...
+}
+
+func ctxLimiterConsumption(ctx context.Context, defaultConsumption int) int {
+	l := ctx.Value(limiterConsumptionCtxKey)
+	if l == nil {
+		return defaultConsumption
+	}
+
+	lc, ok := l.(int)
+	if !ok || lc < 1 {
```
pandeyabs, line 455: Would be good to add an upper bound check too.

src/internal/connector/graph/middleware.go
```diff
@@ -419,13 +424,55 @@ func extractRateLimiterConfig(ctx context.Context) (LimiterCfg, bool) {
...
+	// limit consumption rate for anything permissions related
+	PermissionsLC = 5
+)
+
+// ConsumeNTokens ensures any calls using this context will consume
+// n rate-limiter tokens.  Default is 1, and this value does not need
+// to be established in the context to consume the default tokens.
+// This should only get used on a per-call basis, to avoid cross-pollination.
+func ConsumeNTokens(ctx context.Context, n int) context.Context {
+	return context.WithValue(ctx, limiterConsumptionCtxKey, n)
```
pandeyabs, line 445: Should we validate n here?

pandeyabs (approved):
Left some comments but overall looks good!

src/internal/connector/graph/middleware.go
ryanfkeepers: It's actually not subjective in this case.  Golang context values are a type of `map[any]any` (where, of course, `any` is an alias for the empty interface, `interface{}`).  Using `any` as a key causes the map to store both the value and the type of the key.

This way it's possible to store equally valued strings without collision, but that only works if both values have a unique type.  Go ahead and try to add to a context using `WithValue` and a plain string: you should see a linter error about keying with an untyped string.  

Bit of an ingenious way to ensure that a shared variable with a generic storage interface stays free of common-name collisions.

src/internal/connector/graph/middleware.go
ryanfkeepers: sure

src/internal/connector/graph/middleware.go
ryanfkeepers: We already have lazy validation, which I think is marginally safer in this case, and doesn't need both ends guarded.

src/internal/connector/graph/middleware.go
ryanfkeepers: Upper bounds are naturally checked by the rate limiter.  If a capacity greater than the per-second burst is requested, then the limiter (and thus api call) will error.  Should be clearly identifiable in CI testing, and prevents us from needing to check against N different upper bounds (one for each limiter implementation, should they differ).

src/internal/connector/graph/middleware.go
pandeyabs: Thanks! Yeah, this makes total sense.

PR #3359, opened 2023-05-09
Lint GitHub actions changes

ashmrtn:
Reduce errors when updating actions and workflows
and hopefully stop silent failures for things

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


meain:
Action-Lint step seems to be skipped. We should probably add this as a required check on GitHub for merges.

ashmrtn:
+1 for making it required

Per being skipped, I'm not sure if that's some weird state issue or not. I looked at the output from the run and it seems like the old version of filechecker was run (doesn't output var for changed workflow stuff only) instead of the new version. This caused the linter step to be skipped but it seems like it will probably be picked up once this is merged?

PR #3363, opened 2023-05-09
normalize folders restored in testing

ryanfkeepers:
Normalize all test folders to use a constant prefix: Corso_Test.  This cleans up and centralizes all per- test variations on the restore destination.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/tester/restore_destination.go
```diff
@@ -1,11 +1,26 @@
...
 )
 
-func DefaultTestRestoreDestination() control.RestoreDestination {
-	// Use microsecond granularity to help reduce collisions.
-	return control.DefaultRestoreDestination(dttm.SafeForTesting)
+const RestoreFolderPrefix = "Corso_Test"
+
+func DefaultTestRestoreDestination(namespace string) control.RestoreDestination {
+	var (
+		dest = control.DefaultRestoreDestination(dttm.SafeForTesting)
```
pandeyabs, line 14: Looks like we are overwriting dest on line 23 below. Should we remove this call?

src/internal/connector/sharepoint/restore.go
```diff
@@ -125,7 +125,7 @@ func RestoreCollections(
 }
 
 // restoreListItem utility function restores a List to the siteID.
-// The name is changed to to Corso_Restore_{timeStame}_name
+// The name is changed to to {DestName}_{name}
```
pandeyabs, line 128: This is prod code. Just to confirm, is this change intended?

pandeyabs (approved):
Looks good. Thanks for this!

src/internal/tester/restore_destination.go
ryanfkeepers: Nope, this will populate values beyond just the container name.

PR #3364, opened 2023-05-09
move nested test folder creation to top level

ryanfkeepers:
tests that create items which require automatic cleanup should always create those items at the root directory. We have one bad citizen test which creates them in a subdirectory.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E


pandeyabs (approved):
🚢

PR #3368, opened 2023-05-09
move nightly to us nighttime, disable load tests

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/nightly_test.yml
+++ b/.github/workflows/nightly_test.yml
@@ -3,8 +3,8 @@ on:
   workflow_dispatch:
 
   schedule:
-    # Run every day at 0 minutes and 0 hours (midnight GMT)
-    - cron: "0 0 * * *"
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_test.yml
```diff
@@ -3,8 +3,8 @@ on:
   workflow_dispatch:
 
   schedule:
-    # Run every day at 0 minutes and 0 hours (midnight GMT)
-    - cron: "0 0 * * *"
+    # Run every day at 06:00 GMT (roughly 10pm PST)
+    - cron: "0 6 * * *"
```
ashmrtn, line 7: maybe we should move this a bit earlier so that it's not middle of the day for other folks? 7 or 8pm PT would still be "off hours" to a large extent

ashmrtn (approved):

PR #3369, opened 2023-05-09
release the sensitive-info flag

ryanfkeepers:
Not 100% happy with the flag name, and am open to suggestions.

---

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 
 ### Added
+- Released the --sensitive-info flag, which will automatically obscure private data in logs.
```
ntolia, line 11: Can we make this `--mask-sensitive-data` instead?

src/pkg/logger/logger.go
```diff
@@ -110,7 +107,7 @@ func addFlags(fs *pflag.FlagSet, defaultFile string) {
 		&SensitiveInfoFV,
 		SensitiveInfoFN,
 		PIIPlainText,
-		fmt.Sprintf("set the format for sensitive info in logs to %s|%s|%s", PIIHash, PIIMask, PIIPlainText))
+		fmt.Sprintf("set the format for sensitive info in logs to %s|%s", PIIHash, PIIPlainText))
```
ntolia, line 110: Why do we have an option for this flag? Wouldn't it be easier to just have a `mask-sensitive-info` flag?

website/docs/setup/configuration.md
```diff
@@ -132,7 +132,17 @@ directory within the container.
...
+
 You can also use `stdout` or `stderr` as the `--log-file` location to redirect the logs to "stdout" and "stderr" respectively.
+This setting can cause logs to compete with progress bar displays in the termal.
+We suggest using the `--hide-progress` option if you plan to log to stdout or stderr.
+
+
+Log entries contain plain text information by default, including user and file names.
+If you plan to share your log outputs publically, the `--sensitive-info hash` option will replace all private
+details with a cryptographically secure hash.
+
```
ntolia, line 145:
```suggestion
Log entries, by default, include user names and file names. The `--sensitive-info hash` option can be
used to replace this information with anonymized hashes.
```

ntolia (dismissed):
Can we improve the UX of this feature by changing the flag name and not having an option for the flag?

src/pkg/logger/logger.go
ryanfkeepers: Two reasons: one, it technically supports a third option that is hidden (mask, which replaces sensitive data with a flat `***`).  Two, I've been leaning towards flags with args over boolean flags for ease of future exensibility.  If we decide to formally add a new value here (maybe like letting users choose their hash seed `hash,seed:my_seed`) then we can extend the arguments rather than introduce a new flag.  Loosely held opinions here; so if you feel strongly about a boolean option, we can swap it out for that.

CHANGELOG.md
ryanfkeepers: sure can.

src/pkg/logger/logger.go
ntolia: I believe we can have the best of both worlds here.

- Default (no flag): Plain text
- No option `--mask-sensitive-info` flag: Current hashing (becomes the default masking option)
- Flag w/ options `--mask-sensitive-info <hash/mask>`: Whatever optional behavior we want to support

This supports a good default CLI UX while allowing us to support options in the future.

ryanfkeepers:
@ntolia updated and ready for re-review.

Flag display in `corso --help`
```
Flags:
      --mask-sensitive-data   anonymize personal data in log output
```

website/docs/setup/configuration.md
```diff
@@ -132,7 +132,16 @@ directory within the container.
 Corso generates a unique log file named with its timestamp for every invocation.
 The default location of Corso's log file is shown below but the location can be overridden by using the `--log-file` flag.
 The log file will be appended to if multiple Corso invocations are pointed to the same file.
+
```
ntolia, line 135:
```suggestion
```

website/docs/setup/configuration.md
```diff
@@ -132,7 +132,16 @@ directory within the container.
...
+
 You can also use `stdout` or `stderr` as the `--log-file` location to redirect the logs to "stdout" and "stderr" respectively.
+This setting can cause logs to compete with progress bar displays in the termal.
+We suggest using the `--hide-progress` option if you plan to log to stdout or stderr.
+
```
ntolia, line 140:
```suggestion
```

ntolia (commented):
Docs build will likely complain about double blank lines.

website/docs/setup/configuration.md
```diff
@@ -132,7 +132,14 @@ directory within the container.
...
 The default location of Corso's log file is shown below but the location can be overridden by using the `--log-file` flag.
 The log file will be appended to if multiple Corso invocations are pointed to the same file.
+
 You can also use `stdout` or `stderr` as the `--log-file` location to redirect the logs to "stdout" and "stderr" respectively.
+This setting can cause logs to compete with progress bar displays in the termal.
+We suggest using the `--hide-progress` option if you plan to log to stdout or stderr.
+
+Log entries, by default, include user names and file names. The `--mask-sensitive-data` option can be
+used to replace this information with anonymized hashes.
+
```
ntolia, line 142:
```suggestion
```

ntolia (commented):
LGTM but @gmatev should also take a look. Happy to approve by EOD if he is busy though.

src/pkg/logger/logger.go
ryanfkeepers: Unfortunately, our CLI library requires categorizing flags as expecting arguments or not.  Calling a flag that expects arguments without any will either 1/ treat the next flag as an argument, or 2/ error (if no other flags are described).

We can get very close to that in the future, if needed, by maintaining both a common boolean flag for quick usage and clean UX (`--mask-sensitive-data`, in this case), plus second flag with more granular configuration for power usage (ex: `--mask-config mask:hash,seed:my-seed`).

I've noodled on doing the same for logging.  Maintain both the smaller `--log-level foo` and `--log-file bar` flags, plus a fully extensible flag for more granular config `--logging level:foo,file:bar,format:cli,minimal`.

PR #3378, opened 2023-05-10
fix up per-service token constraints

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/events/events.go
+++ b/src/internal/events/events.go
@@ -188,11 +188,11 @@ func tenantHash(tenID string) string {
 // metrics aggregation
 // ---------------------------------------------------------------------------
 
-type m string
+type metricsCategory string
 
 // metrics collection bucket
 const (
-	APICall m = "api_call"
+	APICall metricsCategory = "api_call"
 )
 
 // configurations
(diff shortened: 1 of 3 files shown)


src/internal/events/events.go
```diff
@@ -261,6 +263,12 @@ func Inc(cat m, keys ...string) {
 	metrics.IncrCounter(cats, 1)
 }
 
+// IncN increments the given category by N.
+func IncN(n int, cat m, keys ...string) {
```
pandeyabs, line 267: nit: Can we change `m` to something like `metricCategory` for better readability?

src/internal/events/events.go
```diff
@@ -191,7 +191,9 @@ func tenantHash(tenID string) string {
 type m string
```
pandeyabs, line 191: nit: Can we change `m` to something like `metricCategory` for better readability?

pandeyabs (approved):

PR #3383, opened 2023-05-10
add repoRef pii transformer

ryanfkeepers:
adds a transformer func in path/elements.go to safely log a repoRef (or any other dir ref) without needing to conceal the entire string.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test


src/pkg/path/elements.go
```diff
@@ -28,7 +28,26 @@ var piiSafePathElems = pii.MapWithPlurals(
 	LibrariesCategory.String(),
 	PagesCategory.String(),
 	DetailsCategory.String(),
-)
+
+	// well known folders
+	// https://learn.microsoft.com/en-us/graph/api/resources/mailfolder?view=graph-rest-1.0
+	"archive",
```
ashmrtn, line 34: hmmm I'm actually not sure if we don't want to hide these. Is the user able to delete folders like this? If they are deleted, do we still get them returned? The presence or absence of things can still leak information so it may be best if we just hide all non-prefix folders

src/pkg/path/elements.go
```diff
@@ -86,3 +105,16 @@ func (el Elements) String() string {
...
 }
+
+// ---------------------------------------------------------------------------
+// helpers
+// ---------------------------------------------------------------------------
+
+// LoggableDir takes in a path reference (of any structure) and conceals any
+// non-standard elements (ids, filenames, foldernames, etc).
+func LoggableDir(ref string) string {
+	elems := Split(ref)
```
ashmrtn, line 116: may want to do some extra trimming on this and/or ensure this is only used for safe call paths. Split has the note that it can't handle unescaped trailing `/`

ryanfkeepers: done

src/pkg/path/elements.go
ryanfkeepers: Not sure I understand how you're approaching this question.  The intent of this func is to add qol to logging repoRef and other path strings, without needing transform the value to a path.  None of this is intended for user facing display, so I'm not sure it matters what the user is able to do, or not, with the folders.  In the meantime, for developer debugging, knowing that a folder matched a well-known graph folder, as opposed to needing to make an estimated guess that may miss the point is something that could make or break resolving a bug.

ashmrtn: >for developer debugging, knowing that a folder matched a well-known graph folder, as opposed to needing to make an estimated guess

This is the crux of the matter. Knowing that a folder exists or not (even if it's a "well known" folder) tells the developer something and could be considered an information leak. Whether this is something we're alright leaking is what I'm wondering

An analogous situation to this sort of information leak is a website telling you that the password you gave for the username is incorrect as opposed to saying the username and/or the password was incorrect. Knowing that a user _does_ exist on the system is a piece of information that can be used/leveraged (i.e. cross-referencing the username with a leaked username/password database for another site and trying passwords to see if they were reused or using this knowledge to build up a profile of what services the user interacts with online based on username reuse)

ryanfkeepers: pinging @gmatev  for input.

I don't find the proposed situation to be analogous in this case.  Simply by the capacity to engage with a drive backup of a given service and category for that user we have established most of the knowledge that these folders provide.  1/ the user is licensed to use the service, 2/ they meet the minimum requirements to back up the given category of data.  What these folderNames tell us is that the system generated set of folders is present, and that depends on 1 and 2 succeeding in the first place.

Imagine a website has already clearly confirmed that you have an existing account with their service.  At that point, is it a leak if they reveal that your account has a password?

gmatev: Interesting debate, but I think you folks might be overthinking this. The goal here is to make it easy for use to get logs from users without them worrying about revealing some personal information they are not comfortable with or something proprietary. It is not to prevent absolute information disclosure that someone an use to derive information. 

With that criteria in mind, I think it is reasonable to to leave the well-known folders unmasked. 

If it makes a difference, I suspect, although I have not tested, that these folders will comeback automatically as needed even if they are deleted (if they could be deleted).

ashmrtn (approved):

PR #3384, opened 2023-05-10
minor updates to the path.Path interface

ryanfkeepers:
Swaps path.Append from a single item method to accept a variadic list of string elements.  Since 95% of all calls to path.Append were items, also adds a shorthand
AppendItem func to the interface for easy clarity. Finally, adds a Last() method to elements for getting the last element in the slice.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/path/resource_path.go
```diff
@@ -253,21 +253,29 @@ func (rp dataLayerResourcePath) Dir() (Path, error) {
...
 		service:  rp.service,
 		category: rp.category,
 		hasItem:  isItem,
 	}, nil
 }
 
+func (rp dataLayerResourcePath) AppendItem(item string) (Path, error) {
+	if rp.hasItem {
+		return nil, clues.New("appending to an item path")
+	}
```
ashmrtn, line 274: do we need to check for this here and return an error? It seems like it would be fine to just rely on the call below to catch this and then we don't have to worry about the error messages drifting apart. There should also a test to make sure this fails as expected so changes down the road should fail the test

ashmrtn (commented):
Need some basic tests for AppendItem (just to make sure it fails/doesn't fail as expected)

src/pkg/path/resource_path.go
ryanfkeepers: fair, fair.  Test added.

ashmrtn (approved):

PR #3398, opened 2023-05-11
separate log level and format config

ryanfkeepers:
Separates the log level and log format.  This will allows sdk consumers using `SeedLogger` to utilize JSON
format logs without sacrificing log level controls.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/pkg/logger/logger.go
+++ b/src/pkg/logger/logger.go
@@ -12,6 +12,9 @@ import (
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/exp/slices"
+
+	"github.com/alcionai/corso/src/internal/common"
 )
 
 // Default location for writing logs, initialized in platform specific files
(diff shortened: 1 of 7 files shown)


src/pkg/logger/logger.go
```diff
@@ -112,9 +133,10 @@ func addFlags(fs *pflag.FlagSet, defaultFile string) {
 
 // Settings records the user's preferred logging settings.
 type Settings struct {
-	File        string // what file to log to (alt: stderr, stdout)
-	Level       string // what level to log at
-	PIIHandling string // how to obscure pii
+	File        string    // what file to log to (alt: stderr, stdout)
+	Format      logFormat // whether to format as text (console) or json (cloud)
+	Level       logLevel  // what level to log at
```
ashmrtn, line 138: these are now both strings and can be passed in, uninitialized via `Seed`. What happens if they're never initialized?

ryanfkeepers: Yeah, that's all getting a bit cluttered, isn't it?  I did a cleanup here: https://github.com/alcionai/corso/pull/3401.  We can squash that into this branch when its ready.

ryanfkeepers:
@ashmrtn cleanup branch has been squashed in, this is ready for a final review, please.

Issue #3388, opened 2023-05-11
[Bug]: repoID is empty in backup events


neha-Gupta1:
### What happened?

repoID is blank/empty in backup start and end events.

### Corso Version?

Corso v0.7.0

### Where are you running Corso?

Local Ubuntu system

vkamra:
Observed pattern was that we get Corso Start/End events _with_ repoIDs but Backup Start/End events *do not* have repoIDs

gmatev:
The relevant recent change is https://github.com/alcionai/corso/pull/3071/files but id does not change where the ID is coming from. In the previous version the repoID was used as userID, but I suspect Rudderstack might be assigning an ID itself in that case so we may not have noticed it before.

vkamra:
@neha-Gupta1  - Adding notes from discussion on this:

- We should add string constants that we initialize the repoID and use to set for error cases (where repoID lookup fails) so it helps to narrow down the issue (e.g. `not_init` , `not_found`) - this will help narrow down what case we're seeing this in
- We should change the code to use the repoID in the config file (similar to what corso_start events do)

neha-Gupta1:
Added a PR- https://github.com/alcionai/corso/pull/3449 for more info

vkamra:
Will close after we do a release and verify.

vkamra:
Need to validate this issue is fixed after the next release - 0.8.0 still had issues

neha-Gupta1:
@gmatev are we still seeing this issue?

PR #3399, opened 2023-05-11
defer end-of-operation events

ryanfkeepers:
move the end-of-op notification event to a defer within backup/restore.Do().  This ensures we'll get the end event data even in case of failure.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3388

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -170,6 +170,22 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 			events.BackupID:  op.Results.BackupID,
 		})
 
+	defer func() {
+		op.bus.Event(
+			ctx,
+			events.BackupEnd,
+			map[string]any{
+				events.BackupID:   op.Results.BackupID,
+				events.DataStored: op.Results.BytesUploaded,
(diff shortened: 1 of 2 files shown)


gmatev:
@ryanfkeepers is this off the right branch? Seems like a bunch of other changes as well.

ryanfkeepers:
> is this off the right branch? Seems like a bunch of other changes as well.

@gmatev there's some git weirdness going on.  This pr should be isolated to only the relevant changes now, but for some reason I need it pointed to the intermediate branch to do that.

gmatev (approved):

src/internal/operations/backup.go
```diff
@@ -170,6 +170,22 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 			events.BackupID:  op.Results.BackupID,
 		})
 
+	defer func() {
+		op.bus.Event(
+			ctx,
+			events.BackupEnd,
+			map[string]any{
+				events.BackupID:   op.Results.BackupID,
```
vkamra, line 178: @ryanfkeepers - I took a look and all the fields referenced here seem same to use even if the backup fails - can you confirm that?

PR #3401, opened 2023-05-12
logger settings cleanup

ryanfkeepers:
better ensurance around log settings falling back
to defaults in cases where population is incomplete.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :muscle: Manual

### DIFF
--- a/src/pkg/logger/logger.go
+++ b/src/pkg/logger/logger.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/exp/slices"
 
 	"github.com/alcionai/corso/src/internal/common"
 )
@@ -23,16 +24,6 @@ var (
 	logCore   *zapcore.Core
 	loggerton *zap.SugaredLogger
 )
-
-// type logLevel int
-
-// const (
-// 	Debug logLevel = iota
-// 	Info
-// 	Warn
-// 	Error
-// 	Disabled
-// )
 
 type logLevel string
 
@@ -53,12 +44,12 @@ const (
 	LFJSON logFormat = "json"
 )
 
-type piiHandling string
+type piiAlg string
 
 const (
-	PIIHash      piiHandling = "hash"
-	PIIMask      piiHandling = "mask"
-	PIIPlainText piiHandling = "plaintext"
+	PIIHash      piiAlg = "hash"
+	PIIMask      piiAlg = "mask"
+	PIIPlainText piiAlg = "plaintext"
 )
 
 // flag names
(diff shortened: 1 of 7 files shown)


src/pkg/logger/logger.go
```diff
@@ -64,27 +74,28 @@ const (
...
-	LogLevelFV          = string(LLInfo)
+	logFileFV           string
+	LogFormatFV         string
+	LogLevelFV          string
 	ReadableLogsFV      bool
 	MaskSensitiveDataFV bool
 
-	LogFile string // logFileFV after processing
+	LogFile     string // logFileFV after processing
+	PIIHandling string // piiHandling after MaskSensitiveDataFV processing
```
ashmrtn, line 84: do these need to be exported? Are they used?

src/pkg/logger/logger.go
```diff
@@ -239,11 +240,47 @@ func GetLogFile(logFileFlagVal string) string {
...
+type Settings struct {
+	File        string      // what file to log to (alt: stderr, stdout)
+	Format      logFormat   // whether to format as text (console) or json (cloud)
+	Level       logLevel    // what level to log at
+	PIIHandling piiHandling // how to obscure pii
+}
+
+// EnsureDefaults sets any non-populated settings to their default value.
+// exported for testing without circular dependencies.
+func (s Settings) EnsureDefaults() Settings {
```
ashmrtn, line 253: this does ensure defaults, but I'm not sure if all things are checked for validity. For example, I don't think there's a fallback for PII handling if it's outside the range of expected values. At a broader level, which functions can expect only valid input data and which functions need to guard against invalid data?

src/pkg/logger/logger.go
ryanfkeepers: guess not.

src/pkg/logger/logger.go
ryanfkeepers: Added checks for validity.

src/pkg/logger/logger_test.go
```diff
@@ -83,3 +83,41 @@ func (suite *LoggerUnitSuite) TestPreloadLoggingFlags() {
 	assert.Equal(t, logger.LFText, settings.Format, "settings.Format")
 	assert.Equal(t, logger.PIIHash, settings.PIIHandling, "settings.PIIHandling")
 }
+
+func (suite *LoggerUnitSuite) TestPreloadLoggingFlags_badArgsEnsureDefault() {
```
ashmrtn, line 87: thanks for the test! 😄

ashmrtn (approved):

src/pkg/logger/logger_test.go
ryanfkeepers: You're welcome, thank you for the nudge!

PR #3416, opened 2023-05-15
migrate exchange api to pkg/services/m365

ryanfkeepers:
No logic changes, just code movement.

Begins the process ofcentralizing the graph package in pkg/services/m365, instead of dividing up the funcs between different connector subpackages.  While the ownership of apis in each connector subpackage does make sense, this movement will reduce duplication, centralize import arrangements, and make it easy
to export usage of the api to sdk consumers.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/api.go
```diff
@@ -19,37 +22,87 @@ import (
...
+// TODO: use selector or path for granularity into specific folders or specific date ranges
+type GraphQuery func(ctx context.Context, userID string) (serialization.Parsable, error)
+
+// GraphRetrievalFunctions are functions from the Microsoft Graph API that retrieve
+// the default associated data of a M365 object. This varies by object. Additional
+// Queries must be run to obtain the omitted fields.
+type GraphRetrievalFunc func(
+	ctx context.Context,
+	user, m365ID string,
+) (serialization.Parsable, error)
```
ashmrtn, line 108: Can these actually be removed? I'm not sure if they're being used anywhere

ashmrtn (approved):

ashmrtn:
I guess the only question I really have is if we want to export this functionality to SDK consumers. I don't really want corso to accidentally become a provider of Graph API access in some strange way as then we'd have more to worry about when it comes to breaking changes. It's unlikely sure, but not impossible

src/pkg/services/m365/api/api.go
ryanfkeepers: I'll check into that as a follow-up.  Wanted to avoid logic/removal changes in this PR.

PR #3419, opened 2023-05-16
Only remove previous deltas from metadata under some conditions

### LINKED ISSUE #2334: Refactor invalid metadata handling during incremental backups

ashmrtn:
Only allow removing previous delta tokens and not
previous paths. This ensures folder deletions will
be picked up if all items are enumerated

Updates tests to match new behavior, mostly
adjusting expectations around del list when
doNotMerge is applied

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #2334

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/onedrive/collections.go
```diff
@@ -199,36 +198,35 @@ func deserializeMetadata(
 				// layers would have caught this. Worst case if we don't handle this
 				// we end up in a situation where we're sourcing items from the wrong
 				// base in kopia wrapper.
-				if errors.Is(err, errExistingMapping) {
-					return nil, nil, clues.Wrap(err, "deserializing metadata file").WithClues(ictx)
+				if err != nil {
```
meain, line 201: We should either keep the old behaviour of failing only for `errExistingMapping` or update the comment to reflect this change.

meain (approved):
Thanks. I guess it makes sense to keep it similar to Exchange. :D

Need a small change, but approving to unblock.

PR #3420, opened 2023-05-16
Place concurrency limiter middleware after retry middlewares

pandeyabs:
This is to safeguard against the scenario where retry handlers might spawn additional requests to graph. Such retries may cause us to exceed concurrent request limits, leading to 429s ( `Application is over its MailboxConcurrency limit`) 

This PR will ensure that such retry requests are also funneled through the concurrency limiter.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph/service.go
```diff
@@ -264,6 +259,14 @@ func kiotaMiddlewares(
...
+
+	// Optionally add concurrency limiter middleware if it has been initialized.
+	if concurrencyLim != nil {
+		mw = append(mw, concurrencyLim)
+	}
+
+	mw = append(mw, []khttp.Middleware{
 		&RateLimiterMiddleware{},
 		&MetricsMiddleware{},
 	}...)
```
ryanfkeepers, line 272: `slice...` isn't needed here, you can just add the individual values. 
```suggestion
	mw = append(
	        mw,
		&RateLimiterMiddleware{},
		&MetricsMiddleware{})
```

ryanfkeepers (approved):
I know this is a draft, but I think it's a sensible change for us to put through whether or not it directly solves the issue.  So you get an approval.

Issue #3392, opened 2023-05-11
[Bug]: Users without given service enabled should generate warning/info message, but not an error


mlipscombe:
### What happened?

When creating a backup for all users using something like `corso backup create exchange --user '*'`, users that do not have the relevant service enabled (e.g. no Exchange license), generate an error in the logs and cause a non-zero exit code:

```
Error: 7 of 11 backups failed:
...
∙ xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx: running backup: producing backup data collections: checking service access: service is not enabled for that resource owner
∙ xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx: running backup: producing backup data collections: checking service access: service is not enabled for that resource owner
```

This occurs with both `exchange` and `onedrive` backups.

These should be warnings or info messages at their highest, because if there is no service, there's also nothing to back up.  They should not cause a non-zero exit code, because that makes catching real errors difficult in scripts using corso.

### Corso Version?

Corso version: v0.7.0

### Where are you running Corso?

All platforms.

vkamra:
Adding to the next dev cycle

neha-Gupta1:
PR- https://github.com/alcionai/corso/pull/3426

neha-Gupta1:
Done as part of https://github.com/alcionai/corso/pull/3426

PR #3426, opened 2023-05-16
don't throw error on backup with service disabled

neha-Gupta1:
If a user does not have a service(like exchange) enabled- don't throw an error but just skip that user and move forward with other backups.

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3392

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E


src/cli/backup/backup.go
```diff
@@ -225,6 +226,13 @@ func runBackups(
 
 		err = bo.Run(ictx)
 		if err != nil {
+			if strings.Contains(err.Error(), graph.ErrServiceNotEnabled.Error()) {
+				logger.Ctx(ctx).Debugf("Service not enabled for creating backup for %s", bo.ResourceOwner.Name())
```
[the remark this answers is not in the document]
neha-Gupta1: That makes sense. I will raise a follow up PR for this. And track it as part of the same [issue](https://github.com/alcionai/corso/issues/3392)

src/cli/backup/backup.go
```diff
@@ -225,6 +226,13 @@ func runBackups(
 
 		err = bo.Run(ictx)
 		if err != nil {
+			if strings.Contains(err.Error(), graph.ErrServiceNotEnabled.Error()) {
```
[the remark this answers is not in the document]
neha-Gupta1: Done!

src/internal/operations/backup.go
```diff
@@ -338,6 +349,44 @@ func (op *BackupOperation) do(
 	return deets, nil
 }
 
+func Precheck(
```
ryanfkeepers, line 352: Don't think this needs to be exported.  Also, please add: a header comment explaining the high level purpose and behavior of this func.  

Name nit:
```suggestion
func CheckIsRunnable(
```

src/internal/operations/backup.go
```diff
@@ -338,6 +349,44 @@ func (op *BackupOperation) do(
...
+	}
+
+	if ui == nil || len(ui.ServicesEnabled) == 0 {
+		return graph.ErrServiceNotEnabled
+	}
+
+	_, ok := ui.ServicesEnabled[service]
+	if !ok {
+		return graph.ErrServiceNotEnabled
+	}
```
ryanfkeepers, line 385: I'd rather move this ownership out of operations and into GC.  After all, the clients being produced here already exist in the connector instance.

To keep in line with the other operations interfaces, I suggest extending the `BackupProducer` interface in `operations/inject.go` with a func like `IsBackupRunnable(ctx, service, resourceOwner) (bool, error)`. That way you can pass GC into this func directly as the interface supplier (or call IsBackupRunnable on line 149 and skip this Precheck func altogether).

Finally, adding testing around the IsBackupRunnable implementation in GC would be valuable.

ryanfkeepers (commented):
Changes are looking good.  Implementation of the precheck should be interface focused, rather than duplicating existing gc code at the operations layer.

neha-Gupta1:
In case of `service unavailable` added an Info level log message saying- `Service not enabled for creating backup for <username>`

src/internal/operations/backup.go
```diff
@@ -140,18 +141,19 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
 
-	//-----
-	// Precheck
-	//-----
-	err = Precheck(ctx, op.account, op.Selectors.PathService(), op.Selectors.DiscreteOwner)
-	if err != nil {
+	// IsBackupRunnable checks if the user has services enabled to run a backup.
+	// it also checks for conditions like mailbox full.
+	runnable, op.canMakeDeltaQueries, err = op.bp.IsBackupRunnable(ctx, op.Selectors.PathService(), op.ResourceOwner.ID())
+	if err != nil && !runnable {
```
ryanfkeepers, line 149: when `err != nil && runnable` this conditional will get skipped, which doesn't feel like the correct behavior.  I'd suggest splitting the error and runnable checks into two separate conditions.

src/internal/operations/backup.go
```diff
@@ -140,18 +141,19 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
 		logger.CtxErr(ctx, err).Error("running backup")
 		op.Errors.Fail(clues.Wrap(err, "running backup"))
 	}
 
+	ctx, flushMetrics := events.NewMetrics(ctx, logger.Writer{Ctx: ctx})
```
ryanfkeepers, line 154: This should be established at the top of the func.  Is there a particular reason it was moved below the runnable check?

src/internal/operations/backup.go
```diff
@@ -56,6 +54,9 @@ type BackupOperation struct {
 
 	// when true, this allows for incremental backups instead of full data pulls
 	incremental bool
+	// when false it indicates we can't run delta queries on the mailbox because
+	// mailbox is full which is indicated by QuotaExceeded.
+	canMakeDeltaQueries bool
```
ryanfkeepers, line 59: BackupOperation isn't the appropriate owner for this value.  Delta queries are a graphAPI construct.  If we introduce a different backup producer, then this flag doesn't translate.  If you're looking to cache the results of `IsBackupRunnable`, then they should get cached within the backup producer (ie: GC), not in the backupOperation.

src/internal/operations/backup.go
```diff
@@ -306,6 +308,7 @@ func (op *BackupOperation) do(
 		mdColls,
 		lastBackupVersion,
 		op.Options,
+		op.canMakeDeltaQueries,
```
ryanfkeepers, line 311: similar to the other comment about ownership, passing a GC-produced flag back into GC from the operations layer doesn't make much sense and upsets dependency principles.  Similarly, it muddies clarity between the query result and the forced no-delta option.

src/internal/connector/data_collections.go
```diff
@@ -61,20 +61,6 @@ func (gc *GraphConnector) ProduceBackupCollections(
 		return nil, nil, clues.Stack(err).WithClues(ctx)
 	}
 
-	serviceEnabled, canMakeDeltaQueries, err := checkServiceEnabled(
```
ryanfkeepers, line 64: I think it's still sensible to make this validation within the backup producer.  Or, at least, suppose the result of the call from IsBackupRunnable gets cached.  In the (unlikely, but possible) event that ProduceBackupCollections cannot find the cached results, it should perform the lookup.

This is primarily to ensure behavior is correct across interface boundaries.  We _assume_ that IsBackupRunnable is called before ProduceBackupCollections.  But that's not a precondition; the code does not require it to have been called.  As a result we can either re-run the check as a fallback, or we could error out if the isRunnable cache is unpopulated, and make it a precondition.

src/internal/operations/backup.go
neha-Gupta1: It was done with intention that we initialise the metrics only if we will be going forward with the backup. This saves initialisation time.

src/internal/operations/backup.go
neha-Gupta1: my bad. Placed `&&` instead of `||` .

src/internal/operations/backup.go
ryanfkeepers: Initializing metrics is lightweight.  we don't need to avoid all initializations.  our core concern should be avoiding unnecessary kopia manifest/snapshot loads.  Since IsBackupRunnable does make graph api calls, we absolutely want metrics to be tracked for those calls.

src/cli/backup/backup.go
```diff
@@ -225,6 +226,12 @@ func runBackups(
 
 		err = bo.Run(ictx)
 		if err != nil {
+			if errors.Is(err, graph.ErrServiceNotEnabled) {
+				logger.Ctx(ctx).Infof("Service not enabled for creating backup for %s", bo.ResourceOwner.Name())
```
ryanfkeepers, line 230: Per logging standards, the message should avoid capitals and formatting.
```suggestion
				logger.Ctx(ctx).Infow("service not enabled", "resource_owner_name", bo.ResourceOwner.Name())
```

src/internal/connector/mock/connector.go
```diff
@@ -43,6 +44,17 @@ func (gc GraphConnector) ProduceBackupCollections(
...
 }
 
+func (gc GraphConnector) IsBackupRunnable(
+	_ context.Context,
+	_ path.ServiceType,
+	_ string,
+) (
+	bool,
+	error,
+) {
```
ryanfkeepers, line 54:
```suggestion
) (bool, error) {
```

src/internal/tester/resource_owners.go
```diff
@@ -197,3 +197,15 @@ func GetM365SiteID(ctx context.Context) string {
...
 }
+
+// UnlicensedM365UserID returns an userID string representing the m365UserID
+// described by either the env var CORSO_M365_TEST_UNLICENSED_USER, the
+// corso_test.toml config file or the default value (in that order of priority).
+// The default is a last-attempt fallback that will only work on alcion's
+// testing org.
+func UnlicensedM365UserID(t *testing.T) string {
+	cfg, err := readTestConfig()
+	require.NoError(t, err, "retrieving unlicensed m365 user id from test configuration", clues.ToCore(err))
```
ryanfkeepers, line 208: You have to add a formatting marker if you add both a message and a errCore.
```suggestion
	require.NoError(t, err, "retrieving unlicensed m365 user id from test configuration: %+v", clues.ToCore(err))
```

src/internal/operations/backup.go
```diff
@@ -138,6 +138,16 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
 
+	var runnable bool
+
+	// IsBackupRunnable checks if the user has services enabled to run a backup.
+	// it also checks for conditions like mailbox full.
+	runnable, err = op.bp.IsBackupRunnable(ctx, op.Selectors.PathService(), op.ResourceOwner.ID())
+	if err != nil || !runnable {
+		logger.CtxErr(ctx, err).Error("running backup")
+		op.Errors.Fail(clues.Wrap(err, "running backup"))
+	}
```
ryanfkeepers, line 149:
```suggestion
		logger.CtxErr(ctx, err).Error("verifying backup is runnable")
		op.Errors.Fail(clues.Wrap(err, "verifying backup is runnable"))
		
		return
	}
```

src/internal/operations/backup.go
```diff
@@ -138,6 +138,16 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		observe.Complete()
 	}()
 
+	var runnable bool
+
+	// IsBackupRunnable checks if the user has services enabled to run a backup.
+	// it also checks for conditions like mailbox full.
+	runnable, err = op.bp.IsBackupRunnable(ctx, op.Selectors.PathService(), op.ResourceOwner.ID())
+	if err != nil || !runnable {
```
ryanfkeepers, line 146: On a response of `false, nil` this is going to log and fail a nil error, which doesn't make much sense.

The two cases should be split into separate conditionals.

ryanfkeepers (changes requested):
error conditional needs fixing, metrics initialization needs to occur before calling IsRunnable.

src/internal/operations/backup.go
neha-Gupta1: @vkamra and I discussed this the other day that it would be better to move the code up. 
We can go with whatever the best fit is here.

src/internal/tester/resource_owners.go
neha-Gupta1: If that is the case we might have to update it across the repo. Fixing it here and will raise another PR for all the other places.

src/internal/operations/backup.go
neha-Gupta1: As of now I have moved the metrics initialisation above to capture  `IsBackupRunnable` graph api calls.

src/internal/operations/backup.go
```diff
@@ -138,18 +138,28 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		logger.CtxErr(ctx, err).Error("verifying backup is runnable")
+		op.Errors.Fail(clues.Wrap(err, "verifying backup is runnable"))
+
+		return
 	}
 
-	ctx, flushMetrics := events.NewMetrics(ctx, logger.Writer{Ctx: ctx})
-	defer flushMetrics()
+	if !runnable {
+		err = clues.New("service unavailable")
```
ryanfkeepers, line 157: please use `graph.ErrServiceNotEnabled` here.

src/internal/operations/backup.go
```diff
@@ -138,18 +138,28 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		logger.CtxErr(ctx, err).Error("verifying backup is runnable")
+		op.Errors.Fail(clues.Wrap(err, "verifying backup is runnable"))
+
+		return
```
ryanfkeepers, line 161: Different log and error messages between conditions help clarify the different cases we hit.  Also, `CtxErr` is for extracting structured error data out of an error that was produced by a child function.  There's no need to call when returning an ad-hoc error.  So I don't think we need the log here at all.
```suggestion
	if !runnable {
		op.Errors.Fail(clues.Wrap(graph.ErrServiceNotEnabled, "checking service enabled"))

		return
```

ryanfkeepers (approved):
looks good: just a couple nits remaining.

PR #3428, opened 2023-05-16
refactor api options

ryanfkeepers:
now that exchange api has been folded in with the rest of the m365 api, it doesn't make sense to maintain an options file with only exchange functionality.  Since all calls in the file were used 1:1 with some api func, those options have been moved into their respective api funcs.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :green_heart: E2E


meain (commented):
The code changes look good, but having `buildOptions` do [supported field check](https://github.com/alcionai/corso/blob/ff6334131aeb8aa1b01ba98e678b1f8b31aeb6b5/src/pkg/services/m365/api/options.go#L190) was kinda useful.

ryanfkeepers:
> The code changes look good, but having buildOptions do supported field check was kinda useful.

I agree with this perspective in concept, but not in implementation.  In reality, it's a fair amount of code that was either not getting used, or only getting used in one-off scenarios.  A comprehensive compliance with that implementation would turn out enormous (note, for instance, the current space does not include drive or library compliance) and this size would have continued to grow with each new service.  All that, and we're still open to very simple edge cases (ex: if graph api released a breaking change on field keys, the buildOptions check here would still succeed, and so would the graph call).

I think we can get similar resilience with less code and easier maintenance.  See this PR as an example of a starting point: https://github.com/alcionai/corso/pull/3441/files#diff-85d660db33a1c43256849cb4ceb617955eb639032f337949678c4963373a699aR59.

meain (approved):

PR #3430, opened 2023-05-16
general api package cleanup

ryanfkeepers:
prior to centralizing api pacakges, the naming conventions and location/ownership of helper functions were scattered across multiple places.  This change focuses on moving code around so that sets of code are clearly owned, easily identified, and deduplicated throughout the package.

Some minor logic changes within, primarily around the funcs, which were moved to common packages.  The rest of the changes are code movement.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/config/account.go
```diff
@@ -64,7 +64,7 @@ func configureAccount(
 
 	m365Cfg = account.M365Config{
 		M365: m365,
-		AzureTenantID: common.First(
+		AzureTenantID: str.First(
```
meain, line 67: Should we give it a more descriptive name like `FirstNonEmpty`?

src/internal/connector/graph/cache_container.go
```diff
@@ -169,20 +169,21 @@ func CreateCalendarDisplayable(entry any, parentID string) *CalendarDisplayable
...
 	}
 
-	if _, ok := ptr.ValOK(c.GetDisplayName()); !ok {
-		return clues.New("container missing display name").With("container_id", id)
+	id := ptr.Val(c.GetId())
+	if len(id) == 0 {
+		return clues.New("container missing ID")
 	}
 
-	if _, ok := ptr.ValOK(c.GetParentFolderId()); !ok {
```
meain, line 184: Was removal of `GetParentForlderId` intentional?

meain (approved):
Few questions, but looks good. Approving to unblock.

src/internal/common/str/str.go
```diff
@@ -0,0 +1,58 @@
...
+func ParseBool(v string) bool {
+	s, err := strconv.ParseBool(v)
+	if err != nil {
+		return false
+	}
+
+	return s
+}
+
+func FromMapToAny(k string, m map[string]any) (string, error) {
```
ashmrtn, line 23: maybe this should have a different name that denotes that it returns a string? There's now two functions named `FromMapToAny`, one in this package and one in `tform`. They're not functions that adhere to an interface (different return types) so it can be a little confusing. It can also confuse GitHub's search and simpler search tools like grep

src/internal/connector/graph/cache_container.go
ashmrtn: we should definitely have a function that checks for ID, display name, and parent ID (it can call this if we want)

src/pkg/services/m365/api/drive_pager.go
```diff
@@ -0,0 +1,317 @@
...
+	pageCount := maxPageSize
+
+	headers := abstractions.NewRequestHeaders()
+	preferHeaderItems := []string{
+		"deltashowremovedasdeleted",
+		"deltatraversepermissiongaps",
+		"deltashowsharingchanges",
+		"hierarchicalsharing",
+	}
+	headers.Add("Prefer", strings.Join(preferHeaderItems, ","))
```
ashmrtn: we should make a note/PR to combine this with the code in Exchange that makes the `Prefer` header kv-pair

src/internal/connector/graph/cache_container.go
ryanfkeepers: Ah, weird diff.  `CheckRequiredValues` wasn't getting called anywhere, so we've never actually checked for the parentID, making this a non-change.  It just looks like I removed it because I moved `CheckIDAndName`, which _is_ getting called, into this file,

src/internal/common/str/str.go
ryanfkeepers: can do

src/pkg/services/m365/api/drive_pager.go
ryanfkeepers: will combine in a follow-up

src/pkg/services/m365/api/query_params.go
```diff
@@ -7,14 +7,12 @@ import (
...
 
-const deltaMaxPageSize = 200
-
 // buildPreferHeaders returns the headers we add to item delta page requests.
 func buildPreferHeaders(pageSize, immutableID bool) *abstractions.RequestHeaders {
 	var allHeaders []string
 
 	if pageSize {
-		allHeaders = append(allHeaders, fmt.Sprintf("odata.maxpagesize=%d", deltaMaxPageSize))
+		allHeaders = append(allHeaders, fmt.Sprintf("odata.maxpagesize=%d", maxPageSize))
```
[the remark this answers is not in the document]
ryanfkeepers: Seems like we can ask for more and it will ceiling?  Since this should be safe, I'll add a refinement in a follow-up PR.

Issue #3407, opened 2023-05-12
Global limiter middleware retry-after handling


ryanfkeepers:
## Issue

Middleware throttling avoidance is handled in multiple places: concurrency pools limit the number of ongoing calls, the rate limiter keeps us within per-minute boundaries, and the graph client and retry middlewares handle 429 responses.  

429 responses are tracked separately from global rate limiting.  The rate limiter assumes that if it does its job, then we won't hit any 429's.  The retry handler assumes that if it does its job (ie: abiding the retry-after on a 429), then it won't worsen throttling issues.

However, microsoft documentation and blogging does suggest that, in case of server issues, rate limiting can be throttled beyond standard limits.  Likewise, the rate limiter can only track local calls, and is unable to account for external consumption of resources.  In either of these cases, we may see 429 responses when a single instance of corso otherwise acts as a good citizen.  In those cases, even though a prior call may have responded with 429, the rate limiter will continue to allow new calls to graph within the retry-after boundary.

## Proposal

The rate limiter middleware should be augmented to identify retry-after headers and block all calls until that period has passe.  Assume the following scenario.

* **time=0**: corso makes 5 graph api calls.  4 succeed, one returns 429 with `retry-after=10`
* **time=2**: the retry handler catches the 429 and sleeps for 10 seconds
* **time=4**:  5 new graph api calls are issued.

In the current design, the 5 new calls will all get sent, and all return 429.  To prevent those excess calls, we can have the rate limiter middleware catch the `retry-after=10` header, and block all calls until that time has passed.  This results in:

* **time=1**: rate limiter middleware sees `retry-after=10`, blocks all calls for 10 seconds. (occurs before token consumption)
* **time=11**: the retry-after block lifts, allowing all calls to proceed (plus token limiting)

This way all calls abide the retry-after of any single response, allowing corso to respond to all cases of throttling even when otherwise within its local rate limits.

PR #3432, opened 2023-05-16
add global throttling fence to graph middleware

ryanfkeepers:
Adds a graph middleware that blocks all calls based on the most recent retry-after header in a 429 response.  This prevents us from failing on 429 requests due to external pressures even while the corso client believes it is within quota on rate limiting and concurrency.

---

- [x] :no_entry: No

#### Type of change

- [x] :robot: Reliability/Supportability/Tests

#### Issue(s)

* #3407

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (approved):
Love the implementation! Just 2 questions:

1. Should we add an upper limit check for `retry-after` header?
2. Given that we are spawning a new timer goroutine in every `RaiseFence` call, there can be a potential explosion of goroutines for non-exchange traffic. I think adding a default concurrency limiter for onedrive/sharepoint should stem this problem, if we see one. 

Not a problem for exchange traffic. # of timer goroutines will be capped at 4 since concurrency limiter only allows 4 requests in flight.

ryanfkeepers:
> Should we add an upper limit check for retry-after header?

I think we can rely on external services to be good citizens here.  Worst case scenario, the context deadline will manage the maximum upper limit for any call blocked at the fence.  Technically we do have contexts without a deadline, so maybe we would want some kind of upper limit?  But I'm hesitant to put down an arbitrary number just to have one for such a niche condition.

> Given that we are spawning a new timer goroutine in every RaiseFence call, there can be a potential explosion of goroutines for non-exchange traffic

That sounds like a small scale performance concern to me.  Channels and timers are lightweight.  We should be able to spawn many thousands of them (and recall that onedrive limits to 1200 calls per minute) without notable impact to the system.  And since we're using an on-close signalling pattern, we don't need to worry about deadlocking or other concurrency hitches.

pandeyabs:
> I think we can rely on external services to be good citizens here.
My worry was that a malformed/malicious `retry-after` header can cause us to indefinitely wait (denial of service). 
But I get your point about adding arbitrary upper limits. I looked into this a bit more. Looks like graph doesn't have an upper limit defined for `retry-after`. Some folks have reported as high as [300+ seconds. ](https://learn.microsoft.com/en-us/answers/questions/648796/retry-after-header-upper-limit-or-range ) 
So we can table this for future. 
I do wish we could capture graph API response metrics over time. That would help us build our own SLOs at some point. 

> Channels and timers are lightweight. We should be able to spawn many thousands of them without notable impact to the system resources.
Sounds good.

Issue #3431, opened 2023-05-16
Add post-test cleanup for the sanity tests


vkamra:
Sanity tests do not run in parallel and adding a post-test cleanup step will reduce the flakiness related to data accumulation in the test accounts.

PR #3433, opened 2023-05-16
purge sanity tests after run

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3431

#### Test Plan

- [x] :green_heart: E2E


.github/workflows/sanity-test.yaml
```diff
@@ -266,7 +265,44 @@ jobs:
 
 ##########################################################################################################################################
 
-# Logging & Notify
+# Post-Run cleanup
+
+      # unlike CI tests, sanity tests are not expected to run concurrently.
+      # we can purge all of the test-generated folders immediately after the run concludes.
+      - name: Set purge boundary
+        if: always()
```
ashmrtn, line 273: Sanity tests may not run concurrently but what's the behavior going to be if runs are cancelled due to more recent pushes? The concurrency control for the workflow is set to cancel the in-flight sanity test and start a newer one

ryanfkeepers: Good question.  I assume that we'll skip out before we run the cleanup.  Thinking about it, the most reliable behavior in that case would be to run the cleanup _prior to_ running the tests.  I'll go make that change.

vkamra (approved):

PR #3439, opened 2023-05-17
Sanity tests improvements

meain:
- Fix how inputs are passed for sanity tests
- Fix timestamp generated
- use go build instead of make when building corso

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/sanity-test.yaml
+++ b/.github/workflows/sanity-test.yaml
@@ -84,12 +84,12 @@ jobs:
       - name: Purge CI-Produced Folders for Users
         uses: ./.github/actions/purge-m365-data
         with:
-          user: ${TEST_USER}
-          folder-prefix: ${RESTORE_DEST_PFX}
+          user: ${{ env.TEST_USER }}
+          folder-prefix: ${{ env.RESTORE_DEST_PFX }}
           older-than: ${{ env.NOW }}
-          azure-client-id: ${AZURE_CLIENT_ID}
-          azure-client-secret: ${AZURE_CLIENT_SECRET}
-          azure-tenant-id: ${AZURE_TENANT_ID}
+          azure-client-id: ${{ env.AZURE_CLIENT_ID }}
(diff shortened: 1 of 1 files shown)


neha-Gupta1 (approved):
Thank for the fix

.github/workflows/sanity-test.yaml
```diff
@@ -79,31 +79,31 @@ jobs:
...
         run: |
-          echo "NOW=$(date +"%Y-%m-%d_%H-%M-%S")" >> $GITHUB_ENV
+          echo "NOW=$(date +"%Y-%m-%dT%H:%M:%SZ")" >> $GITHUB_ENV
 
       - name: Purge CI-Produced Folders for Users
         uses: ./.github/actions/purge-m365-data
         with:
-          user: ${TEST_USER}
-          folder-prefix: ${RESTORE_DEST_PFX}
+          user: ${{ env.TEST_USER }}
```
ryanfkeepers, line 87: We should decide on one of the following patterns and stick with it:

Either this
```
FOO_SEC: ${{ secrets.FOO }}
# later
bar: ${FOO_SEC}
```

Or this
```
bar: ${{ secrets.FOO }}
```

The halfway point of "put it into the env but don't use the shorthand env reference" is indirection for its own sake.

meain: I don't think using `${FOO_SEC}` outside of shell scripts works. That is why I had to switch to `${{env.FOO_SEC}}`.

meain: What about `RESTORE_DST_PRFX`, i.e. things that we don't want in secrets?

PR #3443, opened 2023-05-17
move sharepoint beta service

ryanfkeepers:
relocates the sharepoint beta service files into the correct subfolder: out of m365/api and into connector/sharepoint/api.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/sites.go
```diff
@@ -35,6 +35,20 @@ type Sites struct {
 // methods
 // ---------------------------------------------------------------------------
 
+// GetSite returns a minimal Site with the SiteID and the WebURL
+// TODO: delete in favor of sites.GetByID()
```
ashmrtn, line 39: can this TODO be removed? Seems like we're calling `sites.GetByID`?

ashmrtn (approved):

src/pkg/services/m365/api/sites.go
ryanfkeepers: Ah, there's corso func named `GetByID` for sites, which effectively duplicates this functionality.  I plan to dedupe those while building out the drive client.  While a bit confusing, the todo should stay for now.

PR #3447, opened 2023-05-17
Adding two new apis to corso for validating a user has exchange or one drive is enabled

juleslasarte:
Adds two new apis that expose calls to check whether a user has one drive or exchange enabled

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/m365.go
```diff
@@ -47,12 +51,12 @@ type UserNoInfo struct {
 func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
 	errs := fault.New(true)
 
-	users, err := Users(ctx, acct, errs)
+	usersResult, err := Users(ctx, acct, errs)
```
juleslasarte, line 54: just fixing this to make the linter happy

ryanfkeepers, line 54: suggestion for a more idiomatic golang approach: since this is a small codespace, you can use a shorthand var name like `us` to mean `userSlice`
```suggestion
	us, err := Users(ctx, acct, errs)
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+	usersResult, err := usersNoInfo(ctx, acct, errs)
 	if err != nil {
 		return nil, err
 	}
 
-	return users, errs.Failure()
+	return usersResult, errs.Failure()
+}
+
+// UserHasMailbox returns true if the user has any mailbox enabled
```
ryanfkeepers, line 76:
```suggestion
// UserHasMailbox returns true if the user has an exchange mailbox enabled
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+// false otherwise, and a nil pointer and an error in case of error
+func UserHasMailbox(ctx context.Context, acct account.Account, userID string) (*bool, error) {
```
ryanfkeepers, line 78: avoid pointers on primitive values wherever possible.  Idiomatic golang is to prefer the zero-value of a type whenever possible.  If err is non-nil, the caller should assume the bool to be either false or unusable.
```suggestion
func UserHasMailbox(ctx context.Context, acct account.Account, userID string) (bool, error) {
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+	uapi, err := makeUserAPI(acct)
+	result := false
+
+	if err != nil {
+		return nil, clues.Wrap(err, "getting mailbox").WithClues(ctx)
+	}
```
ryanfkeepers, line 84: styling: please prefer colocating errors with the function call.  Also, default valued variable declaration should use the format `var foo type`
```suggestion
	var result bool

        uapi, err := makeUserAPI(acct)
	if err != nil {
		return nil, clues.Wrap(err, "getting mailbox").WithClues(ctx)
	}
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+
+	options := users.ItemMailFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &requestParameters,
+	}
+
+	_, err = uapi.GetMailFolders(ctx, userID, options)
+	if err != nil {
+		if graph.IsErrUserNotFound(err) {
+			logger.CtxErr(ctx, err).Error("user not found")
```
ryanfkeepers, line 98: this and the other logs in this func (err on line 104, info on 108) can be removed.

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+		}
+
+		logger.Ctx(ctx).Info("resource owner does not have a mailbox enabled")
+
+		return &result, nil
+	}
+
+	result = true
+
+	return &result, nil
```
ryanfkeepers, line 115: Once the return value removes the pointer, this can flatten to a direct value return.
```suggestion
	return true, nil
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+
+	options := users.ItemMailFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &requestParameters,
+	}
+
+	_, err = uapi.GetMailFolders(ctx, userID, options)
+	if err != nil {
+		if graph.IsErrUserNotFound(err) {
+			logger.CtxErr(ctx, err).Error("user not found")
+			return nil, graph.Stack(ctx, clues.Stack(graph.ErrResourceOwnerNotFound, err))
```
ryanfkeepers, line 99: This error response can be reduced.  The called func already handles `graph.Stack/Wrap`.
```suggestion
			return nil, clues.Stack(graph.ErrResourceOwnerNotFound, err)
```

src/pkg/services/m365/m365.go
```diff
@@ -61,12 +65,87 @@ func UsersCompat(ctx context.Context, acct account.Account) ([]*User, error) {
...
+	}
+
+	result = true
+
+	return &result, nil
+}
+
+// UserHasDrives returns true if the user has any drives
+// false otherwise, and a nil pointer and an error in case of error
+func UserHasDrives(ctx context.Context, acct account.Account, userID string) (*bool, error) {
```
ryanfkeepers, line 120: All the above suggestions apply to this func as well.

src/pkg/services/m365/m365_test.go
```diff
@@ -102,6 +102,38 @@ func (suite *M365IntegrationSuite) TestGetUserInfo() {
...
+	var (
+		t    = suite.T()
+		acct = tester.NewM365Account(t)
+		uid  = tester.M365UserID(t)
+	)
+
+	enabled, err := m365.UserHasMailbox(ctx, acct, uid)
+	require.NoError(t, err, clues.ToCore(err))
+	require.NotNil(t, enabled)
+	require.True(t, *enabled)
```
ryanfkeepers, line 118: Prefer assert unless the following checks will be unable to run if the current check fails.
```suggestion
	assert.NotNil(t, enabled)
	assert..True(t, *enabled)
```

ryanfkeepers (commented):
Handful of styling nits, but generally looks good.  Give it a quick touch-up and we can move it through.

ryanfkeepers (approved):
looks good!

PR #3449, opened 2023-05-18
get repoID from config for events

neha-Gupta1:
- Get repoID from config for events for uniformity across repo
- Set a default value if repoID is nil/blank

- [x] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3388

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/repository/repository.go
```diff
@@ -184,6 +184,7 @@ func Connect(
 	ctx context.Context,
 	acct account.Account,
 	s storage.Storage,
+	repo string,
```
vkamra, line 187: nit: This should be named `repoID`

src/pkg/repository/repository.go
```diff
@@ -229,23 +230,21 @@ func Connect(
...
 	}
 
-	rm := &repositoryModel{}
-
 	// Do not query repo ID if metrics are disabled
 	if !opts.DisableMetrics {
-		rm, err = getRepoModel(ctx, ms)
-		if err != nil {
-			return nil, clues.New("retrieving repo info")
+		repoID := "not_found"
```
vkamra, line 235: We shouldn't initialize to `not_found` here. If there aren't any code paths where this is retrieved in this function block - we don't need that

vkamra: I would look into where we initialize `cfg` and initialize to a constant before we retrieve it from the config file

ryanfkeepers (approved):
some nits, approved to unblock

PR #3453, opened 2023-05-18
move ad-hoc attachment api calls to m365/api

ryanfkeepers:
finishes moving ad-hoc exchange api client calls from connector/exchange into m365/api.  Final cleanup for polish to follow.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/exchange/restore_test.go
```diff
@@ -138,6 +138,7 @@ func (suite *ExchangeRestoreIntgSuite) TestRestoreEvent() {
 				ctx,
 				test.bytes,
 				suite.ac.Events(),
+				suite.ac.Events(),
```
ashmrtn, line 141: maybe we should make another interface that is something like `itemAndAttachmentPoster` and just pass in a single instance of the api layer. I'm not really a fan of passing in/having multiple references to the same underlying object. It seems like a good way run into unforeseen problems in the general case

src/pkg/services/m365/api/events.go
```diff
@@ -367,6 +367,59 @@ func (c Events) DeleteItem(
...
+		Users().
+		ByUserId(userID).
+		Calendars().
+		ByCalendarId(containerID).
+		Events().
+		ByEventId(itemID).
+		Attachments().
+		Post(ctx, body, nil)
+	if err != nil {
+		return graph.Wrap(ctx, err, "uploading small mail attachment")
```
ashmrtn, line 390:
```suggestion
		return graph.Wrap(ctx, err, "uploading small event attachment")
```

src/pkg/services/m365/api/events.go
```diff
@@ -367,6 +367,59 @@ func (c Events) DeleteItem(
...
+		ByUserId(userID).
+		Calendars().
+		ByCalendarId(containerID).
+		Events().
+		ByEventId(itemID).
+		Attachments().
+		CreateUploadSession().
+		Post(ctx, session, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "uploading large mail attachment")
```
ashmrtn, line 417:
```suggestion
		return nil, graph.Wrap(ctx, err, "uploading large event attachment")
```

ashmrtn (approved):

src/internal/connector/exchange/restore_test.go
ryanfkeepers: this will change significantly in an upcoming PR, and will reduce the duplicate-value population.

PR #3459, opened 2023-05-19
Add download URL cache for onedrive items

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/url_cache.go
+++ b/src/internal/connector/onedrive/url_cache.go
@@ -5,23 +5,35 @@ import (
 	"sync"
 	"time"
 
-	"github.com/alcionai/corso/src/internal/connector/graph"
-	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/connector/graph"
+	gapi "github.com/alcionai/corso/src/internal/connector/graph/api"
+	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 type itemProperties struct {
-	downloadUrl     string
-	lastRefreshTime time.Time
-	isDeleted       bool
+	downloadURL string
+	isDeleted   bool
+	// temporary
+	driveItem *models.DriveItem
 }
 
+// urlCache caches download URLs for drive items
 type urlCache struct {
 	// Item ID -> download URL map
-	urlMap          map[string]itemProperties
-	cacheLock       sync.RWMutex
+	urlMap map[string]itemProperties
+	// time of last cache  refresh
+	lastRefreshTime time.Time
+	// RW lock for the URL map and lastRefreshTime
+	rwLock sync.RWMutex
+	// semaphore for limiting the number of concurrent cache refreshes to 1
+	refreshSemaphore chan struct{}
+
 	driveID         string
-	driveName       string
 	deltaQueryCount int
 	Errors          *fault.Bus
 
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	uc.deltaQueryCount++
+
+	return nil
+}
+
+// collectDriveItems will enumerate all items in the specified drive and hand
+// them to the provided `collector` method
+// TODO: This is a clone of collectItems call. Refactor collectItems to remove
```
vkamra, line 229: What needs to be changed in `collectItems`?

pandeyabs: oops! just saw this. I actually refactored the cache code to rely on existing `collectItems` as we had discussed. 
this felt like a better approach than refactoring `collectItems`.

ashmrtn (commented):
Generally looks good on the logic side. I'm wondering about the benefit/effort tradeoff of testing against graph directly vs. using mocks for response data

Also may be worth looking into passing in some map of initialized values when making a new cache. Ideally we wouldn't need to run the full delta query again after enumerating items to get the set of Collections we're backing up

pandeyabs:
@ashmrtn 
> I'm wondering about the benefit/effort tradeoff of testing against graph directly vs. using mocks for response data 
I wanted to have a lightweight test which tests directly against graph. This is to catch content url/download url kind of issues we noticed a while back). 

I'm going to use mock responses for all upcoming tests.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	// Hold cache lock in write mode for the entire duration of the refresh.
+	// This is to prevent other threads from reading the cache while it is
+	// being updated page by page
+	uc.cacheLock.Lock()
+	defer uc.cacheLock.Unlock()
+
+	// Issue a delta query to graph
+	logger.Ctx(ctx).Debugw("refreshing url cache")
```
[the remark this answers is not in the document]
pandeyabs: > should add the drive ID to the context so that it's added to the log messages. 

Thanks! I actually added it on line 113, and i'm passing the `ctx` to callees. Let me know if this is not a good pattern.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+type itemProperties struct {
+	downloadURL string
+	isDeleted   bool
+}
+
+// urlCache caches download URLs for drive items
+type urlCache struct {
+	driveID string
+	// urlMap stores Item ID -> item property map
+	urlMap          map[string]itemProperties
```
ryanfkeepers, line 26: Building the map k:v description into the var name does a better job describing how to use it than a comment.  Also, naming a map `fooMap` is a bit redundant.
```suggestion
	idToProps          map[string]itemProperties
```

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	lastRefreshTime time.Time
+	refreshInterval time.Duration
+	// cacheLock protects urlMap and lastRefreshTime
+	cacheLock sync.RWMutex
+	// refreshMutex serializes cache refresh attempts
+	refreshMutex    sync.Mutex
```
ryanfkeepers, line 32: `mu` for standard golang naming expectations.
```suggestion
	cacheMu sync.RWMutex
	// refreshMutex serializes cache refresh attempts
	refreshMu    sync.Mutex
```

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	deltaQueryCount int
+
+	svc           graph.Servicer
+	itemPagerFunc func(
```
ryanfkeepers, line 36: Since this is referenced multiple times, feel free to declare the func signature as a typecast.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+		servicer graph.Servicer,
+		driveID, link string,
+	) itemPager,
+) error {
+	if len(driveID) == 0 {
+		return clues.New("drive id is empty")
+	}
+
+	if refreshInterval < 0 {
```
ryanfkeepers, line 89: is 0 a valid refresh interval?  If so, what is the refresh rate?

Would it make sense for this to be some well defined minimum like, say, 5 minutes?  0 and 1 are relatively indifferent in this case.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Lazy refresh
+	if uc.needsRefresh() {
+		err := uc.refreshCache(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	url, err := uc.readCache(ctx, itemID)
```
ryanfkeepers, line 123:
```suggestion
	props, err := uc.readCache(ctx, itemID)
```

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	// to acquire mutex, return
+	if !uc.needsRefresh() {
+		return nil
+	}
+
+	// Hold cache lock in write mode for the entire duration of the refresh.
+	// This is to prevent other threads from reading the cache while it is
+	// being updated page by page
+	uc.cacheLock.Lock()
+	defer uc.cacheLock.Unlock()
```
ryanfkeepers, line 159: Based on this implementation, I think you could reduce the two mutexes to a single [RWMutex](https://pkg.go.dev/sync#RWMutex).

The lock on 158 could go away.  146 would use the write lock.  134 and 210 would use the read lock (though you might need to have line 151 call the needsRefresh conditional directly, instead of through the locked needsRefresh.  Or, alternatively, move the needsRefresh read lock to 115).

This reduces the implementation to a single mutex, which will help avoid deadlock and minimize maintenance confusion.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Issue a delta query to graph
+	logger.Ctx(ctx).Debug("refreshing url cache")
```
ryanfkeepers, line 162: I think this is significant enough (and, more importantly, infrequent enough) to warrant an info-level.
```suggestion
	logger.Ctx(ctx).Info("refreshing url cache")
```

Also, these logs are effectively duplicates of the `deltaQuery` logs.  Better off just to let that func do the logging.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	_, _, _, err := collectItems(
+		ctx,
+		uc.itemPagerFunc(uc.svc, uc.driveID, ""),
+		uc.driveID,
+		"",
+		uc.updateCache,
+		map[string]string{},
+		"",
+		uc.errors,
+	)
```
ryanfkeepers, line 193:
```suggestion
		uc.errors)
```

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return clues.Wrap(err, "delta query failed").WithClues(ctx)
```
ryanfkeepers, line 195: WithClues not required, the called function should already attach it, since you pass in the ctx.
```suggestion
		return clues.Wrap(err, "delta query failed")
```

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	logger.Ctx(ctx).Debug("finished delta query")
+
+	return nil
+}
+
+// readCache returns the download URL for the specified item
+func (uc *urlCache) readCache(
+	ctx context.Context,
+	itemID string,
+) (*itemProperties, error) {
```
ryanfkeepers, line 209: A pointer to itemProperties feels strange here.  Either the struct warrants being stored as a pointer in the map (thus placing it on the heap as part of map entry), or this func should return a non-pointer value.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+func (uc *urlCache) updateCache(
+	ctx context.Context,
+	_, _ string,
+	items []models.DriveItemable,
+	_ map[string]string,
+	_ map[string]string,
+	_ map[string]struct{},
+	_ map[string]map[string]string,
+	_ bool,
+	errs *fault.Bus,
```
ryanfkeepers, line 234: non-test funcs should only swallow parameters in very rare cases.  All of these ignored inputs produce a smell to me.  The caller won't easily know the difference between required and non-required values, and poor compliance with a currently-swallowed value can become a footgun if a maintainer updates only this func.  Is there something keeping us from removing these unused parameters?

src/internal/connector/onedrive/url_cache.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	return &urlCache{
+			urlMap:          make(map[string]itemProperties),
+			lastRefreshTime: time.Time{},
+			driveID:         driveID,
+			refreshInterval: refreshInterval,
+			svc:             svc,
+			itemPagerFunc:   itemPagerFunc,
+			errors:          fault.New(false),
```
ryanfkeepers, line 70: These errors will get sliced out.  You'll need to pass in a fault.Bus instance, not construct a new one.

src/internal/connector/onedrive/url_cache.go
pandeyabs: > is 0 a valid refresh interval? If so, what is the refresh rate?

Oops. This is a bug. Thanks for pointing out. 

> Would it make sense for this to be some well defined minimum like, say, 5 minutes? 

I want to allow specifying interval in seconds. This would allow us to test for stale cache conditions faster.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Great call out! Those two locks are added by design. I originally had planned to use a single rwMutex lock, but it complicates code quite a bit.
For e.g. consider that `needsRefresh` is lockless, and caller can hold either RLock() or Lock() while calling it. This is needed because we can't acquire RLock() in `needsRefresh` if we are holding Lock().

Now consider line 116 onwards code, copied below with comments. 

```
       // hold RLock()
	if uc.needsRefresh() {
                // RUnlock()
                
		err := uc.refreshCache(ctx) // Assume refreshCache holds Lock() throughout
		if err != nil {
			return nil, err
		}
	} else {
           //RUnlock()
        }
        
	url, err := uc.readCache(ctx, itemID) // Assume readCache holds RLock() throughout.
```

Also, I don't think there is a deadlock possibility here, since we always acquire these 2 locks(refreshMutex, and cacheLock.Lock()) in the same order. Do correct me if I am wrong.

src/internal/connector/onedrive/url_cache.go
pandeyabs: > Also, these logs are effectively duplicates of the deltaQuery logs. Better off just to let that func do the logging.

I would still like to keep the refresh event logs to differentiate between normal delta queries vs delta queries issued by the cache.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Thanks!

src/internal/connector/onedrive/url_cache.go
pandeyabs: Fair point. Context here is that I am reusing `collectItems` func which which takes in a `collector` func which has above signature.

I had a cleaner approach in an [older commit](https://github.com/alcionai/corso/pull/3459/commits/549ae620129c9ec726b0a92cce016b99c0fb5430#diff-307fc0010b8fb64c7aa9cb85a4ac959aaadd612611595075b8a056ec93f8fa87L229)

But that was duplicating `collectItems` call. I looked into refactoring `collectItems` to separate delta query(`itempager`) code from `collector` but I didn't have any fresh ideas there. 

Let me switch back to the older commit, and file a tech debt item for `collectItems` refactoring.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Sounds good. I remember reading our code comments which mentioned that fault.Bus is not used for onedrive yet. Will link it once i find it.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Good point. I'll fix this.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Misread your statement. I'll remove the logs in `deltaQuery`

src/internal/connector/onedrive/url_cache.go
pandeyabs: I put in some more thought for switching to a single rwMutex. I think it's doable without compromising code readability and maintainability. 
I'll add that in a new commit.

ashmrtn: if we're thinking about having more complex locking where some functions may require the mutex to already be held it may be worth adding the [checklocks linter](https://pkg.go.dev/gvisor.dev/gvisor/tools/checklocks#section-readme). It requires annotations in code comments, but it can help make sure we don't accidentally forget to grab a lock when we should have

Unfortunately it's not integrated with golangci-lint, but it seems fairly easy to download/run. If we add this it should be done in a separate PR

pandeyabs: Thanks @ashmrtn! I'll look into this.

src/internal/connector/onedrive/url_cache.go
vkamra: Honestly - I don't think we should duplicate collectItems (let's keep the enumeration logic in one place). Maybe a refactor as a follow up where we move these optional parameters into a struct

pandeyabs: @vkamra sounds good. Let me keep this code as above then. I'll file a tech debt item for `collectItems` refactoring as you mentioned.

src/internal/connector/onedrive/url_cache.go
pandeyabs: Update: I'm keeping two locks for now to finish this PR faster. IMO it's not going to cause deadlocks, or impact cache perf. 

I'll iterate on it in a later PR and move to a single rw lock, along with Ashlie's suggested checklocks lint.

ryanfkeepers: Going to approve: should be good to move ahead with the current implementation and come back to this.

If you would be so kind, can you please create a tech debt ticket to investigate the double-mutex/single-mutex approach?

ryanfkeepers (approved):

PR #3461, opened 2023-05-19
follow-up refinements to m365/api

ryanfkeepers:
some renaming, mostly refactoring the configuration (select and header declaration for each query) by combining the consts.go and query_param.go file into config.go, and making a more composable set of funcs and consts.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/graph/cache_container.go
```diff
@@ -186,5 +186,10 @@ func CheckIDAndName(c Container) error {
 		return clues.New("container missing display name").With("container_id", id)
 	}
 
+	pfid := ptr.Val(c.GetParentFolderId())
```
ashmrtn, line 189: mmm may want to double check this. iirc the folder cache calls some functions like this but due to the difference between calendars and mail/contacts it can only check ID and display name since it doesn't have a concept of parent folders. I think that's why these functions were split to begin with

src/pkg/services/m365/api/config.go
```diff
@@ -0,0 +1,85 @@
...
+
+// header keys
+const (
+	headerKeyConsistencyLevel = "ConsistencyLevel"
+	headerKeyPrefer           = "Prefer"
+)
+
+// header values
+const (
+	idTypeImmutable = `IdType="ImmutableId"`
```
ashmrtn, line 33: should this also be split as a key and a value? Technically it can have different values, we were only interested in immutable IDs though those still have issues that need resolving

src/pkg/services/m365/api/contacts.go
```diff
@@ -81,7 +81,7 @@ func (c Contacts) GetItem(
 	_ *fault.Bus, // no attachments to iterate over, so this goes unused
 ) (serialization.Parsable, *details.ExchangeInfo, error) {
 	options := &users.ItemContactsContactItemRequestBuilderGetRequestConfiguration{
-		Headers: buildPreferHeaders(false, immutableIDs),
+		Headers: newPreferHeaders(idTypeImmutable),
```
ashmrtn, line 84: these lines are not equivalent. The old implementation took a bool (populated as part of the control/options stuff) and decided whether to enable immutable IDs. The new code is always enabling immutable IDs

ashmrtn (changes requested):
Need a change for setting immutable IDs header

src/pkg/services/m365/api/config.go
ryanfkeepers: Going to keep it as is.  Should be easy to update when we want to introduce new values.

ryanfkeepers:
@ashmrtn ready for re-review

ashmrtn (approved):

PR #3464, opened 2023-05-19
More robust folder fetching for container cache

ashmrtn:
Update folder cache population
* try to fetch missing parent folders
* refresh current folder if parent wasn't cached
* only populate paths during populatePaths not during IDToFolder

Manually tested email folder resolver population with
stable reproducer and succeeded

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -31,6 +31,13 @@ type containersEnumerator interface {
 	) error
 }
 
+type containerRefresher interface {
+	refreshContainer(
+		ctx context.Context,
+		dirID string,
+	) (graph.CachedContainer, error)
+}
+
 // ---------------------------------------------------------------------------
 // controller
 // ---------------------------------------------------------------------------
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -68,12 +68,12 @@ func (cr *containerResolver) IDToPath(
 
 	c, ok := cr.cache[folderID]
 	if !ok {
-		return nil, nil, clues.New("folder not cached").WithClues(ctx)
+		return nil, nil, clues.New("container not cached").WithClues(ctx)
 	}
 
 	p := c.Path()
 	if p == nil {
-		return nil, nil, clues.New("folder has no path").WithClues(ctx)
+		return nil, nil, clues.New("cached container has no path").WithClues(ctx)
 	}
 
 	return p, c.Location(), nil
@@ -87,7 +87,7 @@ func (cr *containerResolver) refreshContainer(
 	logger.Ctx(ctx).Debug("refreshing container")
 
 	if cr.refresher == nil {
-		return nil, false, clues.New("nil refresher")
+		return nil, false, clues.New("nil refresher").WithClues(ctx)
 	}
 
 	c, err := cr.refresher.refreshContainer(ctx, id)
@@ -154,7 +154,7 @@ func (cr *containerResolver) idToPath(
 	}
 
 	if !parentCached {
-		logger.Ctx(ctx).Debug("parent folder was refreshed")
+		logger.Ctx(ctx).Debug("parent container was refreshed")
 
 		newContainer, currentShouldDelete, err := cr.refreshContainer(ctx, folderID)
 		if err != nil {
@@ -162,7 +162,7 @@ func (cr *containerResolver) idToPath(
 		}
 
 		if currentShouldDelete {
-			logger.Ctx(ctx).Debug("refreshing folder showed it was deleted")
+			logger.Ctx(ctx).Debug("refreshing container showed it was deleted")
 			delete(cr.cache, folderID)
 
 			return nil, nil, true, true, nil
@@ -185,7 +185,7 @@ func (cr *containerResolver) idToPath(
 	// If the parent wasn't found and refreshing the folder itself showed it
 	// hadn't changed then just delete it.
 	if shouldDelete {
-		logger.Ctx(ctx).Debug("deleting folder since parent was deleted")
+		logger.Ctx(ctx).Debug("deleting container since parent was deleted")
 		delete(cr.cache, folderID)
 
 		return nil, nil, true, true, nil
--- a/src/internal/connector/exchange/container_resolver.go
+++ b/src/internal/connector/exchange/container_resolver.go
@@ -8,6 +8,7 @@ import (
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/connector/graph"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
@@ -47,33 +48,23 @@ type containerRefresher interface {
 // folders if each folder is only a single character.
 const maxIterations = 300
 
-func newContainerResolver() *containerResolver {
+func newContainerResolver(refresher containerRefresher) *containerResolver {
 	return &containerResolver{
-		cache: map[string]graph.CachedContainer{},
+		cache:     map[string]graph.CachedContainer{},
+		refresher: refresher,
 	}
 }
 
 type containerResolver struct {
-	cache map[string]graph.CachedContainer
+	cache     map[string]graph.CachedContainer
+	refresher containerRefresher
 }
 
 func (cr *containerResolver) IDToPath(
 	ctx context.Context,
 	folderID string,
 ) (*path.Builder, *path.Builder, error) {
-	return cr.idToPath(ctx, folderID, 0)
-}
-
-func (cr *containerResolver) idToPath(
-	ctx context.Context,
-	folderID string,
-	depth int,
-) (*path.Builder, *path.Builder, error) {
 	ctx = clues.Add(ctx, "container_id", folderID)
-
-	if depth >= maxIterations {
-		return nil, nil, clues.New("path contains cycle or is too tall").WithClues(ctx)
-	}
 
 	c, ok := cr.cache[folderID]
 	if !ok {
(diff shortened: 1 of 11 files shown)


src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
 func (cr *containerResolver) IDToPath(
 	ctx context.Context,
 	folderID string,
 ) (*path.Builder, *path.Builder, error) {
-	return cr.idToPath(ctx, folderID, 0)
+	ctx = clues.Add(ctx, "container_id", folderID)
+
+	c, ok := cr.cache[folderID]
+	if !ok {
+		return nil, nil, clues.New("folder not cached").WithClues(ctx)
```
ryanfkeepers, line 71: We swap our vocabulary a lot between container, folder, and sometimes calendar.  That makes sense when we know what the item specifically is (ie: a folder vs a calendar).  In the generic case, do we want to call them `containers` or `folders`?
```suggestion
		return nil, nil, clues.New("container not cached").WithClues(ctx)
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+	}
+
+	p := c.Path()
+	if p == nil {
+		return nil, nil, clues.New("folder has no path").WithClues(ctx)
```
ryanfkeepers, line 76: "cached folder" removes ambiguity about whether it's the msoft data or cache data missing a path.
```suggestion
		return nil, nil, clues.New("cached folder has no path").WithClues(ctx)
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+
+func (cr *containerResolver) refreshContainer(
+	ctx context.Context,
+	id string,
+) (graph.CachedContainer, bool, error) {
+	ctx = clues.Add(ctx, "refresh_container_id", id)
+	logger.Ctx(ctx).Debug("refreshing container")
+
+	if cr.refresher == nil {
+		return nil, false, clues.New("nil refresher")
```
ryanfkeepers, line 90:
```suggestion
		return nil, false, clues.New("nil refresher").WithClues(ctx)
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+
+	p := c.Path()
+	if p == nil {
+		return nil, nil, clues.New("folder has no path").WithClues(ctx)
+	}
+
+	return p, c.Location(), nil
+}
+
+func (cr *containerResolver) refreshContainer(
```
ryanfkeepers, line 82: "refresh" could be interpreted to mean many things.  Could this func get a header comment about its expectations/behaviors, please?

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+
+	return c, false, nil
 }
 
 func (cr *containerResolver) idToPath(
 	ctx context.Context,
 	folderID string,
 	depth int,
-) (*path.Builder, *path.Builder, error) {
+) (*path.Builder, *path.Builder, bool, bool, error) {
```
ryanfkeepers, line 109: this return set is getting unwieldy, and the copied types make it extra difficult to grok at face value.  Could this get transitioned to a return struct (such as `resolvedPath{}, error`) instead?

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+			return nil, nil, false, shouldDelete, err
+		}
+
+		if err := cr.addFolder(c); err != nil {
+			return nil, nil, false, false, clues.Wrap(err, "adding new folder").WithClues(ctx)
+		}
+
+		// Retry populating this container's paths.
+		// TODO(ashmrtn): May want to bump the depth here just so we don't get stuck
+		// retrying too much if for some reason things keep moving around?
```
ryanfkeepers, line 134: I doubt we'd get ever get stuck.  slightly delayed, perhaps.  but I think we can swallow that corner.

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+		pth, loc, _, shouldDelete, err := cr.idToPath(ctx, folderID, depth)
+		if err != nil {
+			err = clues.Wrap(err, "retry populating uncached folder")
```
ryanfkeepers, line 137:
```suggestion
			err = clues.Wrap(err, "repopulating uncached folder")
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
 		depth+1)
 	if err != nil {
-		return nil, nil, clues.Wrap(err, "retrieving parent folder")
+		return nil, nil, true, false, clues.Wrap(err, "retrieving parent folder")
+	}
+
+	if !parentCached {
+		logger.Ctx(ctx).Debug("parent folder was refreshed")
+
+		newContainer, currentShouldDelete, err := cr.refreshContainer(ctx, folderID)
```
ryanfkeepers, line 159: not entirely happy with this suggestion. But "currentShouldDelete" is a mouthful, and makes it seem like we're inside a loop here.
```suggestion
		pc, pShouldDelete, err := cr.refreshContainer(ctx, folderID)
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+		if currentShouldDelete {
+			logger.Ctx(ctx).Debug("refreshing folder showed it was deleted")
+			delete(cr.cache, folderID)
+
+			return nil, nil, true, true, nil
+		}
+
+		// TODO(ashmrtn): May want to bump the depth here just so we don't get stuck
+		// retrying too much if for some reason things keep moving around?
+		if ptr.Val(newContainer.GetParentFolderId()) != ptr.Val(c.GetParentFolderId()) ||
```
ryanfkeepers, line 173: might want to explain this conditional with a comment, just for high level objectives.

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
 	c, ok := cr.cache[folderID]
 	if !ok {
-		return nil, nil, clues.New("folder not cached").WithClues(ctx)
+		c, shouldDelete, err := cr.refreshContainer(ctx, folderID)
+		if err != nil {
+			return nil, nil, false, false, clues.Wrap(err, "fetching uncached container")
+		}
+
+		if shouldDelete {
+			logger.Ctx(ctx).Debug("fetching uncached folder showed it was deleted")
```
ryanfkeepers, line 124: this makes it sound like `shouldDelete` == `wasDeleted`.  Is that true?  If so, would that be a better var name?

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
+			if err := cr.addFolder(newContainer); err != nil {
+				return nil, nil, false, false, clues.Wrap(err, "updating cached folder").WithClues(ctx)
+			}
+
+			return cr.idToPath(ctx, folderID, depth)
+		}
+	}
+
+	// If the parent wasn't found and refreshing the folder itself showed it
+	// hadn't changed then just delete it.
```
ryanfkeepers, line 186:
```suggestion
	// If the parent wasn't found and refreshing the current folder produced no
	// diffs, then delete the current folder on the assumption that the parent was
	// deleted, and the current folder will later get deleted via eventual consistency.
	// If we're wrong, then the folder will get picked up again on the next backup.
```

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
 
 	if depth >= maxIterations {
-		return nil, nil, clues.New("path contains cycle or is too tall").WithClues(ctx)
+		return nil, nil, false, false, clues.New("path contains cycle or is too tall").WithClues(ctx)
 	}
 
 	c, ok := cr.cache[folderID]
 	if !ok {
-		return nil, nil, clues.New("folder not cached").WithClues(ctx)
+		c, shouldDelete, err := cr.refreshContainer(ctx, folderID)
```
ryanfkeepers, line 118: This func (idToPath) is a lot to chew through.  Since this conditional block guarantees a return, it could help readability to move the internals to a new func and return a call to that func here.

src/internal/connector/exchange/container_resolver.go
```diff
@@ -40,50 +48,147 @@ type containersEnumerator interface {
...
 	}
 
-	parentPath, parentLoc, err := cr.idToPath(
+	parentPath, parentLoc, parentCached, shouldDelete, err := cr.idToPath(
 		ctx,
 		ptr.Val(c.GetParentFolderId()),
 		depth+1)
 	if err != nil {
-		return nil, nil, clues.Wrap(err, "retrieving parent folder")
+		return nil, nil, true, false, clues.Wrap(err, "retrieving parent folder")
```
ryanfkeepers, line 153: I'm a bit confused by the first bool return val.  on line 148 it's called `parentCached`, implying that the parent was found in the cache (or, maybe, added to the cache?).  But it only gets evaluated to `true` if: an error occurred in idToPath for the parent container, an error occurred refreshing the current container if _its_ parent wasn't cached, or we should delete the current container whether or not its parent was cached.  Or, of course, if we found the parent path, and it was cached, and we shouldn't delete the current container.  _That_ outcome makes sense.  But the prior conditions... not so much.

I'm not sure if the true returns in error/delete cases are unintentional flags that end up benign in the current implementation, or if the value is being intentionally overloaded, contrary to its naming, to achieve a certain outcome.  Either case makes me uneasy, because it feels like something that'll be easy to mix up later on.

ryanfkeepers (approved):
Mostly nits.  A bit concerned about the return value bloat and its coordination across recursive calls.  Going to trust that testing (manual, automated) will uncover any unexpected bugs, and approve to unblock.

src/internal/connector/exchange/container_resolver.go
ashmrtn: so there's a broader question of how consistent we want things to be that this PR doesn't really address. Consider the case where we have `f1/f2/f3` and we've added `f2` and `f3` to the cache. When we go to populate the paths we'll find that `f1` is missing and try to go fetch it. There's a few different outcomes possible given the desired flow we have right now:
* f1 was deleted, f2 was moved
* f1 is there, f2 was moved
* f1 is there, f2 wasn't moved

For all of those cases, fetching only f1 and f2 is enough to fixup things. However, if we fetch `f2` there's now the question of if we should fetch any of it's children and how will we know we need to fetch every child of `f2` (we have no guaranteed order of iteration when populating paths). It's possible the first time we encounter a child of `f2` will be the time we fail for `f1` and thus we'll refresh `f1`, `f2`, and `f3`. However, if we also have `f1/f2/f4` then we won't refresh `f4` because that path is populated in a different iteration and `f1` will be cached for that iteration

`parentCached` is supposed to help determine if the folder we failed on was cached or not. If it wasn't then we fetch the immediate child of that folder but no others. If it was cached then we don't do anything extra

PR #3470, opened 2023-05-22
check for log file location in artifacts

ryanfkeepers:
ensure test artifacts have log files

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/exchange/restore_test.go
+++ b/src/internal/connector/exchange/restore_test.go
@@ -20,23 +20,22 @@ import (
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
 
-type ExchangeRestoreSuite struct {
+type RestoreIntgSuite struct {
 	tester.Suite
 	gs          graph.Servicer
 	credentials account.M365Config
 	ac          api.Client
 }
 
-func TestExchangeRestoreSuite(t *testing.T) {
-	suite.Run(t, &ExchangeRestoreSuite{
+func TestRestoreIntgSuite(t *testing.T) {
+	suite.Run(t, &RestoreIntgSuite{
 		Suite: tester.NewIntegrationSuite(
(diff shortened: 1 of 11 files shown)


ryanfkeepers:
log files are now added to the ci artifacts.  See for example: https://github.com/alcionai/corso/actions/runs/5060181350?pr=3470

src/internal/tester/cli.go
```diff
@@ -32,34 +29,3 @@ func StubRootCmd(args ...string) *cobra.Command {
 
 	return c
 }
-
-func NewContext() (context.Context, func()) {
```
ryanfkeepers, line 36: moved because I keep forgetting what file these are in and always try to look for `/tester/context.go`

src/internal/connector/exchange/restore_test.go
ashmrtn, line 1: this needs to be moved to a different PR if it's not due to being behind main

ashmrtn (approved):
So next question, will we be able to tell which log output belongs to which test after we download an artifact?

src/internal/connector/exchange/restore_test.go
ryanfkeepers: I honestly don't understand this statement.

ryanfkeepers:
> So next question, will we be able to tell which log output belongs to which test after we download an artifact?

@ashmrtn  see: https://github.com/alcionai/corso/pull/3484

src/internal/connector/exchange/restore_test.go
ashmrtn: oh, sorry, if the changes in this file are not due to the PR being behind main (i.e. some diff weirdness in github) then these changes should be moved to another PR because they're not related to getting logging setup at all

PR #3479, opened 2023-05-23
Try to catch concurrent serialization issues

ashmrtn:
Run a bunch of serialization goroutines at the same time to try to catch any buffer reuse errors there may be. Fails on the old version of corso code that didn't make defensive copies.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/item_serialization_test.go
+++ b/src/pkg/services/m365/api/item_serialization_test.go
@@ -1,0 +1,144 @@
+package api_test
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/alcionai/clues"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/connector/support"
+	"github.com/alcionai/corso/src/internal/tester"
+	"github.com/alcionai/corso/src/pkg/services/m365/api"
+)
+
+type ItemSerializationUnitSuite struct {
+	tester.Suite
+}
+
+func TestItemSerializationUnitSuite(t *testing.T) {
+	suite.Run(t, &ItemSerializationUnitSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+func (suite *ItemSerializationUnitSuite) TestConcurrentItemSerialization() {
+	var (
+		user      = "a-user"
+		instances = 100
+	)
+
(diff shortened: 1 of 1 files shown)


src/pkg/services/m365/api/item_serialization_test.go
```diff
@@ -0,0 +1,144 @@
...
+				item.SetGivenName(&name)
+
+				bs, err := api.Contacts{}.Serialize(ctx, item, user, name)
+				require.NoError(t, err, clues.ToCore(err))
+
+				return bs
+			},
+			deserializeAndGetField: func(t *testing.T, bs []byte) string {
+				item, err := support.CreateContactFromBytes(bs)
+				require.NoError(
```
ryanfkeepers, line 105: iirc the `f` is still required for formatting.
```suggestion
				require.NoErrorf(
```

ryanfkeepers (approved):

src/pkg/services/m365/api/item_serialization_test.go
ashmrtn: it actually seems to format it properly without the f suffix. I'm not sure if that's some change in the API or not 🤔 without the format specifier in the string it actually complains that there's extra args when printing

src/pkg/services/m365/api/item_serialization_test.go
```diff
@@ -0,0 +1,144 @@
...
+			t := suite.T()
+			output := make([][]byte, instances)
+			wg := sync.WaitGroup{}
+			wg.Add(instances)
+
+			for i := 0; i < instances; i++ {
+				go func(idx int) {
+					defer wg.Done()
+
+					output[idx] = test.serializer(t, ctx, idx)
```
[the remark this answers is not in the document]
ashmrtn: hmmm actually, let me see if I can repro it without concurrent runs. It may be enough to just try to serialize something a bunch of times since it's a buffer reuse issue linked to the lifetime of the consumption of the returned results not the actual serialization of data

pandeyabs (approved):

src/pkg/services/m365/api/item_serialization_test.go
ryanfkeepers: Oh... that's right.  assert does this ridiculous thing where if it's given `assert.Foo(t, bar, "reason", a, b, c)`, then it'll call `fmt.Printf("reason", a, b, c)` instead of `fmt.Print...`, which would make sense.  So you can pass in a format templated string and vars (`"reason %v %v %v", a, b, c`) without `f` and it'll work.  But if you skip the format templating it'll error.  Which is, yay.

Issue #3304, opened 2023-05-04
[Bug]: Discrepancy in backup output


ntolia:
### What happened?

The obvious disconnect between "0 items to backup" and 2 "Items uploaded" in Corso output is causing cognitive dissonance. Can we fix this?

### Corso Version?

Corso v0.7.0

### Where are you running Corso?

macOS

### Relevant log output

```shell
Discovered 0 items to backup
Backing up data:      0s done
Done

  ID                                    Bytes Uploaded  Items Uploaded  Items Skipped  Errors
  db38d468-a74d-4352-9524-bec8a832d9d6  9.3 kB          2               0              0
```

ntolia:
https://github.com/alcionai/corso/pull/3128#pullrequestreview-1393151333 is related. /cc @ryanfkeepers

ryanfkeepers:
@ntolia Which service was involved in this run?  Exchange, OneDrive, or Sharepoint?  The bug could be related to file item identification in those services.

ashmrtn:
Another thing that could be causing this is the Corso metadata files (previous paths, delta tokens). If the file count is coming from kopia then it would include those in the upload count even though they're not user data that can be restored

ntolia:
> @ntolia Which service was involved in this run? Exchange, OneDrive, or Sharepoint? The bug could be related to file item identification in those services.

I believe it was OD or SP but not 100% sure. This should be a 100% repro though, right?

ntolia:
OneDrive and Exchange shows this behavior. It's not that obvious in Exchange but my guess is we upload 6 objects in my setup for a zero-item backup.

**OneDrive**
```
Discovered 0 items to backup
Backing up data:      0s done
Done

  ID                                    Bytes Uploaded  Items Uploaded  Items Skipped  Errors
  a2273ec7-d8bb-4276-ac64-6f089de7be55  9.3 kB          2               0              0
```

**SharePoint**
```
Discovered 0 items to backup
Backing up data:      0s done
Done

  ID                                    Bytes Uploaded  Items Uploaded  Items Skipped  Errors
  37a2b4a0-bcfd-45de-bee8-fa87d4ea3327  9.3 kB          2               0              0
```

neha-Gupta1:
PR - https://github.com/alcionai/corso/pull/3480

neha-Gupta1:
Completed as part of - https://github.com/alcionai/corso/pull/3480 And the changes impact all the three services. But are most visible in Onedrive

PR #3480, opened 2023-05-23
update files count and total size on CLI for backup

neha-Gupta1:
Currently the backup output show,
- total backup size - size of files + size of metafiles
- no of files - no of files + no of metafile
With this change the output of backup command shows size of only files and count include only file count.

NOTE: all current three services results will be impacted here.

- [x] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3304

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E


src/internal/operations/backup.go
```diff
@@ -676,8 +677,15 @@ func mergeDetails(
 	mans []*kopia.ManifestEntry,
 	dataFromBackup kopia.DetailsMergeInfoer,
 	deets *details.Builder,
+	writeStats *kopia.BackupStats,
 	errs *fault.Bus,
 ) error {
+	// getting the values in writeStats before anything else so that we don't get a return from
+	// conditions like no backup data.
+	noMetaDataFile := deets.Details().DetailsModel.FilterMetaFiles().Entries
+	writeStats.TotalFileCount = len(noMetaDataFile)
```
ashmrtn, line 686: We probably actually want to use the `Items` call instead as that also filters out folders

src/internal/operations/backup.go
```diff
@@ -676,8 +677,15 @@ func mergeDetails(
...
+	writeStats.TotalFileCount = int(deets.Details().DetailsModel.NonMetaFileSizes())
```
ashmrtn, line 687: this should also exclude folders. Also should assign to something other than `TotalFileCount`

neha-Gupta1: Yupp. I pushed that in hurry.

src/internal/operations/backup_integration_test.go
```diff
@@ -1755,7 +1755,7 @@ func runDriveIncrementalTest(
 					makeLocRef(container1))
 			},
 			itemsRead:    0,
-			itemsWritten: 7, // 2*2(data and meta of 2 files) + 3 (dirmeta of two moved folders and target)
+			itemsWritten: 0,
```
ashmrtn, line 1758: this one seems a bit tricky and I'm not sure what the expectation is from the users' perspective. I could see them either expecting to see 2 files moved (since the folder changed) or no files moved (since the file content hasn't changed)

ashmrtn (commented):
From reading this PR it sounds like it's meant to specifically address the fact that we were previously counting OneDrive metadata files (.dirmeta, .meta) as written entries. Can you update the PR description so that's more clear? There's also the "corso metadata files" that contain the previous paths and delta tokens and I don't want to confuse the two. As far as I understand, Exchange output may still be slightly off because of the corso metadata files

src/internal/operations/backup.go
```diff
@@ -676,8 +678,19 @@ func mergeDetails(
 	mans []*kopia.ManifestEntry,
 	dataFromBackup kopia.DetailsMergeInfoer,
 	deets *details.Builder,
+	writeStats *kopia.BackupStats,
+	serviceType path.ServiceType,
 	errs *fault.Bus,
 ) error {
+	detailsModel := deets.Details().DetailsModel
+
+	if serviceType == path.OneDriveService {
```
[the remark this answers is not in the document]
ashmrtn: oh, this should also probably check for SharePoint and do the same for it

src/pkg/backup/details/details.go
```diff
@@ -236,6 +236,21 @@ func (dm DetailsModel) FilterMetaFiles() DetailsModel {
 	return d2
 }
 
+// FilterMetaFiles returns a copy of the Details with all of the
+// .meta files removed from the entries.
+func (dm DetailsModel) NonMetaFileSizes() int64 {
```
ryanfkeepers, line 241: The current name makes it sound like it returns each file size individually.
```suggestion
func (dm DetailsModel) SumNonMetaFileSizes() int64 {
```

src/internal/operations/backup.go
```diff
@@ -676,8 +678,19 @@ func mergeDetails(
...
+		// getting the values in writeStats before anything else so that we don't get a return from
+		// conditions like no backup data.
+		writeStats.TotalFileCount = len(detailsModel.FilterMetaFiles().Items())
+		writeStats.TotalUploadedBytes = detailsModel.NonMetaFileSizes()
+	}
+
```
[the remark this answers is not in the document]
neha-Gupta1: @ryanfkeepers suggested me to move it to `mergeDetails` so that is code is more structured. We can go either ways which fits best here.

src/internal/operations/backup.go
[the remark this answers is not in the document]
neha-Gupta1: yes. I will update the code to handle all the 3 services.

src/internal/operations/backup_integration_test.go
neha-Gupta1: If we see from a backup's perspective we did not backed up anything new. And it shows that we did not backup the same content again.

src/internal/operations/backup.go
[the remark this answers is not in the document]
neha-Gupta1: No. Its not. I had some confusion here. I will handle this code for all services.

src/pkg/backup/backup.go
```diff
@@ -284,12 +284,12 @@ func (b Backup) toStats() backupStats {
...
 		ID:            string(b.ID),
 		BytesRead:     b.BytesRead,
-		BytesUploaded: b.BytesUploaded,
+		BytesUploaded: b.NonMetaBytesUploaded,
 		EndedAt:       b.CompletedAt,
 		ErrorCount:    b.ErrorCount,
 		ItemsRead:     b.ItemsRead,
 		ItemsSkipped:  b.TotalSkippedItems,
-		ItemsWritten:  b.ItemsWritten,
+		ItemsWritten:  b.NonMetaItemsWritten,
```
neha-Gupta1, line 292: A better name suggestion is appreciated here.

src/internal/operations/backup_integration_test.go
```diff
@@ -716,9 +716,9 @@ func (suite *BackupOpIntegrationSuite) TestBackup_Run_incrementalExchange() {
 	testExchangeContinuousBackups(suite, control.Toggles{})
 }
 
-func (suite *BackupOpIntegrationSuite) TestBackup_Run_nonIncrementalExchange() {
-	testExchangeContinuousBackups(suite, control.Toggles{DisableDelta: true})
-}
+// func (suite *BackupOpIntegrationSuite) TestBackup_Run_nonIncrementalExchange() {
```
ryanfkeepers, line 719: This needs to get uncommented.

src/pkg/backup/backup.go
ryanfkeepers: I think this is fine for now.

ryanfkeepers (commented):
Generally looks good.  Holding off on approval until the commented test comes back and passes.

src/internal/operations/backup_integration_test.go
neha-Gupta1: Commented by mistake.

ryanfkeepers (approved):

PR #3484, opened 2023-05-23
add test name to test context

ryanfkeepers:
populate a "test_name" clues field in the
tester context, so that logs can be associated
with the calling test.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/exchange_e2e_test.go
+++ b/src/cli/backup/exchange_e2e_test.go
@@ -58,10 +58,11 @@ func TestNoBackupExchangeE2ESuite(t *testing.T) {
 }
 
 func (suite *NoBackupExchangeE2ESuite) SetupSuite() {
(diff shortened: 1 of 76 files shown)


ashmrtn (approved):
not sure if we want to pursue either of the below, we can add them in a different PR if we want
* log the time of the test with when we get the context
* if we want consolidated test/corso logs would it be possible to switch the log file to be `t.Logf`?

ryanfkeepers:
> log the time of the test with when we get the context
Can do.

> if we want consolidated test/corso logs would it be possible to switch the log file to be t.Logf?
This _should_ be doable by swapping the logfile to `stdout`.  `CORSO_LOG_FILE=stdout go test...`, for example.

PR #3485, opened 2023-05-23
Don't fail on permissions fetch 404

pandeyabs:
This PR fixes sharepoint sanity test failure [here](https://github.com/alcionai/corso/actions/runs/5060632811/jobs/9083768294). 

Context:
It's possible that by the time we read item permissions(for metadata serialization), the item may have been deleted.  In that case, we currently fail with a 404, recoverable error. Since the graph call here [looks up the item by id](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/drive.go#L217), and not the permissions by id, we can assume the item was deleted on a 404

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -513,8 +513,7 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 					oc.driveID,
 					ptr.Val(item.GetId()))
 				if e != nil && !graph.IsErrDeletedInFlight(e) {
-					// something else happened, bail
-					logger.CtxErr(ctx, err).Error("getting item")
+					logger.CtxErr(ctx, e).Error("getting item")
 				} else if graph.IsErrDeletedInFlight(e) ||
 					(e == nil && di.GetDeleted() != nil) {
 					logger.CtxErr(ctx, err).
@@ -529,11 +528,13 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 						graph.ItemInfo(item))
 
 					el.AddSkip(skip)
-				} else {
-					el.AddRecoverable(
-						clues.Wrap(err, "getting item metadata").
-							Label(fault.LabelForceNoBackupCreation))
+
+					return
 				}
+
+				el.AddRecoverable(
+					clues.Wrap(err, "getting item metadata").
+						Label(fault.LabelForceNoBackupCreation))
 
 				return
 			}
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -506,28 +506,27 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				item)
 
 			if err != nil {
-				// Skip if the item was deleted by the time we attempted to get
-				// its permissions
-				if item.GetDeleted() != nil {
+				// Check if the item was deleted
+				di, e := oc.itemGetter(
+					ctx,
+					oc.service,
+					oc.driveID,
+					ptr.Val(item.GetId()))
+				if e != nil && !graph.IsErrDeletedInFlight(e) {
+					// something else happened, bail
+					logger.CtxErr(ctx, err).Error("getting item")
+				} else if graph.IsErrDeletedInFlight(e) ||
+					(e == nil && di.GetDeleted() != nil) {
 					logger.CtxErr(ctx, err).
 						With("skipped_reason", fault.SkipNotFound).
 						Info("item not found")
 
-					skip := fault.ContainerSkip(
+					skip := fault.FileSkip(
 						fault.SkipNotFound,
 						oc.driveID,
 						itemID,
 						itemName,
 						graph.ItemInfo(item))
-
-					if isFile {
-						skip = fault.FileSkip(
-							fault.SkipNotFound,
-							oc.driveID,
-							itemID,
-							itemName,
-							graph.ItemInfo(item))
-					}
 
 					el.AddSkip(skip)
 				} else {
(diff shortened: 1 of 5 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -506,7 +506,36 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
 				item)
 
 			if err != nil {
-				el.AddRecoverable(clues.Wrap(err, "getting item metadata").Label(fault.LabelForceNoBackupCreation))
+				// Skip if the item was deleted by the time we attempted to get
```
pandeyabs, line 509: @meain Quick question: Is the `oc.itemMetaReader` call expected for files only? Line 501 above.

pandeyabs:
Using `GetDeleted()` won't work because the `DriveItemable` object might be stale. 
Doing a fresh GET is the right approach. Will publish this PR once I have that change.

src/internal/connector/onedrive/collection.go
meain: The comment is misleading, it happens for both files and folders. In case of folders, we will only have the metadata and no data.

src/internal/connector/onedrive/collection.go
```diff
@@ -498,15 +498,60 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
...
 
-			// Fetch metadata for the file
 			itemMeta, itemMetaSize, err = oc.itemMetaReader(
 				ctx,
 				oc.service,
 				oc.driveID,
 				item)
 
 			if err != nil {
-				el.AddRecoverable(clues.Wrap(err, "getting item metadata").Label(fault.LabelForceNoBackupCreation))
```
meain, line 509: From the logs, the error that the `itemMetaReader` returns is `Item not found`. I think we can just match on that and assume that the item have been deleted instead of trying to get the item again?

pandeyabs: The `item not found` is coming from [here](https://github.com/alcionai/corso/pull/3485/files#diff-847fd50c281ff780c5a089ba2b3efa718c400a5fc60061c3894298886580fae3R47).

It should actually say `permissions not found`.

src/internal/connector/onedrive/collection.go
```diff
@@ -498,15 +498,60 @@ func (oc *Collection) populateItems(ctx context.Context, errs *fault.Bus) {
...
+				// 2. Item is not nil and has deleted property set
+				// Mark item as skipped and return
+				if clues.HasLabel(e, graph.LabelStatus(http.StatusNotFound)) ||
+					graph.IsErrDeletedInFlight(e) ||
+					(e == nil && di.GetDeleted() != nil) {
+					logger.CtxErr(ctx, err).
+						With("skipped_reason", fault.SkipNotFound).
+						Info("item not found")
+
+					skip := fault.ContainerSkip(
```
ryanfkeepers, line 526: We do not use fault skips for deleted items.  A fault.skip is only for tracking items that are valid, but which we are unable or unwilling to retrieve.  Malware is a good example: it's a valid file, we just don't want to deal with it (and msoft wont let us, anwyay).  A deleted/not-found item is not valid: it doesn't exist.

src/internal/connector/onedrive/collection.go
meain: What I meant was if we get 404 when trying to fetch permissions, can't we just assume that it is because the file was not available and thus just assume the file was deleted without trying to fetch the file again? We could avoid [this](https://github.com/alcionai/corso/blob/100f1eb290a45c047d7789c68a7f7321d147ec8c/src/internal/connector/onedrive/collection.go#L508-L513) completely.

src/internal/connector/onedrive/collection.go
meain: The way we handle 404 is different across different services.I think we silently ignore them in Exchange, but we do add skips for them in OneDrive ([ref](https://github.com/alcionai/corso/blob/7181e2ef90eedcfe7451c9a96949a6978163a8c3/src/internal/connector/onedrive/collection.go#L347)). I believe what Keepers is suggesting is to silently ignore 404 as they are something that "does not exist" at the time of  backup and should be treated as if we never knew about them.

I think we should keep both Exchange and OneDrive consistent, but I'll let Keepers make the final call on what to do here.

src/internal/connector/onedrive/collection.go
pandeyabs: >What I meant was if we get 404 when trying to fetch permissions, can't we just assume that it is because the file was not available and thus just assume the file was deleted without trying to fetch the file again

I see. I'm not sure if it's a valid assumption, since permissions is its own dedicated API and can return 404s. 
I don't know whether ignoring a permissions 404 is the right thing to do though. @ryanfkeepers any thoughts?

src/internal/connector/onedrive/collection.go
ryanfkeepers: Yup!  Also, I forgot that we were "skipping" missing onedrive items.  We've had sync discussions stating the need to remove that skip and silently ignore it, instead.

src/internal/connector/onedrive/collection.go
pandeyabs: > Since the graph call here [looks up the item by id](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/drive.go#L217), and not the permissions by id, I think we can assume the item was deleted on a 404

This is a good point. thanks! I'll remove the recoverable error for permissions 404 then, just an info log should do fine.

src/internal/connector/onedrive/collection.go
pandeyabs: I'll remove the onedrive skip in a new PR

pandeyabs:
Closing this PR as it's stale and I'm running into unnecessary rebase issues. Will open a new one to replace this PR

Issue #3011, opened 2023-04-03
Improving corso restore for OneDrive


meain:
- [ ] #3536
- [x] Investigate throttling (create/upload takes 2 credits compared to 1 for get - [ref](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling))
- [ ] Restoring directories (related: https://github.com/alcionai/corso/issues/2567)
- [ ] #3537
- [x] Potential OOM issues - _as of now only high mem use seems to be in repo connect_

meain:
Closing in favor of individual issues!

PR #3492, opened 2023-05-24
Parallelize restores within a collection for OneDrive

meain:
This should massively speed up when restoring a collection with many
items. Will not impact much if we have a lot of collections with few
items each.

Numbers 🔢 :
- Restoring ~7000 files, mostly small, totaling 1.5GB
  - Sequential: ~70m
  - Parallel: ~50m
- Restoring 1200 50mb files
  - Sequential: 4h 45m
  - Parallel: <40m

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3011
* closes https://github.com/alcionai/corso/issues/3536

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added OwnerName to the backup list json output.  OwnerName holds either a UPN or a WebURL, depending on the resource type.
+- Improve OneDrive restore performance
```
ashmrtn, line 12: may be worth saying how just cause we can
```suggestion
- Improve OneDrive restore performance by parallelizing item restore
```

src/internal/connector/graph/consts.go
```diff
@@ -88,6 +90,14 @@ func (p parallelism) Item() int {
 	return p.item
 }
 
+func (p parallelism) ItemUpload() int {
+	if p.itemUpload == 0 {
```
ashmrtn, line 94: we should probably set an upper bound too. This is going to cause goroutine and buffer creation so we may want to make sure that people don't accidentally set it too high

src/internal/connector/onedrive/restore.go
```diff
@@ -223,62 +261,74 @@ func RestoreCollection(
 
 		case itemData, ok := <-items:
 			if !ok {
-				return metrics, nil
+				complete = true
+				break
```
ashmrtn, line 265: Since it's a little difficult to parse with a select in a for-loop could you please add a comment that this will get us out of the select and then the check above will get us out of the for-loop

src/internal/connector/onedrive/restore.go
```diff
@@ -669,14 +714,70 @@ func createRestoreFolders(
 	return parentFolderID, nil
 }
 
+func copyBufferWithStallCheck(dst io.Writer, src io.Reader, stallTimeout time.Duration) (int64, error) {
+	buffer := make([]byte, copyBufferSize)
```
ashmrtn, line 718: since these buffers are rather large it may actually be worth making a buffer pool for them. We can use sync.Pool, we just need to make sure we don't free them before we're done with them

src/internal/connector/onedrive/restore.go
```diff
@@ -669,14 +714,70 @@ func createRestoreFolders(
...
+		}()
+
+		timer.Reset(stallTimeout)
+
+		select {
+		case <-timer.C:
+			return 0, clues.New("copy stalled")
+		case in := <-cont:
+			if in.err != nil {
+				return 0, in.err
```
ashmrtn, line 760: I think this needs a bit more definition on the error handling. For example, if there's a read error it still tries to upload the data and totalCopied is still updated. However, this will return prior to the upload completing and it also returns 0 as the amount of data copied

We haven't really discussed how to handle partial file uploads, but we should at least make sure we don't exit this function with the goroutine from above still running

src/internal/connector/onedrive/restore.go
```diff
@@ -669,14 +714,70 @@ func createRestoreFolders(
...
+			}
+
+			if in.complete {
+				return totalCopied, nil
+			}
+
+			// Continue copying
+			continue
```
ashmrtn, line 768: this can be removed since it's at the end of the case
```suggestion
```

src/internal/connector/onedrive/restore.go
```diff
@@ -696,26 +797,55 @@ func restoreData(
...
+
+		pname := name
+		iReader := itemData.ToReader()
+
+		if i > 0 {
+			pname = fmt.Sprintf("%s (retry %d)", name, i)
+
+			// If it is not the first try, we have to pull the file
+			// again from kopia. Ideally we could just seek the stream
+			// but we don't have a Seeker available here.
```
ashmrtn, line 822: make an issue for this. Technically kopia returns items that are like the below so we could get a `Seeker` here if we needed. It would be quite a few changes to update the interfaces, but still possible
```go
type Reader interface {
  io.ReadCloser
  io.Seeker
}
```

src/internal/connector/onedrive/restore.go
```diff
@@ -696,26 +797,55 @@ func restoreData(
...
-	iReader := itemData.ToReader()
-	progReader, closer := observe.ItemProgress(
-		ctx,
-		iReader,
-		observe.ItemRestoreMsg,
-		clues.Hide(name),
-		ss.Size())
+	// This is just to retry file upload, the uploadSession creation is not retried here
+	// We need extra logic to retry file upload as we have to pull the file again from kopia
+	for i := 0; i <= maxUploadRetries; i++ {
```
ashmrtn, line 807: not really sure where to put this comment so adding here...

This loop needs clearer definitions of what happens if there's an error. Do we attempt to upload all of the file data again? Do we begin uploading where we left off or near where we left off (maybe reupload the last `copyBufferSize` set of bytes as well as we may not know their state)

It's fine if we don't go for the super optimized solution right now, but we should make it clear what the behavior is. If we want to optimize it later we can track that with tickets

ashmrtn (changes requested):
this generally looks good, though it needs improvements around how things are handled if there's an upload error. It also needs some more comments detailing the behavior when an error occurs

ashmrtn (changes requested):
this generally looks good, though it needs improvements around how things are handled if there's an upload error. It also needs some more comments detailing the behavior when an error occurs

src/internal/connector/graph/concurrency_middleware.go
```diff
@@ -149,6 +149,10 @@ const (
 	SingleGetOrDeltaLC = 1
 	// limit consumption rate for anything permissions related
 	PermissionsLC = 5
+	// https://learn.microsoft.com/en-us/sharepoint/dev/general-development
+	// /how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling
+	// Any create, delete, update or upload operation take 2 tokens
+	CreateLC = 2
```
ryanfkeepers, line 155: These cases should already fall under the `driveDefaultLC`.

src/internal/connector/onedrive/drive.go
```diff
@@ -203,6 +203,8 @@ func CreateItem(
 	rawURL := fmt.Sprintf(itemChildrenRawURLFmt, driveID, parentFolderID)
 	builder := drives.NewItemItemsRequestBuilder(rawURL, service.Adapter())
 
+	ctx = graph.ConsumeNTokens(ctx, graph.CreateLC)
```
ryanfkeepers, line 206: This already consumes 2 tokens by default.

src/internal/connector/onedrive/restore.go
```diff
@@ -669,14 +714,70 @@ func createRestoreFolders(
...
+
+		select {
+		case <-timer.C:
+			return 0, clues.New("copy stalled")
+		case in := <-cont:
+			if in.err != nil {
+				return 0, in.err
+			}
+
+			if in.complete {
```
ryanfkeepers, line 763: if incomplete, it's complete.  😛

src/internal/connector/onedrive/restore.go
```diff
@@ -669,14 +714,70 @@ func createRestoreFolders(
...
+			}
+
+			cont <- ce{complete: false}
+		}()
+
+		timer.Reset(stallTimeout)
+
+		select {
+		case <-timer.C:
+			return 0, clues.New("copy stalled")
```
ryanfkeepers, line 757:
```suggestion
			return 0, clues.New("content buffer stream stalled")
```

src/internal/connector/onedrive/restore.go
```diff
@@ -696,26 +797,55 @@ func restoreData(
...
-		clues.Hide(name),
-		ss.Size())
+	// This is just to retry file upload, the uploadSession creation is not retried here
+	// We need extra logic to retry file upload as we have to pull the file again from kopia
+	for i := 0; i <= maxUploadRetries; i++ {
+		// Get a drive item writer
+		w, err = driveItemWriter(ctx, service, driveID, ptr.Val(newItem.GetId()), ss.Size())
+		if err != nil {
+			return "", details.ItemInfo{}, err
+		}
```
ryanfkeepers, line 812: The driveWriter doesn't need to be re-created on retry, and can be moved out of the loop.

src/internal/connector/onedrive/restore.go
```diff
@@ -696,26 +797,55 @@ func restoreData(
...
+
+		progReader, closer := observe.ItemProgress(
+			ctx,
+			iReader,
+			observe.ItemRestoreMsg,
+			clues.Hide(pname),
+			ss.Size())
 
-	go closer()
+		go closer()
```
ryanfkeepers, line 838: Are we sure this will behave properly across loops?  Seems like you'd stack up multiple displays of the same item.

src/internal/connector/graph/consts.go
meain: Hmm, I personally don't think we need as this is something that we explicitly mention is about parallelism and it is highly unlikely that people will be assigning random values. But in case people set it to higher value, should we panic or just cap it? Or should we just add tests so to check max?

src/internal/connector/onedrive/restore.go
meain: Hmm, that was added mostly to satisfy the linter. Having a block with just a comment was causing lint issues.

src/internal/connector/onedrive/restore.go
meain: https://github.com/alcionai/corso/issues/3520

src/internal/connector/graph/concurrency_middleware.go
meain: Hmm, I didn't think about that. That said, I would like to keep it just to be explicit.

src/internal/connector/onedrive/restore.go
meain: Good catch, I made some last minute changes and missed a few things. 😓

src/internal/connector/onedrive/restore.go
meain: It currently restarts the upload from beginning. I've created an issue (https://github.com/alcionai/corso/issues/3521) to track resuming upload.

src/internal/connector/onedrive/restore.go
meain: Graph does not seem to like reusing the endpoint. Let me look into it along with resuming uploads (created an issue https://github.com/alcionai/corso/issues/3521).

src/internal/connector/onedrive/restore.go
meain: Ahh, somehow missed that possibility. The `copyBufferWithStallCheck` was a result of me not understanding this was what was going on. I'll run a few more tests, but I think we don't need that function and a simple copy would do with the http lib failing in case of any issues.

src/internal/connector/onedrive/restore.go
meain: I'm dropping the function to copy with stall check. It was just me panicking from the result of https://github.com/alcionai/corso/pull/3492#discussion_r1205775330 .

src/internal/connector/graph/consts.go
ashmrtn: I think we still need some cap on it given that we're making 5MB buffers for each thread we use. It can be a fairly high cap, but we don't want to end up OOMing or dying due to memory pressure if the user accidentally adds an extra 0 or something

src/internal/observe/observe.go
```diff
@@ -239,15 +239,15 @@ func ItemProgress(
 	header string,
 	iname any,
 	totalBytes int64,
-) (io.ReadCloser, func()) {
+) (io.ReadCloser, func(), func()) {
```
ashmrtn, line 242: mmm so I know it's a bit more work, but this may be best addressed in a separate PR that goes in prior to this one. I think my biggest concern about returning both an abort and a done function centers around what happens if abort is called and then done is called (i.e. we deferred the done call but ended up with an error and called abort instead)?

There's also some WaitGroup usage in this file that may not be addressed by the abort call which could potentially cause deadlocks if we're not careful

ryanfkeepers: Second this.  We can stick with the current implementation with itemProgress display duplication (maybe hold off on this pr merger until after our next release, if it's on monday?) and fix up the situation after.

Also of note: having a closer func for observe is already a bit heavy.  Ideally, this interface would work similar to a log, where you can fire and forget.  Unfortunately, we don't have exactly that, so it goes, we can make do with a closer.  Beyond that, interaction starts to get really unwieldy.  A clean and minimized interface is important here.

That's not to say we can't have an abort func (validating ash's worries notwithstanding).  But we'll probably want it in the form of returning a struct representing the display persistence, and have multiple control methods available through that.  Or something of that nature.

src/internal/connector/graph/concurrency_middleware.go
ryanfkeepers: It's a nit, but I'd rather it get dropped.  One of two patterns works here: a default value is assumed and only deviation needs to be explicit (what we currently have); or everything needs to be explicit.  This change sits at a halfway point where we have a default, and deviations, _and_ certain places where we call out the same consumption as the default.  That third condition introduces more confusion than it solves.  It implies that either the default might change while this value needs to be protected (doubtful to ever happen), or that cases which don't declare a value are incorrectly configured (but they aren't).

If you feel more comfortable about the implementation if we always declare the token consumption, that's okay.  In that case I'd ask that you ensure that declaration is complete throughout the repo.  Otherwise, things are cleaner without the one-off.

src/internal/connector/onedrive/restore.go
```diff
@@ -212,8 +224,39 @@ func RestoreCollection(
...
+
+		err = deets.Add(repoRef, locationRef, updated, info)
+		if err != nil {
+			// Not critical enough to need to stop restore operation.
+			logger.CtxErr(ctx, err).Infow("adding restored item to details")
+		}
+	}
+
+	// Buffer pool for uploads
+	// TODO: Should we move this up to RestoreCollections?
```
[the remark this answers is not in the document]
meain: I'm moving it up to https://github.com/alcionai/corso/blob/3798fc8f056eaf2d0aabef02ec947b6b50e519cb/src/internal/connector/data_collections.go#L242 at which point we can just send it down to both sharepoint and onedrive with a single init. If we do it in `RestoreCollections` we might have two pool creations, one for OneDrive and one for SharePoint.

src/internal/observe/observe.go
meain: I don't think this will be an issue. What we return in `closer` is [this](https://github.com/alcionai/corso/blob/d00696621f545ea25098f5f69348146bc837f419/src/internal/observe/observe.go#L432), which is essentially wait for the bar to end(abort or complete) , then remove the wg item and print log. That function does not really mark it as "done". It just waits for a done message to do some cleanup and logging. When we call abort, we just happen to simulate the "mark as done" action and stop the bar.

meain: I've created https://github.com/alcionai/corso/pull/3535 where we completely avoid returning closers to the consumer and call that within the progressbar creation code. This should eliminate any potential issues with progressbars.

That PR can be merged into this before this goes into main.

meain: Now that https://github.com/alcionai/corso/pull/3535 is merged, I would like another round of review @ashmrtn @ryanfkeepers

ryanfkeepers (approved):

src/internal/observe/observe.go
ashmrtn: Thanks for the PR on progress bars. I don't mean to unduly hold up this PR or belabor the point, but unfortunately locking/deadlocks is one of the things that are best viewed with as few other changes possible since getting it wrong can be painful and sometimes difficult to reproduce

ashmrtn (approved):

src/internal/observe/observe.go
meain: No worries. I would rather fix this now when we know the issue than later hunting in the dark when something breaks :D

PR #3495, opened 2023-05-24
Do not print all the restored items if there are a lot of them

meain:
For large restore, it just ends up printing a lot of text. This change makes it print a summary instead. We still retain printing the full content if the user asks for a json output as we assume they would be piping it somewhere.

```
Connecting to repository:      1s done
Restoring to folder Corso_Restore_24-May-2023_11-49-44
Connecting to M365:         done
Restoring ∙ 7ceb8e03-bdc5-4509-a136-457526165ec0
Discovered 3 items in backup d5c07808-545d-4906-9ade-f224ca8f13de to restore
Enumerating items in repository:      0s done
Restoring data:     54s done
Restored 304 items.
```

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix Exchange backup issue caused by incorrect json serialization
 
 ### Changed
-- Do not display all the items that we restored at the end if there are more than 15
+- Do not display all the items that we restored at the end if there are more than 15. You can override this with `--verbose`.
 
 ### Known Issues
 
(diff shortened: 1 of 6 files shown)


ryanfkeepers (commented):
Logic looks good.  However, I would like a way to override this limit and print all items without needing the `--json` flag.

meain:
@ryanfkeepers I've added a `--verbose` flag which when used overrides this behaviour. Let me know if we should tweak the name or description of the flag.

ryanfkeepers:
> I've added a --verbose flag which when used overrides this behaviour. Let me know if we should tweak the name or description of the flag.

Thanks!  I'm fine with it for now.  "verbose" is tricky due to being a very generic term.  What, exactly, is verbose?  The logging?  The total output?  We probably don't need more fine tuning beyond this at the moment, but we'll want to make sure we keep user expectations in sight here.

I'll let @gmatev have final say on the flag naming.

ryanfkeepers (approved):

Issue #3202, opened 2023-04-24
Smarter base selection algorithm for incremental backups


ashmrtn:
Currently Corso has a rather simplistic algorithm for selecting bases for incremental backups: it selects the most recently completed data snapshot and the most recently created incomplete data snapshot if there's an incomplete snapshot newer than the completed snapshot

However, the above doesn't take into account other information that's required to have a "complete" backup needed to be considered a base for incremental backups. Other information that is required includes
* the backup model the data snapshot references
* backup details for the backup the data snapshot corresponds to

Corso's ability to complete a kopia snapshot even if there were errors creates a situation where it's possible to find a complete kopia snapshot without a backup model, thus triggering a full backup instead of an incremental backup from an older base backup

Further complicating things, recent complete/incomplete data snapshots should be given to kopia when making the backup so kopia-assisted incrementals is enabled. This puts Corso in a situation where lookups need to be done on both the backup models and the data snapshots to determine what backups to use as a base for incrementals and what snapshots to use for kopia-assisted incrementals

ashmrtn:
related to #2068

ashmrtn:
core logic for selecting bases has been written. Unfortunately wiring it up and updating tests is turning into more work than expected

ashmrtn:
closing code for better base selection has been wired up and merged. There's room for refactoring and further optimizations. These are tracked by #3525

PR #3500, opened 2023-05-24
Move some interfaces to kopia package

ashmrtn:
Searching for new bases is mostly kopia-specific but is going to require the store interface. With the way packages and interfaces are currently laid out it makes it impossible to easily get the interface added to the inject package without a cycle.

This moves functions that the kopia package provides into a subpackage of the kopia package that other things can then import

This change only moves code. No logic is updated

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3202

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ryanfkeepers (approved):
I'm good with this to unblock.  Ideally, the solution would be to replace the `kopia` imports in op/inject with interfaces or generic versions of the same structures.  Needing to specify kopia at that layer is, itself, a coupling anti-pattern.

ryanfkeepers:
per the above comment: https://github.com/alcionai/corso/issues/3510

PR #3501, opened 2023-05-24
migrate last funcs into backupHandler

ryanfkeepers:
migrates the last of the per-type funcs into the onedrive backupHandler.  Or, in some cases, produces a generic func instead of making an addition to the interface.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/drive_pager.go
+++ b/src/pkg/services/m365/api/drive_pager.go
@@ -21,7 +21,7 @@ import (
 // item pager
 // ---------------------------------------------------------------------------
 
-type driveItemPager struct {
+type DriveItemPager struct {
 	gs      graph.Servicer
 	driveID string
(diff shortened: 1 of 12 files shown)


src/pkg/services/m365/api/drive_pager.go
```diff
@@ -21,7 +21,7 @@ import (
 // item pager
 // ---------------------------------------------------------------------------
 
-type driveItemPager struct {
+type DriveItemPager struct {
```
pandeyabs, line 24: nit: Add `var _ DriveItemEnumerator = &DriveItemPager{}` (and others) in this file too?

pandeyabs (approved):
LGTM! Left one nit.

PR #3512, opened 2023-05-25
establish restore handler interfaces

ryanfkeepers:
Complete the ondrive and sharepoint handler creation by introducing restoreHandler interfaces and structs.  

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/connector/onedrive/item_handler.go
```diff
@@ -66,6 +70,72 @@ func (h itemBackupHandler) AugmentItemInfo(
...
+func (h itemRestoreHandler) AugmentItemInfo(
+	dii details.ItemInfo,
+	item models.DriveItemable,
+	size int64,
+	parentPath *path.Builder,
+) details.ItemInfo {
+	return augmentItemInfo(dii, item, size, parentPath)
+}
+
+func (h itemRestoreHandler) FolderByNameGetter() GetFolderByNamer          { return h.ac }
```
pandeyabs, line 123: Curious what are the benefits here of having 6 functions return the same data - `h.ac` ?

pandeyabs (changes requested):

src/internal/connector/onedrive/item_handler.go
ryanfkeepers: I initially thought it would work as a way for tests to piecemeal their mock structures.  IE: you get to compile 6 different small mocks as opposed to one big one.  More modular, easily swappable, etc.

In practice, in later PRs, I found the implementation wasn't as elegant as the concept.  In fact, it caused more work than it solved.  So most/all of these have been transitioned to api client func wrappers instead.

pandeyabs: yeah, that would be mock combo hell. Thanks

PR #3515, opened 2023-05-25
getting to green following refactors, pt 1

ryanfkeepers:
fixing up tests and errors on the way to getting green
again, part 1.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/collection.go
+++ b/src/internal/connector/onedrive/collection.go
@@ -24,6 +24,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
+	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
 
 const (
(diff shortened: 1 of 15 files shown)


src/internal/connector/onedrive/mock/handlers.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/path"
+	"github.com/alcionai/corso/src/pkg/services/m365/api"
+)
+
+// ---------------------------------------------------------------------------
+// Backup Handler
+// ---------------------------------------------------------------------------
+
+type BackupHandler struct {
```
pandeyabs, line 18: We have an interface with the same name. Looks like this struct is implementing that interface. 
Can this be named `MockBackupHandler`?

src/internal/connector/onedrive/collection.go
```diff
@@ -307,16 +308,22 @@ func (oc *Collection) getDriveItemContent(
 	return itemData, nil
 }
 
+type itemGetterAndAugmenter interface {
```
pandeyabs, line 311: Can we add some documentation for this interface?

src/internal/connector/onedrive/item.go
```diff
@@ -23,9 +23,14 @@ var downloadURLKeys = []string{
 	"@content.downloadUrl",
 }
 
+type augmenterAndGetter interface {
```
pandeyabs, line 26: Do we need this? Can we just pass `itemGetterAndAugmenter` to `downloadItem`?

pandeyabs (changes requested):
Left a few review comments.

src/internal/connector/onedrive/mock/handlers.go
ryanfkeepers: That would stutter, since the package name is also mock.  IE: `mock.MockFoo`.  Since the declarations are separated by package, the difference between `onedrive.BackupHandler` and `mock.BackupHandler` helps to produce that clarity.

Per the duplication between a struct and an interface name, I do like that less.  And in production code (ie, non-mock) I'm a little more critical about these overlaps.  But here, I think, it works out to read like "mock.BackupHandler means it mocks the BackupHandler interface".

src/internal/connector/onedrive/collection.go
ryanfkeepers: will do

src/internal/connector/onedrive/item.go
ryanfkeepers: fixed in later PR

PR #3534, opened 2023-05-31
Refactor downloadItem

pandeyabs:
This refactor is being done to allow calling `downloadFile` with an URL obtained from URL cache. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -46,10 +46,6 @@ func downloadItem(
 			}
 		}
 
-		if len(url) == 0 {
-			return nil, clues.New("extracting file url")
-		}
-
 		rc, err = downloadFile(ctx, ag, url)
 		if err != nil {
 			return nil, err
@@ -64,9 +60,13 @@ func downloadFile(
 	ag api.Getter,
 	url string,
 ) (io.ReadCloser, error) {
+	if len(url) == 0 {
+		return nil, clues.New("empty file url")
+	}
+
 	resp, err := ag.Get(ctx, url, nil)
 	if err != nil {
-		return nil, clues.Wrap(err, "getting item")
+		return nil, clues.Wrap(err, "getting file")
 	}
 
 	if graph.IsMalwareResp(ctx, resp) {
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -27,6 +27,10 @@ func downloadItem(
 	ag api.Getter,
 	item models.DriveItemable,
 ) (io.ReadCloser, error) {
+	if item == nil {
+		return nil, clues.New("nil item")
+	}
+
 	var (
 		rc     io.ReadCloser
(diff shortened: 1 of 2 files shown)


src/internal/connector/onedrive/item_test.go
```diff
@@ -256,3 +258,158 @@ func (suite *ItemIntegrationSuite) TestDriveGetFolder() {
...
+			},
+			errorExpected: true,
+			rcExpected:    false,
+		},
+		{
+			name: "success",
+			itemFunc: func() models.DriveItemable {
+				di := newItem("test", false)
+				di.SetAdditionalData(map[string]interface{}{
+					"@microsoft.graph.downloadUrl": "https://example.com",
```
ashmrtn, line 312: could you add one more test that has the URL as the value for a field with key `"@content.downloadUrl"`? It can be the same as this test for all other checks and stuff

src/internal/connector/onedrive/item_test.go
```diff
@@ -256,3 +258,158 @@ func (suite *ItemIntegrationSuite) TestDriveGetFolder() {
...
+
+func (suite *ItemUnitTestSuite) TestDownloadItem() {
+	testRc := ioutil.NopCloser(bytes.NewReader([]byte("test")))
+
+	table := []struct {
+		name          string
+		itemFunc      func() models.DriveItemable
+		GetFunc       func(ctx context.Context, url string) (*http.Response, error)
+		errorExpected bool
+		rcExpected    bool
```
ashmrtn, line 293: you can actually specify assert/require functions as part of the struct. For example, you could have the below types and then just give the function in the struct definitions and call the function in the test itself
```suggestion
		errorExpected require.ErrorAssertionFunc
		rcExpected    require.ValueAssertionFunc // single-arg assertion func
```

in struct definitions:
```go
{
  ...
  errorExpected: require.NoError // or use require.Error if there should be an error
  rcExpected: require.NotNil // or use require.Nil for no rc
}
```

calling in test:
```go
  test.errorExpected(t, err, clues.ToCore(err))
  test.rcExpected(t, rc)
```

src/internal/connector/onedrive/item.go
```diff
@@ -45,31 +50,42 @@ func downloadItem(
...
 
-		if len(url) == 0 {
-			return nil, clues.New("extracting file url")
-		}
-
-		resp, err := ag.Get(ctx, url, nil)
+		rc, err = downloadFile(ctx, ag, url)
 		if err != nil {
-			return nil, clues.Wrap(err, "getting item")
+			return nil, err
```
ashmrtn, line 55: just so we get stack trace information
```suggestion
			return nil, clues.Stack(err)
```

ashmrtn (approved):

Issue #3069, opened 2023-04-07
Improve how expired download URLs are handled


vkamra:
Currently - when a OneDrive backup is initiated - we do a delta based enumeration and request download URLs for every item (this allows us to not have to call GetDriveItem for every item.

However - these URLs are time limited and if the backup operation takes too long they will expire (request will fail with a 401 error) in which case we call GetItem to get a new download URL.

This should be optimized because it's an extra round trip for every item (to get the 401 and then renew).

Possible options could be:
1) Do not enumerate all items up front - enumerate a batch, process it and then enumerate more - the assumption being that the download URLs will be fresher.
2) Track expiry time for these URLs (or trigger expiry on the first 401) and then assume we need to get new ones (i.e. save the 401 round trip)
3) ~Is there something we can do client side to renew these URLs e.g. get a new access token and update URL?~ Update - not possible
4) A version of (1) will be to do a full delta re-enumeration when we hit a 401. It will be cheaper (in terms of Graph requests/tokens) vs calling GetItem for everything left

pandeyabs:
Re #2, did we also explore [configurable token lifetime(CTL)](https://learn.microsoft.com/en-us/azure/active-directory/develop/access-tokens#access-token-lifetime) to extend token lifetime from ~1hr to ~20 hrs?

Although, this may not be ideal from a security standpoint. `Adjusting the lifetime of an access token is a trade-off between improving system performance and increasing the amount of time that the client retains access after the user's account is disabled.`

vkamra:
> Re option 3, did we also explore [configurable token lifetime(CTL)](https://learn.microsoft.com/en-us/azure/active-directory/develop/access-tokens#access-token-lifetime) to extend token lifetime to > 1hr ?

I don't believe that applies to these. The download URLs are [generated](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_get_content?view=odsp-graph-online#response) by the backend service and their lifetime isn't configurable

pandeyabs:
Thanks! Let me explore this a bit more. 

> It also looks like this would be a setting in the customer environment and not something Corso would be able to control?

I found [this page ](https://learn.microsoft.com/en-us/azure/active-directory/develop/active-directory-configurable-token-lifetimes?source=recommendations#application-policies )which seems to suggest that token lifetime policies can be attached to an application.

[...]

PR #3542, opened 2023-05-31
Integrate URL cache into onedrive backup flows

pandeyabs:
Changes:

* Attach URL cache to a collection only if it's parent drive has < 300k items. Higher item count reduces the ROI of using cache, and can even degrade performance in severe cases. We will stick to per item GETs for those scenarios, until we figure out a longer term solution.
* 
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3069

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/collection.go
+++ b/src/internal/m365/onedrive/collection.go
@@ -85,7 +85,7 @@ type Collection struct {
 	// should only be true if the old delta token expired
 	doNotMergeItems bool
 
-	cache urlCacher
+	urlCache getItemPropertyer
 }
 
 func pathToLocation(p path.Path) (*path.Builder, error) {
@@ -111,7 +111,7 @@ func NewCollection(
 	ctrlOpts control.Options,
 	colScope collectionScope,
 	doNotMergeItems bool,
-	cache urlCacher,
+	urlCache getItemPropertyer,
 ) (*Collection, error) {
 	// TODO(ashmrtn): If OneDrive switches to using folder IDs then this will need
 	// to be changed as we won't be able to extract path information from the
@@ -136,7 +136,7 @@ func NewCollection(
 		ctrlOpts,
 		colScope,
 		doNotMergeItems,
-		cache)
+		urlCache)
 
 	c.locPath = locPath
 	c.prevLocPath = prevLocPath
@@ -153,7 +153,7 @@ func newColl(
 	ctrlOpts control.Options,
 	colScope collectionScope,
 	doNotMergeItems bool,
-	cache urlCacher,
+	urlCache getItemPropertyer,
 ) *Collection {
 	c := &Collection{
 		handler:         handler,
@@ -167,7 +167,7 @@ func newColl(
 		state:           data.StateOf(prevPath, currPath),
 		scope:           colScope,
 		doNotMergeItems: doNotMergeItems,
-		cache:           cache,
+		urlCache:        urlCache,
 	}
 
 	return c
@@ -273,7 +273,7 @@ func (oc *Collection) getDriveItemContent(
 		el       = errs.Local()
 	)
 
-	itemData, err := oc.downloadContent(ctx, oc.handler, item, oc.driveID)
+	itemData, err := downloadContent(ctx, oc.handler, oc.urlCache, item, oc.driveID)
 	if err != nil {
 		if clues.HasLabel(err, graph.LabelsMalware) || (item != nil && item.GetMalware() != nil) {
 			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipMalware).Info("item flagged as malware")
@@ -323,9 +323,10 @@ type itemAndAPIGetter interface {
 // downloadContent attempts to fetch the item content.  If the content url
 // is expired (ie, returns a 401), it re-fetches the item to get a new download
 // url and tries again.
-func (oc *Collection) downloadContent(
+func downloadContent(
 	ctx context.Context,
 	iaag itemAndAPIGetter,
+	uc getItemPropertyer,
 	item models.DriveItemable,
 	driveID string,
 ) (io.ReadCloser, error) {
@@ -343,9 +344,9 @@ func (oc *Collection) downloadContent(
 	// token, and that we've overrun the available window to
 	// download the file.  Get a fresh url from the cache and attempt to
 	// download again.
-	content, err = oc.readFromCache(ctx, iaag, itemID)
+	content, err = readItemContents(ctx, iaag, uc, itemID)
 	if err == nil {
-		logger.Ctx(ctx).Debug("found item in cache", itemID)
+		logger.Ctx(ctx).Debug("found item in url cache")
 
 		return content, nil
 	}
@@ -353,7 +354,7 @@ func (oc *Collection) downloadContent(
 	// Consider cache errors(including deleted items) as cache misses. This is
 	// to preserve existing behavior. Fallback to refetching the item using the
 	// API.
-	logger.CtxErr(ctx, err).Info("cache miss. refetching from API")
+	logger.CtxErr(ctx, err).Info("url cache miss: refetching from API")
 
 	di, err := iaag.GetItem(ctx, driveID, ptr.Val(item.GetId()))
 	if err != nil {
@@ -368,18 +369,19 @@ func (oc *Collection) downloadContent(
 	return content, nil
 }
 
-// readFromCache fetches latest download URL from the cache and attempts to
+// readItemContents fetches latest download URL from the cache and attempts to
 // download the file using the new URL.
-func (oc *Collection) readFromCache(
+func readItemContents(
 	ctx context.Context,
 	iaag itemAndAPIGetter,
+	uc getItemPropertyer,
 	itemID string,
 ) (io.ReadCloser, error) {
-	if oc.cache == nil {
-		return nil, clues.New("nil cache")
+	if uc == nil {
+		return nil, clues.New("nil url cache")
 	}
 
-	props, err := oc.cache.getItemProperties(ctx, itemID)
+	props, err := uc.getItemProperties(ctx, itemID)
 	if err != nil {
 		return nil, err
 	}
@@ -387,7 +389,6 @@ func (oc *Collection) readFromCache(
 	// Handle newly deleted items
 	if props.isDeleted {
 		logger.Ctx(ctx).Info("item deleted in cache")
-
 		return nil, graph.ErrDeletedInFlight
 	}
 
--- a/src/internal/m365/onedrive/collection.go
+++ b/src/internal/m365/onedrive/collection.go
@@ -347,7 +347,6 @@ func downloadContent(
 	content, err = readItemContents(ctx, iaag, uc, itemID)
 	if err == nil {
 		logger.Ctx(ctx).Debug("found item in url cache")
-
 		return content, nil
 	}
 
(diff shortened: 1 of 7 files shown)


src/internal/connector/onedrive/collection.go
```diff
@@ -332,8 +341,20 @@ func downloadContent(
...
+	if err == nil {
+		logger.Ctx(ctx).Debug("found item in cache", itemID)
+
+		return content, nil
+	}
+
+	// Consider cache errors(including deleted items) as cache misses. This is
+	// to preserve existing behavior. Fallback to refetching the item using the
+	// API.
+	logger.CtxErr(ctx, err).Info("cache miss. refetching from API")
```
ryanfkeepers, line 356:
```suggestion
	logger.CtxErr(ctx, err).Info("url cache miss: refetching from API")
```

src/internal/connector/onedrive/collection.go
```diff
@@ -332,8 +341,20 @@ func downloadContent(
 
 	// Assume unauthorized requests are a sign of an expired jwt
 	// token, and that we've overrun the available window to
-	// download the actual file.  Re-downloading the item will
-	// refresh that download url.
+	// download the file.  Get a fresh url from the cache and attempt to
+	// download again.
+	content, err = oc.readFromCache(ctx, iaag, itemID)
+	if err == nil {
+		logger.Ctx(ctx).Debug("found item in cache", itemID)
```
ryanfkeepers, line 348: ItemID is already present in the log thanks to the clues in ctx.
```suggestion
		logger.Ctx(ctx).Debug("found item in url cache")
```

src/internal/connector/onedrive/collection.go
```diff
@@ -347,6 +368,41 @@ func downloadContent(
 	return content, nil
 }
 
+// readFromCache fetches latest download URL from the cache and attempts to
+// download the file using the new URL.
+func (oc *Collection) readFromCache(
```
ryanfkeepers, line 373: `cache` can take on many terms, and it's possible we'll have multiple caches.  Best practice is to denote what type of cache you're reading as part of the func name.  Eg: `readFromURLCache`.

src/internal/connector/onedrive/collection.go
```diff
@@ -347,6 +368,41 @@ func downloadContent(
...
+
+	props, err := oc.cache.getItemProperties(ctx, itemID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Handle newly deleted items
+	if props.isDeleted {
+		logger.Ctx(ctx).Info("item deleted in cache")
+
```
ryanfkeepers, line 390:
```suggestion
```

src/internal/connector/onedrive/collection.go
```diff
@@ -347,6 +368,41 @@ func downloadContent(
...
+		return nil, graph.ErrDeletedInFlight
+	}
+
+	rc, err := downloadFile(ctx, iaag, props.downloadURL)
```
ryanfkeepers, line 394: A download isn't a read from a cache, which makes this behavior opaque and unexpected since the func name sounds like you're only reading from a cache.

All in all, I don't think the Collection makes for a sensible owner here.  Presumably, this is a collection method because the receiver already contains the dependencies necessary to perform the logic.  But outside of that, there's little precedent for the Collection being the thing that owns item retrieval wholesale.  And, imo, it's a simpler world if we don't extend that ownership.

I'd recommend turning this into a standalone func like `readItemContents()` that accepts, in addition to its other parameters, the oc.cache.  That'll make testing more simple by minimizing coupling between the collection and this behavior.

src/internal/connector/onedrive/collection.go
```diff
@@ -317,12 +323,15 @@ type itemAndAPIGetter interface {
 // downloadContent attempts to fetch the item content.  If the content url
 // is expired (ie, returns a 401), it re-fetches the item to get a new download
 // url and tries again.
-func downloadContent(
+func (oc *Collection) downloadContent(
```
ryanfkeepers, line 326: As with the other comment, I don't see this as being sensibly owned by the Collection itself.  It makes more sense, and will likely be more conveniently portable and testable, if it remains a function, rather than a method, and extends its parameters to include the url cache.

src/internal/connector/onedrive/url_cache.go
```diff
@@ -14,11 +14,20 @@ import (
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
 
+type urlCacher interface {
```
ryanfkeepers, line 17: This naming convention doesn't follow golang standards for interface naming.  Interfaces should be named after their behavior.  `getItemPropertyer`, would be more appropriate in this case.  That naming convention helps to ensure individual interfaces stay small, and can be composed into more complex aggregates as needed.  Starting off with something like `nameOfStructer` inevitably leads to bloated interfaces that duplicate the surface area of the struct, which is a smell.

src/internal/connector/onedrive/collection_test.go
```diff
@@ -600,6 +605,19 @@ func (suite *GetDriveItemUnitTestSuite) TestGetDriveItem_error() {
 	}
 }
 
+var _ urlCacher = &mockURLCache{}
```
ryanfkeepers, line 608: Is it possible to move this to `onedrive/mocks`?  We want to start thinking about portability for mocks, whenever possible.

src/internal/connector/onedrive/collection_test.go
```diff
@@ -611,6 +629,12 @@ func (suite *GetDriveItemUnitTestSuite) TestDownloadContent() {
 
 	itemWID.SetId(ptr.To("brainhooldy"))
 
+	m := &mockURLCache{
+		Get: func(ctx context.Context, itemID string) (itemProps, error) {
```
ryanfkeepers, line 633: Seems like this could be the default behavior for a `NewMockURLCache`, instead of needing the caller to add it, and risk a nil pointer failure as a gotcha.

src/internal/connector/onedrive/collections.go
```diff
@@ -47,6 +48,12 @@ type folderMatcher interface {
 	Matches(string) bool
 }
 
+const (
```
ryanfkeepers, line 51: These should be declared in `url_cache.go`

src/internal/connector/onedrive/collections.go
```diff
@@ -377,13 +385,30 @@ func (c *Collections) Get(
...
 
 			c.CollectionMap[driveID][fldID] = col
 		}
+
+		numDriveItems := c.NumItems - numPrevItems
+		numPrevItems = c.NumItems
+
+		// Only create a drive cache if there are less than 300k items in the drive.
+		if numDriveItems < urlCacheDriveItemThreshold {
+			logger.Ctx(ictx).Info("adding url cache for drive ", driveID)
```
ryanfkeepers, line 402: Please avoid adding dynamic values to the log.  DriveID should already be embedded in the ctx.
```suggestion
			logger.Ctx(ictx).Info("adding url cache for drive")
```

src/internal/connector/onedrive/collections_test.go
```diff
@@ -2665,3 +2666,83 @@ func (suite *OneDriveCollectionsUnitSuite) TestCollectItems() {
...
+
+				c.CollectionMap[driveID][strconv.Itoa(i)] = coll
+				require.Equal(t, nil, coll.cache, "cache not nil")
+			}
+
+			err := c.addURLCacheToDriveCollections(
+				ctx,
+				driveID,
+				fault.New(true))
+
```
ryanfkeepers, line 2730:
```suggestion
```

src/internal/connector/onedrive/collections_test.go
```diff
@@ -2665,3 +2666,83 @@ func (suite *OneDriveCollectionsUnitSuite) TestCollectItems() {
...
+			require.NoError(t, err, clues.ToCore(err))
+
+			// Check that all collections have the same cache instance attached
+			// to them
+			var uc *urlCache
```
ryanfkeepers, line 2735: Does this need to be the underlying type, or could it be the interface?  The latter seems more resilient across change than type-casting the implementing struct from the coll.cache.

src/internal/connector/onedrive/collections.go
```diff
@@ -445,6 +471,32 @@ func (c *Collections) Get(
 	return collections, canUsePreviousBackup, nil
 }
 
+// addURLCacheToDriveCollections adds a URL cache to all collections belonging to
+// a drive.
+func (c *Collections) addURLCacheToDriveCollections(
```
ryanfkeepers, line 476:
```suggestion
func (c *Collections) setURLCache(
```

ryanfkeepers (approved):
Lots of minor things for design/naming that I'd like to see cleaned up.  But the overall logic and behavior seems sound, so approving to unblock.

src/internal/connector/onedrive/collections.go
pandeyabs: Yeah! This is merge aftermath.

src/internal/connector/onedrive/collections.go
pandeyabs: I prefer the original. It's more explicitly named. Also, `setURLCache` might also mean a (get)/set on URL cache.

src/internal/connector/onedrive/collection.go
pandeyabs: `readItemContents` sounds good. 

> Re: Recommendation on turning it into a standalone function

Hmm, I am torn on this. I see url cache as a property of the collection ( or drive, to be precise). My idea behind `readFromCache` was that it acts as the url cache client and reads from cache, if one is attached to a collection.

Passing the cache around as an argument just seems wrong for some reason. Question to ask is whether we'll ever access the url cache without having a collection(keeping aside tests). 

Decoupling cache from collection can make testing slightly simpler, since it narrows the scope and won't need us to define collections every time we test url cache. 

Let me think a bit more about it. I think either way is fine at the risk of over discussion. If we are going to use static, standalone pattern here, we might have to extend that towards other methods defined in `Collection`.

ryanfkeepers: > Question to ask is whether we'll ever access the url cache without having a collection(keeping aside tests).

mmm, I think that's the wrong question to ask.  The logical claim that question supposes would be: "if A contains B, and a function uses B, then the function should use A."  But the function does not necessarily use A.  It may only use B, and should be able to work independent of A.  By that same reasoning, any top-level client or controller must be passed down into every leaf function (or leaf functions should be methods on the controller), purely because the controller contains the things those leaf functions use. 

This isn't exactly [dependency inversion](https://stackify.com/dependency-inversion-principle/), but SOLID principles feels topical here at large.

pandeyabs: Thanks for the explanation. This makes sense to me. 
Side question: Should i move the `urlCacher` ( or `getItemPropertyer`) into `BackupHandler`? Seems like a natural fit since the url cache is only relevant during backups.

pandeyabs:
Converting to draft since this needs some refactoring

src/internal/connector/onedrive/collections_test.go
pandeyabs: I used the underlying type to confirm that the same URL cache instance is shared by all collections.

src/internal/connector/onedrive/collection_test.go
pandeyabs: I'll move this to mock folder in a later PR

ryanfkeepers (approved):

PR #3543, opened 2023-05-31
post-rebase fixups

ryanfkeepers:
catching up changes to the latest main
commit, including fixes for changes that
added new code.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #1996

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/connector/onedrive/item.go
+++ b/src/internal/connector/onedrive/item.go
@@ -108,18 +108,18 @@ func downloadItemMeta(
 // TODO: @vkamra verify if var session is the desired input
 func driveItemWriter(
 	ctx context.Context,
-	rh RestoreHandler,
+	nicu NewItemContentUploader,
 	driveID, itemID string,
 	itemSize int64,
 ) (io.Writer, error) {
 	ctx = clues.Add(ctx, "upload_item_id", itemID)
 
-	r, err := rh.ItemPoster().PostItem(ctx, driveID, itemID)
+	icu, err := nicu.NewItemContentUpload(ctx, driveID, itemID)
 	if err != nil {
 		return nil, clues.Stack(err)
 	}
 
-	iw := graph.NewLargeItemWriter(itemID, ptr.Val(r.GetUploadUrl()), itemSize)
+	iw := graph.NewLargeItemWriter(itemID, ptr.Val(icu.GetUploadUrl()), itemSize)
 
 	return iw, nil
 }
--- a/src/internal/connector/onedrive/permission.go
+++ b/src/internal/connector/onedrive/permission.go
@@ -129,11 +129,16 @@ func computeParentPermissions(
 	}
 }
 
+type postDeleteItemPermissioner interface {
+	DeleteItemPermissioner
+	UpdateItemPermissioner
+}
+
 // UpdatePermissions takes in the set of permission to be added and
 // removed from an item to bring it to the desired state.
 func UpdatePermissions(
 	ctx context.Context,
-	rh RestoreHandler,
+	pdip postDeleteItemPermissioner,
 	driveID string,
 	itemID string,
 	permAdded, permRemoved []metadata.Permission,
@@ -157,7 +162,7 @@ func UpdatePermissions(
 			return clues.New("no new permission id").WithClues(ctx)
 		}
 
-		err := rh.ItemPermissionDeleter().DeleteItemPermission(
+		err := pdip.DeleteItemPermission(
 			ictx,
 			driveID,
 			itemID,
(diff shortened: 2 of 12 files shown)


src/internal/connector/onedrive/permission.go
```diff
@@ -129,11 +129,16 @@ func computeParentPermissions(
 	}
 }
 
+type postDeleteItemPermissioner interface {
```
pandeyabs, line 132: nit: Interface name is slightly misleading since it both deletes and updates item permisssions. How about `postItemPermissionUpdater`?

Also, on a side note, I feel like we should decouple update from delete in `UpdatePermissions`. Otherwise we will have niche interfaces such as this one. Maybe it's just that I am new to go, but it's a significant cognitive load keeping track of multiple interface names which serve a niche purpose.

src/internal/connector/onedrive/item.go
```diff
@@ -108,18 +108,18 @@ func downloadItemMeta(
 // TODO: @vkamra verify if var session is the desired input
 func driveItemWriter(
 	ctx context.Context,
-	rh RestoreHandler,
+	nicu NewItemContentUploader,
```
pandeyabs, line 111: Could you please explain again how this is better than `rh RestoreHandler`? Applies to all similar changes in this PR. I feel like it's a lot of cognitive load tracking all the new interfaces.  

If the added benefit is mocking, i think we could have that effect with `RestoreHandler` itself, right?

pandeyabs (changes requested):

src/internal/connector/onedrive/item.go
ryanfkeepers: Lots of ink is already spilled on the topic, so I won't rehash much.  Here's a quick read for it: https://www.integralist.co.uk/posts/go-interfaces/#keep-interfaces-small.  And here: https://betterprogramming.pub/a-comprehensive-guide-to-interfaces-in-go-6a945b795db#1fbe.  

Long and short of the philosophy: a function should request only the interface it uses, not the interface that the caller is able to supply.  If it's bad to import packages that aren't utilized, and it's bad to declare variables that aren't used, shouldn't it be bad to declare methods that aren't called?

src/internal/connector/onedrive/permission.go
ryanfkeepers: mmm, have to disagree.  It's a bit clunky as a name, but this interface composition is ultimately the same thing you'll see many places elsewhere in golang code.  Standard example, the io package: https://pkg.go.dev/io#ReadWriter.  Though, you're right, it doesn't have a post method.  Probably should be named `updateDeleteItemPermissioner`.  Or maybe `itemPermissionUpdaterdeleter`.

> Otherwise we will have niche interfaces such as this one.

Niche interfaces, especially those that are compositions of other more standardized interfaces, are an ideal, not a smell.  If a function needs to both delete and upload data in order to complete its process, and a single client is expected to supply both actions, then the shape of the interface required for that func is specifically those two methods: no more and no less.

src/internal/connector/onedrive/item.go
pandeyabs: Thanks for the explanation! Those were excellent articles. I agree with your points, just feel that it can limit readability a bit. Maybe that's just me being a noob go enthusiast. I'll get there.

src/internal/connector/onedrive/permission.go
pandeyabs: `updateDeleteItemPermissioner` sounds good

pandeyabs (approved):
🚢

src/internal/connector/onedrive/item.go
ryanfkeepers: >  Maybe that's just me being a noob go enthusiast.

Nah, this is one of those bikeshedding, endless debate areas for sure.

> If the functions in question were super complex, what you said would make total sense. For smaller get/post type functions, having niche interfaces still feels like an overkill.

I don't disagree, but the principle is very different from its enforcement.  There's no clear way to draw the line between "simple enough" and "complex enough" functions.  That just leads to fuzzy, haphazard implementation.  What _is_ enforceable is to adhere to one side or the other: always declare interfaces that match the needs of the function, or never do it.

Personally, I go with the former, because I feel strongly that it leads to better code at large scale: clearer cognitive boundaries, fewer arbitrary code injections, less hacks, better ease of testing.  Big interfaces, weak abstractions vs small interfaces, strong abstractions, and all that jazz.

pandeyabs: ✍️ Thanks @ryanfkeepers

Issue #3498, opened 2023-05-24
[Bug]: Exchange swallows attachment serialization errors


ryanfkeepers:
### What happened?

https://github.com/alcionai/corso/blob/be4032aec9b19b0c12cdc603e97c9be69d6cc01a/src/internal/connector/exchange/service_restore.go#L300

Exchange item attachment processing swallowing errors instead of reporting them.  When we attempted to remove this code, we started getting "could not deserialize payload" errors in our integration tests.  This implies we might be defaulting on uploads in prod, but not tracking the failures.

Investigation into the case is needed.  There's no explanation for skipping the error or why the code is there in the first place.  Feels like something we should remove, and update the tests to match.

### Corso Version?

Corso v0.8.0

### Where are you running Corso?

any

vkamra:
This was introduced in https://github.com/alcionai/corso/commit/a4b50a1ec05cd44bb361a861b0203dd8dc9a1c92

ashmrtn:
this now seems to be causing failures for `TestRestoreIntgSuite/TestRestoreExchangeObject/Test_Mail:_Item_Attachment_Event` if the line in question is removed. I'm unable to create an attachment of the type in the test so I'm unable to manually reproduce the issue outside the test. The test does consistently fail with a 500 internal server error though

Issue #2425, opened 2023-02-07
GC: itemAttachment Support


dadams39:
Item attachments hold `items` that can be any `Outlookable` object provided by M365. The specific supported list for Outlook Mail can be found [here](https://learn.microsoft.com/en-us/exchange/client-developer/exchange-web-services/folders-and-items-in-ews-in-exchange#items).

Can be closed when `itemAttachments` of the the following types are supported
- [ ] Events
- [ ] Mail
- [ ] Contacts
- [ ] ContactGroup
- [ ] PostItem
- [ ] Task

Until this ticket is closed, the solution provided by <name> will

dadams39:
- [ ] https://github.com/alcionai/corso/issues/2428 
Reconnect after https://github.com/microsoft/kiota-serialization-json-go/issues/61 is patched

PR #3554, opened 2023-06-01
pass location to attachment only if value present

neha-Gupta1:
Scenario - Mail restoration with ItemAttachement(event as attachement to mail)
Set the location for restoration only if not nil.
The same seems to be supported for contacts here - https://github.com/alcionai/corso/pull/2465/files and all other items here - https://github.com/alcionai/corso/issues/2425

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3498

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/exchange/transform.go
```diff
@@ -292,6 +297,19 @@ func sanitizeEvent(orig models.Eventable) (models.Eventable, error) {
 	return newEvent, nil
 }
 
+func isNil(i interface{}) bool {
```
neha-Gupta1, line 300: Is there any existing function in out code which checks for nil interface?

ryanfkeepers, line 300: use `any` in place of `interface{}`
```suggestion
func isNil(i any) bool {
```

ryanfkeepers: nope.  IMO, you should add this to `ptr` for widespread usage.

src/internal/connector/exchange/restore_test.go
```diff
@@ -217,36 +217,38 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
-			category: path.EmailCategory,
-			destination: func(t *testing.T, ctx context.Context) string {
-				folderName := tester.DefaultTestRestoreDestination("mailnestattch").ContainerName
-				folder, err := suite.ac.Mail().CreateMailFolder(ctx, userID, folderName)
-				require.NoError(t, err, clues.ToCore(err))
-
-				return ptr.Val(folder.GetId())
-			},
-		},
+		//TODO: Neha complete as part of https://github.com/alcionai/corso/issues/2428
```
ryanfkeepers, line 221: is this PR still a work in process, then?  If so, you should move it to `draft`.

neha-Gupta1: No. Actually we don't handle the nested attachment as of now. I will take that issue next. Till then we need to comment out the nested attachment test cases.  
I added a todo to make sure the test cases are taken care of in the next issue.

PR #3560, opened 2023-06-02
Retry on econnreset regardless of http response

pandeyabs:
Retry on `ECONNRESET` even if HTTP response is set to 2xx. This is unlikely but it can happen. Currently we fail the backup for this scenario with `connection reset by peer` error.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/connector/graph/middleware.go
```diff
@@ -209,13 +209,23 @@ func (mw RetryMiddleware) Intercept(
...
 
 	resp, err := pipeline.Next(req, middlewareIndex)
-	if err != nil && !IsErrTimeout(err) && !IsErrConnectionReset(err) {
+
+	// Alternate approach
+	retriable := IsErrTimeout(err) || IsErrConnectionReset(err) ||
+		(resp != nil && (resp.StatusCode/100 == 4 || resp.StatusCode/100 == 5))
+
+	if !retriable {
 		return resp, stackReq(ctx, req, resp, err)
```
pandeyabs, line 218: @ryanfkeepers Can I get your thoughts on this? :) I'm seeing `200, no retries` test fail with this alternate approach. 
Even though `stackReq` returns nil if err is nil, for some reason, the error check in middleware_test.go, line 153 fails. Happy to provide more context

Issue #3562, opened 2023-06-02
Advance Restore Options v1


ryanfkeepers:
## Issue
Work tracker for the first pass on introducing advanced restore options to corso.

V1 addition include:
* Item Collision Handling
* In-place restore
* Restore to alternative resource

## Work Tracker
- [x] Refactor RestoreConfig struct to own restore configuration.
- [x] oneDrive collision handling.
- [x] exchange collision handling.
- [x] follow-up: event instances
- [x] oneDrive in-place restore.
- [x] exchange in-place restore.
- [x] CLI flags for in-place restore.
- [x] https://github.com/alcionai/corso/issues/3382
- [x] oneDrive restore-to-resource.
- [x] exchange restore-to-resource
- [x] CLI flags for restore-to-resource.

ryanfkeepers:
Graph api automated handling of file conflicts is either buggy or non-functional for drive items.  I've opened a ticket here: https://github.com/OneDrive/onedrive-api-docs/issues/1702

PR #3563, opened 2023-06-02
cetnralize restoreConfig

ryanfkeepers:
centralizes all restore configuration management within a restoreConfig struct.  This struct is owned by the control package, which allows it to be utilized by both CLI and SDK consumers.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/control/options.go
```diff
@@ -48,44 +47,50 @@ func Defaults() Options {
...
-type RestoreDestination struct {
-	// ResourceOwnerOverride overrides the default resource owner to restore to.
-	// If it is not populated items should be restored under the previous resource
-	// owner of the item.
-	ResourceOwnerOverride string
-	// ContainerName is the name of the root of the restored container hierarchy.
-	// This field must be populated for a restore.
-	ContainerName string
+// RestoreConfig contains
+type RestoreConfig struct {
```
ashmrtn, line 68: what's the long-term plan for this? Based on the struct it looks like we kind of plan to always pass this down to the restore consumer. However I still feel there's a benefit to separating the logic of generating the location something is restored at (assuming no collisions) from the logic that actually restores items/handles collisions

src/pkg/control/options.go
```diff
@@ -48,44 +47,50 @@ func Defaults() Options {
...
+	// Defaults to empty.
+	Drive string
 }
 
-func DefaultRestoreDestination(timeFormat dttm.TimeFormat) RestoreDestination {
-	return RestoreDestination{
-		ContainerName: defaultRestoreLocation + dttm.FormatNow(timeFormat),
+func DefaultRestoreConfig(timeFormat dttm.TimeFormat) RestoreConfig {
+	return RestoreConfig{
+		OnCollision: Skip,
```
ashmrtn, line 92: is this a safe default? It's different than the behavior we have currently. What we have now is closes to copy I think

src/pkg/control/options.go
```diff
@@ -48,44 +47,50 @@ func Defaults() Options {
...
 
-// ---------------------------------------------------------------------------
-// Restore Destination
-// ---------------------------------------------------------------------------
+// CollisionPolicy describes how the datalayer behaves in case of a collision.
+type CollisionPolicy string
 
 const (
-	defaultRestoreLocation = "Corso_Restore_"
+	Unknown CollisionPolicy = ""
```
ashmrtn, line 61: do we ever plan on printing the restore policy variable? If we do it seems like this could cause confusion because all the others will have strings that represent the const but this will be empty

src/pkg/control/options.go
ashmrtn: also, not strictly related to this PR or this chunk of code, but we should probably make a validation function for the restore code to check these fields. Otherwise we could pass bad state down into the restore code

src/pkg/control/options.go
ryanfkeepers: We don't technically have a default, since we always create a new folder, and therefore there can be no collisions.  But, yes, this an appropriately safe default.

src/pkg/control/options.go
ryanfkeepers: I don't think it's unreasonable to equate empty to unset/unknown.

ashmrtn: I'm still a little hesitant about this. Having an exported field be a string where the the empty value is not a valid option (it really shouldn't be) can be error prone. If callers don't realize there's a function to generate a default set of options or if they want some non-default set of options it becomes easy to forget to set the field to something useful. At least with ints it's possible to associate a meaningful default to the field

ashmrtn (approved):

src/internal/operations/restore.go
```diff
@@ -35,11 +35,11 @@ import (
 type RestoreOperation struct {
 	operation
 
-	BackupID    model.StableID             `json:"backupID"`
-	Destination control.RestoreDestination `json:"destination"`
-	Results     RestoreResults             `json:"results"`
-	Selectors   selectors.Selector         `json:"selectors"`
-	Version     string                     `json:"version"`
+	BackupID   model.StableID
```
pandeyabs, line 38: Curious, why did we remove struct tags for `RestoreOperation`?

pandeyabs (approved):
Looks good overall. Left one comment, just for my own curiosity.

src/pkg/control/options.go
ryanfkeepers: `Unknown = iota` is functionally no different.  In the abstract, both cases rely on the zero value to denote an unspecified behavior, and can hit that result in all the same ways.  Likewise, the interpretation of a zero-value int to be something meaningful a naming convention, not something related to its value.  Ultimately, it's all in the handling.

src/internal/operations/restore.go
ryanfkeepers: We don't serialize the model.  The declaration was useless.

src/pkg/control/options.go
ashmrtn: >interpretation of a zero-value int to be something meaningful a naming convention, not something related to its value

I kind of have to push back on this point. Because strings are often printed as-is, having an empty string be associated with some defined behavior that matches a non-empty string description (i.e. default is "skip" but we interpret "" as "skip") doesn't make sense. Ints kind of escape that by having a layer of indirection between the value/behavior association and the printed name. Thus it's much easier and less confusing to associate 0 (or the zero-value) with a meaningful behavior

It also is something associated with the value, not the naming convention. By defining a set of iota "enums" (since golang doesn't technically have enums) one is associated behavior with specific values. The naming conventions certainly help humans determine which behavior is associated with each value, but one could just as easily pass in (possibly type cast) ints or name all the enums something like `enumVal0`, `enumVal1`... and the code would continue to work

ryanfkeepers: > i.e. default is "skip" but we interpret "" as "skip") doesn't make sense.

Fair, I'll give you that one.  In a nitpicky way, we're _falling back_ to skip, not equating empty to skip.  But that argument doesn't clarify the confusion as pointed out.  Its a matter of communication around default behavior and expectations, yeah?

On the other hand, printing iota stringers is only helpful you get to leverage the stringer interface.  In cases of serialization you're left with json like `"{ typeEnum: 3 }"`, and that's not helpful for human evaluation even when properly assigned.

> but one could just as easily pass in (possibly type cast) ints or name all the enums something like enumVal0, enumVal1... and the code would continue to work

Yeah, but one could do that with strings just as easily.  In neither case does the underlying type guarantee usage protections.  It's all "which of these matches in the switch statement". 

I think what I'm taking out of this is that, independent of the underlying type, we want to control for default assignment within structs like this, so that even if the caller passes in a non-standard value, we at least set the property hold a value that matches our expectations.

PR #3578, opened 2023-06-08
fetch repoID from kopia is not found in config

neha-Gupta1:
Some events in have a empty repoID. This could be because of multiple config files used. And new config files might have repoID missing. 
Solution- Fetch repoID from Kopia is not found locally in the config file.

Note: Since the Corso Start event happens before we connect to Kopia, it might have repoID missing if the config file is used for the first time. All the consecutive events will have correct values. Corso start event will start populating correct values if the config file is used once.  

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3388

#### Test Plan

- [ ] :muscle: Manual


src/cli/utils/utils.go
```diff
@@ -32,14 +32,25 @@ func GetAccountAndConnect(ctx context.Context) (repository.Repository, *account.
...
 	if err != nil {
 		return nil, nil, clues.Wrap(err, "Failed to connect to the "+cfg.Storage.Provider.String()+" repository")
 	}
 
+	s3Config, errS3Config := cfg.Storage.S3Config()
+	m365Config, errm365Config := cfg.Account.M365Config()
+
+	// repo config is already set while repo connect and init. This is just to confirm correct values.
+	// So won't fail is the write fails
+	err = config.WriteRepoConfig(ctx, s3Config, m365Config, r.GetID())
```
ryanfkeepers, line 48: Writes occuring within a Get func creates a side effect and a smell.  It will be much cleaner if the two behaviors are separated, giving the caller clear protection against accidental mutation.  That same caller can handle overwriting the value as needed.  Or alternatively, a higher level func can replace the get call, and the get and write can be handled separately.

src/cli/utils/utils.go
```diff
@@ -32,14 +32,25 @@ func GetAccountAndConnect(ctx context.Context) (repository.Repository, *account.
...
+	if err != nil || errS3Config != nil || errm365Config != nil {
+		logger.CtxErr(ctx, err).Info("Failed to write repository configuration")
```
ryanfkeepers, line 50: Multiple error checking is a nonstandard pattern for golang and will cause problems.  For example, logging `ctxErr(ctx, err)` when err is nil and errS3Config is non-nil will still log a nil error.  Errors should be handled at the time of their production, not stacked up for an aggregated check.

src/cli/repo/s3.go
```diff
@@ -193,7 +194,7 @@ func connectS3Cmd(cmd *cobra.Command, args []string) error {
 
 	repoID := cfg.RepoID
 	if len(repoID) == 0 {
-		repoID = "not_found"
+		repoID = events.RepoID
```
ryanfkeepers, line 197: This is an opaque application of a const, since it reads like you're providing a valid RepoID, not the `"not_found"` string.  A clearer implementation would be to add a const like `RepoIDNotFound`, and use that in these locations instead.  This pattern would also make it easier to understand checks like on line 232 in repository.go, which is currently confusing.

ryanfkeepers (changes requested):
Some design concerns, described in the comments.  Also, these changes feel like they should have unit or mocked integration tests for proof.

src/cli/utils/utils.go
```diff
@@ -24,34 +24,53 @@ const (
...
 
-	s3Config, errS3Config := cfg.Storage.S3Config()
-	m365Config, errm365Config := cfg.Account.M365Config()
+	return r, &cfg.Storage, &cfg.Account, nil
+}
+
+func AccountConnectAndWriteRepoConfig(ctx context.Context) (repository.Repository, *account.Account, error) {
+	r, stg, acc, err := GetAccountAndConnect(ctx)
+	if err != nil {
+		logger.CtxErr(ctx, err).Info("Failed to  get and connect account")
```
ryanfkeepers, line 49: Per [logging standards](https://github.com/alcionai/corso/blob/main/src/pkg/logger/example_logger_test.go), avoid prefixes like "failed to" in log messages.
```suggestion
		logger.CtxErr(ctx, err).Info("connecting to repo")
```

src/cli/utils/utils.go
```diff
@@ -24,34 +24,53 @@ const (
...
+		return nil, nil, nil
```
ryanfkeepers, line 50: `(nil, nil)` responses for functions that return `(struct, error)` are an anti-pattern.  The assumption here is that the caller should infer the nil response as an error or unusable state.  But the lack of error makes it unclear why the repository is not provided or is unusable.  It'd be more correct here to return the errors.  Or, if we can continue even if the error is nil, we should return a non-nil repository and account.  One or the other of the return values can be nil, but not both.

ryanfkeepers (approved):

neha-Gupta1:
My habit of merging manually :( Forgot to add the `merge-queue` flag

Issue #3522, opened 2023-05-26
Improve configuration handling of parameters, envars, config file


gmatev:
Context: The current scheme is not super clear and has caused some concerns with users who are struggling to change key parameters that are anchored to envars or config file.

**Proposal: ** -  Use a hierarchical model of overrides along the lines of what is available for the AWS CLI - https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html. In order of precedence:
* Explicit flags
* Envars
* Config file 
* (For storage creds only)
   * Credential provider
   * Instance role

For the above to be effective we also need to:
* Expand the set of parameters available as flags - this should cover everything available in ENVARS and config, in addition to other things that may be flags only. Needs a deeper audit but all evars (including what is users for sensitive items) should have flags
* Available envars can stay as is. Add envar to control config file location. 
* Config file - this may need to be expanded to capture mode flags. At the very least all flags and envars that are required for `init` and/or `connect` today.

Config file and secrets:
* Users can specify a secret in config file (e.g. app client secret, or storage creds), but Corso will never implicitly persist such values on behalf of users 

Corso persisting config file automatically:
* If no file exists, any non-sensitive parameters that are effectively used (including defaults) for init or connect are persisted in the corso config file location. Should include user facing output that the config file is created
* If file exists, parameters follow the precedence rules above, but there are some open questions around what should be overwritten.

Open questions:
* Should an overwrite be allowed on `init` if config exists? If yes, this needs to include requirement to change all parameters that are related to a given repo - all app config settings, bucket settings, passphrase. Any sensitive settings need to be deleted - seems complicated
* Connect is similar to `init` if the settings are actually different

One option is to show an error if the config file exists (with conflicting params) and allow an overwrite based on something like `--overwrite-config`. For simplicity this is treated as just deleting the file and persisting (after validation) all effective settings when `init` or `connect` is executed successfully

neha-Gupta1:
@gmatev need a little more explanation - `For storage creds only`-
- Credential provider
- Instance role
Are we thinking of introducing IAM here? And have roles based accesses?

neha-Gupta1:
Tracked in - https://github.com/alcionai/corso/issues/3586

neha-Gupta1:
Do we want to update `config file` values in case of repo Init command?

neha-Gupta1:
The config vars in - env, flags and config file is done

PR #3590, opened 2023-06-12
add flags for azure and aws

neha-Gupta1:
Flags for all configs-

Azure cred flags- (azure-tenant-id, azure-client-id, azure-client-secret) present in -
- Backup (create, delete, details and list) and restore of Exchange, Onedrive and Sharepoint command
-  S3 repo init and connect command

AWS cred flags - (aws-access-key, aws-secret-access-key, aws-session-token) present in- 
- Backup (create, delete, details and list) and restore of Exchange, Onedrive and Sharepoint command
-  S3 repo init and connect command

Passphrase flag- (--passphrase) present in- 
- Backup (create, delete, details and list) and restore of Exchange, Onedrive and Sharepoint command
-  S3 repo init and connect command

S3 flags- 
--endpoint, --prefix, --bucket, --disable-tls, --disable-tls-verification - flags is for repo init and connect commands
all the S3 env var will also work only in case of repo init and connect command. For all other commands  user first connects to repo. Which will store the config values in config file. And then user can use that config file for other commands.

No cred configs are save in the config file by Corso. 

Config file values added- 
Azure cred - 
- azure_client_id 
- azure_secret 
- azure_tenantid

AWS cred -
- aws_access_key_id
- aws_secret_access_key
- aws_session_token

Passphrase -
- passphrase

**NOTE:** 
- in case of AWS creds all the three values should be provided from same method. Either put all values in env, config file and so on.
- all the S3 env var will also work only in case of repo init and connect command. For all other commands  user first connects to repo. Which will store the config values in config file. And then user can use that config file for other commands.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* https://github.com/alcionai/corso/issues/3522

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/options/options.go
```diff
@@ -50,10 +51,40 @@ var (
...
+
 // AddGlobalOperationFlags adds the global operations flag set.
 func AddGlobalOperationFlags(cmd *cobra.Command) {
 	fs := cmd.PersistentFlags()
 	fs.BoolVar(&noStatsFV, NoStatsFN, false, "disable anonymous usage statistics gathering")
+
+	// Flags addition ordering should follow the order we want them to appear in help and docs:
+	// More generic and more frequently used flags take precedence.
+
+	// S3 flags
```
ryanfkeepers, line 74: s3 and m365 are not global implementations.  In the future we will inevitably support other systems, and the declaration of these flags within a global set will cause more confusion/errors than solution.  Another reason to move these declarations to a different owner.

src/pkg/credentials/m365.go
```diff
@@ -12,6 +13,12 @@ const (
 	AzureClientSecret = "AZURE_CLIENT_SECRET"
 )
 
+// M355 creds from flags
+var (
+	AClientID     string
```
ryanfkeepers, line 18: `A` is not a sufficient prefix.  Shortening variable names like this causes more confusion than it solves.

Besides, it seems like what you're actually doing here is avoiding collision with the similarly named consts in order to export global variables.  Exporting globals vars, especially within a package whose entire purpose is to make those variables safely portable by passing them around inside a struct, produces a smell.  Is it necessary to approach the solution this way?

src/cli/options/options.go
```diff
@@ -50,10 +51,40 @@ var (
...
+	fs.StringVar(&Prefix, "prefix", "", "Repo prefix within bucket.")
+	fs.StringVar(&Endpoint, "endpoint", "s3.amazonaws.com", "S3 service endpoint.")
+	fs.BoolVar(&DoNotUseTLS, "disable-tls", false, "Disable TLS (HTTPS)")
+	fs.BoolVar(&DoNotVerifyTLS, "disable-tls-verification", false, "Disable TLS (HTTPS) certificate verification.")
+	fs.StringVar(&AccessKey, "access-key", "", "S3 access key")
+	fs.StringVar(&SecretAccessKey, "secret-access-key", "", "S3 access key")
+	fs.StringVar(&SessionToken, "session-token", "", "S3 session token")
+
+	// M365 flags
+	fs.StringVar(&credentials.AClientID, "client-id", "", "Azure app client ID")
```
ryanfkeepers, line 85: Sidestepping our current configuration communication patterns to have flags directly mutate global variables in lower level packages is an anti-pattern.

src/pkg/credentials/aws.go
```diff
@@ -22,13 +20,18 @@ type AWS struct {
...
+	var accessKey, secretKey, sessToken string
 	if ovr, ok := override[AWSAccessKeyID]; ok && ovr != "" {
 		accessKey = ovr
 	}
 
-	secretKey := os.Getenv(AWSSecretAccessKey)
-	sessToken := os.Getenv(AWSSessionToken)
+	if ovr, ok := override[AWSSecretAccessKey]; ok && ovr != "" {
+		secretKey = ovr
+	}
```
ryanfkeepers, line 30: the `ok` check on these is unnecessary.  Either they're going to get replaced by a value, or they're going to remain as empty strings.  
```suggestion
	secretKey := override[AWSSecretAccessKey]
```

src/pkg/credentials/corso.go
```diff
@@ -4,23 +4,29 @@ import (
...
 	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/common/str"
 )
 
 // envvar consts
 const (
 	CorsoPassphrase = "CORSO_PASSPHRASE"
+	// Corso Flags
+	CorsoPassphraseFN = "passphrase"
```
ryanfkeepers, line 15: Anything related to flags is purely the concern of the CLI, and should not appear in other layers.  This value, the FV below it should be moved to `cli/utils.go`.  That value can then be passed into this func.

src/pkg/credentials/m365.go
```diff
@@ -12,6 +14,12 @@ const (
 	AzureClientSecret = "AZURE_CLIENT_SECRET"
 )
 
+var (
+	AzureClientTenantFV string
```
ryanfkeepers, line 18: Same thing here: flag values should not exist below the cli layer.  These declarations should get moved to `cli/utils.go` and passed into this func.

src/pkg/storage/s3.go
```diff
@@ -19,9 +21,12 @@ type S3Config struct {
 
 // config key consts
 const (
+	keyS3AccessKey      = "s3_accessKey"
```
ryanfkeepers, line 24: CamelCase here as well, please.
```suggestion
	keyS3AccessKey      = "s3_access_key"
```

src/pkg/credentials/corso.go
neha-Gupta1: I tried that. But getting stuck due to `cyclic dependency`. credentials and cli seems to be quite dependent packages. Also, out of curiosity - why do we have two packages- `options` and `utils packages' flag.go` handling the flags?

src/cli/restore/onedrive.go
```diff
@@ -32,6 +33,9 @@ func addOneDriveCommands(cmd *cobra.Command) *cobra.Command {
 		fs.SortFlags = false
 
 		utils.AddBackupIDFlag(c, true)
+		utils.AddCorsoPassphaseFlags(c)
```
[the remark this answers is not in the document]
neha-Gupta1: Added test for the same.

src/cli/utils/flags.go
```diff
@@ -67,6 +72,14 @@ const (
 	// Maintenance stuff.
 	MaintenanceModeFN  = "mode"
 	ForceMaintenanceFN = "force"
+
+	AccessKeyFN       = "aws-access-key"
+	SecretAccessKeyFN = "aws-secret-access-key"
+	SessionTokenFN    = "aws-session-token"
+
+	AzureClientTenantFN = "azure-tenant-id"
```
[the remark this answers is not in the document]
neha-Gupta1: these are present in credentials package. And it is due cyclic dependency I was getting when adding those vars in utils package.

src/pkg/credentials/m365.go
neha-Gupta1: As mentioned above. Its because of cyclic dependency.

src/pkg/credentials/corso.go
ryanfkeepers: > But getting stuck due to cyclic dependency. credentials and cli seems to have quite dependent packages.

Well, that should only happen if credentials is importing the CLI, yeah?  Which isn't an import direction we should be supporting anyway.  Higher level layers such as apis (cli, sdk) should import from lower levels (storage, credentials), but not the other way around.

I'd try to track down how the cycle is happening first and see if you can resolve that more directly; potentially by passing flag values in through a func parameter instead of referencing the global instance?  Distributing flag ownership into different packages as a solution to circular imports is not the right way to go about it, since that only injects a workaround, and doesn't solve the underlying cause.

> Also, out of curiosity - why do we have two packages- options and utils packages' flag.go handling the flags?

Yeah, I agree, it's a bit odd.  Feel free to write up a tech debt ticket for something like moving options into utils and we can refactor when we have a moment.

src/cli/backup/exchange_e2e_test.go
```diff
@@ -35,6 +35,184 @@ var (
...
+	acct       account.Account
+	st         storage.Storage
+	vpr        *viper.Viper
+	cfgFP      string
+	repo       repository.Repository
+	m365UserID string
+	recorder   strings.Builder
+}
+
+func TestBackupExchangeInvalidFlagValueSuite(t *testing.T) {
```
ryanfkeepers, line 53: This should match the name of the suite being initialized:
```suggestion
func TestExchangeCMDWithFlagsE2ESuite(t *testing.T) {
```

src/cli/backup/exchange_e2e_test.go
```diff
@@ -35,6 +35,184 @@ var (
...
+	cmd.SetErr(&suite.recorder)
+
+	ctx = print.SetRootCmd(ctx, cmd)
+
+	// run the command
+	err := cmd.ExecuteContext(ctx)
+	require.Error(t, err, clues.ToCore(err))
+}
+
+func (suite *NoBackupExchangeE2ESuite) TestExchangeBackupValueFromConfigCmd_empty() {
```
ryanfkeepers, line 101: this and some others are attached to the wrong suite.

src/cli/backup/exchange_e2e_test.go
```diff
@@ -35,6 +35,184 @@ var (
...
+	require.NoError(t, err, clues.ToCore(err))
+
+	result := suite.recorder.String()
+	t.Log("backup results", result)
+
+	// as an offhand check: the result should contain the m365 user id
+	assert.Contains(t, result, suite.m365UserID)
+}
+
+func (suite *NoBackupExchangeE2ESuite) TestExchangeBackupValueFromEnvCmd_empty() {
```
ryanfkeepers, line 131: These are all just variations on what values we accept in order to run a backup, right?  That'll end up running one complete, end-to-end backup for each test, which isn't wanted.  Lets avoid additional e2e tests that aren't focused on error cases or else we'll bloat ci times pretty quickly.  These sorts of conditions are ideally testable with a mocked interface so that we can evaluate all the logic in the cli layer without making any calls to data producers or consumers.

src/cli/repo/s3.go
```diff
@@ -232,12 +240,25 @@ func connectS3Cmd(cmd *cobra.Command, args []string) error {
...
+		credentials.AWSSessionToken:    utils.AWSSessionTokenFV,
+		storage.Bucket:                 bucket,
+		storage.Endpoint:               endpoint,
+		storage.Prefix:                 prefix,
+		storage.DoNotUseTLS:            strconv.FormatBool(doNotUseTLS),
+		storage.DoNotVerifyTLS:         strconv.FormatBool(doNotVerifyTLS),
+	}
+}
+
+func S3UpdateFromEnvVar(s3Flag map[string]string) map[string]string {
```
ryanfkeepers, line 256: I'm a little concerned about how ownership is getting distributed across multiple packages in this change.  For example, env vars have moved [out of the config processing](https://github.com/alcionai/corso/pull/3590/files#diff-f37f5d414f19e717eb2ca53d62ece0c8fa1f4d85d88a1b3ce1bcc3218bc0775aL90) and here into the command itself.  That's not entirely bad, but I'm not sure what's gained here with the move, since no other packages seem to import `S3UpdateFromEnvVar` (which is correct, imo, no other packages should import it).  Seems to me like this could move back to `GetConfigRepoDetails`, and this func could go away, without any repercussions.

src/pkg/credentials/aws.go
```diff
@@ -21,15 +21,15 @@ type AWS struct {
 // GetAWS is a helper for aggregating aws secrets and credentials.
 func GetAWS(override map[string]string) AWS {
 	var accessKey, secretKey, sessToken string
-	if ovr, ok := override[AWSAccessKeyID]; ok && ovr != "" {
+	if ovr := override[AWSAccessKeyID]; ovr != "" {
```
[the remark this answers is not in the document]
neha-Gupta1: Yes. Got it.

src/cli/repo/s3.go
neha-Gupta1: Some S3 config like - Bucket, Endpoint are to be fetched from `envvar` only in case of `init` and `connect` command. And for all other cases it to to be fetched only from config file. So this override is done only in case of connect and init command.

src/cli/backup/exchange_e2e_test.go
neha-Gupta1: Acknowledged.

src/cli/backup/exchange_e2e_test.go
neha-Gupta1: Can you please point me to one such test case?

src/pkg/credentials/corso.go
neha-Gupta1: With some debugging - The import cycle is between cli's config and cli's util package. 
`import cycle not allowed: import stack: [github.com/alcionai/corso/src/cli/restore github.com/alcionai/corso/src/cli/repo github.com/alcionai/corso/src/cli/config github.com/alcionai/corso/src/cli/utils github.com/alcionai/corso/src/cli/config]`

The thing is - its not a single `cli` package. There are multiple small packages like - config, utils inside it. And which is correct for code structuring. But then there should be a fix route like -  config can import utils or vice-versa.

neha-Gupta1:
@ryanfkeepers please revisit this PR. I have tried addressing all the PR review comments. And have kept all the flag related code in CLI folder.

ryanfkeepers (approved):
Looking good.  Thanks for all the work!

As a very minor nit: there's a lot of changes in this one PR.  It'd be nice to slice up changes across multiple PRs in a chain, even if we end up squashing them all together before they get merged.  That'll be much easier for me to review in smaller chunks.  Thanks!

neha-Gupta1:
> Looking good. Thanks for all the work!
> 
> As a very minor nit: there's a lot of changes in this one PR. It'd be nice to slice up changes across multiple PRs in a chain, even if we end up squashing them all together before they get merged. That'll be much easier for me to review in smaller chunks. Thanks!

The PR changes somehow grew with time. I will make sure to have smaller PRs from next time.

PR #3592, opened 2023-06-12
Don't dump response body while checking for malware

pandeyabs:
Corso detects malware in a few places for OD/SP.

1. During item enumeration, using [malware property](https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0#properties) which is set by graph - [code](https://github.com/alcionai/corso/blob/ed47c134b09a19dd6b4869f605f3a384f5763c06/src/internal/connector/onedrive/collections.go#L658) 

2. During item fetch using download URL - [code](https://github.com/alcionai/corso/blob/ed47c134b09a19dd6b4869f605f3a384f5763c06/src/internal/connector/onedrive/item.go#LL53C18-L53C18)
- Graph seems to return 403 with `{"error":{"code":"malwareDetected","message":"Malware detected"}}`. Tested with eicar.

3. By analyzing http response headers in `IsMalwareResp`  Don't see these headers documented by graph. This is for [sharepoint protocol compatibility](https://learn.microsoft.com/en-us/openspecs/sharepoint_protocols/ms-wdvmoduu/6fa6d4a9-ac18-4cd7-b696-8a3b14a98291) it seems.

4. By analyzing response body in `IsMalwareResp` . We check for malwareDetected string inside http response body.

We are accidentally dumping entire file ( http response dump) while checking for malware in 4. This is leading to high mem utilization/OOMs, especially while processing very large files.

I don't think we need 4 at all. If graph is the entity which detects malware, I think we will never get past 2. This PR removes item 4. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/connector/graph/errors.go
+++ b/src/internal/connector/graph/errors.go
@@ -169,11 +169,9 @@ func IsMalware(err error) bool {
 func IsMalwareResp(ctx context.Context, resp *http.Response) bool {
 	// https://learn.microsoft.com/en-us/openspecs/sharepoint_protocols/ms-wsshp/ba4ee7a8-704c-4e9c-ab14-fa44c574bdf4
 	// https://learn.microsoft.com/en-us/openspecs/sharepoint_protocols/ms-wdvmoduu/6fa6d4a9-ac18-4cd7-b696-8a3b14a98291
-	if resp != nil && len(resp.Header) > 0 && resp.Header.Get("X-Virus-Infected") == "true" {
-		return true
-	}
-
-	return false
+	return resp != nil &&
+		len(resp.Header) > 0 &&
+		resp.Header.Get("X-Virus-Infected") == "true"
 }
 
 func IsErrFolderExists(err error) bool {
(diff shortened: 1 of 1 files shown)


src/internal/connector/graph/errors.go
```diff
@@ -169,23 +167,9 @@ func IsMalware(err error) bool {
...
-	}
-
-	respDump, err := httputil.DumpResponse(resp, true)
-	if err != nil {
-		logger.Ctx(ctx).Errorw("dumping http response", "error", err)
-		return false
-	}
-
-	if strings.Contains(string(respDump), string(malwareDetected)) {
+	if resp != nil && len(resp.Header) > 0 && resp.Header.Get("X-Virus-Infected") == "true" {
```
ryanfkeepers, line 172: `if foo { return true } return false` is redundant, and can be reduced to the conditional check.
```suggestion
	return resp != nil &&
	len(resp.Header) > 0 &&
	resp.Header.Get("X-Virus-Infected") == "true"
```

ryanfkeepers (approved):
one small nit, otherwise lgtm

PR #3607, opened 2023-06-13
Skip any attachment fetches that fail with OLE conversion error

meain:
These files from what we can understand are not available and thus can't be fetched. This change ensures that we don't fail the backup just because of this error.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix Exchange folder cache population error when parent folder isn't found.
 - Fix Exchange backup issue caused by incorrect json serialization
 - Fix issues with details model containing duplicate entry for api consumers
+- Handle OLE conversion errors when trying to fetch attachments
 
 ### Changed
 - Do not display all the items that we restored at the end if there are more than 15. You can override this with `--verbose`.
--- a/src/pkg/fault/item.go
+++ b/src/pkg/fault/item.go
@@ -15,6 +15,7 @@ type itemType string
 
 const (
 	FileType          itemType = "file"
+	AttachmentType    itemType = "attachment"
 	ContainerType     itemType = "container"
 	ResourceOwnerType itemType = "resource_owner"
 )
@@ -248,6 +249,11 @@ func ContainerSkip(cause skipCause, namespace, id, name string, addtl map[string
 // FileSkip produces a File-kind Item for tracking skipped items.
 func FileSkip(cause skipCause, namespace, id, name string, addtl map[string]any) *Skipped {
 	return itemSkip(FileType, cause, namespace, id, name, addtl)
+}
+
+// AttachmentSkip produces a Attachment-kind Item for tracking skipped items.
+func AttachmentSkip(cause skipCause, namespace, id, name string, addtl map[string]any) *Skipped {
+	return itemSkip(AttachmentType, cause, namespace, id, name, addtl)
 }
 
 // OnwerSkip produces a ResourceOwner-kind Item for tracking skipped items.
--- a/src/pkg/fault/item.go
+++ b/src/pkg/fault/item.go
@@ -15,7 +15,6 @@ type itemType string
 
 const (
 	FileType          itemType = "file"
-	AttachmentType    itemType = "attachment"
 	ContainerType     itemType = "container"
 	ResourceOwnerType itemType = "resource_owner"
 )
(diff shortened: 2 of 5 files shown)


src/pkg/fault/item.go
```diff
@@ -15,6 +15,7 @@ type itemType string
 
 const (
 	FileType          itemType = "file"
+	AttachmentType    itemType = "attachment"
```
ryanfkeepers, line 18: I'm not entirely confident about the choice of "attachment" as a type here, since it's specific to certain types of data and not others.  That sets us up to have N different types, instead of one type with N different identifiers.  For example, if this were, I dunno, an `ItemPropertyType`, with a category of `Attachment`, then we only need one interface for all ItemProperty skips, all of which can dynamically declare different categories.  Easier to extend and maintain that way.  (similary, we should probably change "fileType" to "itemType", one category of which can be a File.)

Not entirely sure what to call it.  Maybe "attachment" is sufficient for the short term?  Don't want to block on this, since we should be able to rename in a fast follow without much fuss.  But do help me think on the best name, please.

src/pkg/services/m365/api/mail.go
```diff
@@ -406,6 +407,24 @@ func (c Mail) GetItem(
 			ByAttachmentId(ptr.Val(a.GetId())).
 			Get(ctx, attachConfig)
 		if err != nil {
+			if graph.IsErrCannotOpenFileAttachment(err) {
+				logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipNotFound).Info("attachment not found")
+				el.AddSkip(
+					fault.AttachmentSkip(
+						fault.SkipNotFound,
+						itemID,
```
ryanfkeepers, line 415: This is going to be a problem.  Fault items deduplicate [based on the item id](https://github.com/alcionai/corso/blob/main/src/pkg/fault/item.go#L83).  This skip would be read as "we are skipping the entire item", when you mean for it to indicate that we're skipping just the attachment within that item.  If you can use the attachment ID instead, and add the item ID into the extended map, that would solve the issue.

Also, please coordinate with @juleslasarte about the sdk expectations here.  The introduction of subcomponent skips is a new paradigm, and other frameworks need to handle it gracefully.  As a v1, it might be smarter for us to not include a fault skip at all, and instead log and continue.  That will unblock the fix, and we can investigate the fault skip records as a follow-up.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix Exchange folder cache population error when parent folder isn't found.
 - Fix Exchange backup issue caused by incorrect json serialization
 - Fix issues with details model containing duplicate entry for api consumers
+- Handle OLE conversion errors when trying to fetch attachments
```
ryanfkeepers, line 21: So we'll skip the attachment to progress the backup.  I assume we'll need the item to appear in another delta if we want to retry the attachment download, since we have no other way to target a single item for backup, meaning once skipped, we'll never be able to retrieve the attachment.  That actually sounds like a condition we'd want to list in `Known Issues`.  Have you double checked with Georgi and Vaibhav that we're okay with the permanently missing data here?  Or would it be better to record this situation as a recoverable error?

ryanfkeepers (approved):
Generally looks good.  Some concerns about using fault skips to handle this issue (at least, as a v1), and want to double check that we're happy with the overall behavior.  Approving to unblock, and you can chase those issues in the background.

src/pkg/services/m365/api/mail.go
juleslasarte: > Also, please coordinate with @juleslasarte about the sdk expectations here. The introduction of subcomponent skips is a new paradigm, and other frameworks need to handle it gracefully. As a v1, it might be smarter for us to not include a fault skip at all, and instead log and continue. That will unblock the fix, and we can investigate the fault skip records as a follow-up.

I second this -- to simply skip for now, and add the fault records as a follow up so we don't need to coordinate on the work on action details and this.

meain: I've created https://github.com/alcionai/corso/issues/3620 to track this

CHANGELOG.md
meain: > .. I assume we'll need the item to appear in another delta if we want to retry the attachment download ..
>  Have you double checked with Georgi and Vaibhav that we're okay with the permanently missing data here?

I had a conversation with Vaibhav and we finalized on skipping this for now. We want the backup to go through with this skipped. Similar to malware, it looks like this data will never be accessible from what we could find.

ryanfkeepers: > it looks like this data will never be accessible

Hmm, I didn't catch that there's a sense of permanence here, from the code comments.  Maybe you could add a small note about that somewhere in a comment, so it doesn't get confused for a temporary failure?

PR #3613, opened 2023-06-13
add default restore config normalizer

ryanfkeepers:
Adds a func that normalizes the restore config in case inputs don't match accepted or expected values.

Also moves DefaultTestRestoreConfig out of /tester and into /control/testdata to avoid circular dependency issues.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test


src/cmd/factory/impl/common.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/backup/details"
 	"github.com/alcionai/corso/src/pkg/control"
+	ctrlTD "github.com/alcionai/corso/src/pkg/control/testdata"
```
meain, line 27: Should we just use `testdata` here (same thing in a few other files).

src/internal/m365/controller_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/alcionai/corso/src/internal/tester"
 	"github.com/alcionai/corso/src/internal/version"
 	"github.com/alcionai/corso/src/pkg/control"
+	ctrlTD "github.com/alcionai/corso/src/pkg/control/testdata"
```
meain, line 24: Same questions about `testdata`.

src/internal/m365/exchange/contacts_restore_test.go
```diff
@@ -51,7 +52,7 @@ func (suite *ContactsRestoreIntgSuite) TestCreateContainerDestination() {
 		path.EmailCategory,
 		suite.creds.AzureTenantID,
 		suite.userID,
-		tester.DefaultTestRestoreConfig("").Location,
+		testdata.DefaultRestoreConfig("").Location,
```
meain, line 55: If you wanna go with `ctrlTD`, let's change this as well I guess.

src/pkg/control/options.go
```diff
@@ -94,6 +102,27 @@ func DefaultRestoreConfig(timeFormat dttm.TimeFormat) RestoreConfig {
...
+// EnsureRestoreConfigDefaults sets all non-supported values in the config
+// struct to the default value.
+func EnsureRestoreConfigDefaults(
+	ctx context.Context,
+	rc RestoreConfig,
+) RestoreConfig {
+	if !slices.Contains([]CollisionPolicy{Skip, Copy, Replace}, rc.OnCollision) {
+		logger.Ctx(ctx).
+			With("bad_collision_policy", rc.OnCollision).
+			Info("setting collision policy to default")
```
meain, line 114: Might be worth it to mention what the default is.

src/internal/operations/restore.go
```diff
@@ -68,7 +68,7 @@ func NewRestoreOperation(
 		operation:  newOperation(opts, bus, kw, sw),
 		acct:       acct,
 		BackupID:   backupID,
-		RestoreCfg: restoreCfg,
+		RestoreCfg: control.EnsureRestoreConfigDefaults(ctx, restoreCfg),
```
meain, line 71: Isin't this something that we specify in code? What is the possibility that we get the incorrect value? Also, if we do get incorrect value, wouldn't it be better to error out and force us to fix it instead of silently doing potentially the incorrect thing.

src/pkg/control/options_test.go
```diff
@@ -0,0 +1,83 @@
...
+		input  RestoreConfig
+		expect RestoreConfig
+	}{
+		{
+			name: "populated",
+			input: RestoreConfig{
+				OnCollision:       Copy,
+				ProtectedResource: "batman",
+				Location:          "badman",
+				Drive:             "hatman",
```
meain, line 32: 🤣

src/cmd/factory/impl/common.go
ryanfkeepers: I waffled on that here, and can certainly use the default.  For info, I feel there's now a number of `foo/testdata` packages in play now, so the likelihood of collision is higher.  But I should probably stick to the unaliased version until needed.

src/internal/operations/restore.go
ryanfkeepers: I think the larger likelihood is that we get an unpopulated value, and want to ensure it gets set to a default instead of remaining unpopulated.  Yes, this pattern does silently fix bad values without much alert. But since we 1/ export consts for enums and 2/ don't consider other values "improper", I think the potential ramifications here are at a minimum.  A customer would have to deliberately go out of their way and feed in obviously unsupported data for that to happen, which is not likely.

In other situations, where it could be easy for a customer to accidentally feed unsupported values into the configuration, I agree that returning an error is an appropriate pattern.

src/pkg/control/options.go
```diff
@@ -94,6 +102,29 @@ func DefaultRestoreConfig(timeFormat dttm.TimeFormat) RestoreConfig {
...
+		logger.Ctx(ctx).
+			With(
+				"bad_collision_policy", rc.OnCollision,
+				"default_collision_policy", Skip).
+			Info("setting collision policy to default")
+
+		rc.OnCollision = Skip
+	}
+
+	if strings.TrimSpace(rc.Location) == RootLocation {
```
ashmrtn, line 121: is this valid? Is it possible for users to have a folder (very adversarial) that's "/ "? What would it look like if they gave that on the CLI and it got to this point?

meain (approved):
Approving to unblock, but might wanna look into https://github.com/alcionai/corso/pull/3613/files/b1b21babe51831bc6a9c239f748440d9e9908688#r1229922693

Issue #3503, opened 2023-05-24
Factor out slack notification into reusable action


ashmrtn:
We now have multiple things sending slack notifications when they fail, all with the same format and only slightly different information. We should factor the slack notification stuff out into it's own github action and then just reuse it as needed

ashmrtn:
closing this as PR merged

PR #3614, opened 2023-06-13
centralize the slack message action

ryanfkeepers:
centralizes the ci action for slack messages, and adds slack messages for per-resource cleanup failures.

---

- [x] :no_entry: No

#### Type of change

#### Issues

- #3503

### DIFF
--- a/.github/actions/slack-message/action.yml
+++ b/.github/actions/slack-message/action.yml
@@ -26,10 +26,10 @@ runs:
         echo "proof ${{ github.ref }}"
         echo "test 1 ${{env.github_reference}}"
         echo "test 2 ${github_reference}"
-        echo "test 3 ${github_reference#refs}"
+        echo "test 3 ${github_reference#refs} ${ github_reference#refs} ${github_reference#refs } ${ github_reference#refs }"
         echo "test 4 $github_reference"
-        vvv=${{github.ref}} echo "test 5 ${vvv#refs}"
-        echo trimmed_ref=${IN_REF#refs/} >> $GITHUB_ENV
+        vvv=${{github.ref}} echo "test 5 ${ vvv#refs } ${vvv#refs} ${vvv#refs } ${ vvv#refs}"
+        echo trimmed_ref=${ IN_REF#refs/ } >> $GITHUB_ENV
 
     - name: compose links
       shell: bash
--- a/.github/actions/slack-message/action.yml
+++ b/.github/actions/slack-message/action.yml
@@ -22,20 +22,10 @@ runs:
     - name: trim ref
       shell: bash
       run: |
+        echo testing=${ github.ref#refs/ } >> $GITHUB_ENV
+        echo "test ${{env.testing}}"
         echo IN_REF=${{ github.ref }} >> $GITHUB_ENV
         echo trimmed_ref=${IN_REF#refs/} >> $GITHUB_ENV
-
-    - shell: bash
-      run: |
-        echo "-----"
-        echo "j ${{github.job}}"
-        echo "r ${{github.ref}}"
-        echo "t ${{env.trimmed_ref}}"
-        echo "s ${{github.run_id}}"
-        echo "i ${{github.sha}}"
-        echo "u ${{github.server_url}}"
-        echo "e ${{github.repository}}"
-        echo "-----"
 
     - name: compose links
       shell: bash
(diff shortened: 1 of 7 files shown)


.github/actions/slack-message/action.yml
```diff
@@ -0,0 +1,38 @@
...
+    description: The slack message text
+  run_url:
+    description: normally "${{ GITHUB_REF#refs/heads/ }}-${{ GITHUB_SHA }}"
+  sha:
+    description: normally "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+  slack_url:
+    description: passthrough for secrets.SLACK_WEBHOOK_URL
+
+runs:
+  using: composite
```
ashmrtn, line 18: is this not able to be a reusable workflow? According to some [random blog](https://dev.to/n3wt0n/composite-actions-vs-reusable-workflows-what-is-the-difference-github-actions-11kd) reusable workflows can consume secrets

.github/actions/slack-message/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        payload: |
+          {
+            "text": "${{ inputs.text }} :: build: ${{ inputs.job_status }} - SHA: ${{  inputs.sha_info.outputs.SHA }}",
+            "blocks": [
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "${{ inputs.text }} :: <${{  inputs.sha_info.outputs.RUN_URL }}|[Logs]> <${{ inputs.sha_info.outputs.COMMIT_URL }}|[Base]>\nCommit: <${{  inputs.sha_info.outputs.COMMIT_URL }}|${{  inputs.sha_info.outputs.SHA }}>"
```
ashmrtn, line 31: I think this should be `inputs.msg`?
```suggestion
                  "text": "${{ inputs.msg }} :: <${{  inputs.sha_info.outputs.RUN_URL }}|[Logs]> <${{ inputs.sha_info.outputs.COMMIT_URL }}|[Base]>\nCommit: <${{  inputs.sha_info.outputs.COMMIT_URL }}|${{  inputs.sha_info.outputs.SHA }}>"
```

ryanfkeepers: whoops, thanks!

.github/actions/slack-message/action.yml
ryanfkeepers: It could be.  Personally, I like the separation produced by making smaller, one-off actions composites, and leaving workflows for larger, more involved compositions.

.github/actions/slack-message/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        payload: |
+          {
+            "text": "${{ inputs.msg }} :: build: ${{ inputs.job_status }} - SHA: ${{  inputs.sha_info.outputs.SHA }}",
+            "blocks": [
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "${{ inputs.msg }} :: <${{  inputs.sha_info.outputs.RUN_URL }}|[Logs]> <${{ inputs.sha_info.outputs.COMMIT_URL }}|[Base]>\nCommit: <${{  inputs.sha_info.outputs.COMMIT_URL }}|${{  inputs.sha_info.outputs.SHA }}>"
```
neha-Gupta1, line 31: Is there a sample output of this? Specifically for `ci_test_cleanup`. Thinking about the SHA value in that case.

ryanfkeepers: not at the moment.  I assume it'll link to whatever commit is in main?

.github/actions/slack-message/action.yml
ashmrtn: honestly it seems simpler to have all of one type if the differences between composite and reusable workflows are small. I spent a good 5-10min digging around for decent info on what a composite workflow is just to review this PR. There's no need to cause further confusion especially with there being little documentation on some stuff

ashmrtn (approved):
would be good to see a manual test of this working in the pipeline prior to merge since we rely pretty heavily on this alerting to make sure things are passing

.github/actions/slack-message/action.yml
ryanfkeepers: With such a low rate of change for CI tooling stuffs, I don't think the difference adds real burden for development.  For the most part these things are highly stable, with the odd addition of a composite action.  And the composites now have good boilerplate to work off of, both for creation and usage.

PR #3623, opened 2023-06-14
cascade restoreCfg collision policy into onedrive

ryanfkeepers:
Adds collision policy handling to onedrive item posts.  This allows us to override the default "replace" behavior that currently returns a 409 for the creation endpoint, in case we want to use skip (ie: fail) or copy handling.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/restore.go
+++ b/src/internal/m365/onedrive/restore.go
@@ -708,12 +708,16 @@ func createRestoreFolders(
 		}
 
 		// create the folder if not found
+		// the Replace collision policy is used since collisions on that
+		// policy will no-op and return the existing folder.  This has two
+		// benefits: first, we get to treat the post as idempotent; and
+		// second, we don't have to worry about race conditions.
 		folderItem, err = fr.PostItemInContainer(
 			ictx,
 			driveID,
 			parentFolderID,
 			newItem(folder, true),
-			control.Replace) // folders can be replaced without issue.
+			control.Replace)
 		if err != nil {
 			return "", clues.Wrap(err, "creating folder")
 		}
--- a/src/internal/m365/onedrive/restore.go
+++ b/src/internal/m365/onedrive/restore.go
@@ -76,10 +76,7 @@ func ConsumeRestoreCollections(
 		el             = errs.Local()
 	)
 
-	ctx = clues.Add(
-		ctx,
-		"backup_version", backupVersion,
-		"restore_location", restoreCfg.Location)
+	ctx = clues.Add(ctx, "backup_version", backupVersion)
 
 	// Reorder collections so that the parents directories are created
 	// before the child directories; a requirement for permissions.
@@ -97,7 +94,6 @@ func ConsumeRestoreCollections(
 			ictx    = clues.Add(
 				ctx,
 				"category", dc.FullPath().Category(),
-				"destination", clues.Hide(restoreCfg.Location),
 				"resource_owner", clues.Hide(dc.FullPath().ResourceOwner()),
 				"full_path", dc.FullPath())
 		)
@@ -105,10 +101,10 @@ func ConsumeRestoreCollections(
 		metrics, err = RestoreCollection(
 			ictx,
 			rh,
+			restoreCfg,
 			backupVersion,
 			dc,
 			caches,
-			restoreCfg.Location,
 			deets,
 			opts.RestorePermissions,
 			errs)
@@ -141,12 +137,12 @@ func ConsumeRestoreCollections(
 func RestoreCollection(
 	ctx context.Context,
 	rh RestoreHandler,
+	restoreCfg control.RestoreConfig,
 	backupVersion int,
 	dc data.RestoreCollection,
 	caches *restoreCaches,
-	restoreContainerName string,
 	deets *details.Builder,
-	restorePerms bool,
+	restorePerms bool, // TODD: move into restoreConfig
 	errs *fault.Bus,
 ) (support.CollectionMetrics, error) {
 	var (
@@ -181,7 +177,13 @@ func RestoreCollection(
 	// from the backup under this the restore folder instead of root)
 	// i.e. Restore into `<restoreContainerName>/<original folder path>`
 	// the drive into which this folder gets restored is tracked separately in drivePath.
-	restoreDir := path.Builder{}.Append(restoreContainerName).Append(drivePath.Folders...)
+	restoreDir := &path.Builder{}
+
+	if len(restoreCfg.Location) > 0 {
+		restoreDir = restoreDir.Append(restoreCfg.Location)
+	}
+
+	restoreDir = restoreDir.Append(drivePath.Folders...)
 
 	ctx = clues.Add(
 		ctx,
@@ -280,6 +282,7 @@ func RestoreCollection(
 				itemInfo, skipped, err := restoreItem(
 					ictx,
 					rh,
+					restoreCfg,
 					dc,
 					backupVersion,
 					drivePath,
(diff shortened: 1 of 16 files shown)


src/internal/m365/onedrive/restore.go
```diff
@@ -683,7 +708,12 @@ func createRestoreFolders(
 		}
 
 		// create the folder if not found
-		folderItem, err = fr.PostItemInContainer(ictx, driveID, parentFolderID, newItem(folder, true))
+		folderItem, err = fr.PostItemInContainer(
+			ictx,
+			driveID,
+			parentFolderID,
+			newItem(folder, true),
+			control.Replace) // folders can be replaced without issue.
```
ashmrtn, line 716: Could you add a note as well that this will help us avoid race conditions where a folder is created between the `GetFolderByName` call above and here

src/internal/m365/onedrive/restore_test.go
```diff
@@ -315,3 +320,74 @@ func (suite *RestoreUnitSuite) TestAugmentRestorePaths_DifferentRestorePath() {
...
+		{
+			name:        "skip",
+			onCollision: control.Skip,
+			expectErr: func(t *testing.T, err error) {
+				require.NoError(t, err, clues.ToCore(err))
+			},
+			expectSkipped: assert.True,
+		},
+		{
+			name:        "replace",
```
ashmrtn, line 340: should this be "fail"? Ideally "replace" will eventually not return an error (assuming we workaround graph's current weirdness)

src/internal/m365/onedrive/restore_test.go
```diff
@@ -315,3 +320,74 @@ func (suite *RestoreUnitSuite) TestAugmentRestorePaths_DifferentRestorePath() {
...
+			_, skip, err := restoreItem(
+				ctx,
+				rh,
+				restoreCfg,
+				mock.FetchItemByName{
+					Item: &mock.Data{
+						Reader: mock.FileRespReadCloser(mock.DriveFileMetaData),
+					},
+				},
+				version.Backup,
```
ashmrtn, line 380: Should we also be testing lower backup versions?

src/pkg/services/m365/api/drive.go
```diff
@@ -123,21 +124,44 @@ func (c Drives) NewItemContentUpload(
...
 
 // PostItemInContainer creates a new item in the specified folder
 func (c Drives) PostItemInContainer(
 	ctx context.Context,
 	driveID, parentFolderID string,
 	newItem models.DriveItemable,
+	onCollision control.CollisionPolicy,
 ) (models.DriveItemable, error) {
+	// graph api has no policy for Skip; instead we catch
+	// and ignore the same-name failure.
```
ashmrtn, line 144: Ignoring seems to be done at a higher layer

```suggestion
	// and wrap the same-name failure.
```

src/pkg/services/m365/api/drive_test.go
```diff
@@ -4,41 +4,62 @@ import (
...
 }
 
-func (suite *OneDriveAPISuite) SetupSuite() {
+func (suite *DriveAPISuite) SetupSuite() {
 	t := suite.T()
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	userID := tester.LoadTestM365UserID(t)
```
ashmrtn, line 34: does the load test user currently have data purging setup? Does it need to be enabled to ensure these tests don't take too long? From the test below it seems like this would be fine to use the regular CI user

src/pkg/services/m365/api/drive_test.go
```diff
@@ -51,3 +72,195 @@ func (suite *OneDriveAPISuite) TestCreatePagerAndGetPage() {
...
+			postItem:    folder,
+			expectErr: func(t *testing.T, err error) {
+				assert.NoError(t, err, clues.ToCore(err))
+			},
+			expectItem: func(t *testing.T, i models.DriveItemable) {
+				assert.NotEqual(
+					t,
+					ptr.Val(origFolder.GetId()),
+					ptr.Val(i.GetId()),
+					"replaced item should be different from the original")
```
ashmrtn, line 171: Should also check the names aren't the same, after all that's the change the user will see, not the ID change
```suggestion
					"renamed item should be different from the original")
				assert.NotEqual(
					t,
					ptr.Val(origFolder.GetName()),
					ptr.Val(i.GetName()),
					"renamed item should have different name")
```

src/pkg/services/m365/api/drive_test.go
```diff
@@ -51,3 +72,195 @@ func (suite *OneDriveAPISuite) TestCreatePagerAndGetPage() {
...
+			postItem:    file,
+			expectErr: func(t *testing.T, err error) {
+				assert.NoError(t, err, clues.ToCore(err))
+			},
+			expectItem: func(t *testing.T, i models.DriveItemable) {
+				assert.NotEqual(
+					t,
+					ptr.Val(origFile.GetId()),
+					ptr.Val(i.GetId()),
+					"replaced item should be different from the original")
```
ashmrtn, line 212: Should also check the name here too
```suggestion
					"renamed item should be different from the original")
```

src/pkg/services/m365/api/drive_test.go
```diff
@@ -51,3 +72,195 @@ func (suite *OneDriveAPISuite) TestCreatePagerAndGetPage() {
...
+				ptr.Val(i.GetId()))
+			require.NoError(t, err, clues.ToCore(err))
+
+			iw := graph.NewLargeItemWriter(
+				ptr.Val(i.GetId()),
+				ptr.Val(mus.GetUploadUrl()),
+				int64(len(test.name)))
+
+			_, err = iw.Write([]byte(test.name))
+			require.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 263: is there a particular reason to upload data here? It's not really checked (i.e. don't ensure the new file's content changed rather than the old) so it doesn't seem to add much to the test. There's also other tests that check uploading data (albeit more indirectly)

ashmrtn (approved):

src/internal/m365/onedrive/restore.go
ryanfkeepers: Sure, can do.

src/internal/m365/onedrive/restore_test.go
ryanfkeepers: The name was chosen to describe what conflictBehavior the graph api call is using, based on the onCollision value.  Should be appropriate to keep it as is.  Also, yes, ideally we'll not error, whether or not it's fixed by graph.

src/internal/m365/onedrive/restore_test.go
ryanfkeepers: The behavior doesn't change across versions, we should be safe here.

src/pkg/services/m365/api/drive.go
ryanfkeepers: ah, good catch, thanks.

src/pkg/services/m365/api/drive_test.go
ryanfkeepers: Hah!  🤦 I completely forgot about the naming convention and thought to myself, "why yes, i do want to load the test user id."  Swapping that to the standard user now...

src/pkg/services/m365/api/drive_test.go
ryanfkeepers: Oh!  That's leftover from bug investigation.  Thanks for the reminder.

PR #3629, opened 2023-06-15
add per-container item collision enumeration

ryanfkeepers:
populates the item collision cache generators for all exchange data types.  Next we'll use these to compare the items being restored for collision before making post requests, by populating this cache at the top of each collection restore process.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/mail.go
```diff
@@ -540,3 +541,19 @@ func UnwrapEmailAddress(contact models.Recipientable) string {
...
+		return ""
+	}
+
+	sender := "unknown"
+
+	if item.GetSender() != nil && item.GetSender().GetEmailAddress() != nil {
+		sender = ptr.Val(item.GetSender().GetEmailAddress().GetAddress())
+	}
+
+	return dttm.Format(ptr.Val(item.GetCreatedDateTime())) + ptr.Val(item.GetSubject()) + sender
```
ashmrtn, line 558: is this sufficiently unique? What if mail forwarding is setup and dumps multiple emails into the same inbox (i.e. old account from college forwards to this one)? Do we think createdDateTime will be unique enough? Or are things somehow batched on the server so the created date times kind of end up the same if one gets a lot of messages at once?

src/pkg/services/m365/api/mail_pager.go
```diff
@@ -204,6 +224,27 @@ func (p *mailIDPager) valuesIn(pl PageLinker) ([]getIDAndAddtler, error) {
 	return toValues[models.Messageable](pl)
 }
 
+func (c Mail) GetItemsInContainerByCollisionKey(
+	ctx context.Context,
+	userID, containerID string,
+) (map[string]string, error) {
+	ctx = clues.Add(ctx, "container_id", containerID)
+	pager := c.NewMailPager(userID, containerID, idAnd(createdDateTime, "subject")...)
```
ashmrtn, line 232: this is missing the sender

src/pkg/services/m365/api/mail.go
ryanfkeepers: I'm hedging bets that the created time isn't duplicated for any set of messages in a container.  There's not a lot of sufficiently unique properties we can rely on here otherwise that aren't already an ID.

In fact, thinking about it more, the created time might be _too_ unique.  It only identifies collisions where the original item is still in the folder.  That escapes the case where a restore is performed twice to the same folder.  Ideally, we'd recognize the prior restored value as a duplicate of the backup entry.  But the restored mail is going to have a bumped created date, which means no collisions.

I might switch this up to a concatenation of the subject and all recipients (including cc and bcc), and then probably a hash of that to avoid enormous keys.

src/pkg/services/m365/api/contacts_pager_test.go
```diff
@@ -0,0 +1,45 @@
...
+	defer flush()
+
+	results, err := suite.cts.ac.Contacts().GetItemsInContainerByCollisionKey(ctx, suite.cts.userID, "contacts")
+	require.NoError(t, err, clues.ToCore(err))
+	require.Less(t, 0, len(results), "requires at least one result")
+
+	for k, v := range results {
+		assert.NotEmpty(t, k, "all keys should be populated")
+		assert.NotEmpty(t, v, "all values should be populated")
+	}
```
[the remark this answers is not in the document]
ryanfkeepers: Yes, that's the plan.  Completely robust testing (ie, that exercises the use of these maps to detect collisions) will need to wait until we handle in-place restores, since we still can't restore to existing containers.  Until that point you'll see primarily smoke and mock testing in these PRs.

src/pkg/services/m365/api/events_pager.go
```diff
@@ -98,32 +98,74 @@ func (c Events) EnumerateContainers(
...
 //lint:ignore U1000 False Positive
-func (p *eventsPager) valuesIn(pl PageLinker) ([]models.Eventable, error) {
-	// TODO(rkeepers)
-	return nil, nil
+func (c Events) GetItemsInContainerByCollisionKey(
+	ctx context.Context,
+	userID, containerID string,
+) (map[string]string, error) {
+	ctx = clues.Add(ctx, "container_id", containerID)
+	pager := c.NewEventsPager(userID, containerID, idAnd(createdDateTime, "subject")...)
```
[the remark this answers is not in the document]
ryanfkeepers: copy pasta!

src/pkg/services/m365/api/contacts_pager_test.go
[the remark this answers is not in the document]
ashmrtn: given that I found two instances where we weren't fetching enough data to actually populate the unique key as expected I'd say that these should be non-smoke tests. It seems too easy to miss and error-prone to catch in higher layers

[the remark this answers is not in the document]
ryanfkeepers: That's fair, I can bulk them up a bit.

src/pkg/services/m365/api/mail.go
ryanfkeepers: Okay, I'm now three miles down a weird rabbithole.  What I'm going to do for this PR, since at this time we only need the overall structure and not the specific collision identification cases, is reduce the data used in identification to the ID, and come back to refine it in a later PR.  In the meantime, I've opened an async discussion around the design to ensure we're selecting the best data for collision comparison in respect to corner cases.

ryanfkeepers:
@ashmrtn thanks for the review, this is ready for a final check.

src/pkg/services/m365/api/mail.go
ashmrtn: sounds good, thanks for the update

src/pkg/services/m365/api/contacts_pager_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, e := range expect {
+		_, ok := results[e]
+		assert.Truef(t, ok, "expected results to contain collision key: %s", e)
+	}
```
ashmrtn, line 72: I'm a bit split overall on whether we should be making a discrete folder with an item for this. Mostly I'm concerned about flakes when the purge script runs, though I don't recall the precise variables it looks at when purging contacts (or other things) and whether that could result in getting fewer items on the second call than the first. If we do use a dedicated folder, the below could probably be changed to `assert.ElementsMatch`
```suggestion
	assert.Subset(t, maps.Keys(results), expected, "missing collision keys")
```

src/pkg/services/m365/api/events.go
```diff
@@ -679,3 +679,17 @@ func EventFromMap(ev map[string]any) (models.Eventable, error) {
...
+}
+
+// EventCollisionKey constructs a key from the eventable's creation time, subject, and organizer.
+// collision keys are used to identify duplicate item conflicts for handling advanced restoration config.
+func EventCollisionKey(item models.Eventable) string {
+	if item == nil {
+		return ""
+	}
+
+	return ptr.Val(item.GetSubject())
```
ashmrtn, line 694: does this need updated to use only the ID for the moment?

src/pkg/services/m365/api/events_pager_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	results, err := ac.GetItemsInContainerByCollisionKey(ctx, suite.cts.userID, "calendar")
+	require.NoError(t, err, clues.ToCore(err))
+	require.Less(t, 0, len(results), "requires at least one result")
+
+	for k, v := range results {
+		assert.NotEmpty(t, k, "all keys should be populated")
+		assert.NotEmpty(t, v, "all values should be populated")
+	}
```
ashmrtn, line 67: this should be updated to check `expect` so that it will fail when we add more props to compare
```suggestion
	}
	
	assert.Subset(t, results, expected)
```

src/pkg/services/m365/api/mail.go
```diff
@@ -540,3 +540,17 @@ func UnwrapEmailAddress(contact models.Recipientable) string {
...
+}
+
+// MailCollisionKey constructs a key from the messageable's subject, sender, and recipients (to, cc, bcc).
+// collision keys are used to identify duplicate item conflicts for handling advanced restoration config.
+func MailCollisionKey(item models.Messageable) string {
+	if item == nil {
+		return ""
+	}
+
+	return ptr.Val(item.GetSubject())
```
ashmrtn, line 555: also use just ID here?

ashmrtn (approved):

src/pkg/services/m365/api/events.go
ryanfkeepers: It's a stub for the moment, and is a long ways from being used.  Won't affect anything either way.

src/pkg/services/m365/api/contacts_pager_test.go
ryanfkeepers: All test cases are pulling from the canonical root folders (inbox, calendars, contacts) none of which are affected by the purge.  We should be protected across changes here unless someone is manually tampering with those folders, in which case all bets are off for testing anyway.

PR #3630, opened 2023-06-15
increase page size for all exchange pagers

ryanfkeepers:
Noticed during testing that we're still paging 10 items at a time for exchange item pager queries.  This increases that limit to the consts we have set for max delta and non-delta pages, respectively.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/backup_test.go
+++ b/src/internal/m365/exchange/backup_test.go
@@ -382,7 +382,7 @@ func newStatusUpdater(t *testing.T, wg *sync.WaitGroup) func(status *support.Con
 	return updater
 }
 
-type DataCollectionsIntegrationSuite struct {
+type BackupIntgSuite struct {
 	tester.Suite
 	user     string
 	site     string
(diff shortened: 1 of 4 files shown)


ashmrtn (approved):
can we get tests for this that use gock? Seems like something we should have testing for if we have subtle issues like this

ryanfkeepers:
> can we get tests for this that use gock? Seems like something we should have testing for if we have subtle issues like this

Honestly not sure how gock would help us out here.  We thought we were setting page sizes the correct way.  I only found out otherwise because I was printing out result streams while debugging.  Gock would control the response, which might have also hidden the issue.

ashmrtn:
can gock check headers/parameters? Would that help the issue? Presumably whether the header or a param should be set should be in the documentation

ryanfkeepers:
> Would that help the issue?

I don't think it would have caught the issue here.  The bug wasn't that the declaration was absent, but that it wasn't respected by the server.  I'm personally not sure what documentation was followed when setting the prior implementation.  For all I know it was correct at the time, and then graph api changed and we regressed.  Page size is a difficult assertion to check automatically outside of specifically manufacturing a response that pages at the expected quantity.

PR #3634, opened 2023-06-16
Fix uploading large attachments for exchange

meain:
Upload of large attachments were broken previously.

Previously we were relying on the size reported by Graph API, but
that is not reliable and we were not completing uploads because of that.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/attachment.go
+++ b/src/internal/m365/exchange/attachment.go
@@ -95,6 +95,8 @@ func uploadAttachment(
 
 	// for file attachments sized >= 3MB
 	if attachmentType == models.FILE_ATTACHMENTTYPE && size >= largeAttachmentSize {
+		// We expect the entire attachment to fit in memory.
+		// Max attachment size is 150MB.
 		content, err := api.GetAttachmentContent(attachment)
 		if err != nil {
 			return clues.Wrap(err, "serializing attachment content").WithClues(ctx)
--- a/src/internal/m365/exchange/attachment.go
+++ b/src/internal/m365/exchange/attachment.go
@@ -61,16 +61,8 @@ func uploadAttachment(
 		attachmentType = attachmentType(attachment)
 		id             = ptr.Val(attachment.GetId())
 		name           = ptr.Val(attachment.GetName())
+		size           = ptr.Val(attachment.GetSize())
 	)
-
-	content, err := api.GetAttachmentContent(attachment)
-	if err != nil {
-		return clues.Wrap(err, "serializing attachment content").WithClues(ctx)
-	}
-
-	// We cannot rely on attachment.GetSize() as that seems to be
-	// higher than the actual size for some reason
-	size := len(content)
 
 	ctx = clues.Add(
 		ctx,
@@ -103,7 +95,13 @@ func uploadAttachment(
 
 	// for file attachments sized >= 3MB
 	if attachmentType == models.FILE_ATTACHMENTTYPE && size >= largeAttachmentSize {
-		_, err := cli.PostLargeAttachment(ctx, userID, containerID, parentItemID, name, content)
+		content, err := api.GetAttachmentContent(attachment)
+		if err != nil {
+			return clues.Wrap(err, "serializing attachment content").WithClues(ctx)
+		}
+
+		_, err = cli.PostLargeAttachment(ctx, userID, containerID, parentItemID, name, content)
+
 		return err
 	}
 
--- a/src/internal/m365/exchange/attachment.go
+++ b/src/internal/m365/exchange/attachment.go
@@ -22,7 +22,7 @@ type attachmentPoster interface {
 		ctx context.Context,
 		userID, containerID, itemID, name string,
 		content []byte,
-	) (models.UploadSessionable, error)
+	) (string, error)
 }
 
 const (
--- a/src/internal/m365/exchange/attachment.go
+++ b/src/internal/m365/exchange/attachment.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
 
 type attachmentPoster interface {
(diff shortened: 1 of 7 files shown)


src/internal/m365/graph/uploadsession.go
```diff
@@ -75,5 +84,21 @@ func (iw *largeItemWriter) Write(p []byte) (int, error) {
 	// Update last offset
 	iw.lastWrittenOffset = endOffset
 
+	// Once the upload is complete, we get a Location header in the
+	// below format from which we can get the id of the uploaded
+	// item. This will only be available after we have uploaded the
+	// entire content(based on the size in the req header).
+	// https://outlook.office.com/api/v2.0/Users('<user-id>')/Messages('<message-id>')/Attachments('<attachment-id>')
```
ashmrtn, line 91: is there some Microsoft documentation that explains this? It would give some added comfort as presumably it would be some spec they'll stick to going forward

src/internal/m365/graph/uploadsession.go
```diff
@@ -75,5 +84,21 @@ func (iw *largeItemWriter) Write(p []byte) (int, error) {
...
+	loc := resp.Header.Get("Location")
+	if loc != "" {
+		splits := strings.Split(loc, "'")
```
ashmrtn, line 94: splitting on `'` feels pretty brittle. Can we guarantee they'll always quote values like that? Would it be better to split on `/` and then extract information from the final component? If they have a spec that states this then I'd be a bit more comfortable

src/pkg/services/m365/api/events_test.go
```diff
@@ -212,3 +215,73 @@ func (suite *EventsAPIUnitSuite) TestBytesToEventable() {
...
+
+func (suite *EventsAPIIntgSuite) TestRestoreLargeAttachment() {
+	t := suite.T()
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	userID := tester.M365UserID(suite.T())
+
+	folderName := testdata.DefaultRestoreConfig("eventrestoretest").Location
```
ashmrtn, line 253: nit
```suggestion
	folderName := testdata.DefaultRestoreConfig("eventlargeattachmenttest").Location
```

src/pkg/services/m365/api/events_test.go
```diff
@@ -212,3 +215,73 @@ func (suite *EventsAPIUnitSuite) TestBytesToEventable() {
...
+	evts := suite.ac.Events()
+	calendar, err := evts.CreateContainer(ctx, userID, folderName, "")
+	require.NoError(t, err, clues.ToCore(err))
+
+	// Delete on exit
+	defer func() { _ = evts.DeleteContainer(ctx, userID, ptr.Val(calendar.GetId())) }()
+
```
ashmrtn, line 260: as long as the calendar follows the naming scheme for the purge script we can just remove this code

ashmrtn (approved):

src/internal/m365/graph/uploadsession.go
ashmrtn: oh, also is this code shared by events and messages? Does that change anything if it is?

src/internal/m365/graph/uploadsession.go
```diff
@@ -75,5 +84,21 @@ func (iw *largeItemWriter) Write(p []byte) (int, error) {
...
+		if len(splits) != 7 || splits[4] != ")/Attachments(" || splits[5] == "" {
```
ashmrtn, line 95: nit
```suggestion
		if len(splits) != 7 || splits[4] != ")/Attachments(" || len(splits[5]) == 0 {
```

src/internal/m365/exchange/attachment.go
```diff
@@ -95,7 +95,13 @@ func uploadAttachment(
 
 	// for file attachments sized >= 3MB
 	if attachmentType == models.FILE_ATTACHMENTTYPE && size >= largeAttachmentSize {
-		_, err := cli.PostLargeAttachment(ctx, userID, containerID, parentItemID, name, int64(size), attachment)
+		content, err := api.GetAttachmentContent(attachment)
```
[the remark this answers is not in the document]
pandeyabs: Is there a memory upper bound here? Assuming concurrency factor of 4, and a max attachment size of 150mb, we should cap at 600 mb. Is my understanding correct, @meain ?

src/internal/m365/exchange/attachment.go
```diff
@@ -20,15 +21,14 @@ type attachmentPoster interface {
 	PostLargeAttachment(
 		ctx context.Context,
 		userID, containerID, itemID, name string,
-		size int64,
-		body models.Attachmentable,
-	) (models.UploadSessionable, error)
+		content []byte,
```
pandeyabs, line 24: Just curious, how is size reported by graph not reliable? Could you explain a bit more? :)

pandeyabs (commented):
In general looks good to me!

src/internal/m365/exchange/attachment.go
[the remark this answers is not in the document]
meain: > Is there a memory upper bound here? Assuming concurrency factor of 4, and a max attachment size of 150mb, we should cap at 600 mb. Is my understanding correct, @meain ?

Yup, you are right.
Ref: https://learn.microsoft.com/en-us/graph/outlook-large-attachments?tabs=http

src/internal/m365/graph/uploadsession.go
meain: https://learn.microsoft.com/en-us/graph/outlook-large-attachments?tabs=http explains this. I'll add the link to comments. Also, it is same for both mail and event.

src/internal/m365/graph/uploadsession.go
meain: https://learn.microsoft.com/en-us/graph/outlook-large-attachments?tabs=http documents the format. I started of splitting by `/`, but then we will eventually have to split on `'`. Plus the checks [here](https://github.com/alcionai/corso/blob/01e10627630d7c20c3a46f43a9cd80f07fbc43d1/src/internal/m365/graph/uploadsession.go#L95) give me enough confidence.

src/internal/m365/exchange/attachment.go
meain: The size reported by Grpah `attachment.GetSize()` ends up being bigger than `len(attachment.GetContent())`. When we try to upload it with the size bigger than actual mentioned in headers, it never completes the upload.

PR #3635, opened 2023-06-16
handle restore collisions in exchange

ryanfkeepers:
adds item collision handling to exchange restores. Currently an incomplete implementation; the replace setting will skip the restore altogether (no-op) as a first pass.  The next PR will finish out the replace behavior.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/contacts_restore.go
+++ b/src/internal/m365/exchange/contacts_restore.go
@@ -9,7 +9,9 @@ import (
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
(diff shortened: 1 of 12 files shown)


src/internal/m365/exchange/contacts_restore.go
```diff
@@ -73,8 +94,20 @@ func (h contactRestoreHandler) restore(
 	}
 
 	ctx = clues.Add(ctx, "item_id", ptr.Val(contact.GetId()))
+	collisionKey := api.ContactCollisionKey(contact)
+
+	if _, ok := collisionKeyToItemID[collisionKey]; ok {
+		log := logger.Ctx(ctx).With("collision_key", clues.Hide(collisionKey))
+		log.Debug("item collision")
```
meain, line 101: Just in case this was not intentional

```suggestion
		log := logger.Ctx(ctx).With("collision_key", clues.Hide(collisionKey)).Debug("item collision")
```

meain (approved):
LGTM

src/internal/m365/exchange/contacts_restore.go
ryanfkeepers: Ah, `Debug()` doesn't return a logger instance, so that wouldn't work.  It'll change in the next PR or two, anyway.

PR #3636, opened 2023-06-16
properly handle collision replace behavior

ryanfkeepers:
completes the item collision handling behavior in
exchange by turning replace handling into a post-delete process.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/exchange/contacts_restore.go
```diff
@@ -94,22 +99,40 @@ func restoreContact(
 	}
 
 	ctx = clues.Add(ctx, "item_id", ptr.Val(contact.GetId()))
-	collisionKey := api.ContactCollisionKey(contact)
 
-	if _, ok := collisionKeyToItemID[collisionKey]; ok {
+	var (
+		collisionKey = api.ContactCollisionKey(contact)
+		collisionID  string
+		replace      bool
```
meain, line 106: `delete` might be a better name as this is used to decide if we should delete the item. Same in other places.

meain (approved):
The code changes here looks good to me.

That said, I'm a bit concerned about how we deal with restore failures. If we fail a restore and we retry it, we will end up creating the missing item multiple times(one for each restore) as the "original' item is not available in both runs. Should we have something in place to deal with this?

ryanfkeepers:
> we will end up creating the missing item multiple times(one for each restore) as the "original' item is not available in both runs

The collision identifier should be able to track item identity across restorations to avoid duplication.  There is a strange case we could technically hit where the same item across multiple restores fails to delete its original version after creating the backup replacement, true.  But the likelihood of that happening multiple times for the same item is effectively zero unless we generate a pathological bug. 

On the other hand, there's the question of how much trouble the two-step action can cause even without it being the same item each time.  And, yeah, I don't particularly like the pattern either.  Would be much nicer if graph api provided a PUT method.  Kinda stuck between two unhappy options here.  This picks one to get us moving forward, but it's a two-way door and we can always replace it with a patch if this causes distress.

meain:
> The collision identifier should be able to track item identity across restorations to avoid duplication. There is a strange case we could technically hit where the same item across multiple restores fails to delete its original version after creating the backup replacement, true. But the likelihood of that happening multiple times for the same item is effectively zero unless we generate a pathological bug.

My question is a bit different. In this situation, the original file was deleted in OneDrive some time between backup and restore. Now, every time we run a restore, we will end up restoring the file as the original is never preset and our collision id is based on the id of the original file.

src/internal/m365/exchange/contacts_restore.go
```diff
@@ -94,22 +99,40 @@ func restoreContact(
...
 		log.Debug("item collision")
 
-		// TODO(rkeepers): Replace probably shouldn't no-op.  Just a starting point.
-		if collisionPolicy == control.Skip || collisionPolicy == control.Replace {
+		if collisionPolicy == control.Skip {
 			log.Debug("skipping item with collision")
 			return nil, graph.ErrItemAlreadyExistsConflict
 		}
+
+		collisionID = id
```
ashmrtn, line 118: looking at this now, I'm wondering if it's a little simplistic. It feels like we may run afoul of the case where we get some 5xx error but the item is actually created. The run will fail (5xx for post) but now we'll potentially have 2 items with the same collision key: the original original and the one that failed with the 5xx

ryanfkeepers: Yeah, fair point.  We'll have to handle multiple-item collisions gracefully.  After all, for certain category types, that condition can be fairly easily manufactured using the api or webapp.  It'll only turn into a concern when restoring with replace behavior; and in that case we're only going to replace one of the many colliding items.  Hope is that fits around all but the most extreme issues without too much fuss.

ashmrtn: would just keeping a list of IDs in the collision cache (not sure if that's the right name) suffice?

ryanfkeepers:
> My question is a bit different. In this situation, the original file was deleted in OneDrive some time between backup and restore. Now, every time we run a restore, we will end up restoring the file as the original is never preset and our collision id is based on the id of the original file.

Oh, that!  That won't stick.  The id-as-placeholder is only there temporarily while we pin down the correct identifier.  Rest assured it'll be a more stable aggregation of data before we implement this handling.

src/internal/m365/exchange/contacts_restore.go
ryanfkeepers: It'd be a way to identify all the associated items.  But if we're not going to use them for anything, I don't see much purpose in the aggregation.

PR #3638, opened 2023-06-16
minor refactoring before changes

ryanfkeepers:
renaming structs so that they don't follow the interface naming conventions, updating and expanding the test suite setup for api.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/drive.go
+++ b/src/pkg/services/m365/api/drive.go
@@ -8,6 +8,7 @@ import (
 	"github.com/microsoftgraph/msgraph-sdk-go/drives"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
+	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/control"
 )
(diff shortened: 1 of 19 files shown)


meain:
This could probably use a rebase before review as there are quite a few changes from previous PRs.

ryanfkeepers:
> This could probably use a rebase before review as there are quite a few changes from previous PRs.

Yep, planning on it.  CI flakes have made it troublesome to get the prior PR through.  I'll probably do a rebase sweep from here to the tail once that gets merged.

src/pkg/services/m365/api/drive.go
```diff
@@ -305,3 +306,13 @@ func (c Drives) DeleteItemPermission(
...
 }
+
+// DriveItemCollisionKeyy constructs a key from the item name.
+// collision keys are used to identify duplicate item conflicts for handling advanced restoration config.
+func DriveItemCollisionKey(item models.DriveItemable) string {
+	if item == nil {
+		return ""
+	}
+
+	return ptr.Val(item.GetName())
```
ashmrtn, line 317: is the cache going to be per drive or per folder? If it's per drive the collision key will need to include the full (display name) path of the item

ashmrtn (approved):

src/pkg/services/m365/api/drive.go
ryanfkeepers: It will be per folder.

PR #3639, opened 2023-06-16
add the non-delta item pager to drives

ryanfkeepers:
introduces a non-delta item pager to drives_pager, and a func for building an item collision detection cache.  Implementation is coming in the next PR.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/drive_pager.go
+++ b/src/pkg/services/m365/api/drive_pager.go
@@ -18,6 +18,84 @@ import (
 )
 
 // ---------------------------------------------------------------------------
+// non-delta item pager
+// ---------------------------------------------------------------------------
+
+var _ itemPager[models.DriveItemable] = &driveItemPageCtrl{}
+
+type driveItemPageCtrl struct {
+	gs      graph.Servicer
+	builder *drives.ItemItemsItemChildrenRequestBuilder
+	options *drives.ItemItemsItemChildrenRequestBuilderGetRequestConfiguration
+}
+
+func (c Drives) NewDriveItemPager(
+	driveID, containerID string,
+	selectProps ...string,
+) itemPager[models.DriveItemable] {
+	options := &drives.ItemItemsItemChildrenRequestBuilderGetRequestConfiguration{
+		QueryParameters: &drives.ItemItemsItemChildrenRequestBuilderGetQueryParameters{
+			Top: ptr.To(maxNonDeltaPageSize),
+		},
+	}
+
+	if len(selectProps) > 0 {
+		options.QueryParameters.Select = selectProps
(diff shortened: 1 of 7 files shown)


src/pkg/services/m365/api/drive_pager.go
```diff
@@ -17,6 +17,84 @@ import (
...
+		return nil, graph.Wrap(ctx, err, "enumerating drive items")
+	}
+
+	m := map[string]string{}
+
+	for _, item := range items {
+		// folders are returned in this query, we only want the files
+		if item.GetFile() == nil {
+			continue
+		}
```
ashmrtn, line 88: Won't we need both files and folders? It's not possible in OneDrive to have a file and folder with the same name. It's a little difficult to manufacture that situation since the web UI doesn't seem to want to let users change the extensions on things but it can be done by uploading a file to OneDrive. Something like
1. make folder foo in root of OneDrive
2. on local machine `touch ~/Downloads/foo`
3. drag & drop new file foo from above to OneDrive root folder
4. OneDrive pops up dialog saying the file has the same name as another item and gives options to replace or keep both

Keeping both results in `foo (1)` being created. Replace appears to result in an error of "something went wrong so we couldn't upload **foo**"

src/pkg/services/m365/api/drive_pager_test.go
```diff
@@ -0,0 +1,65 @@
...
+			t,
+			[][]string{tester.M365AcctCredEnvs}),
+	})
+}
+
+func (suite *DrivePagerIntgSuite) SetupSuite() {
+	suite.its = newIntegrationTesterSetup(suite.T())
+}
+
+func (suite *DrivePagerIntgSuite) TestGetItemsInContainerByCollisionKey() {
```
ashmrtn, line 31: this should probably be checking for known items since we already know what the collision key will be. You can make a folder with specific items to test with if needed or use something like gock

src/pkg/services/m365/api/drive_pager.go
ryanfkeepers: Good call, will remove the filter.

src/pkg/services/m365/api/drive_pager_test.go
ryanfkeepers: yup, good call.

ryanfkeepers:
@ashmrtn Thanks for the review, this is ready for a final check.

ashmrtn (approved):

PR #3640, opened 2023-06-16
od: add per-container collision caching and checks

ryanfkeepers:
pushes collision checks in onedrive upward from the api layer (due to the bug with ConflictBehavior) and into the restore controller.  Collision resolution in drive follows the same pattern that we use in exchange.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/item_handler.go
+++ b/src/internal/m365/onedrive/item_handler.go
@@ -147,11 +147,11 @@ func (h itemRestoreHandler) AugmentItemInfo(
 	return augmentItemInfo(dii, item, size, parentPath)
 }
 
-func (h itemRestoreHandler) NewItemContentUpload(
+func (h itemRestoreHandler) DeleteItem(
 	ctx context.Context,
 	driveID, itemID string,
-) (models.UploadSessionable, error) {
-	return h.ac.NewItemContentUpload(ctx, driveID, itemID)
+) error {
+	return h.ac.DeleteItem(ctx, driveID, itemID)
 }
 
 func (h itemRestoreHandler) DeleteItemPermission(
@@ -159,6 +159,25 @@ func (h itemRestoreHandler) DeleteItemPermission(
 	driveID, itemID, permissionID string,
 ) error {
 	return h.ac.DeleteItemPermission(ctx, driveID, itemID, permissionID)
+}
+
+func (h itemRestoreHandler) GetItemsInContainerByCollisionKey(
+	ctx context.Context,
+	driveID, containerID string,
+) (map[string]string, error) {
+	m, err := h.ac.GetItemsInContainerByCollisionKey(ctx, driveID, containerID)
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+func (h itemRestoreHandler) NewItemContentUpload(
+	ctx context.Context,
+	driveID, itemID string,
+) (models.UploadSessionable, error) {
+	return h.ac.NewItemContentUpload(ctx, driveID, itemID)
 }
 
 func (h itemRestoreHandler) PostItemPermissionUpdate(
--- a/src/internal/m365/onedrive/mock/handlers.go
+++ b/src/internal/m365/onedrive/mock/handlers.go
@@ -235,8 +235,9 @@ type RestoreHandler struct {
 
 	CollisionKeyMap map[string]string
 
-	CalledDeleteItem bool
-	DeleteItemErr    error
+	CalledDeleteItem   bool
+	CalledDeleteItemOn string
+	DeleteItemErr      error
 
 	CalledPostItem bool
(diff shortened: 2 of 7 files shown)


src/internal/m365/onedrive/item_handler.go
```diff
@@ -147,11 +147,11 @@ func (h itemRestoreHandler) AugmentItemInfo(
...
 }
 
-func (h itemRestoreHandler) NewItemContentUpload(
+func (h itemRestoreHandler) DeleteItem(
 	ctx context.Context,
 	driveID, itemID string,
-) (models.UploadSessionable, error) {
-	return h.ac.NewItemContentUpload(ctx, driveID, itemID)
+) error {
+	return h.ac.DeleteItem(ctx, driveID, itemID)
```
ashmrtn, line 154: I kind of forgot about this until now, but do we need to start worrying about the whole chunked response on delete calls issue that was causing CI to fail for a while? Or was that deleting folders only?

src/internal/m365/onedrive/mock/handlers.go
```diff
@@ -233,11 +233,19 @@ func (m GetsItemPermission) GetItemPermission(
 type RestoreHandler struct {
```
ashmrtn, line 233: I feel like this is increasingly turning into something that would be best suited by generated mocks. We're kind of getting to the point where we want to track which calls are made and return info for some of them. That's something that at least some mock packages support out of the box (iirc the golang mocks package does)

src/internal/m365/onedrive/restore_test.go
```diff
@@ -321,28 +322,73 @@ func (suite *RestoreUnitSuite) TestAugmentRestorePaths_DifferentRestorePath() {
...
 			},
+		},
+		{
+			name:          "collision, replace",
+			collisionKeys: map[string]string{mock.DriveItemFileName: "smarf"},
+			onCollision:   control.Replace,
 			expectSkipped: assert.False,
+			expectMock: func(t *testing.T, rh *mock.RestoreHandler) {
+				assert.True(t, rh.CalledPostItem, "new item posted")
+				assert.True(t, rh.CalledDeleteItem, "new item deleted")
```
ashmrtn, line 380: we should probably be checking the ID of the thing we're deleting

ashmrtn (commented):
this needs rebased prior to final review

src/internal/m365/onedrive/item_handler.go
ryanfkeepers: You mean the empty response body after a delete?  Shouldn't need to worry, that protection is supposed to be built into the api layer funcs (and it is here).

src/internal/m365/onedrive/mock/handlers.go
ryanfkeepers: mm, admittedly, my personal experience is that generated mocks cause more grief than they save.  Lots of time backlogged in fights with extensive generated output overhauls to comply with small changes.  I'm the wrong person to champion this addition.  And, philosophically, I equate an increase in value from generated mocks with a smell.  It suggests that the interface has grown too large and unwieldy, and is a non-idiomatic implementation.  Can't always avoid that, though.

src/internal/m365/onedrive/restore_test.go
ryanfkeepers: can do

ryanfkeepers:
@ashmrtn ready for a final review, thank you

src/internal/m365/onedrive/mock/handlers.go
ashmrtn: >philosophically, I equate an increase in value from generated mocks with a smell. It suggests that the interface has grown too large and unwieldy, and is a non-idiomatic implementation.

I think these can actually be disjoint. The interface that needs tested here isn't exceptionally large, just we want to validate the data passed to said interface is what we expect it to be. That's where mocks can excel because they allow for patterns like `mock.OnFuncX().Expect(param1, param2)` and can be configured to fail the test if there's unexpected calls to functions or more than the expected number of calls

ashmrtn (approved):

src/internal/m365/onedrive/mock/handlers.go
ryanfkeepers: > isn't exceptionally large

heh, a matter of opinion, that.

> That's where mocks can excel

I am, to be fair, fully aware of the functionality provided by generative mocks.  And, also to be fair, I'm quite a fan of the behavior.  My primary disagreement comes from projects where maintenance of those mocks at scale can overtake their granular benefits.  Admittedly, those experiences could have related to poor implementations, and I have a misguided opinion.  Nonetheless, the feeling stands.

Issue #3579, opened 2023-06-09
Improve error message when no items found to restore


meain:
Currently if we try to restore from an empty backup or if the user asks to restore an item not present in the backup, the error that they get is `running restore: primary failure: running restore: producing collections to restore: no restore path given` which is not very useful.

We should probably improve the situation here and let them know what the exact problem is (whether they are trying to restore an empty backup of if we were not able to find the item they requested in the backup) instead of throwing this generic error.

vkamra:
https://github.com/alcionai/corso/pull/3643

vkamra:
@neha-Gupta1 - can this be closed now? #3643 has merged

PR #3643, opened 2023-06-20
Onedrive restore- handle unknown path error

neha-Gupta1:
In case of Onedrive Restore give a proper message if the path provided do not exist or restore is done on empty backup.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3579

#### Test Plan

- [x] :muscle: Manual


src/internal/operations/restore.go
```diff
@@ -167,6 +168,11 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
 	if err != nil {
 		// No return here!  We continue down to persistResults, even in case of failure.
 		logger.CtxErr(ctx, err).Error("running restore")
+
+		if strings.Contains(err.Error(), "no restore path given") {
```
ryanfkeepers, line 172: checking error strings is a smell since the check is brittle and prone to failure when the error message changes.  Is there a way to check for this using errors.Is instead?

neha-Gupta1: as of my understanding errors.Is checks for equality. And here the check is required for sub-string. Is there a way to just check the sub-string and not complete string? Also does clues package has anything to do this check?

neha-Gupta1: Looking into it. Somehow errors.is function isn't able to match the case.

ryanfkeepers: Is there a way for us to wrap the original error with something in corso so that it could match using errors.Is.

neha-Gupta1: clues package itself has a `errors.Is` func which should ideally take care of the stacks too. But is unable to compare the error even when the messages are same. 
Not sure about the reason. Maybe metadata is different.

neha-Gupta1: Done

src/internal/operations/restore.go
```diff
@@ -167,6 +168,11 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
 	if err != nil {
 		// No return here!  We continue down to persistResults, even in case of failure.
 		logger.CtxErr(ctx, err).Error("running restore")
+
+		if errors.Is(err, kopia.ErrNoRestorePath) {
+			op.Errors.Fail(clues.New("running restore: empty backup or unknown path provided"))
```
ryanfkeepers, line 173: Try to avoid prefixing errors with categorizations like `running restore:`.  It's the job of upstream handlers to wrap error messages with that prefix.  Using it in lower level messages can result in messages that read like this, `running restore: foo: bar: running restore: baz: qux`.  And that just ends up being confusing.

ryanfkeepers (approved):
Approving to unblock, but please fix the error return issues as described in the comments.

src/internal/operations/restore.go
neha-Gupta1: done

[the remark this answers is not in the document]
neha-Gupta1: I tried that. But errors.Is is not able to compare them. In fact if the error with same message is defined again `errors.Is` is not equating them. Possible reason could be some metadata which is different in that case.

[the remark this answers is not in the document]
neha-Gupta1: Merging this PR but can continue working on this if you have some suggestions.

PR #3644, opened 2023-06-20
Update attachments for events which drift from series master

meain:
This updates any changes to attachments for individual event instances
for the ones that differ from the series master.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/events_restore.go
+++ b/src/internal/m365/exchange/events_restore.go
@@ -1,6 +1,7 @@
 package exchange
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strings"
@@ -149,7 +150,7 @@ func updateRecurringEvents(
 		return clues.Wrap(err, "update cancelled occurrences")
 	}
 
-	err = updateExceptionOccurrences(ctx, ac, userID, itemID, exceptionOccurrences)
+	err = updateExceptionOccurrences(ctx, ac, userID, containerID, itemID, exceptionOccurrences)
 	if err != nil {
 		return clues.Wrap(err, "update exception occurrences")
 	}
@@ -164,6 +165,7 @@ func updateExceptionOccurrences(
 	ctx context.Context,
 	ac api.Events,
 	userID string,
+	containerID string,
 	itemID string,
 	exceptionOccurrences any,
 ) error {
@@ -210,14 +212,97 @@ func updateExceptionOccurrences(
 
 		evt = toEventSimplified(evt)
 
-		// TODO(meain): Update attachments (might have to diff the
-		// attachments using ids and delete or add). We will have
-		// to get the id of the existing attachments, diff them
-		// with what we need a then create/delete items kinda like
-		// permissions
 		_, err = ac.PatchItem(ctx, userID, ptr.Val(evts[0].GetId()), evt)
 		if err != nil {
 			return clues.Wrap(err, "updating event instance")
+		}
+
+		evt, err = api.EventFromMap(instance)
+		if err != nil {
+			return clues.Wrap(err, "parsing exception event (second)")
+		}
+
+		err = updateAttachments(ctx, ac, userID, containerID, ptr.Val(evts[0].GetId()), evt)
+		if err != nil {
+			return clues.Wrap(err, "updating attachments")
+		}
+	}
+
+	return nil
+}
+
+func updateAttachments(
+	ctx context.Context,
+	client api.Events,
+	userID, containerID, eventID string,
+	event models.Eventable,
+) error {
+	// Attachment can only be deleted or added (validate added) in
+	// modified events and for them, the attachment id will be
+	// different from the original one, both in the backup and during
+	// restore. And so we detect and item using its contentBytes.
+	attachments, err := client.GetAttachments(ctx, false, userID, eventID)
+	if err != nil {
+		return clues.Wrap(err, "getting attachments")
+	}
+
+	// Delete unnecessary attachments
+	for _, att := range attachments {
+		content, err := api.GetAttachmentContent(att)
+		if err != nil {
+			return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(att.GetId()))
+		}
+
+		found := false
+
+		for _, natt := range event.GetAttachments() {
+			bcontent, err := api.GetAttachmentContent(natt)
+			if err != nil {
+				return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(natt.GetId()))
+			}
+
+			if bytes.Equal(content, bcontent) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			err = client.DeleteAttachment(ctx, userID, eventID, ptr.Val(att.GetId()))
+			if err != nil {
+				return clues.Wrap(err, "deleting attachment").
+					With("attachment_id", ptr.Val(att.GetId()))
+			}
+		}
+	}
+
+	// Create missing attachments
+	for _, att := range event.GetAttachments() {
+		content, err := api.GetAttachmentContent(att)
+		if err != nil {
+			return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(att.GetId()))
+		}
+
+		found := false
+
+		for _, natt := range attachments {
+			bcontent, err := api.GetAttachmentContent(natt)
+			if err != nil {
+				return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(natt.GetId()))
+			}
+
+			if bytes.Equal(content, bcontent) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			err = uploadAttachment(ctx, client, userID, containerID, eventID, att)
+			if err != nil {
+				return clues.Wrap(err, "updating attachment").
+					With("attachment_id", ptr.Val(att.GetId()))
+			}
 		}
 	}
 
--- a/src/pkg/services/m365/api/events.go
+++ b/src/pkg/services/m365/api/events.go
@@ -241,9 +241,12 @@ func (c Events) GetItem(
 		return nil, nil, clues.Wrap(err, "fixup exception occurrences")
 	}
 
-	attachments, err := c.getAttachments(ctx, event, immutableIDs, userID, itemID)
-	if err != nil {
-		return nil, nil, err
+	var attachments []models.Attachmentable
+	if ptr.Val(event.GetHasAttachments()) || HasAttachments(event.GetBody()) {
+		attachments, err = c.GetAttachments(ctx, immutableIDs, userID, itemID)
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
 	event.SetAttachments(attachments)
@@ -286,10 +289,14 @@ func fixupExceptionOccurrences(
 
 		// OPTIMIZATION: We don't have to store any of the
 		// attachments that carry over from the original
-		attachments, err := client.getAttachments(ctx, evt, immutableIDs, userID, ptr.Val(evt.GetId()))
-		if err != nil {
-			return clues.Wrap(err, "getting exception attachments").
-				With("exception_event_id", ptr.Val(evt.GetId()))
+
+		var attachments []models.Attachmentable
+		if ptr.Val(event.GetHasAttachments()) || HasAttachments(event.GetBody()) {
+			attachments, err = client.GetAttachments(ctx, immutableIDs, userID, ptr.Val(evt.GetId()))
+			if err != nil {
+				return clues.Wrap(err, "getting exception attachments").
+					With("exception_event_id", ptr.Val(evt.GetId()))
+			}
 		}
 
 		// This odd roundabout way of doing this is required as
@@ -370,17 +377,12 @@ func parseableToMap(att serialization.Parsable) (map[string]any, error) {
 	return item, nil
 }
 
-func (c Events) getAttachments(
+func (c Events) GetAttachments(
 	ctx context.Context,
-	event models.Eventable,
 	immutableIDs bool,
 	userID string,
 	itemID string,
 ) ([]models.Attachmentable, error) {
-	if !ptr.Val(event.GetHasAttachments()) && !HasAttachments(event.GetBody()) {
-		return nil, nil
-	}
-
 	config := &users.ItemEventsItemAttachmentsRequestBuilderGetRequestConfiguration{
 		QueryParameters: &users.ItemEventsItemAttachmentsRequestBuilderGetQueryParameters{
 			Expand: []string{"microsoft.graph.itemattachment/item"},
@@ -401,6 +403,21 @@ func (c Events) getAttachments(
 	}
 
 	return attached.GetValue(), nil
+}
+
+func (c Events) DeleteAttachment(
+	ctx context.Context,
+	userID, eventID, attachmentID string,
+) error {
+	return c.Stable.
+		Client().
+		Users().
+		ByUserId(userID).
+		Events().
+		ByEventId(eventID).
+		Attachments().
+		ByAttachmentId(attachmentID).
+		Delete(ctx, nil)
 }
 
 func (c Events) GetItemInstances(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Do not display all the items that we restored at the end if there are more than 15. You can override this with `--verbose`.
 
-### Known Issues
-- Changes to attachments in instances of recurring events compared to the series master aren't restored
-
 ## [v0.8.0] (beta) - 2023-05-15
 
 ### Added
--- a/website/docs/support/known-issues.md
+++ b/website/docs/support/known-issues.md
@@ -26,5 +26,3 @@ included in backup and restore.
 * SharePoint document library data can't be restored after the library has been deleted.
 
 * Sharing information of items in OneDrive/SharePoint using sharing links aren't backed up and restored.
-
-* Changes to attachments in instances of recurring events compared to the series master aren't restored


src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+	// changed. This change is both present in the data that we backed
+	// up and so we cannot rely on the attachment id to decide if they
+	// are the same. Instead we make use of the contentBytes and name
+	// to ensure they are the same.
+	attachments, err := client.GetAttachments(ctx, false, userID, eventID)
+	if err != nil {
+		return clues.Wrap(err, "getting attachments")
+	}
+
+	// Delete unnecessary attachments
```
ryanfkeepers, line 251: This comment could use more explanation to describe the overall flow and purpose of the loop.  It may be a small and relatively straightforward algorithm, but since it's juggling two different sources with effectively the same data, it can be easy to mix up the state of the two when constructing a mental state flow.

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+	for _, att := range attachments {
```
ryanfkeepers, line 252: Should we actually delete attachments outside of an on-collision=replace policy?  If we're restoring to a copy, we don't want to delete data from the original.  Or is that isolation already handled at a higher level?

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+			if name == bname && bytes.Equal(content, bcontent) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			err = uploadAttachment(ctx, client, userID, containerID, eventID, att)
+			if err != nil {
+				return clues.Wrap(err, "updating attachment").
```
ryanfkeepers, line 313:
```suggestion
				return clues.Wrap(err, "uploading attachment").
```

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+
+		for _, natt := range event.GetAttachments() {
+			bname := ptr.Val(natt.GetName())
+
+			bcontent, err := api.GetAttachmentContent(natt)
+			if err != nil {
+				return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(natt.GetId()))
+			}
+
+			if name == bname && bytes.Equal(content, bcontent) {
```
ryanfkeepers, line 270: Is it possible for there to be a delta across names for the same content?  Ie: the attachment name gets updated, but not its content?

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+		err = updateAttachments(ctx, ac, userID, containerID, ptr.Val(evts[0].GetId()), evt)
+		if err != nil {
+			return clues.Wrap(err, "updating attachments")
+		}
+	}
+
+	return nil
+}
+
+func updateAttachments(
```
ryanfkeepers, line 234: I'm a bit concerned with the higher level algorithm in use here.  So far you have an O(2n^2), which is a bit of a smell, though possibly fine if we're assuming a small n and minimal processing for each step.  If we expect larger sizes of n, or extra processing (such as api calls) within any of these steps, you might want to refactor this to a more streamlined process.  Implementation for a refactor like that looks a little a little messy, so I don't want to push this, so long as performance can stay within reason.

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	calendar, err := handler.ac.CreateContainer(ctx, userID, subject, "")
+	require.NoError(t, err, clues.ToCore(err))
+
+	calendarID := ptr.Val(calendar.GetId())
+
+	defer func() {
+		// Removes calendar containing events created during the test
+		err = suite.ac.Events().DeleteContainer(ctx, userID, calendarID)
+		assert.NoError(t, err, clues.ToCore(err))
+	}()
```
ryanfkeepers, line 398: We should rely on our cleanup process to remove test-generated data.  Is it possible to delete this deferral?

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
 		})
 	}
 }
+
+func (suite *RestoreIntgSuite) TestRestoreAndBackupEventWithRecurringAndAttachments() {
```
ryanfkeepers, line 377: Nit: please follow golang example naming conventions for test names.  ie: `TestPrimaryFunc_variationOfTest`
```suggestion
func (suite *RestoreIntgSuite) TestRestoreAndBackupEvent_recurringInstancesWithAttachments() {
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	require.NotNil(t, event.GetRecurrence(), "recurrence")
+	eo := event.GetAdditionalData()["exceptionOccurrences"]
+	require.NotNil(t, eo, "exceptionOccurrences")
+
+	require.NotEqual(
+		t,
+		ptr.Val(event.GetSubject()),
+		ptr.Val(eo.([]interface{})[0].(map[string]interface{})["subject"].(*string)),
+		"name equal",
+	)
```
ryanfkeepers, line 434:
```suggestion
		"name equal")
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+
+	atts := eo.([]interface{})[0].(map[string]interface{})["attachments"]
+	require.NotEqual(
+		t,
+		ptr.Val(event.GetAttachments()[0].GetName()),
+		ptr.Val(atts.([]interface{})[0].(map[string]interface{})["name"].(*string)),
+		"attachment name equal",
+	)
```
ryanfkeepers, line 442:
```suggestion
		"attachment name equal")
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	assert.NoError(t, err, clues.ToCore(err))
+
+	require.NotNil(t, event.GetRecurrence(), "recurrence")
+	eo := event.GetAdditionalData()["exceptionOccurrences"]
+	require.NotNil(t, eo, "exceptionOccurrences")
+
+	require.NotEqual(
+		t,
+		ptr.Val(event.GetSubject()),
+		ptr.Val(eo.([]interface{})[0].(map[string]interface{})["subject"].(*string)),
```
ryanfkeepers, line 432: Use `any` instead of `interface{}`, please.  (I tried to make a linter for this, and for some reason it's not working).
```suggestion
		ptr.Val(eo.([]any)[0].(map[string]any)["subject"].(*string)),
```

ryanfkeepers (approved):
Comments are mostly stylistic nits., with a couple process-oriented issues on the side.  I'm concerned about the latter, and would like to have clear assurance that those cases won't be an issue.  But none of it is a bug or blocker, and this looks good enough as-is, so I'm approving to unblock.  Please follow up on those corner cases as needed, but feel free to merge when ready.

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+		userID, calendarID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
+
+	ec, err := handler.ac.GetCalendarEvents(ctx, userID, calendarID)
+	assert.NoError(t, err, clues.ToCore(err))
+
+	evts := ec.GetValue()
+	assert.Equal(t, len(evts), 1, "count of events")
```
ashmrtn, line 413:
```suggestion
	assert.Len(t, evts, 1, "count of events")
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	info, err := handler.restore(
+		ctx,
+		bytes,
+		userID, calendarID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
+
+	ec, err := handler.ac.GetCalendarEvents(ctx, userID, calendarID)
+	assert.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 410:
```suggestion
	require.NoError(t, err, clues.ToCore(err))
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+		assert.NoError(t, err, clues.ToCore(err))
+	}()
+
+	bytes := exchMock.EventWithRecurrenceAndExceptionAndAttachmentBytes("Reoccurring event restore and backup test")
+	info, err := handler.restore(
+		ctx,
+		bytes,
+		userID, calendarID,
+		fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 406: nit: general rule of thumb is to use `assert` when checking data for the test but the test can continue otherwise (i.e. comparing expected values). Use `require` when checking for things that should halt the test (i.e. failed to get response from remote server and now there's no data to check for expected values)
```suggestion
	require.NoError(t, err, clues.ToCore(err))
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	require.NoError(t, err, clues.ToCore(err))
+
+	calendarID := ptr.Val(calendar.GetId())
+
+	defer func() {
+		// Removes calendar containing events created during the test
+		err = suite.ac.Events().DeleteContainer(ctx, userID, calendarID)
+		assert.NoError(t, err, clues.ToCore(err))
+	}()
+
```
ashmrtn, line 399: we should be able to skip this. Plus we don't want to fail the test if everything but removing the calendar passes

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	assert.NotNil(t, info, "event item info")
+
+	ec, err := handler.ac.GetCalendarEvents(ctx, userID, calendarID)
+	assert.NoError(t, err, clues.ToCore(err))
+
+	evts := ec.GetValue()
+	assert.Equal(t, len(evts), 1, "count of events")
+
+	sp, info, err := suite.ac.Events().GetItem(ctx, userID, ptr.Val(evts[0].GetId()), false, fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 416:
```suggestion
	require.NoError(t, err, clues.ToCore(err))
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	assert.NotNil(t, info, "event item info")
+
+	body, err := suite.ac.Events().Serialize(ctx, sp, userID, ptr.Val(evts[0].GetId()))
+	assert.NoError(t, err, clues.ToCore(err))
+
+	event, err := api.BytesToEventable(body)
+	assert.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 423:
```suggestion
	require.NoError(t, err, clues.ToCore(err))
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+
+	require.NotNil(t, event.GetRecurrence(), "recurrence")
```
ashmrtn, line 425:
```suggestion
	assert.NotNil(t, event.GetRecurrence(), "recurrence")
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+	eo := event.GetAdditionalData()["exceptionOccurrences"]
+	require.NotNil(t, eo, "exceptionOccurrences")
+
+	require.NotEqual(
```
ashmrtn, line 429:
```suggestion
	assert.NotEqual(
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+		t,
+		ptr.Val(event.GetSubject()),
+		ptr.Val(eo.([]interface{})[0].(map[string]interface{})["subject"].(*string)),
+		"name equal",
+	)
+
+	atts := eo.([]interface{})[0].(map[string]interface{})["attachments"]
+	require.NotEqual(
```
ashmrtn, line 437:
```suggestion
	assert.NotEqual(
```

src/internal/m365/exchange/restore_test.go
```diff
@@ -369,3 +373,71 @@ func (suite *RestoreIntgSuite) TestRestoreExchangeObject() {
...
+
+	evts := ec.GetValue()
+	assert.Equal(t, len(evts), 1, "count of events")
+
+	sp, info, err := suite.ac.Events().GetItem(ctx, userID, ptr.Val(evts[0].GetId()), false, fault.New(true))
+	assert.NoError(t, err, clues.ToCore(err))
+	assert.NotNil(t, info, "event item info")
+
+	body, err := suite.ac.Events().Serialize(ctx, sp, userID, ptr.Val(evts[0].GetId()))
+	assert.NoError(t, err, clues.ToCore(err))
```
ashmrtn, line 420:
```suggestion
	require.NoError(t, err, clues.ToCore(err))
```

src/pkg/services/m365/api/events.go
```diff
@@ -192,6 +192,25 @@ func (c Events) PatchCalendar(
 	return nil
 }
 
+func (c Events) GetCalendarEvents(
```
ashmrtn, line 195: if this is only used for testing it should be defined in the test file as perhaps a wrapper around the `Events` struct (if access to otherwise private data is required). We should generally avoid adding extra functions to a struct/interface if they're only used in testing

ashmrtn (approved):

src/pkg/services/m365/api/events.go
ryanfkeepers: fwiw, this should probably get swapped out for the item enumerator being introduced for advanced restore options, too.

src/internal/m365/exchange/mock/event.go
```diff
@@ -321,7 +324,44 @@ func EventWithRecurrenceAndExceptionBytes(subject string) []byte {
 		defaultEventBody, defaultEventBodyPreview,
 		fmt.Sprintf(originalStartDateFormat, originalStartDate),
 		newTime, newTime, NoRecurrence, attendeesTmpl,
-		false, NoCancelledOccurrences, NoExceptionOccurrences,
+		NoAttachments, NoCancelledOccurrences, NoExceptionOccurrences,
+	)
+	exceptionOccurrences := fmt.Sprintf(
+		exceptionOccurrencesFormat,
+		strings.Join([]string{string(exceptionEvent)}, ","),
```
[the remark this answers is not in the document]
meain: Ahh, yup. I had initially planned for multiple attachments but realized that was unnecessary, but did not remove this bit.

src/internal/m365/exchange/events_restore.go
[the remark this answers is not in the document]
meain: This happens for large attachments as well, but that is still <150MB. This is a limit set by M365.

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+			if name == bname && bytes.Equal(content, bcontent) {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			err = client.DeleteAttachment(ctx, userID, eventID, ptr.Val(att.GetId()))
+			if err != nil {
+				return clues.Wrap(err, "deleting attachment").
```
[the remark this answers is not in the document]
meain: I'm not sure about that. Although they are not "missing" any data, we still will have incorrectly restored data. Any thoughts @ashmrtn @gmatev ?

src/internal/m365/exchange/events_restore.go
meain: It is possible that they upload the same file with a different name.

src/internal/m365/exchange/restore_test.go
meain: Hmm, we have this pattern is other tests in this file. Any reason for not deleting them though?

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
-		// attachments using ids and delete or add). We will have
-		// to get the id of the existing attachments, diff them
-		// with what we need a then create/delete items kinda like
-		// permissions
 		_, err = ac.PatchItem(ctx, userID, ptr.Val(evts[0].GetId()), evt)
 		if err != nil {
 			return clues.Wrap(err, "updating event instance")
 		}
+
+		evt, err = api.EventFromMap(instance)
```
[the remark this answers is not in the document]
meain: Good call!

src/internal/m365/exchange/restore_test.go
meain: > I tried to make a linter for this, and for some reason it's not working

If you tried addding it via forbidigo, it might not work as forbidigo is just for linting identifiers and not patterns.

src/internal/m365/exchange/events_restore.go
```diff
@@ -210,15 +212,108 @@ func updateExceptionOccurrences(
...
+
+		for _, natt := range attachments {
+			bname := ptr.Val(natt.GetName())
+
+			bcontent, err := api.GetAttachmentContent(natt)
+			if err != nil {
+				return clues.Wrap(err, "getting attachment").With("attachment_id", ptr.Val(natt.GetId()))
+			}
+
+			if name == bname && bytes.Equal(content, bcontent) {
```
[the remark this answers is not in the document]
meain: I'm not expecting us to be doing the `bytes.Equal` a lot. We will, in most cases short circuit and get false from `name == bname` and won't even compare the bytes. I didn't think we need to rework this at least from a perf standpoint.

src/internal/m365/exchange/events_restore.go
meain: `n` should be pretty small in most cases as these are attachments to events. Also, the comparison within them will mostly be just comparing two strings which should not be that expensive.
Related: https://github.com/alcionai/corso/pull/3644#discussion_r1235604141

src/internal/m365/exchange/events_restore.go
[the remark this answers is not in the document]
ashmrtn: personally I'd say report it to the user somehow (including event instance date and attachment name) but don't fail. This is data that's already present somewhere else in the system (i.e. other event instances have it) so it doesn't feel a security concern in the sense that we exposed data we shouldn't have (well, at least ignoring pathologic edge cases where all events in the series are modified in the same way)

src/internal/m365/exchange/events_restore.go
ryanfkeepers: copy as in `restore-to-copy` as a restore collision handling option, versus `skip` and `replace`.  It makes sense to delete the original data on a `replace` (on that note, do we need to diff events for instances that have changed in the current version but are not different in the backup?).  But if we're copying the data to a new event entry, should we be deleting the originals?

src/internal/m365/exchange/events_restore.go
ryanfkeepers: But that would be a different attachment.  Is it possible to change an existing attachment name?

meain: Nope, changing the name once you upload is not possible.

src/internal/m365/exchange/events_restore.go
meain: Ahh, I think there has been a misunderstanding here. This is restoring to event exceptions and we are only deleting the ones that we specifically know are deleted from that specific event exception.

src/internal/m365/exchange/events_restore.go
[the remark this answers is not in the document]
meain: I've added it as a recoverable error

Issue #3482, opened 2023-05-23
Move away from doing full delta enumeration during URL cache refresh


pandeyabs:
We have two options for doing delta queries during cache refresh events.

1. Do a full delta enumeration. 
    *  Pros: Simpler. We just do a full query during each cache refresh event. More reliable than 2
    *  Cons: Might be a huge fetch? Might also bloat up cache size. 
    
2. Use the `prevDelta` token as the base.
    * Pros: Fewer # of pages will be returned. Same initial state as original delta query done during `ProduceBackupCollections`.
    * Cons: `prevDelta` token may potentially be invalid ( extreme e.g.: backup lasts for 7+ days). Can fallback to doing full delta query for such scenarios.
    * We can throw away the new delta token obtained during cache refresh, since `prevDelta` is always going to be our base for the entirety of this backup.

Currently we are using 1 for simplicity. Switch to using 2 for better performance.

PR #3649, opened 2023-06-21
Refresh URL cache using previous delta token

pandeyabs:
Currently we do full delta enumeration during cache refresh events. It has some disadvantages as listed in https://github.com/alcionai/corso/issues/3482 .

This PR allows us to instantiate the URL cache with a delta token. This delta token is identical to the one used during onedrive item enumeration.

The URL cache might do multiple delta queries(once every hour) using the same delta token.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)
* https://github.com/alcionai/corso/issues/3482

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/url_cache_test.go
+++ b/src/internal/m365/onedrive/url_cache_test.go
@@ -91,15 +91,16 @@ func (suite *URLCacheIntegrationSuite) TestURLCacheBasic() {
 	nfid := ptr.Val(newFolder.GetId())
 
 	collectorFunc := func(
-		ctx context.Context,
-		driveID, driveName string,
-		driveItems []models.DriveItemable,
-		oldPaths map[string]string,
-		newPaths map[string]string,
-		excluded map[string]struct{},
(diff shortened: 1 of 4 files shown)


src/internal/m365/onedrive/url_cache_test.go
```diff
@@ -89,10 +90,37 @@ func (suite *URLCacheIntegrationSuite) TestURLCacheBasic() {
 
 	nfid := ptr.Val(newFolder.GetId())
 
+	collectorFunc := func(
+		ctx context.Context,
```
ryanfkeepers, line 94: To condense code: since none of these values are used, you can replace their names with `_`, or even take out the var names entirely.
```suggestion
		context.Context,
```

ryanfkeepers (approved):

PR #3660, opened 2023-06-22
exclude flaky integration test assertions

ryanfkeepers:
Comments out some particularly flaky test assertions. We have other assurances around this data already, so the exclusion is relatively safe.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E


meain (approved):
I feel like we should at least actively track fixing this as this is a relatively important check, but let me approve to unblock.

ryanfkeepers:
> I feel like we should at least actively track fixing this as this is a relatively important check, but let me approve to unblock.

Appreciated, and agreed about the fix.  I think we're going to have a difficult time improving the assertion case without also improving our stats gathering process.  But it is something to look forward to.

ashmrtn:
if we no longer have information about when this is happening will we still have a way to debug it?

ryanfkeepers:
> if we no longer have information about when this is happening will we still have a way to debug it?

Have we ever debugged any code based on these assertions?  To my recollection, we have not.

PR #3661, opened 2023-06-22
adds test coverage for events in exchange incrementals

ryanfkeepers:
We still have events excluded from the exchange
incrementals integration tests.  For future safety, these need to be part of the testing group.

However, there's a hitch.  Our primary test user is broken again, and cannot retrieve any calendars
except for the well-known set: calendars, birthdays, holidays.  We'll have to fix that before merging.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup_integration_test.go
+++ b/src/internal/operations/backup_integration_test.go
@@ -749,8 +749,7 @@ func testExchangeContinuousBackups(suite *BackupOpIntegrationSuite, toggles cont
 		categories = map[path.CategoryType][]string{
 			path.EmailCategory:    exchange.MetadataFileNames(path.EmailCategory),
 			path.ContactsCategory: exchange.MetadataFileNames(path.ContactsCategory),
-			// TODO: not currently functioning; cannot retrieve generated calendars
-			// path.EventsCategory:   exchange.MetadataFileNames(path.EventsCategory),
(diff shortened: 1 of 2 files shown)


meain (approved):

src/internal/operations/backup_integration_test.go
```diff
@@ -1262,16 +1266,21 @@ func testExchangeContinuousBackups(suite *BackupOpIntegrationSuite, toggles cont
 			checkMetadataFilesExist(t, ctx, bupID, kw, ms, atid, uidn.ID(), service, categories)
 			deeTD.CheckBackupDetails(t, ctx, bupID, whatSet, ms, ss, expectDeets, true)
 
+			// FIXME: commented tests are flaky due to interference with other tests
```
ashmrtn, line 1269: I don't think it's a concurrency issue where other test runs are interfering with this test. I was doing some debugging on this last week and found that graph is actually returning some items for the delta even when no changes were made. That alone throws off the assertions below. For the investigation I did contacts was returning items

ryanfkeepers: Oh!  That's right, my bad.  Been a bit since I've put my brain on it.  I'll update the comment.

Issue #3671, opened 2023-06-26
[Bug]: Missing sharepoint license handling


ryanfkeepers:
### What happened?

Corso needs to handle the graph api response for a missing sharepoint license.  The error response looks as follows:
```
{
    "error": {
        "code": "BadRequest",
        "message": "Tenant does not have a SPO license.",
        "innerError": {
            "request-id": "uuid",
            "date": "rfc3339"
        }
    }
}
```

### Corso Version?

any

### Where are you running Corso?

any

PR #3673, opened 2023-06-26
handle no sharepoint license err

ryanfkeepers:
adds handling for graph api responses
where the tenant does not have an active
sharepoint license.  For user's drives, this
will fall under the same "service not enabled"
behavior we use to determine if the drive can
get backed up.  For sites, this causes the
"get all sites" call to return a serviceNotEnabled error.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3671

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -183,7 +183,7 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
 	// check whether the user is able to access their onedrive drive.
 	// if they cannot, we can assume they are ineligible for onedrive backups.
 	if _, err := c.GetDefaultDrive(ctx, userID); err != nil {
-		if !clues.HasLabel(err, graph.LabelsMysiteNotFound) {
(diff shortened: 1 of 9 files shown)


juleslasarte (approved):

src/pkg/services/m365/api/users.go
```diff
@@ -183,7 +183,7 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
 	// check whether the user is able to access their onedrive drive.
 	// if they cannot, we can assume they are ineligible for onedrive backups.
 	if _, err := c.GetDefaultDrive(ctx, userID); err != nil {
-		if !clues.HasLabel(err, graph.LabelsMysiteNotFound) {
+		if !clues.HasLabel(err, graph.LabelsMysiteNotFound) || clues.HasLabel(err, graph.LabelsNoSharePointLicense) {
```
pandeyabs, line 186: Quick sanity check: Do we also need to add MysiteURLNotFound check here?

pandeyabs (approved):
Looks good. Just left one comment to sanity check

PR #3677, opened 2023-06-26
remove discovery pkg

ryanfkeepers:
Currently unused, and functionality is largely duplicated by the services/m365 package.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/m365_test.go
```diff
@@ -359,3 +362,240 @@ func (suite *m365UnitSuite) TestGetAllSites() {
...
+
+	sites, err := Sites(ctx, suite.acct, errs)
+	assert.NoError(t, err, clues.ToCore(err))
+
+	ferrs := errs.Errors()
+	assert.Nil(t, ferrs.Failure)
+	assert.Empty(t, ferrs.Recovered)
+	assert.NotEmpty(t, sites)
+}
+
```
ashmrtn, line 455: this is a duplicate of one of the other tests in this file and should be removed

```suggestion
```

src/pkg/services/m365/m365_test.go
```diff
@@ -359,3 +362,240 @@ func (suite *m365UnitSuite) TestGetAllSites() {
...
+			defer flush()
+
+			sites, err := Sites(ctx, test.acct(t), fault.New(true))
+			assert.Empty(t, sites, "returned some sites")
+			assert.NotNil(t, err)
+		})
+	}
+}
+
+func (suite *DiscoveryIntgSuite) TestUserInfo() {
```
ashmrtn, line 503: this is also a near duplicate of another test in this file though in this case the other one (can't comment on directly) should be removed since this has a superset of functionality

ashmrtn (approved):

src/pkg/services/m365/m365_test.go
ryanfkeepers: can do!

Issue #3672, opened 2023-06-26
[Bug]: Inconsistent error code when user doesn't exist


ashmrtn:
### What happened?

Corso attempts to verify if a users exists by querying the users mailbox and drive. It seems that we're getting inconsistent results from the query. Requests to one datacenter are returning a user not found error while requests to another datacenter are returning an invalid user error

Corso should handle both errors and report the user as unavailable for backup

### Corso Version?

all

### Where are you running Corso?

N/A

ashmrtn:
After some digging around it seems like there's a few odd things with the current code and test case. The [failing test](https://github.com/alcionai/corso/blob/2138c69898ba32cf8442d368144b3b0bc1a05796/src/internal/m365/discovery/discovery_test.go#L217) seems to imply we shouldn't be returning an error. The code for getting users is setup to [return an error](https://github.com/alcionai/corso/blob/2138c69898ba32cf8442d368144b3b0bc1a05796/src/pkg/services/m365/api/users.go#L201-L204) saying the user doesn't exist in the tenant. Previously we were matching the case where we [check](https://github.com/alcionai/corso/blob/2138c69898ba32cf8442d368144b3b0bc1a05796/src/pkg/services/m365/api/users.go#L206) if the mail folder is found (thus avoiding the if-block) and just [marking the user as not having an inbox and continuing](https://github.com/alcionai/corso/blob/2138c69898ba32cf8442d368144b3b0bc1a05796/src/pkg/services/m365/api/users.go#L211-L214). With the new error we hit the mail folder if-block and return the error causing the test to fail

After discussion we've decided that we should actually return an error if the user doesn't exist

PR #3678, opened 2023-06-26
Handle broader set of user not found error codes

ashmrtn:
graph has been inconsistent about what error code is returned
when the user doesn't exist. Expand the set of things we
check so that hopefully we'll get them all

Always return an error if the user doesn't exist as determined
by checking for exchange folders during GetInfo calls

Also add a little more stack trace info to returned errors

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #3672

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/graph/errors.go
```diff
@@ -128,11 +132,26 @@ func IsErrQuotaExceeded(err error) bool {
...
-	return hasErrorCode(err, requestResourceNotFound)
+	if hasErrorCode(err, requestResourceNotFound, invalidUser) {
+		return true
+	}
+
+	if hasErrorCode(err, resourceNotFound) {
+		odErr, ok := err.(odataerrors.ODataErrorable)
+		if !ok {
+			return false
+		}
```
ashmrtn, line 147: not really sure what the best strategy here is, but other errors are also checking for `resourceNotFound` so not trying to check what exactly wasn't found could lead to false-positives

src/pkg/services/m365/api/users.go
```diff
@@ -200,12 +200,12 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
 		if graph.IsErrUserNotFound(err) {
 			logger.CtxErr(ctx, err).Error("user not found")
-			return nil, clues.Stack(graph.ErrResourceOwnerNotFound, err)
+			return nil, clues.Wrap(clues.Stack(graph.ErrResourceOwnerNotFound, err), "")
 		}
 
 		if !graph.IsErrExchangeMailFolderNotFound(err) {
 			logger.CtxErr(ctx, err).Error("getting user's mail folder")
-			return nil, err
+			return nil, clues.Wrap(err, "")
```
ryanfkeepers, line 208: Should use Stack if you don't need a message.
```suggestion
			return nil, clues.Stack(err)
```

src/pkg/services/m365/api/users.go
```diff
@@ -200,12 +200,12 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
 
 		if graph.IsErrUserNotFound(err) {
 			logger.CtxErr(ctx, err).Error("user not found")
-			return nil, clues.Stack(graph.ErrResourceOwnerNotFound, err)
+			return nil, clues.Wrap(clues.Stack(graph.ErrResourceOwnerNotFound, err), "")
```
ryanfkeepers, line 203: stack already takes care of the trace; wrap isn't needed here.

src/internal/m365/graph/errors.go
```diff
@@ -38,11 +38,15 @@ const (
...
-	resyncRequired          errorCode = "ResyncRequired" // alt: resyncRequired
-	syncFolderNotFound      errorCode = "ErrorSyncFolderNotFound"
-	syncStateInvalid        errorCode = "SyncStateInvalid"
-	syncStateNotFound       errorCode = "SyncStateNotFound"
+	// Returned when we try to get the inbox of a user that doesn't exist.
+	resourceNotFound errorCode = "ResourceNotFound"
+	// Some datacenters are returning this when we try to get the inbox of a user
+	// that doesn't exist.
+	invalidUser        errorCode = "ErrorInvalidUser"
+	resyncRequired     errorCode = "ResyncRequired" // alt: resyncRequired
```
ryanfkeepers, line 46: I know it's old code, but can you remove the `alt` comment here?  We should be lowering text for comparisons.

src/internal/m365/graph/errors.go
ryanfkeepers: I think this handling is fine, but shouldn't we also now check for the exclusion in those other comparisons?  IE: certain error checks and compare just the code, while others (like this) need to compare both the code and message, and any case that checks both must always check both?

ryanfkeepers (approved):

src/pkg/services/m365/api/users.go
ashmrtn: stack doesn't seem to be giving stack trace information which is what I really want

src/internal/m365/graph/errors.go
ashmrtn: I'll follow up on this in another PR. There's probably a few things that should be adjusted beyond this

PR #3688, opened 2023-06-27
handle file-folder collision cases

ryanfkeepers:
handle onedrive item collision cases where a file to be restored is in conflict with an existing folder. Skip and Copy will function normally.  For Replace behavior, we'll defer to Copy in this situation, so that the file gets restored without changing or deleting the folder.  If a folder creation attempts to replace an item, we do a similar action and make a renamed copy of the folder instead.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/restore.go
+++ b/src/internal/m365/onedrive/restore.go
@@ -472,7 +472,7 @@ func restoreV0File(
 	collisionKeyToItemID map[string]api.DriveCollisionItem,
 	itemData data.Stream,
 ) (details.ItemInfo, error) {
-	_, itemInfo, err := restoreData(
+	_, itemInfo, err := restoreFile(
 		ctx,
 		restoreCfg,
 		rh,
@@ -505,7 +505,7 @@ func restoreV1File(
 ) (details.ItemInfo, error) {
 	trimmedName := strings.TrimSuffix(itemData.UUID(), metadata.DataFileSuffix)
 
-	itemID, itemInfo, err := restoreData(
+	itemID, itemInfo, err := restoreFile(
 		ctx,
 		restoreCfg,
 		rh,
@@ -588,7 +588,7 @@ func restoreV6File(
 		return details.ItemInfo{}, clues.New("item with empty name")
 	}
 
-	itemID, itemInfo, err := restoreData(
+	itemID, itemInfo, err := restoreFile(
 		ctx,
 		restoreCfg,
 		rh,
@@ -711,20 +711,10 @@ func createRestoreFolders(
 			continue
 		}
 
-		folderItem, err := fr.GetFolderByName(ictx, driveID, parentFolderID, folderName)
-		if err != nil && !errors.Is(err, api.ErrFolderNotFound) {
-			return "", clues.Wrap(err, "getting folder by display name")
-		}
-
-		// folder found, moving to next child
-		if err == nil {
-			parentFolderID = ptr.Val(folderItem.GetId())
-			caches.Folders.set(location, folderItem)
-
-			continue
-		}
-
-		folderItem, err = createFolder(
+		// we assume this folder creation always uses the Replace
+		// conflict policy, which means it will act as a GET if the
+		// folder already exists.
+		folderItem, err := createFolder(
 			ictx,
 			fr,
 			driveID,
@@ -794,8 +784,8 @@ type itemRestorer interface {
 	PostItemInContainerer
 }
 
-// restoreData will create a new item in the specified `parentFolderID` and upload the data.Stream
-func restoreData(
+// restoreFile will create a new item in the specified `parentFolderID` and upload the data.Stream
+func restoreFile(
 	ctx context.Context,
 	restoreCfg control.RestoreConfig,
 	ir itemRestorer,
@@ -819,7 +809,6 @@ func restoreData(
 
 	var (
 		item         = newItem(name, false)
-		isFolder     = item.GetFolder() != nil
 		collisionKey = api.DriveItemCollisionKey(item)
 		collision    api.DriveCollisionItem
 		replace      bool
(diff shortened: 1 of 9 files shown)


src/internal/m365/onedrive/restore.go
```diff
@@ -747,6 +743,50 @@ func createRestoreFolders(
...
 }
 
+func createFolder(
+	ctx context.Context,
+	piic PostItemInContainerer,
+	driveID, parentFolderID, folderName string,
+) (models.DriveItemable, error) {
+	// create the folder if not found
+	// the Replace collision policy is used since collisions on that
+	// policy will no-op and return the existing folder.  This has two
```
ashmrtn, line 753: given the amount of churn we've seen on the graph API I kind of want to say we should have a regression test checking specifically that attempting to (re)create an existing folder with items with the replace policy doesn't disturb the items already in the folder. Or if there's explicit documentation of this fact drop a link to it here so we can know the behavior going forward

src/internal/m365/onedrive/restore.go
```diff
@@ -710,7 +711,7 @@ func createRestoreFolders(
 			continue
 		}
 
-		folderItem, err := fr.GetFolderByName(ictx, driveID, parentFolderID, folder)
+		folderItem, err := fr.GetFolderByName(ictx, driveID, parentFolderID, folderName)
```
ashmrtn, line 714: this feels redundant. If we continue with using replace to make folders can you add a TODO on this to remove it?

src/internal/m365/onedrive/restore.go
```diff
@@ -779,12 +819,13 @@ func restoreData(
 
 	var (
 		item         = newItem(name, false)
+		isFolder     = item.GetFolder() != nil
```
ashmrtn, line 822: is this variable needed? The line above makes it seem like this is only restoring items because we only pass `false` to `newItem`

src/internal/m365/onedrive/restore.go
ryanfkeepers: There's no explicit documentation, but from research into the condition a replace conflict is equivalent to a GET request.  I'm trying to parse your request, and am not sure if you're suggesting we cannot trust the server-side replace behavior, or if we need a regression test to ensure we don't accidentally change things on our side.  If the former, I feel like that's stretching the value a little thin.  The latter sounds very useful, though.  Going to add a regression test either way, but am curious about the clarification here.

src/internal/m365/onedrive/restore.go
ashmrtn: it's redundant because we check the cache if we have the folder already, make this request to graph for the folder, and then send a create request with "replace" to make the folder if it doesn't exist. At best we'll send 1 graph request either way (here and find the folder or just outright make the folder). We can remove this if we assume we'll have the create folder code always use replace semantics

src/internal/m365/onedrive/restore.go
ashmrtn: I want to see a test that ensures graph behaves the way we expect it to. I want the test because otherwise if the behavior of graph does change we'll never know we're disappearing items until someone comes and tells us about it, which could take a very long time

For example, imagine the behavior of graph changed such that replace on a folder would delete the original folder (and all items/subfolders in it) and create a new folder. If someone had made a backup from a while back and then tried to do an in-place restore with it we could end up deleting items they may not have a backup for

At least if we have a test we'll know when stuff breaks so we can mitigate it. If it were documented behavior I'd be more open to just having a link to it and pointing to that. However, since it's not documented we need to know what the actual behavior will be in-practice and we need to keep checking the behavior continues to be what we expect

ashmrtn:
oh, there should also be some documentation somewhere about what to expect for file-folder collisions. It seems like behavior that may not be immediately clear to the user so we should have something to point to for it

src/internal/m365/onedrive/restore.go
ryanfkeepers: Yeah, unfortunately docs are less that skeletal when it comes to conflict behavior.  Not much we can reference there for reassurance.

src/internal/m365/onedrive/restore.go
ryanfkeepers: oh, hah, yes.  monday brain.  Will remove, thank you!

src/internal/m365/onedrive/restore.go
ryanfkeepers: @ashmrtn regression test added.  This is ready for another review, please.

ryanfkeepers:
> oh, there should also be some documentation somewhere about what to expect for file-folder collisions. It seems like behavior that may not be immediately clear to the user so we should have something to point to for it

not sure where to put that at this time.  There'll be an end-of-changes documentation PR, as part of making the cli commands available.  I'll think on how to add it as part of that.

src/pkg/services/m365/api/drive_test.go
```diff
@@ -218,3 +220,88 @@ func (suite *DriveAPISuite) TestDrives_PostItemInContainer() {
 		})
 	}
 }
+
+// purpose: ensure that creating a new folder with "replace" conflict behavior
+// makes no changes to the items which exist in that folder.
```
ashmrtn, line 225: thanks for the comment!

ashmrtn (approved):
excellent, thanks for the regression test. I know it's annoying to have to add, but it's the simplest way to make sure we don't miss a change in API behavior going forward

src/internal/m365/onedrive/restore.go
ashmrtn: no worries, the logic for all the pieces is rather scattered and difficult to pick up on

ryanfkeepers:
> I know it's annoying to have to add

Nah, wasn't bad at all.  Glad to have you point out the lack.  Better safe than buggy, yeah?

PR #3689, opened 2023-06-27
make tests less sus

ryanfkeepers:
#### Type of change

#### Test Plan

- [x] :zap: Unit test


ashmrtn (approved):
I'm a bit concerned that these tests weren't failing in the first place. Would it be possible to do a comparison on the complete set of labels in this instance? We have pretty good control over the input data set so it seems like they should really only be getting the label we expect right?

PR #3690, opened 2023-06-27
use errors.As to extract odataError

ryanfkeepers:
casting only the top level error to an odataError
will slice out useful data when the odataError is
wrapped, stacked, or otherwise nested.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/graph/errors.go
+++ b/src/internal/m365/graph/errors.go
@@ -227,7 +227,7 @@ func hasErrorCode(err error, codes ...errorCode) bool {
 		return false
 	}
 
-	var oDataError *odataerrors.ODataError
+	var oDataError odataerrors.ODataErrorable
 	if !errors.As(err, &oDataError) {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


src/internal/m365/graph/errors.go
```diff
@@ -227,7 +227,7 @@ func hasErrorCode(err error, codes ...errorCode) bool {
 		return false
 	}
 
-	var oDataError *odataerrors.ODataError
```
ashmrtn, line 230: but who wouldn't want to (sometimes) use an `***<concrete type>`? 😛

ashmrtn (approved):
could you please add some simple tests for this just so we won't get a regression in the future?

src/internal/m365/graph/errors.go
ryanfkeepers: wtb `*nil`

PR #3655, opened 2023-06-22
Backup and restore link share information

meain:
As of now we are only backing up and restoring actual permissions. This adds support for backing up and restoring link shares as well.

---

#### Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Supportability/Tests
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/3605

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Issue #3605, opened 2023-06-13
Support backing up and restoring link shares


meain:
Currently only direct permissions are backed up and restored. We should also backup and restore link shares. The [beta API](https://learn.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-beta&tabs=http) should let us do that without having email everyone every time they are associated to a link.

PR #3698, opened 2023-06-28
Follow up to handling link shares

meain:
Handle a few edge cases with link shares. Also handle empty link shares, plus inheritence works a bit differently for link shares compared to permissions.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/pull/3655
* https://github.com/alcionai/corso/issues/3605

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - If a link share is created for an item with inheritance disabled
   (via the Graph API), the link shares restored in that item will
   not be inheritable by children
-- Link shares with password protection cannot be restored
+- Link shares with password protection can't be restored
 
 ## [v0.10.0] (beta) - 2023-06-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Return a ServiceNotEnabled error when a tenant has no active SharePoint license.
 
 ### Known issues
-- Partial inheritance of link shares from a parent folder can't be restored.
+* If a link share is created for an item with inheritance disabled
+  (via the Graph API), Corso does not restore any link shares on that
+  item that may be inherited after that
 - Link shares with password cannot be restored
 
 ## [v0.10.0] (beta) - 2023-06-26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] (beta)
 
+### Added
+- Added support for backing up and restoring link shares
+
 ### Fixed
 - Return a ServiceNotEnabled error when a tenant has no active SharePoint license.
+
+### Known issues
+- Partial inheritance of link shares from a parent folder can't be restored.
+- Link shares with password cannot be restored
 
 ## [v0.10.0] (beta) - 2023-06-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,16 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 
 ### Added
-- Added support for backing up and restoring link shares
+- Drive items backup and restore link shares
 
 ### Fixed
 - Return a ServiceNotEnabled error when a tenant has no active SharePoint license.
 
 ### Known issues
-* If a link share is created for an item with inheritance disabled
-  (via the Graph API), Corso does not restore any link shares on that
-  item that may be inherited after that
-- Link shares with password cannot be restored
+- If a link share is created for an item with inheritance disabled
+  (via the Graph API), the link shares restored in that item will
+  not be inheritable by children
+- Link shares with password protection cannot be restored
 
 ## [v0.10.0] (beta) - 2023-06-26
 
--- a/src/internal/m365/onedrive/metadata/permissions.go
+++ b/src/internal/m365/onedrive/metadata/permissions.go
@@ -76,6 +76,8 @@ func (p Permission) Equals(other Permission) bool {
 //     user and so restoring links without users is not useful.
 func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
 	filteredBefore := []LinkShare{}
+	filteredAfter := []LinkShare{}
+
 	for _, ls := range before {
 		if len(ls.Entities) == 0 {
 			continue
@@ -84,7 +86,6 @@ func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
 		filteredBefore = append(filteredBefore, ls)
 	}
 
-	filteredAfter := []LinkShare{}
 	for _, ls := range after {
 		if len(ls.Entities) == 0 {
 			continue
--- a/src/internal/m365/onedrive/metadata/permissions.go
+++ b/src/internal/m365/onedrive/metadata/permissions.go
@@ -90,7 +90,7 @@ func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
 			continue
 		}
 
-		filteredAfter = append(filteredBefore, ls)
+		filteredAfter = append(filteredAfter, ls)
 	}
 
 	return DiffPermissions(filteredBefore, filteredAfter)
--- a/src/internal/m365/onedrive/metadata/permissions.go
+++ b/src/internal/m365/onedrive/metadata/permissions.go
@@ -69,47 +69,47 @@ func (p Permission) Equals(other Permission) bool {
 // filter out link shares which do not have any associated users. This
 // is useful for two reason:
 //   - When a user creates a link share on parent after creating a child
-//     link with `retainInheritedPermisisons`, all the previous link shares
+//     link with `retainInheritedPermissisons`, all the previous link shares
 //     are inherited onto the child but without any users associated with
 //     the share. We have to drop the empty ones to make sure we reset.
 //   - We are restoring link shares so that we can restore permissions for
 //     the user, but restoring links without users is not useful.
-func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
-	filteredBefore := []LinkShare{}
-	filteredAfter := []LinkShare{}
+func DiffLinkShares(current, expected []LinkShare) ([]LinkShare, []LinkShare) {
+	filteredCurrent := []LinkShare{}
+	filteredExpected := []LinkShare{}
 
-	for _, ls := range before {
+	for _, ls := range current {
 		if len(ls.Entities) == 0 {
 			continue
 		}
 
-		filteredBefore = append(filteredBefore, ls)
+		filteredCurrent = append(filteredCurrent, ls)
 	}
 
-	for _, ls := range after {
+	for _, ls := range expected {
 		if len(ls.Entities) == 0 {
 			continue
 		}
 
-		filteredAfter = append(filteredAfter, ls)
+		filteredExpected = append(filteredExpected, ls)
 	}
 
-	return DiffPermissions(filteredBefore, filteredAfter)
+	return DiffPermissions(filteredCurrent, filteredExpected)
 }
 
 // DiffPermissions compares the before and after set, returning
 // the permissions that were added and removed (in that order)
 // in the after set.
-func DiffPermissions[T interface{ Equals(T) bool }](before, after []T) ([]T, []T) {
+func DiffPermissions[T interface{ Equals(T) bool }](current, expected []T) ([]T, []T) {
 	var (
 		added   = []T{}
 		removed = []T{}
 	)
 
-	for _, cp := range after {
+	for _, cp := range expected {
 		found := false
 
-		for _, pp := range before {
+		for _, pp := range current {
 			if cp.Equals(pp) {
 				found = true
 				break
@@ -121,10 +121,10 @@ func DiffPermissions[T interface{ Equals(T) bool }](before, after []T) ([]T, []T
 		}
 	}
 
-	for _, pp := range before {
+	for _, pp := range current {
 		found := false
 
-		for _, cp := range after {
+		for _, cp := range expected {
 			if cp.Equals(pp) {
 				found = true
 				break
--- a/src/internal/m365/onedrive/metadata/permissions.go
+++ b/src/internal/m365/onedrive/metadata/permissions.go
@@ -65,6 +65,37 @@ func (p Permission) Equals(other Permission) bool {
 	return slices.Equal(p1r, p2r)
 }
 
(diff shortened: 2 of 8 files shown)


src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
...
+func computePreviousLinkShares(
+	ctx context.Context,
+	originDir path.Path,
+	parentMetas map[string]metadata.Metadata,
+) ([]metadata.LinkShare, error) {
+	var (
+		parent     path.Path
+		linkShares []metadata.LinkShare
+	)
+
```
ashmrtn, line 97: nit: we should add the starting path so we have it
```suggestion
	)

	ctx = clues.Add("item_path", originalDir)
```

src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
...
+			return nil, clues.New("getting parent").WithClues(ctx)
+		}
+
+		if parent == newParent {
+			break
+		}
+
+		parent = newParent
+
+		ictx := clues.Add(ctx, "parent_dir", parent)
```
ashmrtn, line 112: nit
```suggestion
		ictx := clues.Add(ctx, "current_ancestor_dir", parent)
```

src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
...
+			return nil, clues.New("transforming dir to drivePath").WithClues(ictx)
+		}
+
+		if len(drivePath.Folders) == 0 {
+			break
+		}
+
+		meta, ok := parentMetas[parent.String()]
+		if !ok {
+			return nil, clues.New("no metadata found for parent folder: " + parent.String()).WithClues(ictx)
```
ashmrtn, line 125: do we currently always add metadata entries? For example, if one of the ancestors of the given directory didn't have any permissions associated with it would we still make an entry in `parentMetas`?

src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
...
+		}
+
+		if meta.SharingMode == metadata.SharingModeCustom {
```
ashmrtn, line 128: does adding a link to an item automatically set it to custom sharing mode? (this should probably have a comment saying so too so others have that context)

website/docs/support/known-issues.md
```diff
@@ -28,3 +28,7 @@ included in backup and restore.
 * Sharing information of items in OneDrive/SharePoint using sharing links aren't backed up and restored.
 
 * Permissions/Access given to a site group can't be restored
+
+* If a link share is created for an item with inheritance disabled
+  (via the Graph API), Corso does not restore any link shares on that
+  item that may be inherited after that
```
ryanfkeepers, line 34:
```suggestion
  (via the Graph API), the link shares restored in that item will
  not be inheritable by children
```

src/internal/m365/onedrive/permission.go
```diff
@@ -303,26 +354,49 @@ func UpdateLinkShares(
...
 
-	return nil
+	// It is possible to have empty link shares even though we should
+	// have inherited one if the user creates a link using
+	// `retainInheritedPermissions` as false, but then deleted it.  We
+	// can recreate this by creating an anonymous one and deleting it.
+	if len(lsRemoved) > 0 && len(lsAdded) == 0 {
+		lsbody := drives.NewItemItemsItemCreateLinkPostRequestBody()
+		lsbody.SetType(ptr.To("view"))
+		lsbody.SetScope(ptr.To("anonymous"))
```
ashmrtn, line 383: I think we need to be really careful with this. For example, is it creating a link share that anyone (logged in or not) can access? It's unlikely that anyone could guess the URL, but it's still a security leak if anyone can see it. Plus, there's no way we can really bound the time the link would actually be there for because it's possible corso could crash between making the link and deleting it. Or it could even just fail to delete the link for some reason. If we can't create a link share that nobody can access then maybe we should just add this to the known issues list

Or is it possible to create a link share that only the owner of the file can access? If that's difficult or involves a lot of code add that support in another PR

src/internal/m365/onedrive/permission.go
```diff
@@ -344,36 +418,52 @@ func RestorePermissions(
...
 	}
 
-	err = UpdateLinkShares(
+	permRemoved := []metadata.Permission{}
+	permAdded := current.Permissions
+
+	if !didReset {
+		// In case we did a reset of permissions when restoring link
+		// shares, we have to make sure to restore all the permissions
+		// that an item has as they too will be removed.
```
ashmrtn, line 446: this comment seems at odds with the if-condition. Should the if-condition be flipped?

CHANGELOG.md
```diff
@@ -7,9 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added support for backing up and restoring link shares
+
 ### Fixed
 - Return a ServiceNotEnabled error when a tenant has no active SharePoint license.
 
+### Known issues
+* If a link share is created for an item with inheritance disabled
+  (via the Graph API), Corso does not restore any link shares on that
+  item that may be inherited after that
+- Link shares with password cannot be restored
```
ryanfkeepers, line 20:
```suggestion
- Link shares with password protection cannot be restored
```

CHANGELOG.md
```diff
@@ -7,9 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] (beta)
 
+### Added
+- Added support for backing up and restoring link shares
```
ryanfkeepers, line 11:
```suggestion
- Drive items backup and restore link shares
```

src/internal/m365/onedrive/metadata/permissions.go
```diff
@@ -65,6 +65,38 @@ func (p Permission) Equals(other Permission) bool {
...
 
+// DiffLinkShares is just a wrapper on top of DiffPermissions but we
+// filter out link shares which do not have any associated users. This
+// is useful for two reason:
+//   - When a user creates a link share on parent after creating a child
+//     link with `retainInheritedPermisisons`, all the previous link shares
+//     are inherited onto the child but without any users associated with
+//     the share. We have to drop the empty ones to make sure we reset.
+//   - We are restoring link shares so that we can restore permissions for
+//     user and so restoring links without users is not useful.
```
ryanfkeepers, line 76:
```suggestion
//     the user, but restoring links without users is not useful.
```

ashmrtn (changes requested):
need to figure out removing all link shares (or if we should add that to known issues) prior to merging. See comment about it for more thoughts

src/internal/m365/onedrive/metadata/permissions.go
```diff
@@ -65,6 +65,38 @@ func (p Permission) Equals(other Permission) bool {
...
+func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
+	filteredBefore := []LinkShare{}
+	filteredAfter := []LinkShare{}
+
+	for _, ls := range before {
+		if len(ls.Entities) == 0 {
```
ryanfkeepers, line 82: Can entities be something besides a user?  Such as a group, or a site?  If so, is that something we have to consider?

src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
 	return meta, nil
 }
 
+// Unlike permissions, link shares inherit them from all the parents
```
ryanfkeepers, line 86:
```suggestion
// Unlike permissions, link shares are inherited from all the parents
```

src/internal/m365/onedrive/permission.go
```diff
@@ -82,6 +83,56 @@ func getCollectionMetadata(
...
+
+	for {
+		newParent, err := parent.Dir()
+		if err != nil {
+			return nil, clues.New("getting parent").WithClues(ctx)
+		}
+
+		if parent == newParent {
+			break
+		}
```
ryanfkeepers, line 108: Minor: this method of handling the loop doesn't make it very clear what the state and conditions are.  I'd recommend the following instead:

```suggestion
	parent = originDir.Dir()

	for len(parent.Elements()) > 0 {
	
	// ... down at the bottom  of the loop
	parent, err = parent.Dir()
	if err != nil {
		return nil, clues.New("getting parent").WithClues(ctx)
	}
```

src/internal/m365/onedrive/permission.go
ryanfkeepers, line 125: Error messages should not contain formatting or concatenation.
```suggestion
			return nil, clues.New("no metadata found in parent").WithClues(ictx)
```

src/internal/m365/onedrive/permission.go
```diff
@@ -240,13 +291,13 @@ func UpdateLinkShares(
 	itemID string,
 	lsAdded, lsRemoved []metadata.LinkShare,
 	oldLinkShareIDToNewID map[string]string,
-) error {
-	// You can only delete inherited sharing links the first time you
+) (bool, error) {
+	// You can only delete inherited sharing links the didReset time you
```
ryanfkeepers, line 295: So is `didReset` the second time, or the seventh?

src/internal/m365/onedrive/permission.go
```diff
@@ -240,13 +291,13 @@ func UpdateLinkShares(
...
 	// create a sharing link which is done using
 	// `retainInheritedPermissions`. We get a 204 as a response if we
 	// try to delete an inherited link share, but it does not get
 	// deleted.
-	first := true
+	didReset := false
```
ryanfkeepers, line 300: I think this makes it a little more clear?  I dunno, what do you think?
```suggestion
	alreadyDeleted := false
```

ryanfkeepers (approved):
lgtm, just a bunch of nitpicks.

Oh, aren't there a couple tests that are Skipped due to this?  Can we un-skip those now?

src/internal/m365/onedrive/metadata/permissions.go
```diff
@@ -65,6 +65,38 @@ func (p Permission) Equals(other Permission) bool {
...
+// DiffLinkShares is just a wrapper on top of DiffPermissions but we
+// filter out link shares which do not have any associated users. This
+// is useful for two reason:
+//   - When a user creates a link share on parent after creating a child
+//     link with `retainInheritedPermisisons`, all the previous link shares
+//     are inherited onto the child but without any users associated with
+//     the share. We have to drop the empty ones to make sure we reset.
+//   - We are restoring link shares so that we can restore permissions for
+//     user and so restoring links without users is not useful.
+func DiffLinkShares(before, after []LinkShare) ([]LinkShare, []LinkShare) {
```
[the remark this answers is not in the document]
meain: I guess they are too generic. Let me rename it to `current` (the current state in graph computed from what we have in the backup for parent dirs) and `expected` (what we have to change it to).

src/internal/m365/onedrive/permission.go
meain: Yup, it will still have `parentMeta` but with "inherited" permissions.

src/internal/m365/onedrive/permission.go
meain: Any change in permissions will change the mode to 'custom'. Let me add a note though.

> `custom` is something we use to represent on our end based on the `shared` value in the item response in delta.

src/internal/m365/onedrive/permission.go
meain: Ahh, missed updating the code after flipping the condition. I did this to avoid calling `computePreviousMetadata` and `DiffPermissions` if not needed, but I think they are not that expensive so flipped it back. I think the code reads better this way.

src/internal/m365/onedrive/metadata/permissions.go
meain: It can be others, but I don't think we have to special case them as at the end of the day we are just using the ids from them.

src/internal/m365/onedrive/permission.go
meain: Find and replace fails 😅

src/internal/m365/onedrive/permission.go
meain: I guess slightly improves things although `didReset` makes more sense as a return value. I don't really have any strong opinions though. Let me change it.

website/docs/support/known-issues.md
meain: Ahh, this makes more sense as we don't have to distinguish from one created before and after.

meain:
> Oh, aren't there a couple tests that are Skipped due to this? Can we un-skip those now?

I don't think so. I don't think we had even considered link shares previously. /cc @ryanfkeepers

src/internal/m365/onedrive/permission.go
```diff
@@ -303,26 +350,52 @@ func UpdateLinkShares(
...
 		}
 
 		oldLinkShareIDToNewID[ls.ID] = ptr.Val(newLS.GetId())
 	}
 
-	return nil
+	// It is possible to have empty link shares even though we should
+	// have inherited one if the user creates a link using
+	// `retainInheritedPermissions` as false, but then deleted it.  We
+	// can recreate this by creating an anonymous one and deleting it.
```
ashmrtn, line 375: nit: stale comment
```suggestion
	// can recreate this by creating a link with no users and deleting it.
```

src/internal/m365/onedrive/permission.go
```diff
@@ -303,26 +350,52 @@ func UpdateLinkShares(
...
+	if len(lsRemoved) > 0 && len(lsAdded) == 0 {
+		lsbody := drives.NewItemItemsItemCreateLinkPostRequestBody()
+		lsbody.SetType(ptr.To("view"))
+		// creating a `users` link without any users ensure that even
+		// if we fail to delete the link there are no links lying
+		// around that could be used to access this
```
ashmrtn, line 381: thanks for the comment! Will help others understand why we chose this method

src/internal/m365/onedrive/permission.go
```diff
@@ -344,13 +417,41 @@ func RestorePermissions(
...
 	ctx = clues.Add(ctx, "permission_item_id", itemID)
 
+	previousLinkShares, err := computePreviousLinkShares(ctx, itemPath, caches.ParentDirToMeta)
+	if err != nil {
+		return clues.Wrap(err, "previous link shares")
+	}
+
+	lsAdded, lsRemoved := metadata.DiffLinkShares(previousLinkShares, current.LinkShares)
+
+	didReset, err := UpdateLinkShares(
```
ashmrtn, line 427: it looks like restoring permissions is sensitive to whether we restore link shares first or the old style permissions (based on the comment in the `if didReset` block). Can you leave a comment in the code stating that please?

ashmrtn (approved):

PR #3699, opened 2023-06-28
fix bad user info conditional on spo license

ryanfkeepers:
Primary change is fixing the spo license check in
GetUserInfo from a || to a && !.  Currently the check returns the spo license error instead of treating it as a drive access condition.

Additionally adds a root site check in the m365
isServiceEnabled call, instead of always returning true for sharepoint.  This should catch spo license issues on a per-site basis, if someone makes it that far.

Finally, adds extra testing (plus some test code
consolidation and refactoring) to ensure GetUserInfo is properly tested.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3671

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/backup.go
+++ b/src/internal/m365/backup.go
@@ -158,13 +158,21 @@ func (ctrl *Controller) IsBackupRunnable(
 	resourceOwner string,
 ) (bool, error) {
 	if service == path.SharePointService {
-		// No "enabled" check required for sharepoint
+		_, err := ctrl.AC.Sites().GetRoot(ctx)
+		if err != nil {
+			if clues.HasLabel(err, graph.LabelsNoSharePointLicense) {
+				return false, clues.Stack(graph.ErrServiceNotEnabled, err)
+			}
+
+			return false, err
+		}
+
 		return true, nil
 	}
 
 	info, err := ctrl.AC.Users().GetInfo(ctx, resourceOwner)
 	if err != nil {
-		return false, err
+		return false, clues.Stack(err)
 	}
 
 	if !info.ServiceEnabled(service) {
@@ -208,7 +216,7 @@ func checkServiceEnabled(
 
 	info, err := gi.GetInfo(ctx, resource)
 	if err != nil {
-		return false, false, err
+		return false, false, clues.Stack(err)
 	}
 
 	if !info.ServiceEnabled(service) {
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -95,7 +95,7 @@ func (c Users) GetAll(
 			return false
 		}
 
-		err := validateUser(item)
+		err := ValidateUser(item)
 		if err != nil {
 			el.AddRecoverable(ctx, graph.Wrap(ctx, err, "validating user"))
 		} else {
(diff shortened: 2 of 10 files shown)


src/pkg/services/m365/api/users.go
```diff
@@ -175,17 +175,16 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
...
 
 	// check whether the user is able to access their onedrive drive.
 	// if they cannot, we can assume they are ineligible for onedrive backups.
 	if _, err := c.GetDefaultDrive(ctx, userID); err != nil {
-		if !clues.HasLabel(err, graph.LabelsMysiteNotFound) || clues.HasLabel(err, graph.LabelsNoSharePointLicense) {
-			logger.CtxErr(ctx, err).Error("getting user's drive")
-			return nil, graph.Wrap(ctx, err, "getting user's drive")
+		if !clues.HasLabel(err, graph.LabelsMysiteNotFound) && !clues.HasLabel(err, graph.LabelsNoSharePointLicense) {
+			logger.CtxErr(ctx, err).Error("getting user's default drive")
+			return nil, graph.Wrap(ctx, err, "getting user's default drive info")
```
ashmrtn, line 187: I'm still a little confused by this. Honestly determining if drives are enabled via SP licenses in a call to GetInfo which seems mostly about individual users seems odd

src/internal/m365/backup.go
```diff
@@ -158,13 +158,21 @@ func (ctrl *Controller) IsBackupRunnable(
 	resourceOwner string,
 ) (bool, error) {
 	if service == path.SharePointService {
-		// No "enabled" check required for sharepoint
+		_, err := ctrl.AC.Sites().GetRoot(ctx)
+		if err != nil {
+			if clues.HasLabel(err, graph.LabelsNoSharePointLicense) {
+				return false, clues.Stack(graph.ErrServiceNotEnabled, err)
```
ashmrtn, line 164: I'm kind of worried that we're heading towards a split in how things being disabled is treated between different services. For Exchange we seem to be checking errors directly with functions that return bools. For SharePoint we seem to be adding labels to errors and assuming something upstream from that addition will check labels

pandeyabs (approved):
LGTM!

src/pkg/services/m365/api/users.go
ryanfkeepers: If the tenant has no SP license, then that's the error we get when retrieving user's drives, instead of a MySiteNotFound error.  We're just catching what graph gives us.  In the meantime, it's easier to translate that response to "the user cannot access drives" instead of "this user has a tenant-level error".

src/internal/m365/backup.go
ryanfkeepers: I'd say it's more like:
* api and graph handlers will return errors annotated with sentinels and labels.
* the next layer up (ie: anything in m365) should catch those sentinels and labels and handle them appropriately, returning an error if needed.
* All layers above that (operations/cli/sdk/etc) should only need to check errors.

Is the choice to use a label over a stacked error a little fuzzy?  Yeah, it is.  I don't think we have a clear pattern for prioritizing one or the other.  On the other hand, I don't think it's causing us much harm, either, since the extent to which a layer needs to know the difference is somewhat well contained.  There's outliers we can probably watch for smells; but I don't think we're into the need for refactoring yet.

ashmrtn: > On the other hand, I don't think it's causing us much harm, either, since the extent to which a layer needs to know the difference is somewhat well contained

I'm not sure I completely agree with this. I don't work on the m365 package quite as much so I'm honestly not completely sure where one layer ends and another begins. I think bugs like needing to use `errors.As` instead of a direct type cast are symptoms of not having clear interface definitions and standardized patterns. Because bugs like that are dependent on both internal e2e system behavior and inconsistent behavior by external services it's extra difficult to find and catch them. I only managed to catch the one I did because CI happened to fail

When people work in areas they aren't familiar with they're more likely to copy and modify code near what they're doing (or at least code doing similar things) to fit their needs. Standardization makes this simpler because then whether a solution works or not isn't (as) dependent on which source the "example" code happened to come from

PR #3703, opened 2023-06-28
ensure od and exchange can restore in-place

ryanfkeepers:
Adds changes (mostly to exchange) that allow in- place item restoration.  Onedrive basically already supported in-place restores, however we weren't attempting them.  Exchange needed full support of GetContainerByName across its categories.

Next steps: add CLI integration to simplify manual testing.  Then add automation testing to exercise all in-place restore conditions and configurations.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/contacts.go
+++ b/src/pkg/services/m365/api/contacts.go
@@ -37,8 +37,8 @@ type Contacts struct {
 // If successful, returns the created folder object.
 func (c Contacts) CreateContainer(
 	ctx context.Context,
-	userID, containerName string,
-	_ string, // parentContainerID needed for iface, doesn't apply to contacts
+	// parentContainerID needed for iface, doesn't apply to contacts
+	userID, _, containerName string,
 ) (graph.Container, error) {
 	body := models.NewContactFolder()
 	body.SetDisplayName(ptr.To(containerName))
@@ -120,7 +120,8 @@ func (c Contacts) GetContainerByID(
 // GetContainerByName fetches a folder by name
 func (c Contacts) GetContainerByName(
 	ctx context.Context,
-	userID, containerName string,
+	// parentContainerID needed for iface, doesn't apply to contacts
+	userID, _, containerName string,
 ) (graph.Container, error) {
 	filter := fmt.Sprintf("displayName eq '%s'", containerName)
 	options := &users.ItemContactFoldersRequestBuilderGetRequestConfiguration{
--- a/src/pkg/services/m365/api/contacts.go
+++ b/src/pkg/services/m365/api/contacts.go
@@ -117,6 +117,55 @@ func (c Contacts) GetContainerByID(
 	return c.GetFolder(ctx, userID, containerID)
 }
 
+// GetContainerByName fetches a folder by name
+func (c Contacts) GetContainerByName(
+	ctx context.Context,
+	userID, containerName string,
+) (graph.Container, error) {
+	filter := fmt.Sprintf("displayName eq '%s'", containerName)
+	options := &users.ItemContactFoldersRequestBuilderGetRequestConfiguration{
+		QueryParameters: &users.ItemContactFoldersRequestBuilderGetQueryParameters{
+			Filter: &filter,
+		},
+	}
+
+	ctx = clues.Add(ctx, "container_name", containerName)
+
+	resp, err := c.Stable.
+		Client().
+		Users().
+		ByUserId(userID).
+		ContactFolders().
+		Get(ctx, options)
+	if err != nil {
+		return nil, graph.Stack(ctx, err).WithClues(ctx)
+	}
+
+	gv := resp.GetValue()
+
+	if len(gv) == 0 {
+		return nil, clues.New("container not found").WithClues(ctx)
+	}
+
+	// We only allow the api to match one container with the provided name.
+	// Return an error if multiple container exist (unlikely) or if no container
+	// is found.
+	if len(gv) != 1 {
+		return nil, clues.New("unexpected number of folders returned").
+			With("returned_container_count", len(gv)).
+			WithClues(ctx)
+	}
+
+	// Sanity check ID and name
+	container := gv[0]
+
+	if err := graph.CheckIDAndName(container); err != nil {
+		return nil, clues.Stack(err).WithClues(ctx)
+	}
+
+	return container, nil
+}
+
 func (c Contacts) PatchFolder(
 	ctx context.Context,
 	userID, containerID string,
--- a/src/pkg/services/m365/api/events.go
+++ b/src/pkg/services/m365/api/events.go
@@ -44,8 +44,8 @@ type Events struct {
 // Reference: https://docs.microsoft.com/en-us/graph/api/user-post-calendars?view=graph-rest-1.0&tabs=go
 func (c Events) CreateContainer(
 	ctx context.Context,
-	userID, containerName string,
-	_ string, // parentContainerID needed for iface, doesn't apply to contacts
+	// parentContainerID needed for iface, doesn't apply to events
+	userID, _, containerName string,
 ) (graph.Container, error) {
 	body := models.NewCalendar()
 	body.SetName(&containerName)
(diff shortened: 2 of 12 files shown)


ryanfkeepers:
@pandeyabs I've been isolating reviews to Ash and Abin here to minimize cognitive shuffling.  Brought you in on this one because I know you had done work on the exchange GetContainerByName() behavior in the past.  Don't worry too much about catching up with all the advanced restore options changes, just focus on the GetByName hookups, if you would be so kind.  Thanks!

src/pkg/services/m365/api/events.go
```diff
@@ -153,24 +153,30 @@ func (c Events) GetContainerByName(
...
+	gv := resp.GetValue()
+
+	if len(gv) == 0 {
+		return nil, clues.New("container not found").WithClues(ctx)
+	}
+
+	// We only allow the api to match one calendar with the provided name.
 	// Return an error if multiple calendars exist (unlikely) or if no calendar
 	// is found.
 	if len(resp.GetValue()) != 1 {
```
ashmrtn, line 165: is this going to block restores for users that already have multiple calendars with the same name?

src/pkg/services/m365/api/mail.go
```diff
@@ -165,6 +165,55 @@ func (c Mail) GetContainerByID(
...
+	}
+
+	ctx = clues.Add(ctx, "container_name", containerName)
+
+	resp, err := c.Stable.
+		Client().
+		Users().
+		ByUserId(userID).
+		MailFolders().
+		Get(ctx, options)
```
ashmrtn, line 187: is this a recursive enumeration of all mail folders? Is it meant to be recursive?

src/pkg/services/m365/api/mail.go
```diff
@@ -165,6 +165,55 @@ func (c Mail) GetContainerByID(
...
+	gv := resp.GetValue()
+
+	if len(gv) == 0 {
+		return nil, clues.New("container not found").WithClues(ctx)
+	}
+
+	// We only allow the api to match one container with the provided name.
+	// Return an error if multiple container exist (unlikely) or if no container
+	// is found.
+	if len(gv) != 1 {
```
ashmrtn, line 201: given that mail allows a complete folder hierarchy this doesn't seem like a reasonable assumption to make

src/pkg/services/m365/api/events.go
ryanfkeepers: ooh, good call.  I suppose we're best off just taking the first of many.

src/pkg/services/m365/api/mail.go
ryanfkeepers: It is not recursive.  That works for now because our restores are strictly top-level folders.  I'll add a TODO that the later implementation will need to expand on that.  I don't think we'll want a recursive query at that time either.  But mailFolders within a given parent ID will work.

src/pkg/services/m365/api/mail.go
ryanfkeepers: Since the request is non-recursive and only includes top level folders (and mailfolders cannot have duplicate names), this is safe.  I'll add a comment explaining those assumptions.

src/pkg/services/m365/api/mail.go
ashmrtn: > That works for now because our restores are strictly top-level folders

I don't think this statement fits with how this function is being used. Looking at the code, this function is called in `getOrPopulateContainer` which is called for _any_ container we make in `createDestination`, whether or not it's a top-level container. This will work for events (and probably for contacts) because for those we don't expect a full hierarchy. However, it's quite possible it will result in an incorrect hierarchy for mail because it may end up selecting the wrong folder (or no folder at all), especially if it's supposed to be looking for a nested folder

ryanfkeepers: oh hell, thanks, good catch.

src/pkg/services/m365/api/mail_test.go
```diff
@@ -411,3 +397,32 @@ func (suite *MailAPIIntgSuite) TestRestoreLargeAttachment() {
...
 	require.NotEmpty(t, id, "empty id for large attachment")
 }
+
+func (suite *MailAPIIntgSuite) TestMail_GetContainerByName() {
+	table := []struct {
+		name      string
+		expectErr assert.ErrorAssertionFunc
+	}{
+		{
+			name:      "Inbox",
```
pandeyabs, line 407: Can we add a test to check for type sensitivity?

pandeyabs (commented):
I've reviewed `GetContainerByName` definitions, and they look good to me! 
Just one suggestion to add tests which verify type sensitivity.

src/pkg/services/m365/api/mail.go
ryanfkeepers: Okay, this is now fixed.  The interface takes an additional "parentContainerID", so we'll always try to find folder names within a specific container.

src/pkg/services/m365/api/mail_test.go
ryanfkeepers: These are case insensitive.

ashmrtn:
I think this needs rebasing

src/pkg/services/m365/api/contacts.go
```diff
@@ -117,6 +117,56 @@ func (c Contacts) GetContainerByID(
...
+	gv := resp.GetValue()
+
+	if len(gv) == 0 {
+		return nil, clues.New("container not found").WithClues(ctx)
+	}
+
+	// We only allow the api to match one container with the provided name.
+	// Return an error if multiple container exist (unlikely) or if no container
+	// is found.
+	if len(gv) != 1 {
```
ashmrtn, line 154: are there any (possibly manual) tests we need to do to ensure this won't block restores if the user has contacts folders that we haven't considered at the moment? For example, would something like `contacts/contacts` possible? Would it throw this off?

We can probably punt on folder nesting, but we should add it as a known issues

src/pkg/services/m365/api/events.go
```diff
@@ -153,24 +154,31 @@ func (c Events) GetContainerByName(
...
+	if len(gv) == 0 {
+		return nil, clues.New("container not found").WithClues(ctx)
+	}
+
+	// Return an error if no calendar is found.
+	if len(resp.GetValue()) == 0 {
+		return nil, clues.New("unexpected number of calendars returned").
+			With("returned_container_count", len(gv)).
+			WithClues(ctx)
 	}
```
ashmrtn, line 169: this is redundant now
```suggestion
```

src/pkg/services/m365/api/events.go
```diff
@@ -153,24 +154,31 @@ func (c Events) GetContainerByName(
...
 
+	// We only allow the api to match one calendar with the provided name.
+	// If we match multiples, we'll eagerly return the first one.
```
ashmrtn, line 171: just for the sake of debuggability could you add a log call with the number of results returned. It can be debug level

src/pkg/services/m365/api/mail_test.go
```diff
@@ -411,3 +397,112 @@ func (suite *MailAPIIntgSuite) TestRestoreLargeAttachment() {
 	require.NoError(t, err, clues.ToCore(err))
 	require.NotEmpty(t, id, "empty id for large attachment")
 }
+
+func (suite *MailAPIIntgSuite) TestMail_GetContainerByName() {
```
ashmrtn, line 401: please add a subfolder to `mail_get_container_by_name` which has the same name as the parent and add a test case to fetch the child folder by name

ashmrtn (approved):

src/pkg/services/m365/api/events.go
ryanfkeepers: d'oh, thanks.

src/pkg/services/m365/api/contacts.go
ryanfkeepers: Should be safe.  `filter=displayName eq 'contacts'` doesn't return anything normally.  No reason for it to get confused if there's a folder named 'contacts'.

src/pkg/services/m365/api/mail_test.go
ryanfkeepers: done

PR #3704, opened 2023-06-28
adds cli flags for restore config

ryanfkeepers:
implements cli integration for configuring restore operation collision policy and restore destination.

flags are hidden and will be made visible at a later time.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/utils/restore_config.go
```diff
@@ -0,0 +1,61 @@
...
+func makeRestoreCfgOpts(cmd *cobra.Command) RestoreCfgOpts {
+	return RestoreCfgOpts{
+		Collisions:  flags.CollisionsFV,
+		Destination: flags.DestinationFV,
+
+		Populated: flags.GetPopulatedFlags(cmd),
+	}
+}
+
+// validateRestoreConfigFlags checks common flags for correctness and interdependencies
```
ashmrtn, line 31:
```suggestion
// validateRestoreConfigFlags checks common restore flags for correctness and interdependencies.
```

src/cli/utils/restore_config.go
```diff
@@ -0,0 +1,61 @@
...
+func validateRestoreConfigFlags(opts RestoreCfgOpts) error {
+	_, populated := opts.Populated[flags.CollisionsFN]
+	_, valid := control.ValidCollisionPolicies()[control.CollisionPolicy(flags.CollisionsFV)]
+
+	if populated && valid {
```
ashmrtn, line 36: Seems like this needs adjusted. I'm taking a guess here
```suggestion
	if !populated || !valid {
```

src/cli/utils/restore_config.go
```diff
@@ -0,0 +1,61 @@
...
+		return clues.New("invalid entry for " + flags.CollisionsFN)
```
ashmrtn, line 37: Nit: we should be more specific about what was wrong
```suggestion
		return clues.New("invalid collision value: " + flags.CollisionsFN)
```

src/cli/utils/restore_config.go
```diff
@@ -0,0 +1,61 @@
...
+
+	Populated flags.PopulatedFlags
+}
+
+func makeRestoreCfgOpts(cmd *cobra.Command) RestoreCfgOpts {
+	return RestoreCfgOpts{
+		Collisions:  flags.CollisionsFV,
+		Destination: flags.DestinationFV,
+
+		Populated: flags.GetPopulatedFlags(cmd),
```
ashmrtn, line 27: This needs a comment about what it means. Does it mean there was a value for the flag at all (even if it was the default), the user gave a value (even if it matched the default), or the user gave a value different from the default?

ashmrtn (approved):

src/cli/flags/restore_config.go
```diff
@@ -0,0 +1,30 @@
...
+// AddRestoreConfigFlags adds the restore config flag set.
+func AddRestoreConfigFlags(cmd *cobra.Command) {
+	fs := cmd.Flags()
+	fs.StringVar(
+		&CollisionsFV, CollisionsFN, string(control.Skip),
+		"How to handle item collisions: "+string(control.Skip)+", "+string(control.Copy)+", or "+string(control.Replace))
+	cobra.CheckErr(fs.MarkHidden(CollisionsFN))
+	fs.StringVar(
+		&DestinationFV, DestinationFN, "",
+		"Overrides the destination where items get restored.  '/' places items back in their original location.")
```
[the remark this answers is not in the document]
ryanfkeepers: yes, no.  This will need longer explanation somewhere, but I'm not sure where to put it, and I don't think this is the right place, per se.  Will be working with Georgi to figure that out.

src/cli/utils/restore_config.go
ryanfkeepers: bargle.  `if populated && !valid {`

PR #3713, opened 2023-06-29
fix restore to dest with leading slash

ryanfkeepers:
restoring to "--destination /foo" errors, since
drive folders aren't allowed to contain leading
slashes in their name.  This PR oversteps the
correction a little to remove the leading slash
prefix from all destination declarations.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/control/restore.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+// EnsureRestoreConfigDefaults sets all non-supported values in the config
+// struct to the default value.
+func EnsureRestoreConfigDefaults(
+	ctx context.Context,
+	rc RestoreConfig,
+) RestoreConfig {
+	if !slices.Contains([]CollisionPolicy{Skip, Copy, Replace}, rc.OnCollision) {
```
ashmrtn, line 73: nit: use the function already defined. The fewer places we have to list the valid options the better
```suggestion
	if !slices.Contains(maps.Keys(ValidCollisionPolicies()), rc.OnCollision) {
```

src/pkg/control/restore.go
```diff
@@ -0,0 +1,86 @@
...
+		logger.Ctx(ctx).
+			With(
+				"bad_collision_policy", rc.OnCollision,
+				"default_collision_policy", Skip).
+			Info("setting collision policy to default")
+
+		rc.OnCollision = Skip
```
ashmrtn, line 80: I'm a bit on the fence about this. This PR can go through as is, but would be good to check with others about the behavior. I'm concerned that not returning an error here could lead to unexpected results and difficulty tracking down bugs because everything seems to be working

src/pkg/control/options_test.go
ashmrtn, line 1: this seems like a duplicate of `restore_test.go`? It can be removed if it is. If it isn't the tests here should be moved to the other file since they're really about restore config stuff

src/pkg/control/restore.go
ryanfkeepers: I'm not so concerned, because the only way this'll happen is if: 1. we introduce a new policy (highly unlikely) but forget to add it to the valid set; or 2, if the caller sets the value to a typecast string instead of an enum (also unlikely).  If this were an arbitrary string, instead of a typed enum, I'd have more concern over it.

ashmrtn: forgetting to set it in the struct is also an option. In which case things will work, but may not be what's expected and finding the issue could be difficult

ryanfkeepers: `Forgetting` is always a toss-up for me, since it overlaps conceptually with `intentionally unset`, and I don't recall that we've ever standardized our configuration around required population; opting instead for falling back to defaults.

PR #3715, opened 2023-06-29
fix concurrency limiter init race

ryanfkeepers:
initializes the concurrency limiter as part of CLI pre-run initialization to ensure we don't accidentally skip out on including the limiter in middleware.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3695

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -264,14 +264,18 @@ func runBackups(
 
 // genericDeleteCommand is a helper function that all services can use
 // for the removal of an entry from the repository
-func genericDeleteCommand(cmd *cobra.Command, bID, designation string, args []string) error {
+func genericDeleteCommand(
+	cmd *cobra.Command,
+	pst path.ServiceType,
+	bID, designation string,
+	args []string) error {
 	if utils.HasNoFlagsAndShownHelp(cmd) {
 		return nil
 	}
 
 	ctx := clues.Add(cmd.Context(), "delete_backup_id", bID)
 
-	r, _, _, err := utils.GetAccountAndConnect(ctx, repo.S3Overrides(cmd))
+	r, _, _, err := utils.GetAccountAndConnect(ctx, pst, repo.S3Overrides(cmd))
 	if err != nil {
 		return Only(ctx, err)
 	}
@@ -289,10 +293,15 @@ func genericDeleteCommand(cmd *cobra.Command, bID, designation string, args []st
 
 // genericListCommand is a helper function that all services can use
 // to display the backup IDs saved within the repository
-func genericListCommand(cmd *cobra.Command, bID string, service path.ServiceType, args []string) error {
+func genericListCommand(
+	cmd *cobra.Command,
+	bID string,
+	service path.ServiceType,
+	args []string,
+) error {
 	ctx := cmd.Context()
 
-	r, _, _, err := utils.GetAccountAndConnect(ctx, repo.S3Overrides(cmd))
+	r, _, _, err := utils.GetAccountAndConnect(ctx, service, repo.S3Overrides(cmd))
 	if err != nil {
 		return Only(ctx, err)
 	}
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -268,7 +268,8 @@ func genericDeleteCommand(
 	cmd *cobra.Command,
 	pst path.ServiceType,
 	bID, designation string,
-	args []string) error {
+	args []string,
+) error {
 	if utils.HasNoFlagsAndShownHelp(cmd) {
 		return nil
 	}
--- a/src/cli/utils/options.go
+++ b/src/cli/utils/options.go
@@ -1,10 +1,7 @@
 package utils
 
 import (
-	"golang.org/x/exp/slices"
-
 	"github.com/alcionai/corso/src/cli/flags"
-	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/control"
 )
 
@@ -26,13 +23,4 @@ func Control() control.Options {
 	opt.Parallelism.ItemFetch = flags.FetchParallelismFV
 
 	return opt
-}
-
-func InitConcurrencyControls(args []string) {
-	// FIXME: normally we want to control the limit according to the
-	// control options.  But we're racing for initialization here, so
-	// we'll set it to the default max as a starting point.
-	if slices.Contains(args, "exchange") {
-		graph.InitializeConcurrencyLimiter(4)
-	}
 }
--- a/src/cli/utils/options.go
+++ b/src/cli/utils/options.go
@@ -1,7 +1,10 @@
 package utils
 
 import (
+	"golang.org/x/exp/slices"
+
 	"github.com/alcionai/corso/src/cli/flags"
+	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/control"
 )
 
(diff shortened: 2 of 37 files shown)


src/cli/utils/options.go
```diff
@@ -24,3 +27,12 @@ func Control() control.Options {
 
 	return opt
 }
+
+func InitConcurrencyControls(args []string) {
+	// FIXME: normally we want to control the limit according to the
+	// control options.  But we're racing for initialization here, so
+	// we'll set it to the default max as a starting point.
+	if slices.Contains(args, "exchange") {
+		graph.InitializeConcurrencyLimiter(4)
```
pandeyabs, line 36: We won't be able to change limiter capacity later, since this is a singleton and can only be initialized once?

ryanfkeepers: I've updated this to ensure we pass along the control options before initializing, so we shouldn't have any need to update the semaphore capacity afterward.

src/cli/utils/utils.go
```diff
@@ -38,14 +39,19 @@ func GetAccountAndConnect(
 		return nil, nil, nil, clues.Wrap(err, "connecting to the "+cfg.Storage.Provider.String()+" repository")
 	}
 
+	if _, err := r.ConnectToM365(ctx, pst); err != nil {
```
pandeyabs, line 42: What's the purpose of this call? Is this like a dry run/sanity check on m365 connection?

src/pkg/repository/repository.go
```diff
@@ -585,6 +590,19 @@ func deleteBackup(
 	return sw.DeleteBackup(ctx, model.StableID(id))
 }
 
+// BackupDetails returns the specified backup.Details
```
pandeyabs, line 593: nit: Func description is wrong

src/internal/m365/graph/concurrency_middleware.go
```diff
@@ -20,34 +20,41 @@ import (
...
-	return &concurrencyLimiter{
+	return &concurrencyLimiterMiddleware{
 		semaphore: make(chan struct{}, capacity),
 	}
 }
 
 func InitializeConcurrencyLimiter(capacity int) {
 	once.Do(func() {
-		concurrencyLim = generateConcurrencyLimiter(capacity)
+		concurrencyLimitMiddlewareSingleton.semaphore = generateConcurrencyLimiter(capacity).semaphore
```
pandeyabs, line 46: Can we add a `turning on concurrency limiter` log here?

src/internal/m365/graph/concurrency_middleware.go
```diff
@@ -20,34 +20,41 @@ import (
...
 	})
 }
 
-func (cl *concurrencyLimiter) Intercept(
+func TurnOffConcurrencyLimiter(ctx context.Context) {
```
pandeyabs, line 50: Having two functions modify the `concurrencyLimitMiddlewareSingleton` feels wrong. It gives reader an illusion that they can toggle the singleton state during a run, from on to off & vice versa. 

Alternate proposal: What if we supply an `enabled` flag to `InitializeConcurrencyLimiter`. That way the singleton will be only initialized once, based on flag state?

src/internal/m365/sharepoint/backup_test.go
```diff
@@ -174,11 +175,14 @@ func TestSharePointPagesSuite(t *testing.T) {
...
 		Suite: tester.NewIntegrationSuite(
 			t,
-			[][]string{tester.M365AcctCredEnvs},
-		),
+			[][]string{tester.M365AcctCredEnvs}),
 	})
 }
 
+func (suite *SharePointPagesSuite) SetupSuite() {
+	graph.InitializeConcurrencyLimiter(4)
```
pandeyabs, line 183: What's the reason behind specifying limiter initialization for various tests? It feels like a bad pattern. 

Also, side note: We need to disable limiter for sharepoint.

pandeyabs (changes requested):

pandeyabs:
I like the overall placement of concurrency limiter initialization, but i have reservations against 
- having two functions initialize the same singleton instance. 
- CL initialize calls in individual test suites

src/internal/m365/sharepoint/backup_test.go
pandeyabs: nvm, looks like you are testing whether we turned off the limiter for sharepoint/OD?
A comment would help add some context.

src/internal/m365/graph/concurrency_middleware.go
```diff
@@ -20,34 +20,41 @@ import (
...
-	once                  sync.Once
-	concurrencyLim        *concurrencyLimiter
-	maxConcurrentRequests = 4
+	once                                sync.Once
+	concurrencyLimitMiddlewareSingleton = &concurrencyLimiterMiddleware{}
+	maxConcurrentRequests               = 4
 )
 
-func generateConcurrencyLimiter(capacity int) *concurrencyLimiter {
+func generateConcurrencyLimiter(capacity int) *concurrencyLimiterMiddleware {
```
ashmrtn, line 34: this function should be reworked to return just the channel. I don't see a reason given the code we have right now to return a full middleware and then pull out one of the internal fields of it while discarding the rest

src/internal/m365/controller_test.go
```diff
@@ -283,7 +283,7 @@ func (suite *ControllerIntegrationSuite) SetupSuite() {
 	ctx, flush := tester.NewContext(t)
 	defer flush()
 
-	suite.ctrl = newController(ctx, t, resource.Users)
+	suite.ctrl = newController(ctx, t, resource.Users, path.ExchangeService)
```
pandeyabs, line 286: Not a blocker, but we should add integration tests with OD & SP services too

src/cli/backup/backup.go
```diff
@@ -264,14 +264,18 @@ func runBackups(
 
 // genericDeleteCommand is a helper function that all services can use
 // for the removal of an entry from the repository
-func genericDeleteCommand(cmd *cobra.Command, bID, designation string, args []string) error {
+func genericDeleteCommand(
+	cmd *cobra.Command,
+	pst path.ServiceType,
+	bID, designation string,
```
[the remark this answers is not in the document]
ryanfkeepers: going to keep it in just for user display, since the serviceType const isn't necessarily correctly formatted.

src/cli/utils/utils.go
ryanfkeepers: It will initialize the graph api clients, including enabling the concurrency limiter.  I'll add a comment for clarification.

src/internal/m365/graph/concurrency_middleware.go
ryanfkeepers: done

src/internal/m365/sharepoint/backup_test.go
ryanfkeepers: > What's the reason behind specifying limiter initialization for various tests? It feels like a bad pattern & can be accidentally missed for newer tests.

If we don't initialize it, the test will error  out.  When I have some time, I'm going to standardize around a test setup factory [like this](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/helper_test.go#L85), and we'll be able to include it there easily.  In the meantime, I'm just going to hack in the calls.

> Also, side note: We need to disable limiter for sharepoint.

In tests?  I don't think that's a need so much as a want.

src/internal/m365/graph/concurrency_middleware.go
ryanfkeepers: I'm not against the rework, but it does give us a weird sort of regression protection in this state.  Since we initialize a non-nil middleware struct holding a nil channel, we can signal for an unconfigured (trinary) state to catch occasions where we forget or slice out initialization of the middleware.  An  absent signal is what led to this bug in the first place.

src/internal/m365/controller_test.go
ryanfkeepers: I don't think that's needed, imo.  It'd be testing a single conditional, and we'd have to jump through hoops to extract signal on the different results.

ryanfkeepers:
@pandeyabs this is ready for a final review, please and thank you!

pandeyabs (approved):
LGTM! Thanks for finding & fixing this!

PR #3722, opened 2023-06-29
add counter for skip-collision counting

ryanfkeepers:
Introduces a counting bus, and threads it into restore operations so that we can count the number of
collision skips that occur.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cmd/factory/impl/common.go
```diff
@@ -63,6 +64,7 @@ func generateAndRestoreItems(
 	dbf dataBuilderFunc,
 	opts control.Options,
 	errs *fault.Bus,
+	cb *count.Bus,
```
ashmrtn, line 67: nit: use a different variable name. `cb` often represents `callback` which can make skimming confusing

src/pkg/count/count.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if b.parent != nil {
+		return b.parent.Total(k)
+	}
+
+	return b.Get(k)
+}
+
+// Values returns a map of all local values.
```
ashmrtn, line 79: nit: should note this is not a consistent snapshot of values (same with `TotalValues`)

src/pkg/count/keys.go
```diff
@@ -0,0 +1,8 @@
+package count
+
+type key string
+
+const (
+	CollisionSkip key = "collision-skip"
+	Test          key = "just-for-testing"
```
ashmrtn, line 7: this should be defined only in the test file

ashmrtn (approved):

src/cmd/factory/impl/common.go
ryanfkeepers: good thinking, will do

PR #3726, opened 2023-06-30
don't match config in init repo

neha-Gupta1:
In case of init repo - don't match values from config file.

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual

### DIFF
--- a/src/cli/config/account.go
+++ b/src/cli/config/account.go
@@ -40,6 +40,7 @@ func m365Overrides(in map[string]string) map[string]string {
 func configureAccount(
 	vpr *viper.Viper,
 	readConfigFromViper bool,
+	matchFromConfig bool,
 	overrides map[string]string,
 ) (account.Account, error) {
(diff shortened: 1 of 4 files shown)


meain:
Looks like the CI is running into some trouble with this change.

neha-Gupta1:
> Looks like the CI is running into some trouble with this change.

Its fixed now

src/cli/config/account.go
```diff
@@ -40,6 +40,7 @@ func m365Overrides(in map[string]string) map[string]string {
 func configureAccount(
 	vpr *viper.Viper,
 	readConfigFromViper bool,
```
ryanfkeepers, line 42: this is unused now, yeah?

ryanfkeepers (approved):

PR #3727, opened 2023-06-30
Test cases for fetching configs from config file and flags

neha-Gupta1:
Add test-cases to fetch config values from config file or flags

- [ ] :no_entry: No

#### Type of change

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :zap: Unit test


src/cli/config/config_test.go
```diff
@@ -26,15 +29,20 @@ const (
...
+` + AzureClientID + ` = '%s'
+` + AzureSecret + ` = '%s'
+` + DisableTLSKey + ` = '%s'
+` + DisableTLSVerificationKey + ` = '%s'
 `
 )
 
+// b, tID, accKey, secret,
+// 		token, passphrase, azureClientID, azureSecret,
+
```
ryanfkeepers, line 45: not sure what this comment is for.  can it be removed?

src/cli/config/config_test.go
```diff
@@ -223,6 +243,104 @@ func (suite *ConfigSuite) TestMustMatchConfig() {
...
+	assert.Equal(t, str.ParseBool(overrides[storage.DoNotVerifyTLS]), s3Cfg.DoNotVerifyTLS)
+
+	assert.Equal(t, flags.AzureClientIDFV, m365Config.AzureClientID)
+	assert.Equal(t, flags.AzureClientSecretFV, m365Config.AzureClientSecret)
+	assert.Equal(t, flags.AzureClientTenantFV, m365Config.AzureTenantID)
+
+	assert.Equal(t, flags.CorsoPassphraseFV, pass)
+
+	// reset values
+	flags.AzureClientTenantFV = ""
```
ryanfkeepers, line 333: this makes me a little uneasy, because it can be easy to cross-contaminate other tests unintentionally.  In the long, long term we'll probably not want to use globals for flag values.  But that's out of scope of this change.

For this test, can you please move these resets into a deferred func on line 290?  That'll ensure that this gets changed even in the event of an early failure  Or maybe better than that, the tester library has an "after every test" hook that you can put these into.

ryanfkeepers (approved):

src/cli/config/config_test.go
neha-Gupta1: Yes. both these things make sense. Global flag values is not that great idea. Meanwhile I will move those statements to defer.

PR #3729, opened 2023-06-30
Add Corso extensions package

pandeyabs:
Introduces interfaces for corso extensions, which is basically wrappers around `io.ReadCloser`.  Will add integration in follow up PRs.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+
+		extRc, err := factory(ctx, rc, info, extInfo)
+		if err != nil {
+			return nil, nil, clues.Wrap(err, "calling extension factory")
+		}
+
+		rc = extRc
+	}
+
+	logger.Ctx(ctx).Info("added extensions")
```
zackrossman, line 123: nit: "added len(factories) extension(s)"

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (l *loggerExtension) Read(p []byte) (int, error) {
+	n, err := l.innerRc.Read(p)
+	if err != nil && err != io.EOF {
+		logger.CtxErr(l.ctx, err).Error("inner read")
+		return n, err
+	}
+
+	if err == io.EOF {
+		logger.Ctx(l.ctx).Debug("corso extensions: EOF")
```
zackrossman, line 55: could you also add a debug log for `if n == 0  `

src/pkg/backup/details/details.go
```diff
@@ -1044,3 +1044,8 @@ func updateFolderWithinDrive(
 
 	return nil
 }
+
+// ExtensionInfo describes extension data associated with an item
+type ExtensionInfo struct {
+	Data map[string]any `json:"data,omitempty"`
```
zackrossman, line 1050: Is the plan to add an `ExtensionInfo` field to the core item info structs? Doesn't need to be done in this PR but would like to understand exactly how this struct is going to be linked to the item info (e.g. OneDrive)

pandeyabs: yep! I'll link `ExtensionInfo` to item info in a follow up PR. Trying to keep PRs short.

src/pkg/extensions/extensions.go
pandeyabs: This [line](https://github.com/alcionai/corso/pull/3729/files#diff-472771baf60efc2ad0acf325d3459a1c00bb0df873c2641ca0663c998bee3151R104) should add it for all following logs using that `ctx`

ryanfkeepers, line 123: Can we drop this to debug?

src/pkg/extensions/extensions_test.go
```diff
@@ -0,0 +1,184 @@
...
+			ctx, flush := tester.NewContext(t)
+			defer flush()
+			ith := &ItemExtensionHandler{}
+
+			extRc, extInfo, err := ith.AddItemExtensions(
+				ctx,
+				test.rc,
+				testItemInfo,
+				test.factories)
+			require.True(t, test.validateOutputs(extRc, extInfo, err))
```
ryanfkeepers, line 178: For full coverage, this test should read and close the reader on each run (which means you'll also need to construct a new base reader).  That will allow you to:
1. ensure the item is read properly.
2. exercise tests for different read chunks and ensure that works properly.
3. exercise the close func.

src/pkg/extensions/extensions_test.go
```diff
@@ -0,0 +1,184 @@
...
+		})
+	}
+}
+
+// TODO: tests for loggerExtension
+// TODO: Tests to verify RC wrapper ordering by AddItemExtensioner
```
ryanfkeepers, line 184:
```suggestion
// TODO(pandeyabs): tests for loggerExtension
// TODO(pandeyabs): Tests to verify RC wrapper ordering by AddItemExtensioner
```

ryanfkeepers (approved):
Approving to unblock, since I don't see anything that will be outwardly problematic, but I also have the following concerns:
1. (major) the current tests are too weak, and should be stressing not only the application of extensions, but also the use them for reading and closing.
2. I think the implementation could be cut down by half by removing some of the interfaces/structs in favor of plain functions with no loss to testing or functionality.
3. (minor) I'm dubious about the gains we're getting out of the logging extension.  Your instinct to add it is great.  But ultimately I think it'll produce mostly chaff and minimal signal.

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+// Extension client interface
+type CorsoItemExtension interface {
+	io.ReadCloser
+}
+
+type CorsoItemExtensionFactory func(
+	context.Context,
+	io.ReadCloser,
+	details.ItemInfo,
+	*details.ExtensionInfo,
```
ryanfkeepers, line 22: imo, this should be an interface, not a pointer to the struct itself.  That way you can provide the user something simple like this, and no one needs to worry about concrete implementation.  Also useful for testing, in this case.
```
type ExtensionKVStorer interface {
func StoreExtensionKV(k string, v any)
}
```

src/pkg/backup/details/details.go
ryanfkeepers, line 1050: Suggestion: check out the `xsync` package I'm using for synchronous mapping in [this pr](https://github.com/alcionai/corso/pull/3722/files#diff-f0179d0b53f0e6169d90c6b0904c7988aa126b0ed82135bf2c0950473e9679beR4) ([github](https://github.com/puzpuzpuz/xsync)).  It might save us some concurrency handling concerns.

ryanfkeepers (commented):
Oh, another concern: ExtensionInfo would be better off with an interface, instead of a direct reference (there's a comment with suggestions).

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+	"io"
+
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/logger"
+)
+
+// Extension client interface
+type CorsoItemExtension interface {
```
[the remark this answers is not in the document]
pandeyabs: > Lets keep Corso* out of our variable naming

Sounds good. 

> Nested interfaces are good for extending the set of behaviors (ie, when you need it to do foo/bar/baz, plus some extra behavior)

Thanks for the detailed overall review, first of all! 
I nested the `io.ReadCloser` so that we can have the flexibility to add extra behavior later on. For e.g. we could add `GetIdentifier()` to `CorsoItemExtension` to fetch the extension UUID/other properties. 
This could then be then used towards - 
a. Logging
b. Tracking dedicated scratch space for extensions ( uuid -> `ExtensionInfo`) sort of thing.

That being said, I don't see any current need to expose `GetIdentifier` or similar extra behavior. So I can remove `CorsoItemExtension` and just use `io.ReadCloser`.

src/pkg/extensions/extensions.go
pandeyabs: Great suggestion! I did have later plans on exposing `details.ItemInfo` via a getter and `ExtensionInfo` via a setter interface(one you recommended above). But I wanted to get out something quick.

Let me add the `ExtensionKVStorer` right away, as it's a great pattern &  will save extension authors some refactoring time later.

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (l *loggerExtension) Read(p []byte) (int, error) {
+	n, err := l.innerRc.Read(p)
+	if err != nil && err != io.EOF {
+		logger.CtxErr(l.ctx, err).Error("inner read")
+		return n, err
+	}
+
+	if err == io.EOF {
```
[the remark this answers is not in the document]
pandeyabs: Great catch!

src/pkg/extensions/extensions_test.go
pandeyabs: yes! This is covered in a later PR. `mock_extensions.go` has some code which lays the groundwork for read/close testing.
I've only added tests here for  `AddItemExtensions` which only initializes the extensions.

Thanks for raising it though! Really appreciate it!

src/pkg/extensions/extensions.go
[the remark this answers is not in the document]
pandeyabs: Let me address a few questions around verbose logging &  `loggingExtension` ROI here.

I was thinking in terms of E2E observability. Since extension lifetimes are per item, it's going to generate some logging chaf. 

Specifically, we should be able to answer below questions using logs. 

1. If the extensions were supplied, were they enabled?
2. If the extensions were enabled, did they run?
3. Did extensions run for all items, or just some of them? If some items were skipped, was it corso at fault or was it the extension?
4. If a read/close failed, was it because of extension code or corso code?
5. If the extensions produced data, was it persisted by corso?

Having a `loggingExtension` wrapper can help us answer 2, 3, and to a certain extent 4.

Additionally, my north star for adding `loggingExtension` was to have an upper `loggingExtension` & lower `loggingExtension` with third party extensions sandwiched in between. It can help us isolate metrics around extension runtimes, enforce read only checks ( compare lower.crc with upper.crc) etc. No current need here though.

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+		logger.CtxErr(l.ctx, err).Error("inner close")
+		return err
+	}
+
+	logger.Ctx(l.ctx).Info("corso extensions: closed")
+
+	return nil
+}
+
+type AddItemExtensioner interface {
```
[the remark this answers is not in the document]
pandeyabs: I did have this as a plain func in an earlier commit. My reasons for adding an interface here were
-  Have a shim layer for abstracting the contract with extensions. It does feel over-engineered given current functionality.
- Capability to mock `AddItemExtensions` and test caller code against extension initialization failures.

You're right, its not adding much value. I'll move to a plain func. Thank you!

src/pkg/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+	rc io.ReadCloser,
+	info details.ItemInfo,
+	factories []CorsoItemExtensionFactory,
+) (io.ReadCloser, *details.ExtensionInfo, error) {
+	if rc == nil {
+		return nil, nil, clues.New("nil inner readcloser")
+	}
+
+	if len(factories) == 0 {
+		return nil, nil, clues.New("no extensions supplied")
```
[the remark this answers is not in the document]
pandeyabs: My thought process was that if the caller called this function, they have 100% intent to add extensions. So we should fail in caller rather than silently skipping extensions.

What do you think?

src/pkg/extensions/extensions.go
pandeyabs: It's verbose but it'll help us answer below question. 

Did extensions run for all items, or just some of them? If some items were skipped, was it corso at fault or was it the extension?

src/pkg/backup/details/details.go
pandeyabs: Missed this one. 100%. Thanks for the suggestion!

src/pkg/extensions/extensions.go
ryanfkeepers: > for each item

Which will flood our logs with, potentially, thousands of messages.

Imo, there shouldn't need to be any deliberation about "did we wrap readers with extensions?".  That should be a given.  The caller gave us wrappers.  We then apply them to the readers.  Whether or not the caller wants to run a conditional passthrough in their extensions, or something heavier, is entirely up to them.  But it should be the caller making that choice within their extension.  If we cascade that decision making onto corso, we're going to ingest a whole lot of concern that doesn't belong to us.

src/pkg/extensions/extensions.go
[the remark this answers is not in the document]
ryanfkeepers: > If the extensions were supplied, were they enabled?

IMO, every extension that's supplied should be enabled.  The caller can decide what gets enabled or not from within the extension implementation.

> If the extensions were enabled, did they run?

Same as above, that shouldn't be corso's concern (outside of basic smoke testing).  

> Did extensions run for all items, or just some of them? If some items were skipped, was it corso at fault or was it the extension?

io.Readers give us a lot of stability here that alleviates fault from corso.  One of the reasons for pushing toward a design centralized around wrapping the reader interface instead of middle-manning copies of buffers is that we, in corso, cannot be "at fault".  At least, not beyond any behavior that would also fundamentally break corso.

> If a read/close failed, was it because of extension code or corso code?

Given the nature of golang errors, we'd find out either way by receiving the error at the topmost reader (corso, in this case).  That error should contain a record of its causes from which we can extract reviewable details.

> If the extensions produced data, was it persisted by corso?

In controlled testing, that assertion is better verified by review of the details entries post-processing.  Yes, there's a lot of assertions we want to make about the behavior.  But those are all assertions we can make in a generic case, by focusing on the system handling of extensions, not the external user's application of them.

Overall, again, I think the instinct here is good.  But it's a little early in the project for us to need this scale of granularity.  Especially when it comes to details like `compare lower.crc with upper.crc`.

I honestly won't be mad if you keep it in.  I may grumble a bit when I go to filter out the messages.  But ultimately it will add more value than it detracts.  I'd only ask that we defer to either Debug or Error logging, and avoid Info.

src/pkg/extensions/extensions.go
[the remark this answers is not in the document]
ryanfkeepers: The way I read the code, it seems that we'd be the ones instantiating the call, not the extension provider.  In that case the ball is in our court, and a no-op here saves us a len conditional.

src/pkg/extensions/extensions.go
ashmrtn: knowing how many extensions were applied can be useful. As a middle ground, would it be possible to log it once in another function (it should apply to all items anyway) instead of for every item? This could be done in a separate PR if that would fit better

pandeyabs:
@ryanfkeepers Could you please do a final review on this? :)

PR #3735, opened 2023-07-01
tweaks and fixes found while testing

ryanfkeepers:
a variety of small updates that came
from manual testing of restore with various
collision and destination combinations.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual


src/pkg/backup/details/details.go
```diff
@@ -503,14 +503,14 @@ func (ents entrySet) PrintEntries(ctx context.Context) {
...
-		!print.DisplayJSONFormat() &&
-		!print.DisplayVerbose() {
-		// TODO: Should we detect if the user is piping the output and
-		// print if that is the case?
-		print.Outf(ctx, "Restored %d items.", len(ents))
-	} else {
+func (ents entrySet) MaybePrintEntries(ctx context.Context, isRestore bool) {
+	if isRestore {
+		print.Outf(ctx, "Restored %d items", len(ents))
+	}
```
meain, line 509: I don't think we should introduce a bool. This print should be just a part of the restore code before calling the function.

src/pkg/backup/details/details.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // Max number of items for which we will print details. If there are
 // more than this, then we just show a summary.
-const maxPrintLimit = 15
+const maxPrintLimit = 100
```
meain, line 25: 100 seems a bit much. 15 might be small, but I feel like with 100, the information will be too much to be useful as console output.

src/internal/m365/graph/middleware.go
```diff
@@ -148,7 +148,8 @@ func (mw *LoggingMiddleware) Intercept(
 		logBody = logger.DebugAPIFV ||
 			os.Getenv(logGraphRequestsEnvKey) != "" ||
 			resp.StatusCode == http.StatusBadRequest ||
-			resp.StatusCode == http.StatusForbidden
+			resp.StatusCode == http.StatusForbidden ||
+			resp.StatusCode == http.StatusConflict
```
meain, line 152: Might wanna update the comment to reflect this.

meain (approved):
Have a few small comments, approving to unblock.

src/pkg/backup/details/details.go
ryanfkeepers: reduced to 50

src/internal/m365/graph/middleware.go
ryanfkeepers: doesn't seem like a very helpful comment.  I've removed the 4xx enumeration part of it.

src/pkg/backup/details/details.go
ryanfkeepers: fair, moved it to the calling func.

src/cli/restore/restore.go
```diff
@@ -82,14 +82,17 @@ func runRestore(
 		return Only(ctx, clues.Wrap(err, "Failed to run "+serviceName+" restore"))
 	}
 
-	Info(ctx, "Completed Restore:")
+	Info(ctx, "Restore Complete")
```
ashmrtn, line 85: Is this going to break sanity tests and other similar github workflows? I know some of those check for specific output strings

src/pkg/services/m365/api/events.go
```diff
@@ -748,5 +753,5 @@ func EventCollisionKey(item models.Eventable) string {
 
 	// this result gets hashed to ensure that an enormous list of attendees
 	// doesn't generate a multi-kb collision key.
-	return clues.ConcealWith(clues.SHA256, subject+a+t+s+e+r)
+	return subject + t + s + e + r + strconv.FormatBool(draft) + strconv.FormatBool(cancelled)
```
ashmrtn, line 756: these need better names. It's way too cryptic to be easily readable so it's difficult to actually determine what's happening and if the code is doing what we think it's doing

src/cli/restore/restore.go
ryanfkeepers: Hopefully not!  I'll go check on that.

src/internal/m365/exchange/contacts_restore.go
```diff
@@ -100,6 +99,14 @@ func restoreContact(
 	errs *fault.Bus,
 	cb *count.Bus,
 ) (*details.ExchangeInfo, error) {
+	// contacts has a weird relationship with its default
+	// folder, which is that the folder is treated as invisible
+	// in many cases.  If we're restoring to a blank location,
+	// we can interpret that as the root.
```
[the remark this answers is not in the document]
ryanfkeepers: It's graph, yeah, the problem lies in a difference between the folder lookup and item destination.  It's kinda like how with emails, the folders aren't all under `Inbox`, they're often under `msgfolderroot`.  The difference there is that the messages themselves are only ever inside a folder that's a child of msgfolderroot.  Contacts can attach contact items directly to its root, instead of requiring a child folder inbetween.

There might be a way for us to tackle the handling more directly, but I haven't taken an eye to the code with that goal in mind yet.

Issue #3679, opened 2023-06-27
Longevity Test Automation


vkamra:
This tracks creating a framework to run longevity tests for Corso.

Current requirements:

Setup:
- Create a longevity test corso repository

Backup:
- Run backup nightly (Exchange, OD, Sharepoint, Teams (in future))
- Use release builds for this (this will guard against any repo corruption for changes that merge into mainline that aren't fully tested)
- Important : Must disable telemetry when using this build to avoid affecting usage stats

Restore:
- TBD (perhaps we wait for Export/DryRun features and enable those))

Backup Deletion:
- Delete any backups older than XXXX days

Maintenance:
- Run quick maintenance every night
- Run full maintenance every week

ashmrtn:
backups should run against a user that has some amount of data churn so we can better exercise the maintenance logic

neha-Gupta1:
@ashmrtn the data here should be fixed( so that we are aware of input provided) or should it be any data that the user have(to make it more inclusive and have wide range of data).

ashmrtn:
we can just use any data the user has for this. These tests are mostly to make sure that over the long run things work as intended. Ideally we'd have a controlled data set, but we also need a fair amount of churn in the data set (so we can track how maintenance behaves). That would require quite a bit of work to setup

neha-Gupta1:
https://github.com/alcionai/corso/pull/3752 covers- setup, backup and maintenance

neha-Gupta1:
Will track restore in - https://github.com/alcionai/corso/issues/3825

vkamra:
Closing this ticket now.

PR #3740, opened 2023-07-03
Longevity test

neha-Gupta1:
Longevity test which runs every night on same repo with the latest release code.  

- [ ] :no_entry: No

#### Type of change

#### Issue(s)

* [#<issue>](https://github.com/alcionai/corso/issues/3679)

#### Test Plan

- [ ] :muscle: Manual


.github/workflows/longetivity_test.yml
```diff
@@ -0,0 +1,118 @@
...
+        id: repo-init
+        run: |
+          set -euo pipefail
+          prefix='longevity'
+          echo -e "\nRepo init test\n" >> ${{ env.CORSO_LOG_FILE }}
+          ./corso repo init s3 \
+            --no-stats \
+            --hide-progress \
+            --prefix $prefix \
+            --bucket ${{ secrets.CI_TESTS_S3_BUCKET }} \
```
ashmrtn, line 89: I think we may need to adjust the [bucket cleanup script](https://github.com/alcionai/corso/blob/main/.github/workflows/weekly_cleanup.yml). I think right now it deletes all contents in the bucket. If we want a long-lived repo we'll need to exclude this prefix

ashmrtn (approved):
looks good so far. This can be merged and other additions can be made in separate PRs

ashmrtn:
we should also send notifications about failures here to slack

.github/workflows/longetivity_test.yml
neha-Gupta1: I will take care of this in next PR. And also add notification to slack.

PR #3746, opened 2023-07-03
Ctrl extensions

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/control/options.go
+++ b/src/pkg/control/options.go
@@ -8,18 +8,20 @@ import (
 
 	"github.com/alcionai/corso/src/internal/common/dttm"
 	"github.com/alcionai/corso/src/pkg/control/repository"
+	"github.com/alcionai/corso/src/pkg/extensions"
 	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 // Options holds the optional configurations for a process
 type Options struct {
-	DisableMetrics     bool               `json:"disableMetrics"`
-	FailureHandling    FailurePolicy      `json:"failureHandling"`
-	RestorePermissions bool               `json:"restorePermissions"`
-	SkipReduce         bool               `json:"skipReduce"`
-	ToggleFeatures     Toggles            `json:"toggleFeatures"`
-	Parallelism        Parallelism        `json:"parallelism"`
-	Repo               repository.Options `json:"repo"`
+	DisableMetrics       bool                               `json:"disableMetrics"`
+	FailureHandling      FailurePolicy                      `json:"failureHandling"`
+	RestorePermissions   bool                               `json:"restorePermissions"`
+	SkipReduce           bool                               `json:"skipReduce"`
+	ToggleFeatures       Toggles                            `json:"toggleFeatures"`
+	Parallelism          Parallelism                        `json:"parallelism"`
+	Repo                 repository.Options                 `json:"repo"`
+	ItemExtensionFactory []extensions.CreateItemExtensioner `json:"-"`
 }
 
 type Parallelism struct {
--- a/src/pkg/repository/repository_test.go
+++ b/src/pkg/repository/repository_test.go
@@ -1,4 +1,4 @@
-package repository_test
+package repository
 
 import (
 	"os"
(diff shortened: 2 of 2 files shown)


pandeyabs:
@ashmrtn @ryanfkeepers could you please review this PR? :)

src/pkg/repository/repository_test.go
```diff
@@ -284,13 +284,88 @@ func (suite *RepositoryIntegrationSuite) TestConnect_DisableMetrics() {
...
 	assert.Equal(t, r.GetID(), r.GetID())
 }
+
+// Test_Options tests that the options are passed through to the repository
+// correctly
+func (suite *RepositoryIntegrationSuite) Test_Options() {
+	table := []struct {
+		name       string
+		opts       func() control.Options
+		assertFunc func(t *testing.T, opts control.Options)
```
ryanfkeepers, line 304: since you're only using `len` checks here, this could be simplified to an `expectedLen` int instead.

src/pkg/control/options.go
```diff
@@ -8,18 +8,20 @@ import (
...
-	Parallelism        Parallelism        `json:"parallelism"`
-	Repo               repository.Options `json:"repo"`
+	DisableMetrics       bool                               `json:"disableMetrics"`
+	FailureHandling      FailurePolicy                      `json:"failureHandling"`
+	RestorePermissions   bool                               `json:"restorePermissions"`
+	SkipReduce           bool                               `json:"skipReduce"`
+	ToggleFeatures       Toggles                            `json:"toggleFeatures"`
+	Parallelism          Parallelism                        `json:"parallelism"`
+	Repo                 repository.Options                 `json:"repo"`
+	ItemExtensionFactory []extensions.CreateItemExtensioner `json:"-"`
```
ryanfkeepers, line 24: Thoughts on a `type Extensions []extensions.CreateItemExtensioner` with a method like `Extensions.Add(a, b, c)`?

src/pkg/repository/repository_test.go
```diff
@@ -284,13 +284,88 @@ func (suite *RepositoryIntegrationSuite) TestConnect_DisableMetrics() {
...
+			},
+		},
+		{
+			name: "options with an extension factory",
+			opts: func() control.Options {
+				o := control.Defaults()
+				f := []extensions.CreateItemExtensioner{
+					&extensions.MockItemExtensionFactory{},
+				}
+				o.ItemExtensionFactory = f
```
ryanfkeepers, line 322: alternatively, you can:
```suggestion
				o.ItemExtensionFactory = append(
				    o.ItemExtensionFactory,
				    &extensions.MockItemExtensionFactory{})
```

src/pkg/repository/repository_test.go
```diff
@@ -284,13 +284,88 @@ func (suite *RepositoryIntegrationSuite) TestConnect_DisableMetrics() {
...
+	r, err := Connect(ctx, account.Account{}, st, repo.GetID(), control.Options{DisableMetrics: true})
 	assert.NoError(t, err)
 
 	// now we have repoID beforehand
 	assert.Equal(t, r.GetID(), r.GetID())
 }
+
+// Test_Options tests that the options are passed through to the repository
+// correctly
+func (suite *RepositoryIntegrationSuite) Test_Options() {
```
ryanfkeepers, line 300: I'm undecided on whether this is the correct location and owner for this test.  Do we expect repo connection to modify options or extensions?  Do we expect repo connection to interact with those details at all?

This feels like a basic test for existence within a consumer.  "I added foo to bar.  Does bar contain foo?  Yes."  It's testing, yes, but I don't see the assurance guarantees we're getting out of it.  Perhaps there's a lower-level test, more associated with usage and interaction of the options, that would make more sense?

ryanfkeepers (approved):
Not sure about the tests, but nothing blocking here.

src/pkg/repository/repository_test.go
pandeyabs: >  "I added foo to bar. Does bar contain foo? Yes."

Yep. This is the intent of this particular test. I'll be adding more tests which exercise options usage

src/pkg/control/options.go
pandeyabs: Thanks! This is better! I'll consider this in a later PR.

src/pkg/repository/repository_test.go
ryanfkeepers: > This is the intent of this particular test

Right, so you're testing struct population.  I don't see that as providing value; it's too foundational of a behavior to need to test.  Unless, of course, the consumer you're handing the struct into is mutating it in some way.

pandeyabs: This is just to protect against future regressions in `Initialize` and `Connect`

PR #3747, opened 2023-07-03
remove goreleaser replacements property

ryanfkeepers:
Goreleaser [deprecated](https://goreleaser.com/deprecations/?h=replacement#archivesreplacements) the replacements yaml property, which has been causing our binary publications to fail for a while now.  This fixes the yaml and introduces a slack message for binary publication failures.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -442,7 +442,8 @@ jobs:
     needs: [Test-Suite-Trusted, Source-Code-Linting, Website-Linting, SetEnv]
     environment: ${{ needs.SetEnv.outputs.environment }}
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'
+    # TODO: reinstate before merge
+    # if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'
 
     steps:
       - uses: actions/checkout@v3
@@ -454,6 +455,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           rudderstack_write_key: ${{ secrets.RUDDERSTACK_CORSO_WRITE_KEY }}
           rudderstack_data_plane_url: ${{ secrets.RUDDERSTACK_CORSO_DATA_PLANE_URL }}
+          slack_webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
 
   Publish-Image:
     needs: [Test-Suite-Trusted, Source-Code-Linting, Website-Linting, SetEnv]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -442,8 +442,7 @@ jobs:
     needs: [Test-Suite-Trusted, Source-Code-Linting, Website-Linting, SetEnv]
     environment: ${{ needs.SetEnv.outputs.environment }}
     runs-on: ubuntu-latest
-    # TODO: reinstate before merge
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -442,7 +442,8 @@ jobs:
     needs: [Test-Suite-Trusted, Source-Code-Linting, Website-Linting, SetEnv]
     environment: ${{ needs.SetEnv.outputs.environment }}
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'
+    # TODO: reinstate before merge
+    # if: startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'
```
ashmrtn: I'm guessing this TODO still needs addressed? 😉

src/.goreleaser.yaml
```diff
@@ -18,13 +18,15 @@ builds:
...
 archives:
-  - name_template: "{{ .ProjectName }}_{{ .Tag }}_{{ .Os }}_{{ .Arch }}"
-    replacements:
-      darwin: Darwin
-      linux: Linux
-      windows: Windows
-      386: i386
-      amd64: x86_64
+  # this name template makes the OS and Arch compatible with the results of uname.
+  - name_template: >-
```
ashmrtn, line 22: is this leading line going to behave as expected? I'm not sure if it's GitHub workflow formatting magic or not. It doesn't seem like golang templating DSL

ashmrtn (approved):

src/.goreleaser.yaml
pandeyabs: The deprecation notice [seems to propose](https://goreleaser.com/deprecations/?h=replacement#archivesreplacements) this as the alternative to replacements.

ashmrtn: the overall text yes, but I'm wondering about the `>-` specifically which I don't think is part of golang's DSL for templates

pandeyabs: i see. 

I do see `>-` being used in several other packages. [Example](https://github.com/aquasecurity/trivy/blob/a9badeaba8bb9e908bce2f6f6defecf2f8515aeb/goreleaser.yml#L105)

src/.goreleaser.yaml
```diff
@@ -18,13 +18,15 @@ builds:
...
+      {{ .ProjectName }}_
+      {{- .Tag }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
```
pandeyabs, line 27: We should remove support for 32 bit archs in a future PR, and we should add arm64 support.

src/.goreleaser.yaml
ryanfkeepers: This came out of goreleasers autogenerated yaml file, so, I take that as a sign that it _must_ work.

src/.goreleaser.yaml
ryanfkeepers: That's a topic to take up with Georgi and Vaibhav.  I don't recall the original conversations around this, but I assume we retained it intentionally.  Maybe you can make a tech debt ticket for it?

pandeyabs: https://github.com/golang/go/wiki/GoArm

PR #3749, opened 2023-07-04
Add concurrency init to api

ryanfkeepers:
expose middleware init via api layer

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/client.go
+++ b/src/pkg/services/m365/api/client.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/account"
+	"github.com/alcionai/corso/src/pkg/path"
 )
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/pkg/services/m365/api/client.go
```diff
@@ -55,6 +56,14 @@ func NewClient(creds account.M365Config) (Client, error) {
 	return Client{creds, s, li, rqr}, nil
 }
 
+// initConcurrencyLimit ensures that the graph concurrency limiter is
+// initialized, so that calls do not step over graph api's service limits.
+// Limits are derived from the provided servie type.
+// Callers will need to call this func before making api calls an api client.
+func InitConcurrencyLimit(ctx context.Context, pst path.ServiceType) {
+	graph.InitializeConcurrencyLimiter(ctx, pst == path.ExchangeService, 4)
```
pandeyabs, line 64: We should source parallelism from `control.Options` here?

ryanfkeepers: No; the callers here don't provide control options at this time.  We can add that at a later date if needed.

pandeyabs: works

src/pkg/services/m365/m365.go
```diff
@@ -209,7 +210,7 @@ func GetUserInfo(
 	acct account.Account,
 	userID string,
 ) (*api.UserInfo, error) {
-	ac, err := makeAC(acct)
+	ac, err := makeAC(ctx, acct, path.ExchangeService)
```
pandeyabs, line 213: Sanity check: Is `path.ExchangeService` correct here?

pandeyabs (approved):

src/pkg/services/m365/m365.go
ryanfkeepers: `GetInfo` calls `GetMailbox` which works under exchange concurrency constraints.

PR #3750, opened 2023-07-04
Integrate extensions into drive item reader

pandeyabs:
* Integrates corso extensions into onedrive/sharepoint library item backup flows.
* Also includes integration tests for OD/SP, unit tests for collections.
* Includes a small fix for `rw *backupStreamReader Close()`

Remaining things which will be covered in later PRs:
* extension tests with incremental backups
* Observability related changes for extensions

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* internal
#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/onedrive/collection.go
```diff
@@ -531,11 +534,29 @@ func (oc *Collection) populateDriveItem(
...
+					rc,
+					itemInfo,
+					itemExtensionFactory)
+
+				if err != nil {
+					return nil, clues.Stack(err).WithClues(ctx)
+				}
+
+				itemData = extRc
+				itemInfo.OneDrive.Extension = extInfo
```
pandeyabs, line 556: @ryanfkeepers Quick question for you: Since sharepoint libraries also utilize onedrive collections, this code is currently failing for sharepoint backups with nil reference. 

To fix it, I have a few ideas. Wanted to run them by you quickly :) 

1. Currently, `Collection` doesn't distinguish between OneDrive vs Sharepoint library items. Add `path.ServiceType` and/or `path.Category` to onedrive `Collections` & `Collection`. Use that here to decide between filling `Sharepoint.Extension` vs `OneDrive.Extension`. 
2. Expose `[Extension](https://github.com/alcionai/corso/pull/3750/commits/98a3f3abb330a64c9618845dbf3864b805a41049#diff-47f6ad08905a9164dd5eeedd907a7559345c803d7775970e88e9121a0f50a033R991)` directly in `ItemInfo` rather than inside `ItemInfo.OneDrive` or `ItemInfo.Sharepoint`

I am leaning towards 1, since 2 mangles scope of nested structures. Also, the extension is tied to the item. Ideally, if it's a onedrive item, everything should be under `ItemInfo.OneDrive` scope.

ryanfkeepers: I'd say `2`.  The reason being that:

> Ideally, if it's a onedrive item, everything should be under ItemInfo.OneDrive scope

Is slightly off the mark.  If it's a onedrive item, all item data that is _specific to onedrive_ should be under ItemInfo.OneDrive.  Data that can generically represent any item can be promoted to the top level ItemInfo.  

Extensions are not specific to onedrive.  They're just starting their implementation in onedrive.  So it makes sense to promote them.

pandeyabs: Hmm. This loose coupling was not sitting well with me. If we do 2),  ideally the extension should have some identifier to say that this is a OneDrive extension vs Sharepoint vs ... But for adding that identifier, we'd need 1) again from my previous comment. 

Let me go with 2). It will push some burden on to the ItemInfo consumer to associate extension with OneDrive/Sharepoint, but benefit here is that we can keep onedrive.`Collection` generic.

src/internal/kopia/upload.go
```diff
@@ -73,7 +73,10 @@ func (rw *backupStreamReader) Close() error {
 	rw.combined = nil
 
 	for _, r := range rw.readers {
-		r.Close()
+		err := r.Close()
```
pandeyabs, line 76: @ashmrtn I think this should be harmless, but could you please review this change? :)

src/internal/m365/onedrive/mock/handlers.go
```diff
@@ -47,7 +47,7 @@ type BackupHandler struct {
 
 func DefaultOneDriveBH() *BackupHandler {
 	return &BackupHandler{
-		ItemInfo:     details.ItemInfo{OneDrive: &details.OneDriveInfo{}},
+		ItemInfo:     details.ItemInfo{OneDrive: &details.OneDriveInfo{}, Extension: &details.ExtensionInfo{}},
```
ryanfkeepers, line 50: It may be a good time to introduce an ItemInfo factory, to avoid gotchas from forgetting to initialize this or that property.

src/internal/m365/onedrive/collection.go
```diff
@@ -531,11 +534,29 @@ func (oc *Collection) populateDriveItem(
...
+			// Add per item extensions if available
+			if len(itemExtensionFactory) > 0 {
+				extRc, extInfo, err := extensions.AddItemExtensions(
+					ctx,
+					rc,
+					itemInfo,
+					itemExtensionFactory)
+				if err != nil {
+					el.AddRecoverable(ctx, clues.Wrap(err, "adding extensions").Label(fault.LabelForceNoBackupCreation))
+					return nil, clues.Stack(err).WithClues(ctx)
```
ryanfkeepers, line 553: Shouldn't need `WithClues(ctx)` here, since the error producer accepted the `ctx` already.  This could unintentionally overwrite some downstream values.

ryanfkeepers, line 553:
```suggestion
					err := clues.Wrap(err, "adding extensions").Label(fault.LabelForceNoBackupCreation)
					el.AddRecoverable(ctx, err)

					return nil, err
```

src/internal/operations/test/onedrive_test.go
```diff
@@ -822,3 +823,58 @@ func (suite *OneDriveBackupIntgSuite) TestBackup_Run_oneDriveOwnerMigration() {
...
+		bod.sss,
+		expectDeets,
+		false)
+
+	// Check that the extensions are in the backup
+	for _, ent := range deets.Entries {
+		if ent.Folder == nil {
+			verifyExtensionData(t, ent.ItemInfo, path.OneDriveService)
+		}
+	}
```
ryanfkeepers, line 879: I'm curious if we could build this into the standard `Backup_Run_oneDrive` test (same with sharepoint) instead of running an entirely separate test.  Running another backup entirely for this feels pretty heavy.  Thoughts on folding them together?

src/internal/operations/test/helper_test.go
```diff
@@ -625,3 +623,25 @@ func newIntegrationTesterSetup(t *testing.T) intgTesterSetup {
...
+	t *testing.T,
+	itemInfo details.ItemInfo,
+	p path.ServiceType,
+) {
+	assert.NotNil(t, itemInfo.Extension, "nil extension")
+	assert.NotNil(t, itemInfo.Extension.Data["NumBytes"], "key not found in extension")
+	eSize := int64(itemInfo.Extension.Data["NumBytes"].(float64))
+
+	if p == path.SharePointService {
+		assert.Equal(t, itemInfo.SharePoint.Size, eSize, "incorrect data in extension")
```
ryanfkeepers, line 643: For consistency throughout our code, assertions should be written as `assert.Foo(t, expected, actual)`.  This has it backwards.

src/internal/operations/test/helper_test.go
```diff
@@ -625,3 +623,25 @@ func newIntegrationTesterSetup(t *testing.T) intgTesterSetup {
...
+		&extensions.MockItemExtensionFactory{},
+	}
+}
+
+func verifyExtensionData(
+	t *testing.T,
+	itemInfo details.ItemInfo,
+	p path.ServiceType,
+) {
+	assert.NotNil(t, itemInfo.Extension, "nil extension")
```
ryanfkeepers, line 638: require here will save you a nil panic if the extension is nil.
```suggestion
	require.NotNil(t, itemInfo.Extension, "nil extension")
```

ryanfkeepers (approved):
Generally looks good.  Just a bunch of nits.

src/pkg/extensions/mock_extensions.go
```diff
@@ -14,60 +14,61 @@ import (
...
 func (me *MockExtension) Close() error {
-	if me.failOnClose {
+	if me.FailOnClose {
 		return clues.New("mock close error")
 	}
 
-	err := me.innerRc.Close()
+	err := me.InnerRc.Close()
 	if err != nil {
 		return err
```
ashmrtn, line 56: nit: stack for line numbers
```suggestion
		return clues.Stack(err)
```

src/pkg/extensions/mock_extensions.go
```diff
@@ -14,60 +14,61 @@ import (
...
 
-	me.extInfo.Data["numBytes"] = me.numBytes
-	me.extInfo.Data["crc32"] = me.crc32
-	logger.Ctx(me.ctx).Infow(
+	me.ExtInfo.Data["NumBytes"] = me.NumBytes
+	me.ExtInfo.Data["Crc32"] = me.Crc32
+	logger.Ctx(me.Ctx).Infow(
 		"mock extension closed",
-		"numBytes", me.numBytes, "crc32", me.crc32)
+		"NumBytes", me.NumBytes, "Crc32", me.Crc32)
```
ashmrtn, line 63: we should probably make these constants in the file with the mock definition since they're used in several different packages

src/pkg/extensions/mock_extensions.go
```diff
@@ -14,60 +14,61 @@ import (
...
-	failOnRead  bool
-	failOnClose bool
+	NumBytes    int64
+	Crc32       uint32
+	Info        details.ItemInfo
+	ExtInfo     *details.ExtensionInfo
+	InnerRc     io.ReadCloser
+	Ctx         context.Context
+	FailOnRead  bool
+	FailOnClose bool
```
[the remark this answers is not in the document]
pandeyabs: Thanks! Since these are bool values, we will have the same effect here with caller provided flags.

src/pkg/extensions/mock_extensions.go
```diff
@@ -14,60 +14,61 @@ import (
...
+	n, err := me.InnerRc.Read(p)
 	if err != nil && err != io.EOF {
-		logger.CtxErr(me.ctx, err).Error("inner read error")
+		logger.CtxErr(me.Ctx, err).Error("inner read error")
 		return n, err
 	}
 
-	me.numBytes += n
-	me.crc32 = crc32.Update(me.crc32, crc32.IEEETable, p[:n])
+	me.NumBytes += int64(n)
```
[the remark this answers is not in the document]
pandeyabs: Yeah! We are safe here though, since the read flow is single threaded.
But i'll add atomics

src/pkg/extensions/mock_extensions.go
```diff
@@ -14,60 +14,61 @@ import (
...
-	innerRc     io.ReadCloser
-	ctx         context.Context
-	failOnRead  bool
-	failOnClose bool
+	NumBytes    int64
+	Crc32       uint32
+	Info        details.ItemInfo
+	ExtInfo     *details.ExtensionInfo
+	InnerRc     io.ReadCloser
+	Ctx         context.Context
```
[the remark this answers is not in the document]
pandeyabs: Mock is just a replica of how actual extensions will work. We are passing `ctx` through `CreateItemExtension` which is eventually stored here.

Not sure if there is an easy way to provide it as a runtime func param, since this is an `io.ReadCloser` wrapper

src/pkg/backup/details/details.go
```diff
@@ -708,6 +708,7 @@ type ItemInfo struct {
 	Exchange   *ExchangeInfo   `json:"exchange,omitempty"`
 	SharePoint *SharePointInfo `json:"sharePoint,omitempty"`
 	OneDrive   *OneDriveInfo   `json:"oneDrive,omitempty"`
+	Extension  *ExtensionInfo  `json:"extension,omitempty"`
```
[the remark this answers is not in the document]
pandeyabs: Can do! I'll also leave a comment explaining what `ExtensionInfo` is

src/internal/operations/test/helper_test.go
pandeyabs: I think the order is correct? `itemInfo.SharePoint.Size` is the expected size.

src/pkg/backup/details/details.go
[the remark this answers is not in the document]
ashmrtn: I would feel more comfortable if the renaming happened prior to merging to main. The tag may not need updated, but for peace of mind we should be careful about how we handle things that get persisted

src/internal/m365/onedrive/mock/handlers.go
pandeyabs: Sounds good. Let me add this in a later PR.

src/internal/m365/onedrive/collection_test.go
```diff
@@ -771,3 +773,210 @@ func (suite *GetDriveItemUnitTestSuite) TestDownloadContent() {
...
+				payload []byte,
+			) {
+				verifyExtensionData(t,
+					info.Extension,
+					int64(len(payload)),
+					crc32.ChecksumIEEE(payload))
+			},
+		},
+		{
+			name: "one extension fails on read",
```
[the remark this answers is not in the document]
pandeyabs: This is just testing whether the extension errors propagate correctly by corso code (with an assumption that the extensions are doing the right thing). We are not verifying state of extension data here.

Re semantics, i think we should make no promises about state of extension data for all extensions.

src/internal/operations/test/helper_test.go
ryanfkeepers: Oh, is it?  I just read `eSize` and thought `expectedSize`.

pandeyabs: :) that was supposed to be `extensionSize`

src/pkg/extensions/extensions_test.go
```diff
@@ -166,4 +168,69 @@ func (suite *ExtensionsUnitSuite) TestAddItemExtensions() {
...
+	for {
+		_, err := rc.Read(p)
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
pandeyabs: I kept it this as is, since it has identical functionality.

PR #3756, opened 2023-07-05
final few changes from manual testing

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/contacts_restore_test.go
+++ b/src/internal/m365/exchange/contacts_restore_test.go
@@ -176,6 +176,19 @@ func (suite *ContactsRestoreIntgSuite) TestRestoreContact() {
 				assert.True(t, m.calledDelete, "old item deleted")
 			},
 		},
+		{
+			name:         "collision: replace - err already deleted",
+			apiMock:      &contactRestoreMock{deleteItemErr: graph.ErrDeletedInFlight},
+			collisionMap: map[string]string{collisionKey: "smarf"},
+			onCollision:  control.Replace,
+			expectErr: func(t *testing.T, err error) {
+				assert.NoError(t, err, clues.ToCore(err))
+			},
+			expectMock: func(t *testing.T, m *contactRestoreMock) {
+				assert.True(t, m.calledPost, "new item posted")
+				assert.True(t, m.calledDelete, "old item deleted")
+			},
+		},
 	}
 	for _, test := range table {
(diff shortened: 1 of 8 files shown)


src/internal/m365/exchange/contacts_restore_test.go
```diff
@@ -175,6 +175,19 @@ func (suite *ContactsRestoreIntgSuite) TestRestoreContact() {
...
+			name:         "collision: replace - err already deleted",
+			apiMock:      &contactRestoreMock{deleteItemErr: graph.ErrDeletedInFlight},
+			collisionMap: map[string]string{collisionKey: "smarf"},
+			onCollision:  control.Replace,
+			expectErr: func(t *testing.T, err error) {
+				assert.NoError(t, err, clues.ToCore(err))
+			},
+			expectMock: func(t *testing.T, m *contactRestoreMock) {
+				assert.True(t, m.calledPost, "new item posted")
+				assert.True(t, m.calledDelete, "old item deleted")
```
ashmrtn, line 188: this should be checking the ID of the deleted item (same with other data types too)

src/internal/m365/onedrive/restore.go
```diff
@@ -851,7 +851,7 @@ func restoreData(
 		}
 
 		collision = dci
-		replace = restoreCfg.OnCollision == control.Replace && isFolder == dci.IsFolder
+		shouldDeleteOriginal = restoreCfg.OnCollision == control.Replace && isFolder == dci.IsFolder
```
ashmrtn, line 854: this may disappear when rebasing, but this doesn't need the `isFolder` check because it'll always be false

ashmrtn (approved):

src/internal/m365/onedrive/restore.go
ryanfkeepers: Ahah, fixed that in exchange and forgot onedrive.  Thanks!

src/internal/m365/exchange/contacts_restore_test.go
ryanfkeepers: good catch, thanks.

PR #3764, opened 2023-07-07
avoid provider check in init

neha-Gupta1:
do not check for provider in init command

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :zap: Unit test


src/pkg/storage/testdata/storage.go
```diff
@@ -58,7 +57,7 @@ func NewPrefixedS3Storage(t *testing.T) storage.Storage {
 func GetAndInsertCorso(passphase string) credentials.Corso {
 	// fetch data from flag, env var or func param giving priority to func param
 	// Func param generally will be value fetched from config file using viper.
-	corsoPassph := str.First(flags.CorsoPassphraseFV, os.Getenv(credentials.CorsoPassphrase), passphase)
```
neha-Gupta1, line 61: This one is a hack just to confirm that its the root case.

src/cli/backup/backup.go
```diff
@@ -236,7 +236,7 @@ func runBackups(
 			Infof(ctx, "Done\n")
 			printBackupStats(ctx, r, string(bo.Results.BackupID))
 		} else {
-			Infof(ctx, "Done - ID: %v\n", bo.Results.BackupID)
```
meain, line 239: I'm guessing this was to fix the issue with not being able to find user's name in output. I think the problem is probably from us changing where the output is written to. Don't think we should be changing he output format.
/cc @ryanfkeepers in case you have any thoughts

src/cli/repo/s3_e2e_test.go
```diff
@@ -144,7 +144,8 @@ func (suite *S3E2ESuite) TestInitS3Cmd_missingBucket() {
 
 	// run the command
 	err = cmd.ExecuteContext(ctx)
-	require.Error(t, err, clues.ToCore(err))
+	// init command can now pull data from config file too
+	require.NoError(t, err, clues.ToCore(err))
```
meain, line 148: The test is to make sure we error on missing bucket from the name. We should probably not change the semantics and ensure that we have something that will error out when we do not have the bucket even in the config file.

src/pkg/storage/testdata/storage.go
meain: In case you missed dropping this hack after testing.

meain (changes requested):

src/cli/repo/s3_e2e_test.go
neha-Gupta1: Yes. Let me try overriding bucket values.

src/pkg/storage/testdata/storage.go
neha-Gupta1: This confirms that `CorsoPassphraseFV` has some value. Need to check how.

src/cli/backup/backup.go
ryanfkeepers: Conceptually I'm okay with the change; but I don't recall exactly what it looks like.  For full assurance, I'd like to see a before/after comparison of the different outputs.  That'd be the best way to ensure we're happy with the result.

also, if we keep this change, the duplicate infof should be moved out of the conditional statement.

src/pkg/storage/testdata/storage.go
ryanfkeepers: I'm confused: is this a TODO/FIXME that needs removal?  Or do we assume the caller is passing FV into the func as the param?  I like the latter better, and it means we wouldn't need to concern ourselves with the FV.  Or, at least, the caller can be in charge of that.

neha-Gupta1: flags.CorsoPassphraseFV is to fetch the values from flags. And since this test helper function receives value from envar or config file, flags.CorsoPassphraseFV could be removed. Also flags.CorsoPassphraseFV is a global flag. The value here could be creating conflict.

src/cli/backup/exchange_e2e_test.go
```diff
@@ -117,7 +117,7 @@ func (suite *ExchangeCMDWithFlagsE2ESuite) TestBackupCreateExchange_azureIDFromC
 		"--config-file", suite.cfgFP)
 	cli.BuildCommandTree(cmd)
 
-	cmd.SetErr(&suite.recorder)
+	cmd.SetOut(&suite.recorder)
```
meain, line 120: Did we change where we print out the output to? Do you know what changed caused this nightly test to fail? I just wanted to make sure we did not accidentally change the behaviour of the print.
/cc @ryanfkeepers

neha-Gupta1: These were newly added test cases. Implementation was not correct. We haven't changes any behaviour of the print.

meain (approved):

PR #3779, opened 2023-07-10
Add tests that boomerang items

### LINKED ISSUE #3755: Make integration test to boomerang an email between folders
When we attempted to turn on immutable IDs in M365 for Exchange we found that the following didn't work properly
1. make backup and get delta token
2. move email from folder A to folder B (can be any folders)
3. move email from folder B to folder A (same folders in step 2)
4. make another backup that uses incrementals with the backup from 1/ as the base

In the above, the email that was moved (boomeranged) doesn't appear in the second backup. We should have a test to ensure this case works properly. It will also keep us from accidentally enabling features that could break this behavior

A more complete writeup on the situation that causes this problem is on the [Microsoft Q&A forum](https://learn.microsoft.com/en-us/answers/questions/1257719/inconsistent-graph-api-exchange-results-between-im) (no responses yet)

ashmrtn:
Add a test that will fail if something is wrong with tracking items that are moved from a folder to a different folder and back to the original folder between incremental backups. This is important as it will help us ensure we don't accidentally turn on Exchange immutable IDs before they're fixed on the Graph side

Only tests Exchange mail and OneDrive/SharePoint as other Exchange data types don't support move operations via Graph API (it is possible to move events between calendars in the web UI though)

Manually tested to ensure it does actually fail if immutable IDs are enabled

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3755

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/mail.go
+++ b/src/pkg/services/m365/api/mail.go
@@ -446,7 +446,7 @@ func (c Mail) PostItem(
 
 func (c Mail) MoveItem(
 	ctx context.Context,
-	userID, oldContainerID, itemID, newContainerID string,
+	userID, oldContainerID, newContainerID, itemID string,
 ) (string, error) {
 	body := users.NewItemMailFoldersItemMessagesItemMovePostRequestBody()
 	body.SetDestinationId(ptr.To(newContainerID))
--- a/src/pkg/services/m365/api/mail.go
+++ b/src/pkg/services/m365/api/mail.go
@@ -444,6 +444,30 @@ func (c Mail) PostItem(
 	return itm, nil
 }
 
+func (c Mail) MoveItem(
+	ctx context.Context,
+	userID, oldContainerID, itemID, newContainerID string,
+) (string, error) {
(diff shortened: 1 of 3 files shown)


src/pkg/services/m365/api/mail.go
```diff
@@ -444,6 +444,30 @@ func (c Mail) PostItem(
 	return itm, nil
 }
 
+func (c Mail) MoveItem(
+	ctx context.Context,
+	userID, oldContainerID, itemID, newContainerID string,
```
ryanfkeepers, line 449: My ocd is undecided between "move this whole thing to that container" vs stepping parameters in order as they would appear in the hierarchy.  Either way works, but figured I'd put up the nit.
```suggestion
	userID, oldContainerID, newContainerID, itemID string,
```

ryanfkeepers (approved):

PR #3783, opened 2023-07-11
add operations test for adv rest

ryanfkeepers:
ads operations level tests for advanced restore
configuration on all three services.
Code is largely boilerplate between each service, but with just enough quirks that full consolidation would require excess jumping through hoops.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3562

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/test/exchange_test.go
+++ b/src/internal/operations/test/exchange_test.go
@@ -929,14 +929,17 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			path.EmailCategory:    "",
 			path.EventsCategory:   "",
 		}
+		collKeys               = map[path.CategoryType]map[string]string{}
 		countContactsInRestore int
+		acCont                 = suite.its.ac.Contacts()
+		contactIDs             map[string]struct{}
 		countEmailsInRestore   int
+		acMail                 = suite.its.ac.Mail()
+		mailIDs                map[string]struct{}
+		countItemsInRestore    int
 		// countEventsInRestore int
-		countItemsInRestore int
-		collKeys            = map[path.CategoryType]map[string]string{}
-		acCont              = suite.its.ac.Contacts()
-		acMail              = suite.its.ac.Mail()
 		// acEvts   = suite.its.ac.Events()
+		// eventIDs = []string{}
 	)
 
 	// initial restore
@@ -967,6 +970,9 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 
 		// get all files in folder, use these as the base
 		// set of files to compare against.
+
+		// --- contacts
+
 		contGC, err := acCont.GetContainerByName(ctx, userID, "", restoreCfg.Location)
 		require.NoError(t, err, clues.ToCore(err))
 
@@ -979,6 +985,11 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 		require.NoError(t, err, clues.ToCore(err))
 		countContactsInRestore = len(collKeys[path.ContactsCategory])
 		t.Log(countContactsInRestore, "contacts restored")
+
+		contactIDs, err = acCont.GetItemIDsInContainer(ctx, userID, cIDs[path.ContactsCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		// --- events
 
 		// gc, err = acEvts.GetContainerByName(ctx, userID, "", restoreCfg.Location)
 		// require.NoError(t, err, clues.ToCore(err))
@@ -1001,6 +1012,8 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 
 		cIDs[path.EmailCategory] = ptr.Val(mailGC.GetId())
 
+		// --- mail
+
 		collKeys[path.EmailCategory], err = acMail.GetItemsInContainerByCollisionKey(
 			ctx,
 			userID,
@@ -1008,6 +1021,9 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 		require.NoError(t, err, clues.ToCore(err))
 		countEmailsInRestore = len(collKeys[path.EmailCategory])
 		t.Log(countContactsInRestore, "emails restored")
+
+		mailIDs, err = acMail.GetItemIDsInContainer(ctx, userID, cIDs[path.EmailCategory])
+		require.NoError(t, err, clues.ToCore(err))
 
 		countItemsInRestore = countContactsInRestore + countEmailsInRestore // + countEventsInRestore
 		checkRestoreCounts(t, ctr1, 0, 0, countItemsInRestore)
@@ -1046,6 +1062,8 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 
 		checkRestoreCounts(t, ctr2, countItemsInRestore, 0, 0)
 
+		// --- contacts
+
 		// get all files in folder, use these as the base
 		// set of files to compare against.
 		result := filterCollisionKeyResults(
@@ -1056,8 +1074,17 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			GetItemsInContainerByCollisionKeyer[string](acCont),
 			collKeys[path.ContactsCategory])
 
+		currentContactIDs, err := acCont.GetItemIDsInContainer(ctx, userID, cIDs[path.ContactsCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		assert.Equal(t, contactIDs, currentContactIDs, "ids are equal")
+
+		// --- events
+
 		// m = checkCollisionKeyResults(t, ctx, userID, cIDs[path.EventsCategory], acEvts, collKeys[path.EventsCategory])
 		// maps.Copy(result, m)
+
+		// --- mail
 
 		m := filterCollisionKeyResults(
 			t,
@@ -1067,6 +1094,11 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			GetItemsInContainerByCollisionKeyer[string](acMail),
 			collKeys[path.EmailCategory])
 		maps.Copy(result, m)
+
+		currentMailIDs, err := acMail.GetItemIDsInContainer(ctx, userID, cIDs[path.EmailCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		assert.Equal(t, mailIDs, currentMailIDs, "ids are equal")
 
 		assert.Len(t, result, 0, "no new items should get added")
 	})
@@ -1110,6 +1142,8 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			countItemsInRestore,
 			"every item should have been replaced")
 
+		// --- contacts
+
 		result := filterCollisionKeyResults(
 			t,
 			ctx,
@@ -1118,8 +1152,22 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			GetItemsInContainerByCollisionKeyer[string](acCont),
 			collKeys[path.ContactsCategory])
 
+		currentContactIDs, err := acCont.GetItemIDsInContainer(ctx, userID, cIDs[path.ContactsCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		assert.Equal(t, len(contactIDs), len(currentContactIDs), "count of ids are equal")
+		for orig := range contactIDs {
+			assert.NotContains(t, currentContactIDs, orig, "original item should not exist after replacement")
+		}
+
+		contactIDs = currentContactIDs
+
+		// --- events
+
 		// m = checkCollisionKeyResults(t, ctx, userID, cIDs[path.EventsCategory], acEvts, collKeys[path.EventsCategory])
 		// maps.Copy(result, m)
+
+		// --- mail
 
 		m := filterCollisionKeyResults(
 			t,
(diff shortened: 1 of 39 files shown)


src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+	t := suite.T()
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	// a backup is required to run restores
+
+	baseSel := selectors.NewExchangeBackup([]string{suite.its.userID})
+	baseSel.Include(
+		// cannot be run, for the same reason as incremental backups
```
ashmrtn, line 876: nit: put the data type that can't be run so the comment is complete without having to determine what the commented out code does
```suggestion
		// Events cannot be run, for the same reason as incremental backups
```

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+			t,
+			[][]string{tconfig.M365AcctCredEnvs, storeTD.AWSStorageCredEnvs}),
+	})
+}
+
+func (suite *ExchangeRestoreIntgSuite) SetupSuite() {
+	suite.its = newIntegrationTesterSetup(suite.T())
+}
+
+func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptions() {
```
ashmrtn, line 866: I think this test is good for checking broad strokes of behavior but it doesn't ensure that corso takes the proper action on a per item basis. For example, assume no prior knowledge about how corso operated under the hood. In that case, an implementation that that looked at folder names and decided to take the collision policy action would also pass this test

To ensure proper behavior for individual items we should add a test that restores a subset of items in a folder to start with and have the different policies restore a (possibly disjoint) set of the remaining items in the folder

For example, backup a folder with items 1, 2, 3, and 4. Start by restoring item 1. Then, for each policy restore the next as-of-yet unrestored item and the first item (so [1, 2], [1, 3]...). Or if you didn't want to deal with the fact that item 1's ID would change for the replace policy, backup a folder with 6 items and restore 3 to start with and do [1, 4], [2, 5], [3, 6] for the different policies

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	closer func()
+}
+
+func (rod *restoreOpDependencies) close(
+	t *testing.T,
+	ctx context.Context, //revive:disable-line:context-as-argument
+) {
+	fmt.Printf("\n-----\nREST CLOSING\n-----\n")
+
```
ashmrtn, line 50:
```suggestion
```

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if rod.kms != nil {
+		err := rod.kw.Close(ctx)
+		assert.NoErrorf(t, err, "kms close: %+v", clues.ToCore(err))
+	}
+}
+
+// prepNewTestRestoreOp generates all clients required to run a restore operation,
+// returning both a resttore operation created with those clients, as well as
```
ashmrtn, line 65:
```suggestion
// returning both a restore operation created with those clients, as well as
```

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	err := k.Connect(ctx, repository.Options{})
+	require.NoError(t, err, clues.ToCore(err))
+
+	// kopiaRef comes with a count of 1 and Wrapper bumps it again
+	// we're so safe to close here.
+	rod.closer = func() {
+		err := k.Close(ctx)
+		assert.NoErrorf(t, err, "k close: %+v", clues.ToCore(err))
+	}
```
ashmrtn, line 97: no matter what it's safe to close k at the end of the function since it's not a part of the restore dependencies struct
```suggestion
	defer k.Close(ctx)
```

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	rod.kw, err = kopia.NewWrapper(k)
+	if !assert.NoError(t, err, clues.ToCore(err)) {
+		return operations.RestoreOperation{}, nil
+	}
+
+	rod.kms, err = kopia.NewModelStore(k)
+	if !assert.NoError(t, err, clues.ToCore(err)) {
+		return operations.RestoreOperation{}, nil
```
ashmrtn, line 106: this should return `rod` so the wrapper can be closed if it was successfully initialized
```suggestion
		return operations.RestoreOperation{}, rod
```

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+	t *testing.T,
+	ctx context.Context, //revive:disable-line:context-as-argument
+	ro *operations.RestoreOperation,
+	mb *evmock.Bus,
+	expectItemsWritten int, // can be -1 to skip the check
+	acceptNoData bool,
+) *details.Details {
+	deets, err := ro.Run(ctx)
+	require.NoError(t, err, clues.ToCore(err))
+	require.NotEmpty(t, ro.Results, "the restore had non-zero results")
```
ashmrtn, line 189: add a comment about why this check is alright. At first glance it seems like a restore that restores no data would fail this check

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if expectItemsWritten >= 0 {
+		require.Equal(t, ro.Results.ItemsWritten, expectItemsWritten, "count of items written")
+	}
+
+	assert.NotZero(t, ro.Results.ItemsRead, "count of items read")
+	assert.NotZero(t, ro.Results.BytesRead, "bytes read")
+	assert.Equal(t, 1, ro.Results.ResourceOwners, "count of resource owners")
+	assert.NoError(t, ro.Errors.Failure(), "incremental non-recoverable error", clues.ToCore(ro.Errors.Failure()))
+	assert.Empty(t, ro.Errors.Recovered(), "incremental recoverable/iteration errors")
```
ashmrtn, line 213:
```suggestion
	assert.NoError(t, ro.Errors.Failure(), "non-recoverable error", clues.ToCore(ro.Errors.Failure()))
	assert.Empty(t, ro.Errors.Recovered(), "recoverable/iteration errors")
```

src/internal/operations/test/restore_helper_test.go
```diff
@@ -0,0 +1,245 @@
...
+	assert.Equal(t, 1, mb.TimesCalled[events.RestoreStart], "restore-start events")
+	assert.Equal(t, 1, mb.TimesCalled[events.RestoreEnd], "restore-end events")
+
+	return deets
+}
+
+type giicbcker[T any] interface {
```
ashmrtn, line 220: this needs a more descriptive name

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+		m := filterCollisionKeyResults(
+			t,
+			ctx,
+			userID,
+			cIDs[path.EmailCategory],
+			giicbcker[string](acMail),
+			collKeys)
+		maps.Copy(result, m)
+
+		assert.Len(t, result, 0, "no new items should get added")
```
ashmrtn, line 1036: nit
```suggestion
		assert.Empty(t, result, "no new items should get added")
```

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+			t,
+			ctx,
+			userID,
+			cIDs[path.EmailCategory],
+			giicbcker[string](acMail),
+			collKeys)
+		maps.Copy(result, m)
+
+		// TODO: we have the option of modifying copy creations in exchange
+		// so that the results don't collid.  But we haven't made that
```
ashmrtn, line 1173:
```suggestion
		// so that the results don't collide.  But we haven't made that
```

src/internal/operations/test/exchange_test.go
ashmrtn: also, this test is not sufficient to say we didn't duplicate or replace items since there's no item ID comparison happening

src/internal/operations/test/sharepoint_test.go
```diff
@@ -177,3 +181,260 @@ func (suite *SharePointBackupIntgSuite) TestBackup_Run_sharePointExtensions() {
...
+			t,
+			[][]string{tconfig.M365AcctCredEnvs, storeTD.AWSStorageCredEnvs}),
+	})
+}
+
+func (suite *SharePointRestoreIntgSuite) SetupSuite() {
+	suite.its = newIntegrationTesterSetup(suite.T())
+}
+
+func (suite *SharePointRestoreIntgSuite) TestRestore_Run_sharepointWithAdvancedOptions() {
```
ashmrtn, line 202: it seems like much of the later part of this test could be pulled out into a helper function and used to run both OneDrive and SharePoint tests?

src/internal/operations/test/exchange_test.go
ryanfkeepers: I'm a little hesitant to run with that implementation as it ends up being a lot of lift and complexity for a relatively small gain.  The primary difference from restoring distinct subsets with expanding data would be to ensure that the collision case handling can pivot on a _per-item_ basis, as opposed to a _per-restore_ basis.  That's a level of granularity that's easyto guarantee with a unit test or a mocked integration test.  I can go look and see if we're adequately testing for that sort of complexity at a lower level, and add those tests if they're missing.  As for here, in the operations layer, I think it's acceptable to work with broader strokes, provided we have those granular tests covered elsewhere.

src/internal/operations/test/restore_helper_test.go
ryanfkeepers: whoops, thanks!

src/internal/operations/test/restore_helper_test.go
ryanfkeepers: It's not even just the `t` anymore, I'm getting this with many of my keys. :sob:

src/internal/operations/test/restore_helper_test.go
ryanfkeepers: good idea

src/internal/operations/test/restore_helper_test.go
ryanfkeepers: Heh, I was hoping the initialism would sufficiently pass for its func, to avoid the long name.  `GetItemsInContainerByCollisionKeyer` here we go!

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+	var (
+		restoreCfg = ctrlTD.DefaultRestoreConfig("exchange_adv_restore")
+		sel        = rsel.Selector
+		userID     = sel.ID()
+		cIDs       = map[path.CategoryType]string{
+			path.ContactsCategory: "",
+			path.EmailCategory:    "",
+			path.EventsCategory:   "",
+		}
+		collKeys = map[string]string{}
```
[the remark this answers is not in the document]
ryanfkeepers: If there's collisions across data types then we'll fail on checks that assert an equal quantity of details entries and collision keys.  Plus, I don't think we have any real possibility of cross-category collision (currently).  Still, I'll add a note here as a warning, just in case.

src/internal/operations/test/exchange_test.go
ryanfkeepers: I'll add support for this in a branch, and will link the pr when it's ready.

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+			cIDs[path.EmailCategory],
+			giicbcker[string](acMail),
+			collKeys)
+		maps.Copy(result, m)
+
+		assert.Len(t, result, 0, "all items should have been replaced")
+
+		for k, v := range result {
+			assert.NotEqual(t, v, collKeys[k], "replaced items should have new IDs")
+		}
```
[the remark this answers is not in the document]
ryanfkeepers: oops, check.

src/internal/operations/test/exchange_test.go
```diff
@@ -842,3 +845,333 @@ func testExchangeContinuousBackups(suite *ExchangeBackupIntgSuite, toggles contr
...
+			userID,
+			cIDs[path.EmailCategory],
+			giicbcker[string](acMail),
+			collKeys)
+		maps.Copy(result, m)
+
+		// TODO: we have the option of modifying copy creations in exchange
+		// so that the results don't collid.  But we haven't made that
+		// decision yet.
+		assert.Len(t, result, 0, "no items should have been added as copies")
```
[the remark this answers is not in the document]
ryanfkeepers: Yup, will also fix this in the next update.

src/internal/operations/test/sharepoint_test.go
ryanfkeepers: very possible.  Going to save that for a patch on this PR.

src/internal/operations/test/exchange_test.go
ashmrtn: were you able to find any more granular tests?

src/internal/operations/test/exchange_test.go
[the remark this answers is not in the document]
ashmrtn: is there a plan for checks like this? The subsequent PRs don't seem to address it and I'm not sure if there's some other plan for checking the items have or haven't been replaced/duplicated as expected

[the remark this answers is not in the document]
ryanfkeepers: subsequent PRs aren't wrapped yet.

src/internal/operations/test/exchange_test.go
ryanfkeepers: Yup, each exchange data category has a per-collision set of tests like [this one](https://github.com/alcionai/corso/blob/main/src/internal/m365/exchange/mail_restore_test.go#L102).  And drive restores have a similar test case [here](https://github.com/alcionai/corso/blob/main/src/internal/m365/onedrive/restore_test.go#L329).  I need to go hook up the new counters to those as assurance.  Otherwise, I think that covers us on the implementation regression in a per-item basis.

src/internal/operations/test/exchange_test.go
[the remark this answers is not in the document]
ryanfkeepers: @ashmrtn how's this: https://github.com/alcionai/corso/pull/3805?

src/internal/operations/test/exchange_test.go
ashmrtn: after some out of band discussion the plan is to get this PR merged along with ID checks for items. After that we can go back and fill out testing a bit more by adding some test cases where only some of the items being restored have the collision policy applied

ashmrtn (approved):
thanks for all the work on this PR and for bearing with me through all the different asks I had 😅

PR #3788, opened 2023-07-12
add counting for replace and new items

ryanfkeepers:
adds counters during restoration for
replaced and newly created (copy and
non-colliding) items.  The initial value is
to facilitate test assurances.  But in this we
begin a slow movement towards using
a standardized bus for operations counts instead
of consumer/producer structs.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/contacts_restore.go
+++ b/src/internal/m365/exchange/contacts_restore.go
@@ -146,17 +146,19 @@ func restoreContact(
 	// at least we'll have accidentally over-produced data instead of deleting
 	// the user's data.
 	if shouldDeleteOriginal {
-		ctr.Inc(count.CollisionReplace)
-
 		if err := cr.DeleteItem(ctx, userID, collisionID); err != nil && !graph.IsErrDeletedInFlight(err) {
 			return nil, graph.Wrap(ctx, err, "deleting colliding contact")
 		}
-	} else {
-		ctr.Inc(count.NewItemCreated)
 	}
 
 	info := api.ContactInfo(item)
 	info.Size = int64(len(body))
+
+	if shouldDeleteOriginal {
+		ctr.Inc(count.CollisionReplace)
+	} else {
+		ctr.Inc(count.NewItemCreated)
+	}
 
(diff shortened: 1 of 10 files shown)


src/internal/operations/test/exchange_test.go
```diff
@@ -971,6 +975,9 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			userID,
 			cIDs[path.EmailCategory])
 		require.NoError(t, err, clues.ToCore(err))
+		countItemsInRestore += len(collKeys[path.EmailCategory])
+
+		checkRestoreCounts(t, ctr, 0, 0, countItemsInRestore)
```
ashmrtn, line 980: I can't comment on it directly, but it seems like we should be making a new counter for every restore we do right? Otherwise the counter will have the cumulative count from all restores and that seems like it would be a bit difficult for readers to reason through if just skimming the code

ashmrtn (approved):

src/internal/m365/exchange/contacts_restore.go
```diff
@@ -146,9 +146,13 @@ func restoreContact(
 	// at least we'll have accidentally over-produced data instead of deleting
 	// the user's data.
 	if shouldDeleteOriginal {
+		ctr.Inc(count.CollisionReplace)
+
```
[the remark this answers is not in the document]
ryanfkeepers: Hmm, good thought.  Probably the latter.  I'll move that.

src/internal/operations/test/exchange_test.go
ryanfkeepers: The counters are unique to each run.  I could rename them if that'll ease confusion.

ryanfkeepers: no prob!

PR #3790, opened 2023-07-12
Longevity test- delete backup after n days

neha-Gupta1:
Longevity Test- Delete the backups older than n no of days. Currently n's value is 5. 

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* https://github.com/alcionai/corso/issues/3679

#### Test Plan

- [x] :zap: Unit test


src/cmd/longevity_test/longevity.go
ashmrtn, line 1: I'm a bit split on how we're calling the delete here. I think there's benefit to exercising the CLI-layer delete/backup list commands, but I feel that using it here is a bit too fragile

I think it'd be much simpler (and easier to deal with errors) if the `pkg/repository` package interface was used instead of running commands via golang. The interface for looking up backups and deleting backups doesn't require many parameters so it should be fairly simple to access once a repository struct instance is obtained. Since we're not using any special flags to denote the bucket/prefix or anything like that something simple like the below should work
```go
r, _, _, err := utils.GetAccountAndConnect(ctx, service, nil)
if err != nil {...}
defer r.Close(ctx)

backups, err := r.BackupsByTag(ctx, store.Service(service))

// logic to select backups by timestamp

for _, backupID := range idsToDelete {
  if err := r.DeleteBackup(ctx, id); err != nil {...}
}
```

The example code I've given is modeled closely after the [backup list CLI command](https://github.com/alcionai/corso/blob/4d4f5dc93460966040487ceb76d339026383bfa0/src/cli/backup/backup.go#L297)

Given that it's also important to test the CLI functionality of deleting backups, perhaps the sanity tests would be a good place to put those tests. There I believe the backup ID is already known so there won't be a need for additional parsing of output

.github/workflows/longevity_test.yml
```diff
@@ -194,6 +191,48 @@ jobs:
 
 ##########################################################################################################################################
 
+# Backup Exchange Deletion test
+      - name: Backup Delete exchange test
+        id: delete-exchange-test
+        env:
+          SERVICE: "exchange"
+          DELETION_DAYS: 5
```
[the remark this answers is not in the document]
neha-Gupta1: Kept it to 10 now.

.github/workflows/longevity_test.yml
```diff
@@ -194,6 +191,48 @@ jobs:
...
+# Backup Exchange Deletion test
+      - name: Backup Delete exchange test
+        id: delete-exchange-test
+        env:
+          SERVICE: "exchange"
+          DELETION_DAYS: 10
+        run: |
+          set -euo pipefail
+          echo -e "\nDelete Backup Exchange test\n" >> ${CORSO_LOG_FILE}
+          CORSO_LOG_FILE=${{ env.CORSO_LOG_DIR }}/backup_exchange.log
```
ashmrtn, line 203: will this env var assignment be used in the `./longevity-test` command? I'm a little rusty on bash, but I thought either it would need to be like `export CORSO_LOG_FILE=...` (to end up in the environment from that point forward) or be on the same line as the command being run (to be used as part of the environment only for the command being run)

ashmrtn (approved):

Issue #3519, opened 2023-05-26
Immutable Backups


vkamra:
Leverage S3 object locking/versioning and retention to provide guarantees that backups cannot be removed or tampered with for the specified duration

Additional work on testing:
* #3799

vkamra:
Test

ashmrtn:
non-CLI support for this is done. There's some CLI code that I can push that should be kept hidden but we should use a separate ticket to track that

ashmrtn:
closing this as SDK support has been merged to main

PR #3792, opened 2023-07-12
Kopia wrapper code to set retention parameters

ashmrtn:
Code at the kopia wrapper layer that allows configuring kopia
for retention if desired. Includes some tests that require
running against a bucket with object locking enabled

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3519

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/ci.yml
ashmrtn, line 1: I merged things in the wrong order 😞 This is from #3813 so can be ignored

src/pkg/control/repository/repo.go
```diff
@@ -51,3 +51,25 @@ const (
 	// https://github.com/kopia/kopia/blob/f9de453efc198b6e993af8922f953a7e5322dc5f/repo/maintenance/maintenance_safety.go#L42
 	NoMaintenanceSafety
 )
+
+type RetentionMode int
+
+// Can't be reordered as we rely on iota for numbering.
```
ryanfkeepers, line 57: I'm (still) increasingly wary of iota, for exactly this reason.  I just wrapped another PR where the author tried to slip a value into the middle of an iota enumeration.  Purely an accident, but the sort of accident that's easy to make and difficult to catch if it goes through (short of writing equivalence unit tests).

At minimum, I think an equivalence unit test is warranted for any iota enumeration, for protection.  Or, alternatively, we could use hardcoded ints instead of the iota.

src/internal/kopia/wrapper.go
```diff
@@ -710,3 +713,184 @@ func (w Wrapper) setMaintenanceParams(
...
+) error {
+	if retention.Mode == nil && retention.Duration == nil && retention.Extend == nil {
+		return nil
+	}
+
+	// Somewhat confusing case, when we have no retention but a non-zero duration
+	// it acts like we passed in only the duration and returns an error about
+	// having to set both. Return a clearer error here instead.
+	if ptr.Val(retention.Mode) == repository.NoRetention && ptr.Val(retention.Duration) != 0 {
+		return clues.New("retention duration must be 0 or unset if disabling retention")
```
ryanfkeepers, line 732: Nit: smaller error
```suggestion
		return clues.New("duration must be 0 if retention is disabled").WithClues(ctx)
```

src/internal/kopia/wrapper.go
```diff
@@ -710,3 +713,184 @@ func (w Wrapper) setMaintenanceParams(
...
+	}
+
+	dr, ok := w.c.Repository.(repo.DirectRepository)
+	if !ok {
+		return clues.New("unable to get valid handle to repo").WithClues(ctx)
```
ryanfkeepers, line 737: Per error standards, please avoid phrases like "unable to foo"
```suggestion
		return clues.New("getting handle to repo").WithClues(ctx)
```

src/internal/kopia/wrapper.go
```diff
@@ -710,3 +713,184 @@ func (w Wrapper) setMaintenanceParams(
...
+		return clues.Stack(err)
+	}
+
+	// Update blob config information.
+	blobChanged, err := w.setRetentionMode(retention.Mode, blobCfg)
+	if err != nil {
+		return clues.Wrap(err, "setting retention mode")
+	}
+
+	if retention.Duration != nil && blobCfg.RetentionPeriod != *retention.Duration {
```
ryanfkeepers, line 751: This opens up the case where RetentionPeriod is non-zero, and Duration is nil, which sets RetentionPeriod to zero.  Not sure if that'd be expected/wanted.  Disregard if needed.
```suggestion
	if blobCfg.RetentionPeriod != ptr.Val(retention.Duration) {
```

src/internal/kopia/wrapper.go
```diff
@@ -710,3 +713,184 @@ func (w Wrapper) setMaintenanceParams(
...
+
+	blobCfg, params, err := getRetentionConfigs(ctx, dr)
+	if err != nil {
+		return clues.Stack(err)
+	}
+
+	// Update blob config information.
+	blobChanged, err := w.setRetentionMode(retention.Mode, blobCfg)
+	if err != nil {
+		return clues.Wrap(err, "setting retention mode")
```
ryanfkeepers, line 748:
```suggestion
		return clues.Wrap(err, "setting retention mode").WithClues(ctx)
```

src/internal/kopia/wrapper.go
```diff
@@ -710,3 +713,184 @@ func (w Wrapper) setMaintenanceParams(
...
+	maintenanceChanged bool,
+) error {
+	// Persist changes.
+	if !blobChanged && !maintenanceChanged {
+		return nil
+	}
+
+	mp, err := dr.FormatManager().GetMutableParameters()
+	if err != nil {
+		return clues.Wrap(err, "getting mutable parameters")
```
ryanfkeepers, line 813: This is needed for many of the errors in this func.
```suggestion
		return clues.Wrap(err, "getting mutable parameters").WithClues(ctx)
```

src/internal/kopia/wrapper_test.go
```diff
@@ -277,6 +281,325 @@ func (suite *BasicKopiaIntegrationSuite) TestMaintenance_WrongUser_Force_Succeed
...
+
+	w := &Wrapper{k}
+
+	// Enable retention.
+	err = w.SetRetentionParameters(ctx, repository.Retention{
+		Mode:     ptr.To(repository.GovernanceRetention),
+		Duration: ptr.To(time.Hour * 48),
+		Extend:   ptr.To(true),
+	})
+	require.Error(t, err)
```
ryanfkeepers, line 304: Nit: standard pattern
```suggestion
	require.Error(t, err, clues.ToCore(err))
```

src/internal/kopia/wrapper_test.go
```diff
@@ -277,6 +281,325 @@ func (suite *BasicKopiaIntegrationSuite) TestMaintenance_WrongUser_Force_Succeed
...
+			opts: repository.Retention{
+				Mode:     ptr.To(repository.GovernanceRetention),
+				Duration: ptr.To(time.Hour * 48),
+			},
+			expectErr:      assert.NoError,
+			expectMode:     blob.Governance,
+			expectDuration: time.Hour * 48,
+			expectExtend:   assert.False,
+		},
+		//{
```
ryanfkeepers, line 455: Could you please add a comment here stating why this case is skipped?

ryanfkeepers (approved):
LGTM.  Just nits.  @pandeyabs should probably review as well just for a second set of eyes.

src/pkg/control/repository/repo.go
```diff
@@ -51,3 +51,25 @@ const (
...
+//
+//go:generate stringer -type=RetentionMode -linecomment
+const (
+	UnknownRetention RetentionMode = iota
+	NoRetention
+	GovernanceRetention
```
pandeyabs, line 63: Any reason we didn't reuse kopia definitions?

src/internal/kopia/wrapper.go
pandeyabs, line 751: nit: Could lines 751 to 753 be merged into `setRetentionMode`? Maybe call it `setRetentionParams` or similar.

We are modifying `blobChanged` we received from 746. It'd probably be better consolidate all blob change detection logic into a single function.

pandeyabs (approved):
Looks good! left a few comments & nits.

src/pkg/control/repository/repo.go
pandeyabs: Oh yeah! Thanks, didn't consider direct dependency aspect. Totally agree

PR #3793, opened 2023-07-13
add api funcs for creating documentLibs

ryanfkeepers:
Adds api handlers for creating document libraries in sharepoint. This is the first step in allowing us to restore drives that were deleted between backup and restore.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/lists.go
```diff
@@ -0,0 +1,64 @@
...
+
+// PostDrive creates a new list of type drive.  Specifically used to create
+// documentLibraries for SharePoint Sites.
+func (c Lists) PostDrive(
+	ctx context.Context,
+	siteID, driveName string,
+) (models.Driveable, error) {
+	list := models.NewList()
+	list.SetDisplayName(&driveName)
+	list.SetDescription(ptr.To("corso auto-generated restore destination"))
```
meain, line 34: Just a question for the future. Should we have the name in here or provide a way to customize the name here in case SDK consumer wants change the description. For example to "Ark auto-generated restore destination".

meain (approved):
LGTM

src/pkg/services/m365/api/lists.go
ryanfkeepers: Good question.  I was going to ping Georgi about that later.  For now I think we're safe following standard folder naming conventions.  Though, now that I think about it, that doesn't allow SDK customers to override the corso naming conventions.  I'll probably do a follow-up PR where this gets added to the RestoreConfig.

src/internal/m365/onedrive/handlers.go
```diff
@@ -86,7 +90,9 @@ type RestoreHandler interface {
 	GetItemsByCollisionKeyser
 	GetRootFolderer
 	ItemInfoAugmenter
+	NewDrivePagerer
```
pandeyabs, line 93: Can we add short descriptions for these interfaces?

src/internal/m365/onedrive/item_handler.go
```diff
@@ -133,6 +134,19 @@ func NewRestoreHandler(ac api.Client) *itemRestoreHandler {
 	return &itemRestoreHandler{ac.Drives()}
 }
 
+func (h itemRestoreHandler) PostDrive(
+	context.Context,
+	string, string,
+) (models.Driveable, error) {
+	return nil, clues.New("creating drives in oneDrive is not supported")
```
pandeyabs, line 141: Curious, what's our plan of action if a onedrive drive is deleted between backup & restore?

src/pkg/services/m365/api/lists.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, graph.Wrap(ctx, err, "creating documentLibrary list")
+	}
+
+	// drive information is not returned by the list creation.
+	drive, err := builder.
+		ByListId(ptr.Val(newList.GetId())).
+		Drive().
+		Get(ctx, nil)
+
+	return drive, graph.Wrap(ctx, err, "fetching created documentLibrary").OrNil()
```
pandeyabs, line 63: nit: First time seeing `OrNil` usage in our codebase. Do we need to use `OrNil` here? Can we switch to the `if err != nil` pattern to be consistent?

pandeyabs (approved):
Looks good! Left some comments

src/internal/m365/onedrive/handlers.go
ryanfkeepers: I can add descriptions to the interface declarations, yes.

src/internal/m365/onedrive/item_handler.go
ryanfkeepers: I don't know that it's traditionally possible to delete a onedrive drive.  It's already a very strange case that a user's onedrive can contain more than one... drive...  As a result, I think this handling is appropriate.

That said, this implementation will have to change a bit when we introduce support for moving data between users, since the current handling will error instead of placing the data in the destination user's drive.  That's a change for another PR, however.

src/pkg/services/m365/api/lists.go
ryanfkeepers: `OrNil` is a recent addition to Clues, which is why you're only seeing it now.  Either pattern is fine.  Personally I'm going to start transitioning over to the use of `OrNil`, where appropriate.  No plans for a single-pass refactor at this time.

PR #3795, opened 2023-07-13
create missing drives on restore

ryanfkeepers:
when restoring sharepoint, if a document library
was deleted between the time of backup and restore, create a new drive to hold the restored data.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/restore.go
+++ b/src/internal/m365/onedrive/restore.go
@@ -255,7 +255,7 @@ func RestoreCollection(
 		protectedResourceID,
 		fallbackDriveName)
 	if err != nil {
-		return metrics, clues.Wrap(err, "ensureing drive exists")
+		return metrics, clues.Wrap(err, "ensuring drive exists")
 	}
 
 	// Assemble folder hierarchy we're going to restore into (we recreate the folder hierarchy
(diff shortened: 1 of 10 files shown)


src/internal/m365/onedrive/restore.go
```diff
@@ -174,13 +247,15 @@ func RestoreCollection(
...
-		caches.DriveIDToRootFolderID[drivePath.DriveID] = ptr.Val(root.GetId())
+	err = ensureDriveExists(
+		ctx,
+		rh,
+		caches,
+		drivePath,
+		protectedResourceID,
+		fallbackDriveName)
+	if err != nil {
+		return metrics, clues.Wrap(err, "ensureing drive exists")
```
pandeyabs, line 258: nit: `ensuring`

src/internal/m365/onedrive/restore.go
```diff
@@ -1107,3 +1182,70 @@ func AugmentRestorePaths(
...
+
+	var (
+		newDriveName = driveName
+		newDrive     models.Driveable
+		err          error
+		i            = 1
+	)
+
+	if _, ok := caches.DriveNameToDriveInfo[newDriveName]; ok {
+		newDriveName = fmt.Sprintf("%s %d", driveName, 1)
```
pandeyabs, line 1218: nit: this could take `i` as a param, instead of 1

pandeyabs (approved):
LGTM! Left some nits

src/internal/m365/onedrive/restore.go
ryanfkeepers: done

src/internal/m365/onedrive/restore.go
ryanfkeepers: switched it around so that i is only in the for loop handler, instead of being a variable outside of it.  Makes it clearer how it's being used.

PR #3802, opened 2023-07-13
retry on http/2 stream errors

ryanfkeepers:
http/2 stream error failures don't allow
us to leverage client middleware.  As a
result we're lacking a retry handler.  This
change adds retries for http_wrapper
in the explicit case of a http/2 stream
error.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/graph/http_wrapper.go
+++ b/src/internal/m365/graph/http_wrapper.go
@@ -4,11 +4,15 @@ import (
 	"context"
 	"io"
 	"net/http"
+	"regexp"
+	"time"
 
 	"github.com/alcionai/clues"
 	khttp "github.com/microsoft/kiota-http-go"
 
+	"github.com/alcionai/corso/src/internal/events"
 	"github.com/alcionai/corso/src/internal/version"
+	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/internal/m365/graph/http_wrapper.go
```diff
@@ -135,6 +169,7 @@ func (pl pipeline) Next(req *http.Request, idx int) (*http.Response, error) {
 func defaultTransport() http.RoundTripper {
 	defaultTransport := http.DefaultTransport.(*http.Transport).Clone()
 	defaultTransport.ForceAttemptHTTP2 = true
+	defaultTransport.IdleConnTimeout = 0 // no limit
```
pandeyabs, line 172: Do we need this particular change given we are retrying http2 errors now? What is the benefit of this change?

I'm a bit concerned if this can have unintended side effects. Not exhaustive, but some scenarios come to mind.

1. Idle connections can become stale, especially if we are doing long running backups. 
2. If idle connections don't get reused for some reason, we may hit max connection limits with graph (assuming there is one).

pandeyabs (changes requested):

src/internal/m365/graph/http_wrapper.go
ryanfkeepers: Ah, that was from an earlier test.  I think we can take it out safely.

pandeyabs (approved):

PR #3805, opened 2023-07-14
add extra restore checks

ryanfkeepers:
Adds extra checks to the restore op tests that
retrieves and compares the lists of item ids
after each restore step.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/test/exchange_test.go
+++ b/src/internal/operations/test/exchange_test.go
@@ -1238,7 +1238,7 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 		require.NoError(t, err, clues.ToCore(err))
 
 		assert.Equal(t, 2*len(mailIDs), len(currentMailIDs), "count of ids should be double from before")
-		assert.Subset(t, maps.Keys(mailIDs), maps.Keys(currentMailIDs), "original item should exist after copy")
+		assert.Subset(t, maps.Keys(currentMailIDs), maps.Keys(mailIDs), "original item should exist after copy")
 
 		// TODO: we have the option of modifying copy creations in exchange
 		// so that the results don't collide.  But we haven't made that
--- a/src/internal/operations/test/exchange_test.go
+++ b/src/internal/operations/test/exchange_test.go
@@ -903,14 +903,17 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			path.EmailCategory:    "",
 			path.EventsCategory:   "",
 		}
+		collKeys               = map[path.CategoryType]map[string]string{}
 		countContactsInRestore int
+		acCont                 = suite.its.ac.Contacts()
+		contactIDs             map[string]struct{}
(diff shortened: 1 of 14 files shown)


src/internal/operations/test/exchange_test.go
```diff
@@ -1092,9 +1126,24 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
 			GetItemsInContainerByCollisionKeyer[string](acCont),
 			collKeys[path.ContactsCategory])
 
+		currentContactIDs, err := acCont.GetItemIDsInContainer(ctx, userID, cIDs[path.ContactsCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		assert.Equal(t, len(contactIDs), len(currentContactIDs), "count of ids ids are equal")
```
ashmrtn, line 1132: nit
```suggestion
		assert.Equal(t, len(contactIDs), len(currentContactIDs), "count of ids are equal")
```

src/internal/operations/test/exchange_test.go
```diff
@@ -1092,9 +1126,24 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
...
+		for orig := range contactIDs {
+			_, ok := currentContactIDs[orig]
+			assert.False(t, ok, "original item should not exist after replacement")
```
ashmrtn, line 1135: you could maybe try `assert.NotContains` is you want. Not sure if the test output from that would be any clearer though

src/internal/operations/test/exchange_test.go
```diff
@@ -1157,6 +1218,15 @@ func (suite *ExchangeRestoreIntgSuite) TestRestore_Run_exchangeWithAdvancedOptio
...
 			collKeys[path.ContactsCategory])
 
+		currentContactIDs, err := acCont.GetItemIDsInContainer(ctx, userID, cIDs[path.ContactsCategory])
+		require.NoError(t, err, clues.ToCore(err))
+
+		assert.Equal(t, 2*len(contactIDs), len(currentContactIDs), "count of ids should be double from before")
+		for orig := range contactIDs {
+			_, ok := currentContactIDs[orig]
+			assert.True(t, ok, "original item should exist after copy")
+		}
```
ashmrtn, line 1228: I think a call to `assert.Subset(t, currentContactIDs, contactIDs)` would work here?

ashmrtn (approved):
just a few minor things about other possible functions that could be used. Thanks for adding this!

src/internal/operations/test/exchange_test.go
ryanfkeepers: Can't do it with maps, but `maps.Keys()` works just fine.

src/internal/operations/test/exchange_test.go
ryanfkeepers: Worked!

PR #3812, opened 2023-07-14
add test for 503 resp body

### LINKED ISSUE #3811: [Bug]: 503 retry failure loses response body
### What happened?

Failing out of all middleware retries after consecutive 503's can erase the response body, causing the graph client to return a `content is empty` error instead of a 503 error.

### Corso Version?

Corso v0.10.0

any

### Relevant log output

_No response_

ryanfkeepers:
#### Type of change

#### Issue(s)

* closes #3811

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/m365/graph/errors_test.go
+++ b/src/internal/m365/graph/errors_test.go
@@ -55,7 +55,7 @@ func parseableToMap(t *testing.T, thing serialization.Parsable) map[string]any {
 	require.NoError(t, err, "serialize")
 
 	content, err := sw.GetSerializedContent()
-	require.NoError(t, err, "serialize")
+	require.NoError(t, err, "deserialize")
 
 	var out map[string]any
 	err = json.Unmarshal([]byte(content), &out)
--- a/src/internal/m365/graph/errors_test.go
+++ b/src/internal/m365/graph/errors_test.go
@@ -2,14 +2,18 @@ package graph
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
 	"syscall"
 	"testing"
 
 	"github.com/alcionai/clues"
+	"github.com/microsoft/kiota-abstractions-go/serialization"
+	kjson "github.com/microsoft/kiota-serialization-json-go"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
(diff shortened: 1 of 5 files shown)


src/internal/m365/graph/errors_test.go
```diff
@@ -44,6 +48,22 @@ func odErrMsg(code, message string) *odataerrors.ODataError {
...
 }
 
+func parseableToMap(t *testing.T, thing serialization.Parsable) map[string]any {
+	sw := kjson.NewJsonSerializationWriter()
+
+	err := sw.WriteObjectValue("", thing)
+	require.NoError(t, err, "serialize")
+
+	content, err := sw.GetSerializedContent()
+	require.NoError(t, err, "serialize")
```
ashmrtn, line 58: nit
```suggestion
	require.NoError(t, err, "deserialize")
```

src/internal/m365/graph/middleware.go
```diff
@@ -252,93 +245,88 @@ func (mw RetryMiddleware) retryRequest(
...
 
-	ctx = clues.Add(
-		ctx,
-		"prev_resp_status", status,
-		"retry_count", executionCount)
-
-	// only retry under certain conditions:
-	// 1, there was an error.  2, the resp and/or status code match retriable conditions.
+	// only retry if all the following conditions are met:
+	// 1, there was a prior error.
```
ashmrtn, line 255: I'm a little confused by this and the lead comment on the previous line. It seems like we want "there was a prior error or the below"

src/internal/m365/graph/middleware.go
```diff
@@ -252,93 +245,88 @@ func (mw RetryMiddleware) retryRequest(
...
-			nextResp,
-			executionCount,
-			cumulativeDelay,
-			exponentialBackoff,
-			err)
+	// we have to reset the original body reader for each retry, or else the graph
+	// compressor will produce a 0 length body following an error response such
+	// as a 500.
+	if req.Body != nil {
+		if s, ok := req.Body.(io.Seeker); ok {
```
ashmrtn, line 289: I know this isn't code directly modified in this PR, but should we be logging an error or something if we don't have an `io.Seeker`?

src/internal/m365/graph/middleware.go
```diff
@@ -252,93 +245,88 @@ func (mw RetryMiddleware) retryRequest(
...
+			}
+		}
 	}
 
-	if priorErr != nil {
-		return nil, stackReq(ctx, req, nil, priorErr)
+	nextResp, err := pipeline.Next(req, middlewareIndex)
+	if err != nil && !IsErrTimeout(err) && !IsErrConnectionReset(err) {
+		return nextResp, stackReq(ctx, req, nextResp, err)
 	}
```
ashmrtn, line 299: this also isn't directly modified in this PR but it does feel rather weird to check for a subset of errors and explicitly return on those rather than calling this function again with the response which could then handle it at the start of the function (or may already handle it)

ashmrtn (approved):

pandeyabs (approved):

src/internal/m365/graph/middleware.go
ryanfkeepers: ah, good catch, will fix

src/internal/m365/graph/middleware.go
ryanfkeepers: I don't see any harm in it.

src/internal/m365/graph/middleware.go
ryanfkeepers: heh, no disagreement there.  I think that's a case which would fit more nicely into a func refactor, which we can handle some other time.

PR #3797, opened 2023-07-13
Add option to export data for OneDrive

meain:
This adds the option to export data for OneDrive. This can be used to export the full backup, a single file or a subset of files.

Here is what the cli usage looks like:

```
corso export <service> [flags] [--archive] <destination>
```

---

#### Does this PR need a docs update or release note?

- [x] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Supportability/Tests
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/3670
* fixes https://github.com/alcionai/corso/issues/2536

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

Issue #3670, opened 2023-06-26
File export from OneDrive backup


meain:
https://github.com/alcionai/corso-roadmap/issues/43

Ability to export files/foldres from OneDrive directly to local filesystem instead of restoring. SDK users should also be able to extract files through corso.

breisig:
This would be a excellent feature to have.

meain:
> This would be a excellent feature to have.
This is currently being worked on in https://github.com/alcionai/corso/pull/3797. No final timelines yet, but should have something in the coming weeks.

meain:
This is now available in `main`.

PR #3819, opened 2023-07-17
Export data from OneDrive

meain:
Core logic for exporting data from OneDrive

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3670
* https://github.com/alcionai/corso/pull/3797

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/onedrive/export_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+			fitems := []data.ExportItem{}
+			for item := range items {
+				fitems = append(fitems, item)
+			}
+
+			assert.Len(t, fitems, len(test.expectedItems), "num of items")
+
+			for i, item := range fitems {
+				assert.Equal(t, test.expectedItems[i].ID, item.ID, "id")
```
ashmrtn, line 342: I'm a bit on the fence about expecting things in the same order. Nothing actually depends on the returned order of the items (and really we don't want to enforce some ordering in the exported API if we don't have to). Unfortunately, not relying on ordering seems like it would just make the test more complex since we need non-trivial comparisons for things like errors

src/internal/m365/onedrive/export_test.go
```diff
@@ -0,0 +1,401 @@
...
+			version: 4,
+			name:    "name",
+			errFunc: assert.NoError,
+		},
+		{
+			tname:   "name in metadata",
+			id:      "name.data",
+			version: version.Backup,
+			name:    "name",
+			fin:     finD{id: "name.data", name: "name"},
```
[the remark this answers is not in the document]
meain: Good catch, I was not checking this in the mock struct before returning.

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,160 @@
...
+					continue
+				}
+
+				name, err := getItemName(ctx, itemUUID, ec.version, c)
+
+				ch <- data.ExportItem{
+					ID: itemUUID,
+					Data: data.ExportItemData{
+						Name: name,
+						Body: item.ToReader(),
```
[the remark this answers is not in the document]
meain: Since this is reading from an unbuffered channel and writing to another unbuffered channel just when it is called, I feel like we don't have to worry about that.

src/internal/m365/onedrive/export_test.go
```diff
@@ -0,0 +1,401 @@
...
+					FetchItemByNamer: finD{err: assert.AnError},
+				},
+			},
+			expectedItems: []data.ExportItem{
+				{
+					ID:    "id1.data",
+					Error: assert.AnError,
+				},
+			},
+		},
```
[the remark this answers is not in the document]
meain: Hmm, this brought up a hole in my implementation. I was not returning recoverable errors like the ones that could happen [here](https://github.com/alcionai/corso/blob/3efb67e110870db817ef31c5384f91d8c8c5eb9e/src/internal/kopia/data_collection.go#L45). I could return something like below for now but they won't have id's associated with them. Wondering if we should be doing more than that. 🤔 

```go
		for _, ec := range errs.Errors().Recovered {
			ch <- export.Item{
				// Convert recovered errors to simpler errors. These
				// will not have an ID associated with them.
				Error: errors.New(ec.String()),
			}
		}
```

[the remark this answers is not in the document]
meain: /cc @ryanfkeepers Should clues have some way to bind id of an item.

src/internal/m365/onedrive/export_test.go
meain: I've left a note in there for now. We can update it later if necessary I guess.

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (ec exportCollection) GetBasePath() string {
+	return ec.baseDir
+}
+
+func (ec exportCollection) GetItems(ctx context.Context) <-chan export.Item {
+	ch := make(chan export.Item)
+
+	go func() {
```
ryanfkeepers, line 46: This pyramid gets pretty deep.  While it may not seem like much, could you split out this inner func to it's own function?  Then you can call `go items(ctx...)` here.  That gives us two benefits: 1, unnests the func a little; 2, gives us a non-async func that we can call directly for unit and integration testing.

src/internal/m365/onedrive/export.go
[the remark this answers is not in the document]
ryanfkeepers: Ash has a point here.  `ch` is unbuffered, and we have no guarantees about the consumer keeping up with the channel.  That means we can load many potentially very large items into memory without limit.  A lazy reader is one way to solve this.  Limiting the channel is another.

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+						Name: name,
+						Body: item.ToReader(),
+					},
+					Error: err,
+				}
+			}
+		}
+
+		// Return all the items that we failed to get from kopia at the end
+		for _, err := range errs.Errors().Items {
```
ryanfkeepers, line 74: `errs.Errors()` is expensive, and should only get called once.

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+				ID:    err.ID,
+				Error: &err,
+			}
+		}
+
+		for _, ec := range errs.Errors().Recovered {
+			ch <- export.Item{
+				// Convert recovered errors to simpler errors. These
+				// will not have an ID associated with them.
+				Error: errors.New(ec.String()),
```
ryanfkeepers, line 85: Hmm, I don't like this pattern at all.  It slices out all of the valuable structured data and traces.  Given the full handling of errors here, instead of calling `errs.Errors()`, I suggest you add a new func to fault similar to the `itemsIn` func, which iterates over the RecoveredErrors list, deduplicating and dividing the entries into items and non-items, but returning a canonical `errors` value.  That'll maximize our capacity for debugging code here.

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+	}()
+
+	return ch
+}
+
+// isMetadataFile is used to determine if a path corresponds to a
+// metadata file.  This is OneDrive specific logic and depends on the
+// version of the backup unlike metadata.IsMetadataFile which only has
+// to be concerned about the current version.
+func isMetadataFile(id string, backupVersion int) bool {
```
ryanfkeepers, line 97: doesn't this func already exist in `onedrive/metadata`?

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+	if backupVersion < version.OneDrive1DataAndMetaFiles {
+		return false
+	}
+
+	return strings.HasSuffix(id, metadata.MetaFileSuffix) ||
+		strings.HasSuffix(id, metadata.DirMetaFileSuffix)
+}
+
+// getItemName is used to get the name of the item as how we get the
+// name depends on the version of the backup.
```
ryanfkeepers, line 107: Nit: easier to read.
```suggestion
// getItemName is used to get the name of the item.
// how we get the name depends on the version of the backup.
```

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+	backupVersion int,
+	exportCfg control.ExportConfig,
+	opts control.Options,
+	dcs []data.RestoreCollection,
+	deets *details.Builder,
+	errs *fault.Bus,
+) ([]export.Collection, error) {
+	el := errs.Local()
+
+	ec := make([]export.Collection, 0, len(dcs))
```
ryanfkeepers, line 150:
```suggestion
        var (
	  el = errs.Local()
	  colls = make([]export.Collection, 0, len(dcs))
	)
```

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+
+	for _, dc := range dcs {
+		drivePath, err := path.ToDrivePath(dc.FullPath())
+		if err != nil {
+			return nil, clues.Wrap(err, "creating drive path").WithClues(ctx)
```
ryanfkeepers, line 155: Since you're not creating a new path here.
```suggestion
			return nil, clues.Wrap(err, "transforming path to drive path").WithClues(ctx)
```

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		baseDir := &path.Builder{}
+		baseDir = baseDir.Append(drivePath.Folders...)
```
ryanfkeepers, line 159:
```suggestion
		baseDir := path.Builder{}.Append(drivePath.Folders...)
```

src/internal/m365/onedrive/export_test.go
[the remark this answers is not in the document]
ryanfkeepers: Clues is purely informational, and should not be used as a canonical container of data.  If you want to associate an error with a specific item you should use `fault/item`, which complies with the error interface and can be cast back to the item struct for handling.

src/pkg/export/export.go
```diff
@@ -0,0 +1,45 @@
...
+	"io"
+)
+
+// Collection is the interface that is returned to the SDK consumer
+type Collection interface {
+	// GetBasePath gets the base path of the collection
+	GetBasePath() string
+
+	// GetItems gets the items within the collection(folder)
+	GetItems(context.Context) <-chan Item
```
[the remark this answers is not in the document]
ashmrtn: no strong opinions in this case. I have found that having functions named the same across different interfaces can make using tools like grep/GitHub's code explorer more difficult because they don't always take into account type information. In this case though I'm not sure what a better name would be and I have no strong opinion on including `Get` in function names

src/internal/m365/onedrive/export.go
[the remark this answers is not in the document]
ashmrtn: unrelated to whether we use a lazy reader or not, but in our case it's actually the small items that will lead to high memory usage because kopia will load all of the item's data when making the reader. For large items kopia will lazily load data as it goes along

ashmrtn (commented):
no blocking issues remain for me. Just cleanup of existing comments where needed. Not approving since I'm not sure what the back and forth will be for some stuff like error handling

src/internal/m365/onedrive/export.go
```diff
@@ -0,0 +1,169 @@
...
+
+// exportCollection is the implementation of export.ExportCollection for OneDrive
+type exportCollection struct {
+	// baseDir contains the path of the collection
+	baseDir string
+
+	// backingCollections will, in most cases contain just one
+	// collection. However, in cases where we have to combine multiple
+	// collections, like when generating pst files for outlook, this
+	// will contain multiple collections.
```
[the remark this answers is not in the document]
meain: Hmm, now that you point it out, we can just have a separate one for exchange and there is not need to have backing collections as a list for onedrive.

src/pkg/export/export.go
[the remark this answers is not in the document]
meain: Hmm, I was kinda on the fence about it. I've renamed `GetBasePath` to `BasePath` and `GetItems` to `Items`.

src/internal/m365/onedrive/export.go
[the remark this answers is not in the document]
meain: IIUC unbuffered channels does not keep any items in the channel and so we will be block and won't create any items unless the user reads from it. That said, I can add in a lazy reader if we should be concerned about user read the data and not using it.

src/internal/m365/onedrive/export.go
meain: It does, but the usecase is a bit different. We can maybe combine both though. I had left a not in there for now.

src/internal/m365/onedrive/export.go
meain: Ahh, yup. I should have done that from the start. I was a bit reluctant to introduce something in the `fault` package earlier for some reason.

src/internal/m365/onedrive/export.go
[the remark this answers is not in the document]
ryanfkeepers: oh, doh, you're right.  unbuffered is capacity 1, my bad.

src/internal/m365/onedrive/export.go
ryanfkeepers: no worries!

src/pkg/fault/fault.go
```diff
@@ -208,6 +208,32 @@ func (e *Bus) Errors() *Errors {
 	}
 }
 
+// ErrorEntries returns the items that failed along with other
+// recoverable errors
+func (e *Bus) ErrorEntries() ([]Item, []error) {
```
ryanfkeepers, line 213: Nit: this name is a little too generic.  From a certain perspective, all errors in the fault are entries.
```suggestion
func (e *Bus) ItemsAndRecovered() ([]Item, []error) {
```

ryanfkeepers (approved):
looks good, thank you for the changes!

PR #3820, opened 2023-07-17
Operations level changes for OneDrive exports

meain:
This mostly mimics what we have for restore, just modified to call the
export consumer instead of the restore consumer.

Prev: https://github.com/alcionai/corso/pull/3819
Next: https://github.com/alcionai/corso/pull/3821

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/pull/3797
* https://github.com/alcionai/corso/issues/3670

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/export.go
+++ b/src/internal/operations/export.go
@@ -1,17 +1,15 @@
 package operations
 
 import (
-	"archive/zip"
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"path"
 	"time"
 
 	"github.com/alcionai/clues"
 	"github.com/google/uuid"
 
+	"github.com/alcionai/corso/src/internal/archive"
 	"github.com/alcionai/corso/src/internal/common/crash"
 	"github.com/alcionai/corso/src/internal/common/dttm"
 	"github.com/alcionai/corso/src/internal/data"
@@ -198,85 +196,6 @@ func (op *ExportOperation) Run(ctx context.Context) (
 	return expCollections, nil
 }
 
-type zipExportCol struct {
-	reader io.ReadCloser
-}
-
-func (z zipExportCol) BasePath() string {
-	return ""
-}
-
-func (z zipExportCol) Items(ctx context.Context) <-chan export.Item {
-	rc := make(chan export.Item, 1)
-	defer close(rc)
-
-	rc <- export.Item{
-		Data: export.ItemData{
-			Name: "Corso_Export_" + dttm.FormatNow(dttm.HumanReadable) + ".zip",
-			Body: z.reader,
-		},
-	}
-
-	return rc
-}
-
-// zipExportCollection takes a list of export collections and zips
-// them into a single collection.
-func zipExportCollection(
-	ctx context.Context,
-	expCollections []export.Collection,
-) (export.Collection, error) {
-	if len(expCollections) == 0 {
-		return nil, clues.New("no export collections provided")
-	}
-
-	reader, writer := io.Pipe()
-	wr := zip.NewWriter(writer)
-
-	go func() {
-		defer writer.Close()
-		defer wr.Close()
-
-		buf := make([]byte, CopyBufferSize)
-
-		for _, ec := range expCollections {
-			folder := ec.BasePath()
-			items := ec.Items(ctx)
-
-			for item := range items {
-				err := item.Error
-				if err != nil {
-					writer.CloseWithError(clues.Wrap(err, "getting export item").With("id", item.ID))
-					return
-				}
-
-				name := item.Data.Name
-
-				// We assume folder and name to not contain any path separators.
-				// Also, this should always use `/` as this is
-				// created within a zip file and not written to disk.
-				// TODO(meain): Exchange paths might contain a path
-				// separator and will have to have special handling.
-
-				//nolint:forbidigo
-				f, err := wr.Create(path.Join(folder, name))
-				if err != nil {
-					writer.CloseWithError(clues.Wrap(err, "creating zip entry").With("name", name).With("id", item.ID))
-					return
-				}
-
-				_, err = io.CopyBuffer(f, item.Data.Body, buf)
-				if err != nil {
-					writer.CloseWithError(clues.Wrap(err, "writing zip entry").With("name", name).With("id", item.ID))
-					return
-				}
-			}
-		}
-	}()
-
-	return zipExportCol{reader}, nil
-}
-
 func (op *ExportOperation) do(
 	ctx context.Context,
 	opStats *exportStats,
@@ -359,7 +278,7 @@ func (op *ExportOperation) do(
 	logger.Ctx(ctx).Debug(opStats.ctrl)
 
 	if op.ExportCfg.Archive {
-		zc, err := zipExportCollection(ctx, expCollections)
+		zc, err := archive.ZipExportCollection(ctx, expCollections)
 		if err != nil {
 			return nil, clues.Wrap(err, "zipping export collections")
 		}
@@ -405,7 +324,7 @@ func (op *ExportOperation) finalizeMetrics(
 }
 
 // ---------------------------------------------------------------------------
-// Exportr funcs
+// Exporter funcs
 // ---------------------------------------------------------------------------
 
 func exportRestoreCollections(
--- a/src/internal/operations/export.go
+++ b/src/internal/operations/export.go
@@ -33,6 +33,13 @@ import (
 	"github.com/alcionai/corso/src/pkg/store"
 )
 
+const (
+	// CopyBufferSize is the size of the copy buffer for disk
+	// write operations
+	// TODO(meain): tweak this value
+	CopyBufferSize = 5 * 1024 * 1024
+)
+
 // ExportOperation wraps an operation with export-specific props.
 type ExportOperation struct {
 	operation
@@ -230,6 +237,8 @@ func zipExportCollection(
 		defer writer.Close()
 		defer wr.Close()
 
+		buf := make([]byte, CopyBufferSize)
+
 		for _, ec := range expCollections {
 			folder := ec.BasePath()
 			items := ec.Items(ctx)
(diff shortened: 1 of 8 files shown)


src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+			for item := range items {
+				err := item.Error
+				if err != nil {
+					writer.CloseWithError(clues.Wrap(err, "getting export item").With("id", item.ID))
+					return
+				}
+
+				name := item.Data.Name
+
+				f, err := wr.Create(folder + "/" + name)
```
ashmrtn, line 245: hmmm this actually made me think of an interesting issue. Exchange folders can have `/` in them which may cause problems here in the future depending on the output formats we allow. It doesn't need to be addressed in this PR, but do make a comment stating that this assumes the path has no `/` in it except for separators. We may also need a known issues entry when this feature goes out

src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+				if err != nil {
+					writer.CloseWithError(clues.Wrap(err, "creating zip entry").With("name", name).With("id", item.ID))
+					return
+				}
+
+				_, err = io.Copy(f, item.Data.Body)
```
ashmrtn, line 251: we may want to allocate a temp buffer at the start of this function and use `CopyBuffer` instead. The [godoc](https://pkg.go.dev/io#CopyBuffer) for `CopyBuffer` makes it sound like just `Copy` may cause many allocations for buffers. If we go that route always providing a buffer will at least give us some performance guarantees without having to worry about the underlying implementations of the readers/writers (i.e. if they implement other interfaces)

src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+		for _, ec := range expCollections {
+			folder := ec.GetBasePath()
+			items := ec.GetItems(ctx)
+
+			for item := range items {
+				err := item.Error
+				if err != nil {
+					writer.CloseWithError(clues.Wrap(err, "getting export item").With("id", item.ID))
+					return
+				}
```
ashmrtn, line 241: this pattern raises an interesting question about fail fast vs. best effort execution and how we handle that with our layered implementation. Here since we're exiting earlier there's nothing to drain the underlying collection channels. I think we can get away with it for now since Corso will be exiting soon anyway, but we should probably be more careful about it going forward

src/internal/operations/export.go
ashmrtn: also could you please add a comment to this stating it needs to use `/` instead of platform-specific path separators? And we should probably just use `path.Join` from the stdlib package (you'll need to add a linter exception for this; it'll be similar to [this](https://github.com/alcionai/corso/blob/57125f3ea3e35796fd780dbe305c0a50de3dfd69/src/.golangci.yml#L148))

src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+		ctx,
+		op.BackupID,
+		op.store,
+		detailsStore,
+		op.Errors)
+	if err != nil {
+		return nil, clues.Wrap(err, "getting backup and details")
+	}
+
+	observe.Message(ctx, "Restoring", observe.Bullet, clues.Hide(bup.Selector.DiscreteOwner))
```
ashmrtn, line 283: nit: seems like this should be export-specific wording?
```suggestion
	observe.Message(ctx, "Exporting", observe.Bullet, clues.Hide(bup.Selector.DiscreteOwner))
```

src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+		op.ExportCfg,
+		op.Options,
+		dcs,
+		op.Errors)
+	if err != nil {
+		return nil, clues.Wrap(err, "restoring collections")
+	}
+
+	opStats.ctrl = op.ec.Wait()
+
```
ashmrtn, line 341: I'm not completely up to date on all the m365 package stuff. Is this actually adding anything to the implementation? I thought it was to make sure we emptied all the collections the m365 package makes during backup?

src/internal/operations/export.go
```diff
@@ -0,0 +1,422 @@
...
+	return ""
+}
+
+func (z zipExportCol) GetItems(ctx context.Context) <-chan data.ExportItem {
+	rc := make(chan data.ExportItem, 1)
+	defer close(rc)
+
+	rc <- data.ExportItem{
+		Data: data.ExportItemData{
+			Name: "export.zip",
```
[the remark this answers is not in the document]
meain: Let me change the default name to `"Corso_Export_"+dttm.FormatNow(dttm.HumanReadable)+".zip"`. As for making it configurable, this is merely a suggested name and since there will only be one item to restore, they can just choose to ignore the name. Or are you talking about cli users being able to edit the name of the archive being created.

Related: https://github.com/alcionai/corso/pull/3793#discussion_r1262244736 /cc @ryanfkeepers

src/internal/operations/export.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	finalizeErrorHandling(ctx, op.Options, op.Errors, "running export")
+	LogFaultErrors(ctx, op.Errors.Errors(), "running export")
+
+	// -----
+	// Persistence
+	// -----
+
+	err = op.persistResults(ctx, start, &opStats)
```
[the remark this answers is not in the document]
meain: We do the same for restores [here](https://github.com/alcionai/corso/blob/7bc35d3928778f9ec61d1ec247a01f665dd1972a/src/internal/operations/restore.go#L189). In case of exports, we don't really have the data about if the writes succeeded though as that happens after we persist for now. I have not given much thought about moving this to happen after that.

src/internal/operations/export.go
meain: Nothing much besides gathering metrics. I was mostly just mimicking what restore was doing.

src/internal/operations/export.go
meain: Since they are unbuffered channels, I don't think we have to drain them.

src/internal/operations/export.go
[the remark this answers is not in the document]
ryanfkeepers: `persist` in restore is the wrong word, right?  That func is just finalizing metrics.  IIRC, I named it that on the belief that we'd eventually persist restores.  If we're not actually persisting anything here, feel free to rename the func.

src/internal/operations/export.go
```diff
@@ -0,0 +1,440 @@
...
+		op.Errors.Fail(clues.Wrap(err, "persisting export results"))
+		return nil, op.Errors.Failure()
+	}
+
+	logger.Ctx(ctx).Infow("completed export", "results", op.Results)
+
+	return expCollections, nil
+}
+
+type zipExportCol struct {
```
ryanfkeepers, line 201: I think the code will turn out much clearner in the long term if we move anything related to archive processing into its own file, as a way to separate concerns.  So, `/export/zip.go`, in this case. 

Ideally what I'd like to see is for archival to have its own package in internal, completely separate from the export package.  Something like `internal/archive/zip.go` with `package archive`, or some such.  That would more closely follow our current implementations with kopia and m365, as well as what we might imagine from other producers and consumers.  

Separately, this name feels like it's dodging a conflict with the func `zipExportCollection`, and I think it's coming out worse for that (`col` as a suffix, specifically.)  Another benefit of moving this to an internal package and replicating current patterns: instead of having `zip` specific functions and collections: make a plain `Collection` struct in the archive package; let that collection define its methodology / categorization (zip, in this case); centralize processing around standard package behavior while separating per-category specifics into different handlers.

src/internal/operations/export.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+
+	// We don't have data on what all items were written
+	// op.Results.ItemsWritten = opStats.ctrl.Successes
+
+	return op.Errors.Failure()
+}
+
+// ---------------------------------------------------------------------------
+// Exportr funcs
```
ryanfkeepers, line 408:
```suggestion
// Exporter funcs
```

src/internal/operations/export_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+			require.NoError(t, err, "error")
+
+			assert.Empty(t, zc.BasePath(), "base path")
+
+			count := 0
+			for item := range zc.Items(ctx) {
+				assert.Equal(t, "export.zip", item.Data.Name, "name")
+
+				_, err := io.Copy(io.Discard, item.Data.Body)
```
ashmrtn, line 251: how difficult would it be to actually check the items in the generated archive? Right now we don't have any automated tests that ensure the archive content is actually what we expect. Either a test here or something in the sanity tests or otherwise would help put my mind at ease

src/internal/operations/export.go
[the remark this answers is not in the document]
meain: @ryanfkeepers I'm not sure if I fully understand you. By "when given a file", did you mean in the case where we are archiving only a single file?

src/internal/operations/export.go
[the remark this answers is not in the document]
meain: I've renamed it to `finalizeMetrics`.

src/internal/operations/export.go
meain: I've created a separate package under `internal/archive`. It still work on `export.Collection` instead of `data.Collection` as this still runs as the very last step. That said, if we want to change it in future if we decide to as the implementation is not exposed to the consumer.

src/internal/operations/export_test.go
meain: Let me update the test to verify the data within the zip file.

src/internal/operations/export.go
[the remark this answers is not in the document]
ryanfkeepers: as in, depending on the user's input.  `--archive /foo/bar/baz`.  If baz is a directory, archive to `.../baz/standard_name`.  Otherwise assume baz is a non-existing file.

ryanfkeepers (approved):
looking good, thank you!

ashmrtn (approved):
LGTM, thanks for the test expansion

src/internal/operations/export.go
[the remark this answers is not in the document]
meain: Hmm, I feel a bit uneasy about that. I feel it kinda makes the operation a bit unpredictable. Any thoughts @gmatev ?

PR #3822, opened 2023-07-17
CLI for exporting data from OneDrive

meain:
This adds the final piece to enable OneDrive exports. The CLI interface which consumes NewExport interface from Repository and lets the user to export and write the contents in a backup to the local filesystem.

Prev: https://github.com/alcionai/corso/pull/3821
Next: https://github.com/alcionai/corso/pull/3824

---

#### Does this PR need a docs update or release note?

- [x] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Supportability/Tests
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

#### Issue(s)

* https://github.com/alcionai/corso/pull/3797
* https://github.com/alcionai/corso/issues/3670

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #3824, opened 2023-07-17
Export data from SharePoint

meain:
This borrows a lot of the core logic from OneDrive as the internal structure is mostly the same.

Prev: https://github.com/alcionai/corso/pull/3822

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/3823

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/export/export.go
+++ b/src/cli/export/export.go
@@ -6,6 +6,7 @@ import (
 
 var exportCommands = []func(cmd *cobra.Command) *cobra.Command{
 	addOneDriveCommands,
+	addSharePointCommands,
 }
 
 // AddCommands attaches all `corso export * *` commands to the parent.
--- a/src/cli/export/export.go
+++ b/src/cli/export/export.go
@@ -1,0 +1,41 @@
+package export
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var exportCommands = []func(cmd *cobra.Command) *cobra.Command{
+	addOneDriveCommands,
+	// addExchangeCommands,
+	// addSharePointCommands,
+}
+
+// AddCommands attaches all `corso export * *` commands to the parent.
+func AddCommands(cmd *cobra.Command) {
+	exportC := exportCmd()
+	cmd.AddCommand(exportC)
+
+	for _, addExportTo := range exportCommands {
+		addExportTo(exportC)
+	}
+}
+
+const exportCommand = "export"
+
+// The export category of commands.
+// `corso export [<subcommand>] [<flag>...]`
+func exportCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   exportCommand,
+		Short: "Export your service data",
+		Long:  `Export the data stored in one of your M365 services.`,
+		RunE:  handleExportCmd,
+		Args:  cobra.NoArgs,
+	}
+}
+
+// Handler for flat calls to `corso export`.
+// Produces the same output as `corso export --help`.
+func handleExportCmd(cmd *cobra.Command, args []string) error {
+	return cmd.Help()
+}
--- a/src/cli/export/export.go
+++ b/src/cli/export/export.go
@@ -3,9 +3,6 @@ package export
 import (
 	"context"
 	"errors"
-	"io"
-	"os"
-	"path/filepath"
 
 	"github.com/alcionai/clues"
 	"github.com/spf13/cobra"
@@ -72,7 +69,7 @@ func runExport(
 	defer utils.CloseRepo(ctx, r)
 
 	exportLocation := args[0]
-	if exportLocation == "" {
+	if len(exportLocation) == 0 {
 		// This should not be possible, but adding it just in case.
 		exportLocation = control.DefaultRestoreLocation + dttm.FormatNow(dttm.HumanReadableDriveItem)
 	}
@@ -83,8 +80,7 @@ func runExport(
 		ctx,
 		backupID,
 		sel,
-		utils.MakeExportConfig(ctx, ueco),
-	)
+		utils.MakeExportConfig(ctx, ueco))
 	if err != nil {
 		return Only(ctx, clues.Wrap(err, "Failed to initialize "+serviceName+" export"))
 	}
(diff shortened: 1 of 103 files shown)


src/cli/export/sharepoint.go
```diff
@@ -0,0 +1,155 @@
...
+		flags.AddFailFastFlag(c)
+		flags.AddCorsoPassphaseFlags(c)
+		flags.AddAWSCredsFlags(c)
+	}
+
+	return c
+}
+
+const (
+	sharePointServiceCommand          = "sharepoint"
```
ryanfkeepers, line 49: hmm, we should probably move these to utils as consts.  Not in this PR.  But sometime.

src/cli/export/sharepoint.go
```diff
@@ -0,0 +1,155 @@
...
+
+	sel := utils.IncludeSharePointRestoreDataSelectors(ctx, opts)
+	utils.FilterSharePointRestoreInfoSelectors(sel, opts)
+
+	eo, err := r.NewExport(
+		ctx,
+		flags.BackupIDFV,
+		sel.Selector,
+		utils.MakeExportConfig(ctx, opts.ExportCfg),
+	)
```
ryanfkeepers, line 127:
```suggestion
		utils.MakeExportConfig(ctx, opts.ExportCfg))
```

src/cli/export/sharepoint.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			return nil
+		},
+		Example: sharePointServiceCommandExportExamples,
+	}
+}
+
+// processes an sharepoint service export.
+func exportSharePointCmd(cmd *cobra.Command, args []string) error {
```
ryanfkeepers, line 87: Now that we're adding more than one export handler, can you look at making a centralized export runner, much like we have for [restores](https://github.com/alcionai/corso/blob/main/src/cli/restore/restore.go#L90) and backup?

src/cli/export/sharepoint_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+				"--" + flags.AWSAccessKeyFN, testdata.AWSAccessKeyID,
+				"--" + flags.AWSSecretAccessKeyFN, testdata.AWSSecretAccessKey,
+				"--" + flags.AWSSessionTokenFN, testdata.AWSSessionToken,
+
+				"--" + flags.CorsoPassphraseFN, testdata.CorsoPassphrase,
+
+				"--" + flags.ArchiveFN,
+
+				testdata.RestoreDestination,
```
ryanfkeepers, line 84: In line with the other comments about flag/command ordering, can you please move this to line 62?

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+		return Only(ctx, clues.Wrap(err, "Failed to run "+serviceName+" export"))
+	}
+
+	// It would be better to give a progressbar than a spinner, but we
+	// have know way of knowing how many files are available as of now.
+	diskWriteComplete := observe.MessageWithCompletion(ctx, "Writing data to disk")
+	defer func() {
+		diskWriteComplete <- struct{}{}
+		close(diskWriteComplete)
+	}()
```
ryanfkeepers, line 107: I have other fixes like this in a separate PR, but since this is new, might as well fix it here.
```suggestion
	defer close(diskWriteComplete)
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+
+	err = writeExportCollections(ctx, exportLocation, expColl)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
ryanfkeepers, line 114: the conditional is useful if you want to `clues.Stack(err)`.  But here I think we can return directly.
```suggestion
	return writeExportCollections(ctx, exportLocation, expColl)
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+}
+
+func writeExportCollections(
```
ryanfkeepers, line 117: The CLI layer is the wrong owner for this logic, as well as `writeExportItem`.  That should probably move into the `export` package.  That will ensure the functionality 1/ exists independently of the top level api layers,  2/ is accessible to SDK consumers, and 3/ will likely be easier to test in the long term.

Also, naming nit.  That will keep the func name in line with our other naming conventions.
```suggestion
func ConsumeExportCollections(
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+	ctx context.Context,
+	exportLocation string,
+	expColl []export.Collection,
+) error {
+	for _, col := range expColl {
+		folder := filepath.Join(exportLocation, col.BasePath())
```
ryanfkeepers, line 123: it's easy to confuse the `BasePath` here the fully prefixed `FullPath`.  You might want to add a comment that this only contains folder info.

Also, looking at https://github.com/alcionai/corso/blob/main/src/internal/m365/onedrive/export.go#L156, wouldn't the `basePath` contain the folder IDs, not their names?  Does this need to change to `LocationPath` instead of `FullPath().Folders`?

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+
+	return nil
+}
+
+func writeExportCollections(
+	ctx context.Context,
+	exportLocation string,
+	expColl []export.Collection,
+) error {
+	for _, col := range expColl {
```
ryanfkeepers, line 122: Can you please modify this func to accept a fault.Bus and control.Options, and to scrutinize failFast settings?  That will allow us to handle recoverable failures as part of export item creation.

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+		folder := filepath.Join(exportLocation, col.BasePath())
+
+		for item := range col.Items(ctx) {
+			err := item.Error
+			if err != nil {
```
ryanfkeepers, line 127:
```suggestion
			if err := item.Error; err != nil {
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+				return Only(ctx, clues.Wrap(err, "getting item").With("dir_name", folder))
+			}
+
+			err = writeExportItem(ctx, item, folder)
+			if err != nil {
```
ryanfkeepers, line 132:
```suggestion
			if err = writeExportItem(ctx, item, folder); err != nil {
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+		item.Data.Body,
+		observe.ItemExportMsg,
+		clues.Hide(name))
+
+	defer item.Data.Body.Close()
+	defer pclose()
+
+	err := os.MkdirAll(folder, os.ModePerm)
+	if err != nil {
+		return Only(ctx, clues.Wrap(err, "creating directory").With("dir_name", folder))
```
ryanfkeepers, line 157: please add `.WithCtx(ctx)` to all of these errors, since ctx isn't passed into the called func.
```suggestion
		return Only(ctx, clues.Wrap(err, "creating directory").With("dir_name", folder).WithClues(ctx))
```

src/cli/export/export.go
```diff
@@ -37,3 +55,119 @@ func exportCmd() *cobra.Command {
...
+
+	defer utils.CloseRepo(ctx, r)
+
+	exportLocation := args[0]
+	if exportLocation == "" {
+		// This should not be possible, but adding it just in case.
+		exportLocation = control.DefaultRestoreLocation + dttm.FormatNow(dttm.HumanReadableDriveItem)
+	}
+
+	Infof(ctx, "Exporting to folder %s", exportLocation)
```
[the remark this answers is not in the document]
meain: I was mimicking the behaviour for export. I think it would be better to keep them the same.
https://github.com/alcionai/corso/blob/a973f69d182c161d6c30ad3376f43335f20b7c41/src/cli/utils/restore_config.go#L77

src/cli/export/export.go
meain: You are right in that we need `LocationPath`, but that is not available for `RestoreCollections` which is what we have at this stage. That said, we get back names(at least of OneDrive and SharePoint) as the folder names in kopia are the names and not ids.

src/pkg/export/consume.go
```diff
@@ -0,0 +1,80 @@
...
+		for item := range col.Items(ctx) {
+			if item.Error != nil {
+				el.AddRecoverable(ictx, clues.Wrap(item.Error, "getting item").WithClues(ctx))
+			}
+
+			if err := writeItem(ictx, item, folder); err != nil {
+				el.AddRecoverable(
+					ictx,
+					clues.Wrap(err, "writing item").With("file_name", item.Data.Name).WithClues(ctx),
+				)
```
ryanfkeepers, line 40:
```suggestion
					clues.Wrap(err, "writing item").With("file_name", item.Data.Name).WithClues(ctx))
```

ryanfkeepers (approved):

PR #3843, opened 2023-07-18
utilize old drive name when restoring

ryanfkeepers:
utilizes the old drive name in case the drive
was deleted between backup and restore.
Priority is first to use drives whose ids match the backup id; second, to use drives whose names
match the backup drive name; third, to fall
back to a new, arbitrary name.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/onedrive/restore.go
+++ b/src/internal/m365/onedrive/restore.go
@@ -258,7 +258,7 @@ func RestoreCollection(
 		return metrics, clues.Wrap(err, "creating drive path").WithClues(ctx)
 	}
 
-	err = ensureDriveExists(
+	di, err := ensureDriveExists(
 		ctx,
 		rh,
 		caches,
@@ -268,6 +268,12 @@ func RestoreCollection(
 	if err != nil {
 		return metrics, clues.Wrap(err, "ensuring drive exists")
 	}
+
+	// clobber the drivePath details with the details retrieved
+	// in the ensure func, as they might have changed to reflect
+	// a different drive as a restore location.
+	drivePath.DriveID = di.id
+	drivePath.Root = di.rootFolderID
 
 	// Assemble folder hierarchy we're going to restore into (we recreate the folder hierarchy
 	// from the backup under this the restore folder instead of root)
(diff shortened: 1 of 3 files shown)


src/internal/m365/onedrive/restore.go
```diff
@@ -1215,51 +1221,63 @@ func ensureDriveExists(
...
 		newDrive     models.Driveable
 		err          error
 	)
 
-	if _, ok := caches.DriveNameToDriveInfo[newDriveName]; ok {
-		newDriveName = fmt.Sprintf("%s %d", driveName, 1)
+	// if the drive wasn't found by ID, maybe we can find a
+	// drive with the same name but different ID.
+	// start by looking up the old drive's name
+	oldName, ok := caches.BackupDriveIDName.NameOf(driveID)
```
meain, line 1244: IIUC `caches.BackupDriveIDName` will only contain the drives that were available at the time of backup, right? Should we get the latest set of drives before we do this check?

src/internal/m365/onedrive/restore.go
```diff
@@ -1215,51 +1221,63 @@ func ensureDriveExists(
...
 	// item types beyond just drive.  Lists, for example, cannot share
-	// names with document libraries.  In those cases it's not enough
-	// to compare the names of drives; we also need to continue this
-	// loop until we can create a drive without error.
-	for i := 2; ; i++ {
-		ictx := clues.Add(ctx, "new_drive_name", clues.Hide(newDriveName))
+	// names with document libraries (they're the same type, actually).
+	// In those cases we need to rename the drive until we can create
+	// one without a collision.
+	for i := 1; ; i++ {
```
meain, line 1263: I know this code is not new, but should we bound this (even to a big number)?

src/internal/m365/onedrive/restore.go
meain, line 1: Just wanted to make sure we have the behaviour right. IIRC from yesterday's discussion, we had decided to create a library with the original name in case we could not find by id or name but this one creates with a different name.

src/internal/m365/onedrive/restore.go
ryanfkeepers: The latest set of drives are already cached in `caches.DriveIDToDriveInfo`.

src/internal/m365/onedrive/restore.go
ryanfkeepers: I think you're missing the flow a little.  It works like this (referring to line nums):
1231. if the drive still exists by ID, use that, even if the name differs.
1247. if a drive exists with the same name, use that one.  Requires a backup drive cache entry.
1256. if no drive exists by either id or name, make a new one.  Name that drive either the fallback name (1224) or the prior name (1253).

So in order to use the fallback name, we must have: no drives matching the id, no drives matching the name, and no backup cache entry for that ID (which, technically, should only happen if the backup version is older than our storage of drive names).

meain (approved):

Issue #3836, opened 2023-07-18
Discovery APIs for teams


neha-Gupta1:
Expose APIs to get list of teams available.

neha-Gupta1:
Done.

PR #3850, opened 2023-07-19
teams discovery API

neha-Gupta1:
Teams discovery API
- fetch all groups from endpoint and filter teams out of it.
- fetch team with ID. 

Permission added- 
- Application- Team.ReadBasic.All
- Group.Read.All

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* https://github.com/alcionai/corso/issues/3836

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


neha-Gupta1:
Testcases and few error cases still to be handled.

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+			if s == teamService {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// GetID retrieves team by groupID/teamID.
+func (c Groups) GetByID(
```
ryanfkeepers, line 117: On the topic of supporting WebURLs as a teams identifier, we'll need an api for looking up teams by either id or weburl.  That change should be saved for a separate PR, just wanted to make the need clear.

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx context.Context,
+	identifier string,
```
ryanfkeepers, line 119: Can this fetch a team using a team ID, or does it need a group ID that is different from the team id?

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := service.Client().Groups().ByGroupId(identifier).Get(ctx, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "getting group by ID")
+	}
+
+	if err != nil {
```
ryanfkeepers, line 131: duplicate conditionals

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+		err := graph.Wrap(ctx, err, "getting teams by id")
```
ryanfkeepers, line 132: You're not necessarily getting teams here (unless you change the func, as suggested above).
```suggestion
		err := graph.Wrap(ctx, err, "getting groups by id")
```

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+
+		// TODO: check if its applicable here
+		if graph.IsErrItemNotFound(err) {
+			err = clues.Stack(graph.ErrResourceOwnerNotFound, err)
+		}
```
ryanfkeepers, line 137: Let's skip the addition of any code that we're not certain will be used.

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+
+		return nil, err
+	}
+
+	return resp, err
```
ryanfkeepers, line 142: Returns should be one of:
```go
return resp, nil
return nil, resp
return resp, graph.Stack(ctx, err).OrNil()
```

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ValidateGroup ensures the item is a Groupable, and contains the necessary
+// identifiers that we handle with all groups.
+func ValidateGroup(item models.Groupable) error {
+	if item.GetId() == nil {
+		return clues.New("missing ID")
+	}
+
+	if item.GetDisplayName() == nil {
+		return clues.New("missing principalName")
```
ryanfkeepers, line 157:
```suggestion
		return clues.New("missing display name")
```

src/pkg/services/m365/m365.go
ryanfkeepers, line 1: Please move this file to a separate PR.  It'll keep each review smaller, and give more space to focus on the addition of testing as necessary.

src/pkg/services/m365/api/team_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type TeamsUnitSuite struct {
+	tester.Suite
+}
+
+func TestTeamsUnitSuite(t *testing.T) {
+	suite.Run(t, &TeamsUnitSuite{Suite: tester.NewUnitSuite(t)})
+}
+
```
ryanfkeepers, line 22: this isn't getting used.

src/pkg/services/m365/api/team_test.go
```diff
@@ -0,0 +1,55 @@
...
+	t := suite.T()
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	teams, err := suite.its.ac.
+		Groups().
+		GetAll(ctx, true, fault.New(true))
+	require.NoError(t, err)
+	require.NotZero(t, len(teams), "must have at least one team")
```
ryanfkeepers, line 50: This is a weak test that doesn't validate much for us beyond our ability to call the func in the first place.  I'd really like to see the default test team set up in the tester package as part of this PR, and for the tests here to ensure that team can be found when we call `GetAll`, and also when we call `GetByID`.

src/pkg/services/m365/api/team_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestTeamsIntgSuite(t *testing.T) {
+	suite.Run(t, &TeamsIntgSuite{
+		Suite: tester.NewIntegrationSuite(
+			t,
+			[][]string{tconfig.M365AcctCredEnvs}),
+	})
+}
+
+func (suite *TeamsIntgSuite) SetupSuite() {
+	suite.its = newIntegrationTesterSetup(suite.T())
```
ryanfkeepers, line 37: integrationTesterSetup should have teams (and groups?) details added to it, so that the data is available for all tests as needed.

ryanfkeepers (changes requested):
Generally looks good, though some minor refactoring is wanted.  The two blocking changes are:

1. the PR should be split into two PRs.  One for the api file.  One for api/teams.go, one for the addition of teams to services/m365.
2. Testing setup needs to be complete and the api tests for groups needs to be expanded.

src/pkg/services/m365/api/teams.go
neha-Gupta1: Done

src/pkg/services/m365/api/teams.go
neha-Gupta1: With my understanding/investigation - value of groupID created for the team is same as TeamID.

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+// ---------------------------------------------------------------------------
+
+func (c Client) Groups() Groups {
+	return Groups{c}
+}
+
+// On creation of each Microsoft Teams a corrsponding group gets created from them.
+// Most of the information like events, drive and mail info will be fetched directly
+// from groups. So we pull in group and process only the once which are associated with
+// a team for further proccessing of teams.
```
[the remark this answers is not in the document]
neha-Gupta1: I have updated the comments.

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/alcionai/clues"
+	"github.com/alcionai/corso/src/internal/common/str"
+	"github.com/alcionai/corso/src/internal/common/tform"
+	"github.com/alcionai/corso/src/internal/m365/graph"
+	"github.com/alcionai/corso/src/pkg/fault"
+	msgraphgocore "github.com/microsoftgraph/msgraph-sdk-go-core"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+)
+
+const teamService = "Team"
```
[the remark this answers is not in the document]
neha-Gupta1: `Team` is the value endpoint provides. So will keep it same. Updated the const name.

src/pkg/services/m365/m365.go
neha-Gupta1: Moved to - https://github.com/alcionai/corso/pull/3864

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+
+	return getGroups(ctx, filterTeams, errs, service)
+}
+
+// GetAll retrieves all groups.
+func getGroups(
+	ctx context.Context,
+	filterTeams bool,
```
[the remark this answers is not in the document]
neha-Gupta1: Exposed two funcs now- 
- func (c Groups) GetAllTeams() - gets all the groups which have a corresponding team with them.
- func (c Groups) GetAll() - gets all the groups in org

src/pkg/services/m365/api/teams.go
neha-Gupta1: Will handle it in -- https://github.com/alcionai/corso/pull/3864

src/pkg/services/m365/m365.go
ryanfkeepers: thank you!

src/pkg/services/m365/api/teams.go
ryanfkeepers: Sounds good.  In that case, we'll want this func to accept any variation of identifier we use for the team.  For example: sites can provide either the site id or the webURL, users can provide either user id or principal name.  Given that not all groups are teams, we might want to make a `GetTeamByID` func instead.

Though, now that I think about it, maybe betwen this and the GetAllTeams suggestion, this is a good time to swap this client from `Groups` to `Teams`, and have the teams api call groups under the hood?  It'll probably be clearer for us to make calls like `ac.Teams().GetByID()` and `ac.Teams().GetGroupByID()` than it would be to have lots of `ac.Groups().GetFooByID()` variations.

src/pkg/services/m365/api/teams.go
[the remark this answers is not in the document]
ryanfkeepers: Ah, good catch, thanks.

src/pkg/services/m365/api/teams.go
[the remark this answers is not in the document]
ryanfkeepers: See my other comment: I think we'll want two clients.  One for teams, and one for groups, and the teams client can call groups api under the hood.  For this PR, since we don't have a need for generalized groups yet, can you just shift the api client naming to teams?  Thanks for handling all the-back-and-forth.

src/pkg/services/m365/api/teams.go
ryanfkeepers: hmm, I think this PR is the correct place to handle it.  Or, at least, the api layer should own that handling, not the discovery layer.  This change can move to a follow-up PR, though we'll need a TODO for it.

src/pkg/services/m365/api/teams.go
neha-Gupta1: Done.

src/pkg/services/m365/api/teams.go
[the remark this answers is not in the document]
neha-Gupta1: As of now I have added the Teams client with uses group api under the hood.

src/pkg/services/m365/api/team_test.go
neha-Gupta1: Done

src/pkg/services/m365/api/team_test.go
neha-Gupta1: Updated it.

src/pkg/services/m365/api/team_test.go
neha-Gupta1: Added helper functions.

src/pkg/services/m365/api/groups.go
```diff
@@ -0,0 +1,165 @@
...
+func (c Teams) GetAll(
+	ctx context.Context,
+	errs *fault.Bus,
+) ([]models.Groupable, error) {
+	service, err := c.Service()
+	if err != nil {
+		return nil, err
+	}
+
+	return getGroups(ctx, true, errs, service)
```
ryanfkeepers, line 49: I'm  playing around with graph api and it looks like the call to `/teams` works just fine as an enumerator, instead of calling `/groups` and filtering down to teams.  

On that note, this file is still conflicted about its identity.  It's called `Groups`, but has a client named `Teams`.  The teams client has a `GetAll` query that calls a `Groups` endpoint.  This all compiles and works out fine, but it's going to be a mess for us to maintain in the long term.  Lets keep it simple for now and have something like this:

```
// in /api/teams.go

func (c Client) Teams() Teams

func (c Teams) GetAll(ctx, errs)
    // should call `/teams`
    
func (c Teams) GetByID(ctx, id, errs)
    // should call `/teams/<id>`
    
func ValidateTeam(...)
```

Simple as that.  Ditch anything groups related until we have clear reason to need it.

neha-Gupta1: Yes. That was my first try. But I could not find the GET method for teams in the sdk. Here is the ticket for same - https://github.com/microsoftgraph/msgraph-sdk-go/issues/539. 
Till then I am try to use the raw API.

neha-Gupta1: With latest msgraph SDK it works fine. Updated the code.

neha-Gupta1: Will require https://github.com/alcionai/corso/pull/3879 to be merged.

src/internal/tester/tconfig/config.go
```diff
@@ -38,6 +40,8 @@ const (
 const (
 	EnvCorsoM365TestSiteID           = "CORSO_M365_TEST_SITE_ID"
 	EnvCorsoM365TestSiteURL          = "CORSO_M365_TEST_SITE_URL"
+	EnvCorsoM365TestTeamID           = "CORSO_M365_TEST_TEAM_ID"
+	EnvCorsoM365TestGroupID          = "CORSO_M365_TEST_Group_ID"
```
ryanfkeepers, line 44:
```suggestion
	EnvCorsoM365TestGroupID          = "CORSO_M365_TEST_GROUP_ID"
```

src/internal/tester/tconfig/config.go
```diff
@@ -148,6 +152,18 @@ func ReadTestConfig() (map[string]string, error) {
 		os.Getenv(EnvCorsoM365TestSiteID),
 		vpr.GetString(TestCfgSiteID),
 		"10rqc2.sharepoint.com,4892edf5-2ebf-46be-a6e5-a40b2cbf1c1a,38ab6d06-fc82-4417-af93-22d8733c22be")
+	fallbackTo(
+		testEnv,
+		TestCfgTeamID,
+		os.Getenv(EnvCorsoM365TestTeamID),
+		vpr.GetString(TestCfgTeamID),
+		"d288d6bc-4595-4ff5-87a1-6e7fd750aa42")
```
ryanfkeepers, line 160: any particular reason to have a separate group and team?

Either way, for the team, can we please use `6f24b40d-b13d-4752-980f-f5fb9fba7aa0` (CorsoCITeam).

src/internal/tester/tconfig/protected_resources.go
```diff
@@ -209,3 +209,29 @@ func UnlicensedM365UserID(t *testing.T) string {
...
+	cfg, err := ReadTestConfig()
+	require.NoError(t, err, "retrieving m365 team id from test configuration: %+v", clues.ToCore(err))
+
+	return strings.ToLower(cfg[TestCfgTeamID])
+}
+
+// Groups
+
+// M365GroupID returns a groupID string representing the m365GroupID described
+// by either the env var CORSO_M365_TEST_Group_ID, the corso_test.toml config
```
ryanfkeepers, line 229:
```suggestion
// by either the env var CORSO_M365_TEST_GROUP_ID, the corso_test.toml config
```

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,165 @@
...
+// The group acts as the protected resource, and all teams data like events,
+// drive and mail messages are owned by that group.
+
+// Teams is an interface-compliant provider of the client.
+type Teams struct {
+	Client
+}
+
+// GetAllTeams retrieves all groups.
+func (c Teams) GetAll(
```
ryanfkeepers, line 40: Apologies again on the design thrashing, but per our sync conversations, can you please change this package, file, and client backup to `Groups` instead of `Teams`?  That seems like it'll be a more appropriate starting point for the work we plan to implement.  If we move from there to needing a teams specific api, we can introduce another client for that.  Once the client naming is squared away, this is good to go.
```suggestion
func (c Groups) GetAll(
```

src/pkg/services/m365/api/teams.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+
+	return getGroups(ctx, true, errs, service)
+}
+
+// GetAll retrieves all groups.
+func getGroups(
+	ctx context.Context,
+	getOnlyTeams bool,
```
ryanfkeepers, line 55: As mentioned in the other comment, let's ditch this for now and get all groups instead.

That said, not all groups have access to conversations (and, I assume, groups without access to conversations also don't have access to a calendar).  Instead of packing that filter into the pagination, how about a separate func that can accept a `[]models.Groupable` as a parameter and filter it down to only specific groups, such as only those which can access conversations?  This change can be made in a follow-up PR.

src/pkg/services/m365/api/teams.go
neha-Gupta1: done

src/pkg/services/m365/api/teams.go
neha-Gupta1: Done. Currently we have only one such func implemented for teams. But it can be extended in future.

src/pkg/services/m365/api/groups.go
```diff
@@ -46,13 +46,26 @@ func (c Teams) GetAll(
...
+	}
+
+	return getGroups(ctx, FetchOnlyTeams, errs, service)
 }
 
 // GetAll retrieves all groups.
 func getGroups(
 	ctx context.Context,
-	getOnlyTeams bool,
+	filterGroupsData func(ctx context.Context, g models.Groupable) bool,
```
ryanfkeepers, line 68: Hmm, this isn't my preferred implementation.  Ad-hoc funcs for filtering are genrally a more complicated implementation of a standalone transformer func.  I think a more simple implementation is to split this out to the following funcs.  These are each more simple to grok, more simple to test, and can be exported if we need their behavior elsewhere.

```go
// gets all groups, does no filtering

func (c Client) getGroups(ctx, errs) ([]models.Groupable, error) {...}

// callers can pass the results of getGroups() here.

func removeTeams([]models.Groupable) []models.Groupable {...}

func removeNonTeams([]models.Groupable) []models.Groupable {...}
```

ryanfkeepers (approved):
one last design nit.  approving to unblock.

src/pkg/services/m365/api/groups.go
neha-Gupta1: Got it. Something similar to what I implemented initially.

src/pkg/services/m365/api/groups.go
```diff
@@ -111,25 +113,28 @@ func getGroups(
...
-				log.Debug("could not be converted to string value: ", ResourceProvisioningOptions)
-				return false
-			}
-
-			if s == teamsAdditionalDataLabel {
-				return true
+	var teams []models.Groupable
+
+	for _, g := range groups {
+		if g.GetAdditionalData()[ResourceProvisioningOptions] != nil {
```
ryanfkeepers, line 121:
```suggestion
		if g.GetAdditionalData()[ResourceProvisioningOptions] != nil {
		  continue
		}
		// the rest
```

src/pkg/services/m365/api/groups.go
```diff
@@ -111,25 +113,28 @@ func getGroups(
...
+			for _, v := range val {
+				s, err := str.AnyToString(v)
+				if err != nil {
+					log.Debug("could not be converted to string value: ", ResourceProvisioningOptions)
+					continue
+				}
+
+				if s == teamsAdditionalDataLabel {
+					teams = append(teams, g)
+				}
```
ryanfkeepers, line 132: I forget, do we already have this in multiple spots?  If so, can it be moved into a helper func?

src/pkg/services/m365/api/groups_test.go
```diff
@@ -116,15 +116,25 @@ func (suite *GroupsIntgSuite) TestGetAllTeams() {
 	ctx, flush := tester.NewContext(t)
 	defer flush()
 
-	groups, err := suite.its.ac.
+	teams, err := suite.its.ac.
 		Groups().
 		GetTeams(ctx, fault.New(true))
 	require.NoError(t, err)
+	require.NotZero(t, len(teams), "must have at least one group")
```
ryanfkeepers, line 123:
```suggestion
	require.NotZero(t, len(teams), "must have at least one team")
```

src/pkg/services/m365/api/groups_test.go
```diff
@@ -116,15 +116,25 @@ func (suite *GroupsIntgSuite) TestGetAllTeams() {
...
 
-	for _, team := range groups {
-		assert.True(t, api.FetchOnlyTeams(ctx, team), "must not return non groups groups")
+	var isTeam bool
+
+	if len(groups) > len(teams) {
+		isTeam = true
 	}
+
+	assert.True(t, isTeam, "must not return non groups ")
```
ryanfkeepers:
```suggestion
	assert.True(t, len(groups) > len(teams), "must only return teams")
```

src/pkg/services/m365/api/groups.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if err := iter.Iterate(ctx, iterator); err != nil {
+		return nil, graph.Wrap(ctx, err, "iterating all groups")
+	}
+
+	return groups, el.Failure()
+}
+
+func FetchOnlyTeams(ctx context.Context, groups []models.Groupable) []models.Groupable {
```
ryanfkeepers, line 116:
```suggestion
func OnlyTeams(ctx context.Context, groups []models.Groupable) []models.Groupable {
```

src/pkg/services/m365/api/groups.go
neha-Gupta1: I am little confused here. We want to continue is there is some value in `g.GetAdditionalData()[ResourceProvisioningOptions]` ?

src/pkg/services/m365/api/groups.go
neha-Gupta1: its used only here and not at multiple places.

PR #3853, opened 2023-07-19
look up restore resource if specified

ryanfkeepers:
If the restore configuration specifies a protected resource as a restore target, use that as the destination for the restore.  First step is to ensure the provided target can be retrieved and identified.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test


src/internal/operations/inject/inject.go
```diff
@@ -59,6 +60,25 @@ type (
 		CacheItemInfo(v details.ItemInfo)
 	}
 
+	PopulateProtectedResourceIDAndNamer interface {
+		// PopulateProtectedResourceIDAndName takes the provided owner identifier and produces
+		// the owner's name and ID from that value.  Returns an error if the owner is
+		// not recognized by the current tenant.
+		//
+		// The id-name swapper should be optional.  Some processes will look up all owners in
```
pandeyabs, line 68: nit: Did you mean to say `idname.Cacher should be optional`?

src/internal/operations/restore.go
```diff
@@ -321,6 +332,24 @@ func (op *RestoreOperation) persistResults(
...
+	if len(restoreCfg.ProtectedResource) == 0 {
+		return orig, nil
+	}
+
+	id, name, err := pprian.PopulateProtectedResourceIDAndName(
+		ctx,
+		restoreCfg.ProtectedResource,
+		nil)
+
+	return idname.NewProvider(id, name), clues.Stack(err).OrNil()
```
pandeyabs, line 350: One concern with `OrNil()` usage here is that returned params(`idname.Provider`) may get populated unintentionally. This can cause problems in 
* callee ( `idname.NewProvider`) if callee modifies state or is passed nil ptrs, and/or
* caller of `chooseRestoreResource` if it doesn't do a proper error check

I think `if err != nil` fits better here, let me know what you think.

pandeyabs (commented):
Generally looks good to me! Please see review comments. Once those are addressed, I can approve.

src/internal/operations/inject/inject.go
ryanfkeepers: fixed, thank you

src/internal/operations/restore.go
ryanfkeepers: As an abstract pattern, your point applies to any and all appearances of `OrNil()`.  The concern revolves around what happens if, like you said, the caller `doesn't do a proper error check`.   This is why much of boilerplate golang usage involves returning either a value or an error, but not both.

It's not, I think, a concern worth putting effort into worrying over.  Fundamentally, we always have a flag to tell us whether the value is safe to use: whether or not the error is nil.  When not nil, the caller shouldn't use the provider.  It's a simple procedure to follow, and leaves very little room for footguns or gotchas.

pandeyabs: > As an abstract pattern, your point applies to any and all appearances of OrNil()

Not always true. [This usage](https://github.com/alcionai/corso/blob/d2dda001950ad4211b2e0d4e2fb097be72e223b6/src/internal/kopia/conn.go#L447) for e.g. is fine.

> The concern revolves around what happens if, like you said, the caller doesn't do a proper error check

This is just one part of my concern. The other part is that we are calling `idname.NewProvider` which is a throwaway. It's fine in this particular case, but can be problematic if the callee modifies state or uses uninitialized args.

ryanfkeepers: > Not always true. [This usage](https://github.com/alcionai/corso/blob/d2dda001950ad4211b2e0d4e2fb097be72e223b6/src/internal/kopia/conn.go#L447) for e.g. is fine.

Hah, okay, you got me.  Any usage involving the return of multiple values.

> It's fine in this particular case,

If you're concerned about patterns around the presence or usage of `OrNil()`, that's totally fine.  But by this admission, it's not a concern that needs to block the current PR from approval.  We do have a system in place for the identifying and settling code pattern concerns, which we can use for this topic.  In the meantime, are you sufficiently confident in the safety of the situation to disagree and move forward?

pandeyabs: Sounds good

pandeyabs (approved):

PR #3858, opened 2023-07-20
add restore to alternate resource

ryanfkeepers:
adds support for restoring to a resource that
differs from the one whose data appears in the backup.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/exchange/restore.go
+++ b/src/internal/m365/exchange/restore.go
@@ -40,15 +40,12 @@ func ConsumeRestoreCollections(
 	}
 
 	var (
-		userID         = dcs[0].FullPath().ResourceOwner()
+		userID         = rcc.ProtectedResource.ID()
 		directoryCache = make(map[path.CategoryType]graph.ContainerResolver)
 		handlers       = restoreHandlers(ac)
 		metrics        support.CollectionMetrics
 		el             = errs.Local()
 	)
-
-	// FIXME: should be user name
-	ctx = clues.Add(ctx, "resource_owner", clues.Hide(userID))
 
 	for _, dc := range dcs {
 		if el.Failure() != nil {
--- a/src/internal/m365/exchange/restore.go
+++ b/src/internal/m365/exchange/restore.go
@@ -40,7 +40,7 @@ func ConsumeRestoreCollections(
 	}
 
 	var (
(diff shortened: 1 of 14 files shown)


src/internal/m365/onedrive/restore_caches.go
```diff
@@ -0,0 +1,116 @@
+package onedrive
```
ryanfkeepers, line 1: This file is purely lift-and-shift. it was moved from `restore.go` since it has grown from a small container of maps to a struct with constructors, populators, and other methods.  No changes to this code have been made in this PR outside of the movement.

src/internal/m365/exchange/restore.go
```diff
@@ -40,16 +40,13 @@ func ConsumeRestoreCollections(
...
 	var (
-		userID         = dcs[0].FullPath().ResourceOwner()
+		userID         = rcc.ProtectedResource.ID()
 		directoryCache = make(map[path.CategoryType]graph.ContainerResolver)
 		handlers       = restoreHandlers(ac)
 		metrics        support.CollectionMetrics
 		el             = errs.Local()
 	)
 
-	// FIXME: should be user name
```
pandeyabs, line 50: Curious, why was this removed?

src/internal/m365/exchange/restore.go
```diff
@@ -40,16 +40,13 @@ func ConsumeRestoreCollections(
 	}
 
 	var (
-		userID         = dcs[0].FullPath().ResourceOwner()
+		userID         = rcc.ProtectedResource.ID()
```
pandeyabs, line 43: nit: Call it `protectedResourceID` instead of `userID`?

pandeyabs (approved):

src/internal/m365/exchange/restore.go
ryanfkeepers: It's a duplicate, and is already present in clues from a parent caller.

src/internal/m365/exchange/restore.go
ryanfkeepers: I'll go with `resourceID` for a middle ground.

PR #3865, opened 2023-07-20
Retention update op

ashmrtn:
Create new operation type to update retention parameters
for immutable backups

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/operations/retention_config.go
```diff
@@ -0,0 +1,77 @@
...
+func (op *RetentionConfigOperation) Run(ctx context.Context) (err error) {
+	defer func() {
+		if crErr := crash.Recovery(ctx, recover(), "retention_config"); crErr != nil {
+			err = crErr
+		}
+	}()
+
+	op.Results.StartedAt = time.Now()
+
+	// TODO(ashmrtn): Send telemetry?
```
meain, line 58: I'm guessing this is for a follow up PR.

src/internal/operations/retention_config.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/control/repository"
+	"github.com/alcionai/corso/src/pkg/count"
+)
+
+// RetentionConfigOperation wraps an operation with restore-specific props.
+type RetentionConfigOperation struct {
+	operation
+	Results RetentionConfigResults
+	rcOpts  repository.Retention
```
meain, line 22: Should we add a level of indirection here? You might be a better judge of the decision, but just wanted to float the idea out there.

meain (approved):
Had a thought about the `RetentionConfigOperation` struct, but LGTM.

src/internal/operations/retention_config.go
ashmrtn: I think we'll be alright (famous last works 😂). The struct is meant to just carry data without functionality. It will also allow us to add more fields if needed without changing the API

src/internal/operations/retention_config.go
ashmrtn: let's hold off for the moment. I'm not sure if we want metrics on this, but if we do adding it isn't too big a lift

PR #3866, opened 2023-07-20
add cli flag for to-resource

ryanfkeepers:
adds a cli flag for restoring to a different resource (mailbox, user, site) from the one stored in the
backup.

---

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3562

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] (beta)
 
 ### Added
-- Restore commands now accept an optional resource override with the `--to-resource` flag.  This allows restores to recreate backup data witthin different mailboxes, sites, and users.  
+- Restore commands now accept an optional resource override with the `--to-resource` flag.  This allows restores to recreate backup data within different mailboxes, sites, and users.  
 
 ### Fixed
 - SharePoint document libraries deleted after the last backup can now be restored.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] (beta)
 
+### Added
+- Restore commands now accept an optional resource override with the `--to-resource` flag.  This allows restores to recreate backup data witthin different mailboxes, sites, and users.  
+
 ### Fixed
 - SharePoint document libraries deleted after the last backup can now be restored.
+- Restore requires the protected resource to have access to the service being restored.
 
 ## [v0.11.1] (beta) - 2023-07-20
 
--- a/src/cli/flags/restore_config.go
+++ b/src/cli/flags/restore_config.go
@@ -30,5 +30,5 @@ func AddRestoreConfigFlags(cmd *cobra.Command) {
 		"Overrides the folder where items get restored; '/' places items into their original location")
 	fs.StringVar(
 		&ToResourceFV, ToResourceFN, "",
-		"Overrides the protected resource (mailbox, site, etc) where data gets restored")
+		"Overrides the protected resource (mailbox, site, user, etc) where data gets restored")
 }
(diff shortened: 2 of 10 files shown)


ryanfkeepers:
@gmatev added you to the review for checks on flag names and documentation.

website/docs/setup/restore-options.md
```diff
@@ -108,3 +108,18 @@ the copy of`reports.txt` is named `reports 1.txt`.
...
 version. If multiple existing items collide with the backup item, only one of the
 existing items is replaced.
+
+## To resource
+
+The `--to-resource` flag lets you select which resource will receive the restored data.
+A resource can be a mailbox, user, or sharepoint site.  
+
+<CodeBlock language="bash">{
+    `corso restore onedrive --backup abcd --to-resource adelev@alcion.ai`
```
pandeyabs, line 118: Suggest changing `abcd` to a uuid format instead

CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] (beta)
 
+### Added
+- Restore commands now accept an optional resource override with the `--to-resource` flag.  This allows restores to recreate backup data witthin different mailboxes, sites, and users.
```
pandeyabs, line 11:
```suggestion
- Restore commands now accept an optional resource override with the `--to-resource` flag.  This allows restores to recreate backup data within different mailboxes, sites, and users.  
```

src/cli/flags/restore_config.go
```diff
@@ -25,5 +27,8 @@ func AddRestoreConfigFlags(cmd *cobra.Command) {
 		"Sets the behavior for existing item collisions: "+string(control.Skip)+", "+string(control.Copy)+", or "+string(control.Replace))
 	fs.StringVar(
 		&DestinationFV, DestinationFN, "",
-		"Overrides the destination where items get restored; '/' places items into their original location")
+		"Overrides the folder where items get restored; '/' places items into their original location")
+	fs.StringVar(
+		&ToResourceFV, ToResourceFN, "",
+		"Overrides the protected resource (mailbox, site, etc) where data gets restored")
```
pandeyabs, line 33:
```suggestion
		"Overrides the protected resource (mailbox, site, user, etc) where data gets restored")
```

pandeyabs (approved):
Looks good! Left some nits

website/docs/setup/restore-options.md
ryanfkeepers: Going to keep this, as it currently matches our other examples.

src/cli/flags/restore_config.go
ryanfkeepers: Isn't that covered in `etc`? 😆

pandeyabs: IMO, it's important enough to not be in `etc`

website/docs/setup/restore-options.md
gmatev: The structure of this page is a little off beyond just this flag.
* The Item collision handling sub options probably should move to H3 `###` under `## Item collision handling`
* The heading `To resource` while describing the flag correctly is probably not descriptive enough for a heading. Better option is probably `Restore resource target`. For consistency you could rename `Destination` above to `Restore folder target` or `Restore folder destination`
* Probably need to cover the interactions of `to-resource` with `destination`, if any. Assuming that the example restores to an alternate resource in a folder `Corso_Restore_<>` this should probably be mentioned for clarity. If we support combing the parameters, an extra example can be useful as well

Not sure, if related to the functionality here, but some of the behavior around restoring to a resource that may have been deleted at some point for `original location` should be covered somewhere in the docs.

gmatev (approved):
Ideally reorganize the restore options page a bit based on the feedback. Otherwise LGTM

website/docs/setup/restore-options.md
ryanfkeepers: fixed the suggested comments.

> some of the behavior around restoring to a resource that may have been deleted at some point for original location should be covered somewhere in the docs

There's the `limitation` at the bottom which states we won't restore to a missing or deleted resource.  Is that sufficient?

Issue #2360, opened 2023-02-01
Flag to disable kopia-assisted incrementals


ashmrtn:
Occasionally we just need a really big hammer. A good hammer to have would be to disable kopia assisted incrementals so we can get back to a known good state. This does not necessarily have to disable delta token-based incremtentals though we should hash out exactly what semantics we want

PR #3869, opened 2023-07-20
Have kopia package generate tags during backup based on Reasons

ashmrtn:
Move tag generation from the backup op to the
kopia package. This makes it match the pattern
that finding base backups uses where a set of
Reasons and a separate set of additional tags
are provided

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2360

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/kopia/inject/inject.go
```diff
@@ -15,6 +15,7 @@ type (
 	BackupConsumer interface {
 		ConsumeBackupCollections(
 			ctx context.Context,
+			backupReasons []kopia.Reason,
```
ryanfkeepers, line 18: Is the plan still to move the `Reasoner` interface out of kopia?  I think relocation will be needed to avoid circular dependencies, right?

src/internal/kopia/wrapper.go
```diff
@@ -190,6 +192,19 @@ func (w Wrapper) ConsumeBackupCollections(
 		return nil, nil, nil, clues.Wrap(err, "building kopia directories")
 	}
 
+	// Add some extra tags so we can look things up by reason.
```
ryanfkeepers, line 195: 🎉

ryanfkeepers (approved):

pandeyabs (approved):

src/internal/kopia/inject/inject.go
ashmrtn: I think we can avoid circular dependencies for the moment even if `Reasoner` stays in the kopia package

src/internal/kopia/wrapper.go
ashmrtn: technically this doesn't add anything extra from what we had previously. It just ensures things keep working but is commented a bit better

PR #3875, opened 2023-07-21
allow users to limit page size

ryanfkeepers:
allows cli users to limit the page size of delta queries by calling a new hidden flag: --delta-page-size.
This also adds the control.Options struct to the api client, so that configurations such as this can be easily handed into, and used by, the client.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn (approved):
do we need a ticket to have this be a dynamically determined value in the future?

ryanfkeepers:
> do we need a ticket to have this be a dynamically determined value in the future?

Good question.  Let's see if it solves the issue, first.

PR #3884, opened 2023-07-21
centralize graph api panic and conn retry

ryanfkeepers:
Adds a middleware to the graph api adapter which
watches all http request responses for both panics and connection reset issues.  Panics are
transformed into standard errors.  Connection
resets get retried up to three times.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test


src/internal/m365/graph/service.go
```diff
@@ -275,3 +288,70 @@ func kiotaMiddlewares(
...
+})
+
+func (aw *adapterWrap) Send(
+	ctx context.Context,
+	requestInfo *abstractions.RequestInformation,
+	constructor serialization.ParsableFactory,
+	errorMappings abstractions.ErrorMappings,
+) (sp serialization.Parsable, err error) {
+	defer func() {
+		if crErr := crash.Recovery(ctx, recover(), "graph adapter request"); crErr != nil {
```
pandeyabs, line 325: Awesome!

src/internal/m365/graph/http_wrapper.go
```diff
@@ -99,15 +99,13 @@ func (hw httpWrapper) Request(
...
-	i := 0
-
 	// stream errors from http/2 will fail before we reach
 	// client middleware handling, therefore we don't get to
 	// make use of the retry middleware.  This external
 	// retry wrapper is unsophisticated, but should only
 	// retry in the event of a `stream error`, which is not
 	// a common expectation.
-	for i < 3 {
+	for i := 0; i < 3; i++ {
```
pandeyabs, line 108: Do we need retries here, now that we are also retrying in graph adapter wrapper?
Could we move `streamErrRE` match logic to the wrapper?

pandeyabs: Sorry, nevermind. This won't be using the graph adapter.

src/internal/m365/graph/service.go
```diff
@@ -63,18 +71,18 @@ type Servicer interface {
 var _ Servicer = &Service{}
 
 type Service struct {
-	adapter *msgraphsdkgo.GraphRequestAdapter
+	adapter abstractions.RequestAdapter
```
pandeyabs, line 74: Quick sanity check on this: Is the removal of pointer intentional here? Looks fine to me, but just want to confirm

src/internal/m365/graph/service.go
```diff
@@ -275,3 +288,70 @@ func kiotaMiddlewares(
...
+	// make use of the retry middleware.  This external
+	// retry wrapper is unsophisticated, but should only
+	// retry in the event of a `stream error`, which is not
+	// a common expectation.
+	for i := 0; i < 3; i++ {
+		ictx := clues.Add(ctx, "request_retry_iter", i)
+
+		sp, err = aw.RequestAdapter.Send(ctx, requestInfo, constructor, errorMappings)
+		if err != nil &&
+			!(errors.Is(err, syscall.ECONNRESET) ||
```
pandeyabs, line 341: We do have a helper `IsErrConnectionReset` for this

src/internal/m365/graph/service_test.go
```diff
@@ -96,3 +102,74 @@ func (suite *GraphUnitSuite) TestSerializationEndPoint() {
...
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	url := "https://graph.microsoft.com/fnords/beaux/regard"
+
+	// the panics should get caught and returned as errors
+	alwaysPanicMiddleware := mwForceResp{
+		alternate: func(req *http.Request) (bool, *http.Response, error) {
+			panic(clues.New("intentional panic"))
```
pandeyabs, line 117: :panic:

pandeyabs (approved):
Looks good! Great fix!

src/internal/m365/graph/service.go
ryanfkeepers: Yup, but I'm not currently certain that I want to catch context cancellations at this location, since those _should_ get handled by the middleware.  In the meantime I wanted to avoid 9x retries per call if we pathologically hit a ctx cancel.

PR #3905, opened 2023-07-26
Persist backup models for assist backups

pandeyabs:
This PR adds changes to persist backup models for assist backups. 	Primary reason for doing this is to ensure we don't lose corso extension data(deets) for items which were downloaded and processed by kopia during this backup op.

Note: This is 1/n PRs, and will not be merged into main until all n are reviewed, given we are touching state here.

**Changes in this PR:**
1. Error handling changes to ensure assist backups get persisted
2. Introduce tags for assist/merge backups. These tags will be used for filtering during base selection process.
3. Don't persist backups if they are not assist backup, and they have non zero errors.
4. Remove `LabelForceNoBackupCreation` criteria for not persisting a backup. More details are in [PR comment](https://github.com/alcionai/corso/pull/3905/files#r1278030140).
5. Switch kopia failure from non-recoverable to recoverable. More details are in [PR comment](https://github.com/alcionai/corso/pull/3905/files#r1278019992).

**What are assist backups?**
Below conditions must be true for a backup operation to qualify as an assist backup.
	1. new deets were produced during this backup operation
	2. we have a valid snapshot ID
	3. we don't have any non-recoverable errors
	4. we have recoverable errors

**A merge backup satisfies the following conditions:**
	1. Is not an assist backup
	2. No primary or recoverable errors occurred during the backup operation.

Tests will be added in later PRs.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* internal

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/operations/backup.go
```diff
@@ -502,7 +526,7 @@ func consumeBackupCollections(
 	if kopiaStats.ErrorCount > 0 {
 		err = clues.New("building kopia snapshot").WithClues(ctx)
 	} else if kopiaStats.IgnoredErrorCount > kopiaStats.ExpectedIgnoredErrorCount {
-		err = clues.New("downloading items for persistence").WithClues(ctx)
+		logger.Ctx(ctx).Info("recoverable errors were seen during backup")
```
pandeyabs, line 529: Note: I've removed primary failure from here. Thought process behind this:
1. We set[ kopia error handling policy](https://github.com/alcionai/corso/blob/34d6c18d74698b3ec8e0318e855dd498d4e93bf7/src/internal/kopia/wrapper.go#L284) to ignore file & directory errors.
2. Ignored errors are recoverable, [as marked here](https://github.com/alcionai/corso/blob/34d6c18d74698b3ec8e0318e855dd498d4e93bf7/src/internal/kopia/upload.go#L290) in corso.
3. Recoverable errors should not be counted as a primary failure. 
4. Presence of recoverable errors (and absence of primary failure) is a key requirement for persisting assist backups.

src/internal/operations/backup.go
```diff
@@ -239,35 +256,24 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		return op.Errors.Failure()
 	}
 
-	// force exit without backup in certain cases.
-	// see: https://github.com/alcionai/corso/pull/2510#discussion_r1113532530
-	for _, e := range op.Errors.Recovered() {
```
pandeyabs, line 244: Note: I've removed this section. Reasons:
1. There is 1:1 mapping between `AddRecoverable` and `LabelForceNoBackupCreation`, i.e. we always attach this label to recoverable failures. Now that we *want* to persist backups which have recoverable failures, we need to reassess places where this label should be added. 
2. We don't hit this code path currently. Assuming `FailAfterRecovery` policy is in place(which is default), `finalizeErrorHandling` will [fold recoverable failures](https://github.com/alcionai/corso/blob/e0368ba45af72be2d97e7e82285f76bba84c0636/src/internal/operations/helpers.go#L40) into non-recoverable. Later on, `persistResults` is going to return this folded `Failure()` and we will exit at line 239.

src/pkg/backup/backup.go
```diff
@@ -70,7 +71,8 @@ type Backup struct {
 // interface compliance checks
 var _ print.Printable = &Backup{}
 
-func New(
+func CreateNewBackup(
```
ryanfkeepers, line 74: I'm lukewarm on this rename.  The package name is already `backup`, so there's a redundancy in calling `backup.CreateNewBackup`.  And `backup.CreateNew` is a longer version of `backup.Create` or `backup.New`.  Of the latter two, either one seems good to me.

src/pkg/backup/backup.go
```diff
@@ -111,12 +114,31 @@ func New(
 		}
 	}
 
+	ctx = clues.Add(ctx, "is_assist_backup", isAssistBackup)
+	ctx = clues.Add(ctx, "error_count", errCount)
+	ctx = clues.Add(ctx, "skip_count", skipCount)
```
ryanfkeepers, line 119: Add is variadic:
```suggestion
	ctx = clues.Add(
	  ctx,
	  "is_assist_backup", isAssistBackup,
	  "error_count", errCount,
	  "skip_count", skipCount)
```

src/pkg/backup/backup.go
```diff
@@ -143,7 +165,7 @@ func New(
 			SkippedNotFound:           notFound,
 			SkippedInvalidOneNoteFile: invalidONFile,
 		},
-	}
+	}, nil
```
ryanfkeepers, line 168: The backup struct is a bit large to be part of a return value in a multi-value return.  Can you swap it to something like this, please?

```
bup := Backup{...}

return &bup, nil
```

src/pkg/backup/backup.go
```diff
@@ -111,12 +114,31 @@ func New(
...
+	tags := map[string]string{
+		model.ServiceTag: selector.PathService().String(),
+	}
+
+	// Add tags to mark this backup as either assist or merge. This is used to:
+	// 1. Filter assist backups by tag during base selection process
+	// 2. Differentiate assist backups from merge backups
+	if isAssistBackup {
+		tags[model.AssistBackupTag] = ""
+	} else if errCount == 0 {
```
ryanfkeepers, line 130: The mix of ownership here worries me a bit.  This is both externally asking the caller to evaluate the state of the backup, and internally asserting conditions that may or may not conflict with the caller's decision.  I feel like one or the other is best for long term maintenance: let the caller tell you whether its an assist or a merge, or make that decision on their behalf.

src/internal/operations/backup.go
```diff
@@ -739,6 +765,10 @@ func (op *BackupOperation) persistResults(
 
 	op.Results.ItemsRead = opStats.ctrl.Successes
 
+	if op.Status == Completed {
```
ryanfkeepers, line 768: Is it possible for a backup to be completed and also have a non-nil op.Errors.Failure()?

src/internal/operations/backup.go
```diff
@@ -117,9 +117,26 @@ func (op BackupOperation) validate() error {
...
+	resourceCount  int
+	assistBackupOp bool
+}
+
+// To qualify as an assist backup op, all of the following must be true:
+// 1. new deets were produced during this operation
+// 2. we have a valid snapshot ID
+// 3. we don't have any non-recoverable errors
+// 4. we recorded recoverable errors
+func isAssistBackupOp(
```
ryanfkeepers, line 131: Nit, I think adding `op` to the name adds confusion.  There's no such thing as an `assist operation`.  There's only an assist-category backup at the end of a backup operation.

```suggestion
func isAssistBackup(
```

src/internal/operations/backup.go
```diff
@@ -117,9 +117,26 @@ func (op BackupOperation) validate() error {
 // pointer wrapping the values, while those values
 // get populated asynchronously.
 type backupStats struct {
-	k             *kopia.BackupStats
-	ctrl          *data.CollectionStats
-	resourceCount int
+	k              *kopia.BackupStats
+	ctrl           *data.CollectionStats
+	resourceCount  int
+	assistBackupOp bool
```
ryanfkeepers, line 123:
```suggestion
	tagAsAssistBackup bool
```

src/internal/operations/backup.go
```diff
@@ -239,35 +256,24 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
...
+		deets.Details(),
+		opStats.assistBackupOp)
 	if err != nil {
-		op.Errors.Fail(clues.Wrap(err, "persisting backup"))
+		op.Errors.Fail(clues.Wrap(err, "persisting backup models"))
 		return op.Errors.Failure()
 	}
 
-	logger.Ctx(ctx).Infow("completed backup", "results", op.Results)
+	if op.Errors.Failure() == nil {
```
ryanfkeepers, line 272: Can we even get to this point if op.Errors.Failure() is non-nil?

src/internal/operations/backup.go
ashmrtn: also a question: do we only want success/failure states? Are backups that succeeded with errors considered successful? How will we display that to the users?

src/pkg/backup/backup.go
```diff
@@ -111,12 +112,21 @@ func New(
...
+	tags := map[string]string{
+		model.ServiceTag: selector.PathService().String(),
+	}
+
+	// Add tags to mark this backup as either assist or merge. This is used to:
+	// 1. Filter assist backups by tag during base selection process
+	// 2. Differentiate assist backups from merge backups
+	if isAssistBackup {
+		tags[model.AssistBackupTag] = ""
+	}
```
ashmrtn, line 124: it might actually be easier to use the same key (e.x. "status" or something) and add a value as either "assist-base" or "merge-base" (naming TBD, there for example purposes)

The reasoning for this is that it can make selecting backups to display to the user and/or sanity checking a bit easier. We can't switch to nice clean logic that operates on one or the other tag yet since there's still old backups, but if we ever break backwards compatibility it'll give us a leg up. For sanity checking, it helps enforce the idea that everything will be tagged and everything is highly likely to have only one or the other tag because operators like `+=` are unlikely to be used for a single field

ashmrtn (commented):
I think this PR looks pretty good (mostly just nits here and there). I'm happy with the logic overall and don't see the need for major changes

However, given that there's going to be more PRs in the chain I'm going to hold off on approving for the moment just so things don't go in accidentally

src/internal/operations/backup.go
pandeyabs: > Is it possible for a backup to be completed and also have a non-nil op.Errors.Failure()?
- I have changed this logic a bit. Could you please re-review? 

> Are backups that succeeded with errors considered successful? How will we display that to the users?
- This needs some discussion. I'll prepare some material to discuss on Monday.

src/internal/operations/backup.go
pandeyabs: Yes, because `finalizeErrorHandling` folds recoverable errors into non-recoverable errors if `FailAfterRecovery` policy is being used. 

Also, I've shuffled the code here a bit, but above is still true.

src/internal/operations/backup.go
```diff
@@ -117,9 +117,65 @@ func (op BackupOperation) validate() error {
...
+		return false
+	}
+
+	if err.Failure() != nil {
+		return false
+	}
+
+	if failurePolicy == control.BestEffort {
+		return true
+	}
```
ashmrtn, line 176: Honestly, I think we can remove this. Incremental base selection is a black box to end users and it's not really safe to have backups that could have errors in the merge base set. We don't expect many people to be using this anyway
```suggestion
```

src/internal/operations/backup.go
```diff
@@ -117,9 +117,65 @@ func (op BackupOperation) validate() error {
...
+	snapID, ssid string,
+	failurePolicy control.FailurePolicy,
+	err *fault.Bus,
+) bool {
+	return newDeetsProduced &&
+		snapID != "" &&
+		ssid != "" &&
+		err.Failure() == nil &&
+		len(err.Recovered()) > 0 &&
+		failurePolicy != control.BestEffort
```
ashmrtn, line 153: if the best effort mode check is removed from isMergeBackup it should be removed from here too

src/internal/operations/backup.go
```diff
@@ -117,9 +117,65 @@ func (op BackupOperation) validate() error {
...
+
+// To qualify as an assist backup, all of the following must be true:
+// 1. new detail entries were produced during this operation
+// 2. The updated details file was persisted without error
+// 3. we have a valid snapshot ID
+// 4. we don't have any non-recoverable errors
+// 5. we recorded recoverable errors
+// 6. We are not running in best effort mode. Reason being that there is
+// no way to distinguish assist backups from merge backups in best effort mode
+// Primary reason for persisting assist backup models is to ensure we don't
```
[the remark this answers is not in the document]
pandeyabs: Ah yes. This has become too verbose. I'll trim it down a bit.

src/internal/operations/backup.go
pandeyabs: Great! Sounds good

pandeyabs: Do note that after removing BestEffort checks in `isMergeBackup` &`isAssistBackup`, here is the new behavior for users who may continue using BestEffort mode.

1. If user runs into recoverable errors, we will persist an assist backup.
2. Say there is a file which always fails to download with a recoverable error, we will never be able to persist a merge backup for this user. 

Given above, i think we should just remove support for BestEffort mode during backup. I'll create a ticket for this.

ashmrtn: >we will never be able to persist a merge backup for this user. So no restore possibility.

I'm not certain this is the case right now. The backup list logic and restore logic don't check tags as far as I know, they just check for the existence of backup models. If they only check if the backup model exists, has an item data snapshot and a backup details snapshot then a restore will technically be possible

pandeyabs: That makes sense. Should we also add a PR to only report merge backups during `backup list`?
Returning assist backups in that list seems wrong, since assist backups are an intermediate state with recoverable errors, and we should probably prevent restoring from them?

ashmrtn: yea, there should be a PR to update how list works. As far as what we filter out, I'm still debating that a little. As you've pointed out, we should probably technically display best effort backups as completed since they well, completed. Unfortunately this may put us at odds between tags and display :/ Open to discussion on this point, and the discussion doesn't necessarily have to happen in this thread

pandeyabs: Sounds good. I propose the following:

1.  We keep the best effort mode checks in `isMergeBackup` & `isAssistBackup`. So we will continue to persist merge backups for best effort modes. 
2. We add changes to `list backup` to only return merge backups i.e. they are not tagged with `AssistBackup` tag. 
3. Regarding future of best effort mode, I'm creating an internal doc for discussion.

Does this sound good to you? :)

ashmrtn: yup, sounds good!

src/internal/operations/backup.go
```diff
@@ -345,6 +389,7 @@ func (op *BackupOperation) do(
 		return nil, clues.Wrap(err, "persisting collection backups")
 	}
 
+	opStats.hasNewDetailEntries = deets != nil && !deets.Empty()
```
ashmrtn, line 392: nit: per one of our discussions this needs updated to take into account the fact that details entries from assist bases are now in the `DetailsMergeInfoer`. The check should probably be something like the below (we can refine it as we go forward if we need to, but for now this will suffice)
```suggestion
	// tbh, not sure if it can be another && operator instead of ||
	opStats.hasNewDetailEntries = (deets != nil && !deets.Empty()) || (toMerge != nil && !toMerge.Empty())
```

pandeyabs: Ah yes! Thanks for catching this! I guess we would use `ItemsToMerge()` check for `toMerge` here, right?

But this opens up a question. How do we distinguish between assist detail entries in `toMerge` compared to merge detail entries? Do we even need to anymore? Consider a scenario where `deets` is empty, and `toMerge` only has details sourced from a merge base. We might end up persisting an assist backup in this case, but we dont really have newly added deets?

Also re: `&&` vs `||`. I think we would need `||`. Consider the case where it's the very first backup, `toMerge` would be nil here.

ashmrtn: Good question!

I think there's a few things that together decide the behavior we need overall:
1. it's safe to be conservative and force detail merging even if there's no assist base items merged (it's not necessarily efficient, but it won't break anything)
2. if there are any assist details items we do need to merge them even if `deets` is empty. If we don't they won't be there the next time we use the current backup as an assist base. The backup will instead just have an empty details file which won't cause it to not be an assist base but it will break our assumption that the assist base contains details item info for every item in it

Going back to your questions/comments with the above context,

>How do we distinguish between assist detail entries in toMerge compared to merge detail entries? Do we even need to anymore?

Right now we don't have a way to tell assist details entries and merge details entries apart. If we really want to be efficient about things we would figure out a way to tell them apart. For now we can just always do merging to get this wrapped up and not over-optimize. If we find we need to be more efficient about it down the line we can revisit this

>Also re: && vs ||. I think we would need ||. Consider the case where it's the very first backup, toMerge would be nil here.

Excellent point, I didn't think that far ahead when I wrote the comment heh

pandeyabs: Agree with your thought process here! I think we are good to go. Thanks again!

PR #3907, opened 2023-07-26
Return assist backup models during base selection process

pandeyabs:
This PR adds support for selecting assist backup models. The intent here is to use the assist backups during merge details process. 

Changelist:
1. Assist bases now have a requirement to have a) an associated assist backup model b) details ssid. Please see [this code comment](https://github.com/alcionai/corso/pull/3907/files#diff-f9b6b7ab52c8e1148147909fd071d9b80b816d072203d8e9d2fa34ab93185db1R286) for more.
2. Any incomplete assist bases or assist bases not matching 1) are now discarded, i.e. they are not fed to kopia for kopia assisted incrementals. The impact here is that files cached by kopia might be now redownloaded if the backup failed without qualifying for an assist backup.

Unit tests added.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/backup_bases.go
```diff
@@ -381,7 +432,41 @@ func (bb *backupBases) fixupAndVerify(ctx context.Context) {
...
+				"dropping assist backup, invalid snapshot or details ID",
+				"backup_id", bup.ID)
+
+			continue
+		}
+
+		assistBackupsToKeep = append(assistBackupsToKeep, bup)
+	}
+
+	// TODO(pandeyabs): Go through all reasons & ensure that the assist backups
```
pandeyabs, line 463: @ashmrtn Could you please comment on this? :) Should I get this TODO fixed before we merge this PR?

src/internal/kopia/backup_bases.go
```diff
@@ -180,9 +187,10 @@ func (bb *backupBases) MergeBackupBases(
 	}
 
 	res := &backupBases{
-		backups:     bb.Backups(),
-		mergeBases:  bb.MergeBases(),
-		assistBases: bb.AssistBases(),
+		backups:       bb.Backups(),
+		mergeBases:    bb.MergeBases(),
+		assistBases:   bb.AssistBases(),
+		assistBackups: bb.AssistBackups(),
```
ashmrtn, line 193: I think the code below in this function (can't comment directly) needs some touchups for selecting assist backups too. The code will be similar to lines 189/197-209/217 except we'll only add to assist. I guess we should also be adding the merge backups to the assist backup set? Or do we need to completely separate the assist set and the merge set?

src/internal/kopia/backup_bases.go
ashmrtn: huh good callout, I didn't think to check that when I wrote this function. I would say it would be nice to have, but not critical to getting this feature enabled. There's already a fair amount of testing for the base selection and merge code that should hopefully ensure that doesn't happen. We can pick this up when we have some spare time

src/internal/kopia/base_finder.go
```diff
@@ -251,6 +252,8 @@ func (b *baseFinder) findBasesInSet(
 				logger.Ctx(ictx).Info("found incomplete backup")
```
ashmrtn, line 252: I think this if-block also needs updated since we're possibly adding to the set of assist backups here. Maybe removing this block and checking below for a non-empty incomplete reason (or requiring an empty incomplete reason) would lead to somewhat cleaner code?

src/internal/kopia/base_finder.go
```diff
@@ -304,30 +307,108 @@ func (b *baseFinder) findBasesInSet(
 		// If we've made it to this point then we're considering the backup
```
ashmrtn, line 307: I can't comment on them directly but lines 283/286-302/305 should be changed to just `continue`. We can't add anything to the assist snapshot set anymore unless it fits the (new) definition of an assist base

The key to remember with this code is that it's selecting both the backups that will be used to source details during details merging and the snapshots that will be used either for hierarchy merging (merge bases) or passed to kopia to check for cached files (assist base)

src/internal/kopia/base_finder.go
```diff
@@ -304,30 +307,108 @@ func (b *baseFinder) findBasesInSet(
...
+) *BackupEntry {
+	if bup == nil {
+		return nil
+	}
+
+	allTags := map[string]string{
+		model.BackupTypeTag: model.AssistBackup,
+	}
+
+	allTags = normalizeTagKVs(allTags)
```
ashmrtn, line 375: I don't think manifest tags are normalized like snapshot tags are. The code for them was written at different times and I was trying to be fancy and get kopia CLI filters working with snapshot tags 😓

ashmrtn (commented):
I'm a little  confused about how the new code is meant to interact with existing code. Part of this could stem from the fact that the existing code is not setup the best for the changes you're making now. The questions that I have are:
1. what is `assistBases` meant to represent in `BackupBases`? Just the snapshots that match with the assist backups? The snapshots for assist backups and merge backups? Something else?
2. what are the expected possible outputs from `findBasesInSet`? The existing code will select a merge base if there's a valid one, a snapshot if it's incomplete or the backup model for it is missing some information (i.e. something that doesn't necessarily match our new definition of an assist base). The merge base backup's snapshot is also added as an assist base. One of my other comments notes that this code is selecting backup models for details merging and snapshots for kopia hierarchy merging/kopia cached file checks. With the stricter requirements for assist bases we should end up with a 1:1 mapping of backup model -> kopia snapshot, regardless of it's an assist base or a merge base but it doesn't seem like the added code is following that

src/internal/kopia/base_finder.go
pandeyabs: Thanks! I'll remove this.

pandeyabs:
Thanks @ashmrtn! Answers to your questions below:

> what is assistBases meant to represent in BackupBases? Just the snapshots that match with the assist backups? The snapshots for assist backups and merge backups? Something else?

* I tried to retain existing definition of assistBases. These are kopia assist bases which may or may not have a backup model associated with them. 
* assistBackups are backup models(manifests) with a valid ssid & assistBase snapshot ID(one of the assistBases). I only intended to use this during `mergeDetails`. These are not fed into kopia.

> With the stricter requirements for assist bases we should end up with a 1:1 mapping of backup model -> kopia snapshot, regardless of it's an assist base or a merge base but it doesn't seem like the added code is following that

Yeah. I think this is where I should have communicated better. I know we have talked about discarding assist bases if they don't have a backup model(& associated details ssid). I had documented this as a future task (will reshare the doc). 

Now I am thinking about it, I think we can add 1:1 mapping change right away.

src/internal/kopia/backup_bases.go
ashmrtn: looking a this again the code does need expanded to handle the assist backup models. Otherwise we'll accidentally drop them from the set when this function is called

pandeyabs: Thanks! So I thought about this some more, and removed some migration code which merges legacy assist bases. 

**Summary:**
1. Bases tagged with `fallbackReasons` can only be merge bases, not assist bases. One exception being merge base reported as an assist base. This will be fixed in #3943 
2. Assist backups can never exist for `fallbackReasons`, because we don't do UPN based tagging anymore.

Impact of 1) is minimal. This would impact below scenario.

A) User updates corso to latest, and has a legacy merge base + legacy assist base. Delta token will be sourced from legacy merge base. Assuming its valid, items will be enumerated & due to 1) those cached items will be redownloaded by kopia from graph. 
B) Similar to A) but only has a legacy assist base. We'll do a full backup in this case.

src/internal/kopia/base_finder.go
```diff
@@ -204,17 +204,25 @@ func (b *baseFinder) getBackupModel(
 	return bup, nil
 }
 
+type backupBase struct {
+	*BackupEntry
+	*ManifestEntry
```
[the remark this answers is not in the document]
pandeyabs: Thank you! Will fix this.

src/internal/kopia/base_finder.go
```diff
@@ -285,49 +272,125 @@ func (b *baseFinder) findBasesInSet(
...
-		// complete as it has both an item data snapshot and a backup details
-		// snapshot.
+		// Found the most recent merge backup. We can stop here.
+		//
+		// TODO(pandeyabs): Ideally we would do some sanity checks here, e.g.
+		// 1. Check for MergeBackup tag. Not safe, since we may have merge
+		// backups with older versions of corso which won't have the tag.
+		// 2. Check for ErrorCount == 0. We will enforce this after we remove
+		// support for BestEffort mode.
 		logger.Ctx(ictx).Infow("found complete backup", "base_backup_id", bup.ID)
```
ashmrtn, line 318: there needs to be some other check prior to reaching this point. Based on the above we'll reach this point if we've already found an assist base even if the base we're looking at in this loop iteration is an assist base. I think just checking for tags would suffice?

src/internal/kopia/base_finder.go
```diff
@@ -285,49 +272,125 @@ func (b *baseFinder) findBasesInSet(
...
+		mergeBackup = &backupBase{
+			BackupEntry:   mergeModel,
+			ManifestEntry: &mergeSnap,
+		}
+
+		break
 	}
 
-	logger.Ctx(ctx).Info("no base backups for reason")
+	logger.Ctx(ctx).Info("no merge or assist backup for reason")
```
ashmrtn, line 340: nit: could you please update this log message? We'll reach here even if we do find valid bases

src/internal/kopia/base_finder.go
pandeyabs: Good catch! Fixed.

src/internal/kopia/backup_bases.go
```diff
@@ -381,7 +407,41 @@ func (bb *backupBases) fixupAndVerify(ctx context.Context) {
 		assistToKeep = append(assistToKeep, man)
 	}
 
+	// Now do the same for assist backups. Discard any that have overlapping
+	// reasons or if missing details or snapshot IDs.
+	assistBackupDrops := findNonUniqueAssistBackups(ctx, bb.assistBackups)
+
+	for _, bup := range bb.assistBackups {
```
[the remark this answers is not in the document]
pandeyabs: Thanks! This was a remnant of old logic. I've updated this code.

src/internal/kopia/backup_bases.go
```diff
@@ -373,15 +358,63 @@ func (bb *backupBases) fixupAndVerify(ctx context.Context) {
 
 	var assistToKeep []ManifestEntry
 
-	for _, man := range bb.assistBases {
+	// Every merge base is also a kopia assist base.
+	// TODO(pandeyabs): This should be removed as part of #3943.
+	for _, man := range bb.mergeBases {
```
pandeyabs, line 363: @ashmrtn Could you please review the PR again? :) I have added a few changes + tests.

One change of note here is that I am adding merge bases to assist bases here. It was earlier at `findBasesInSet`. But that was making `fixupAndVerify` more complex, since we lose 1:1 mapping between assist backup & assist snapshot. 

This doesn't change behavior from `FindBase` caller's perspective.  Given that we are going to stop adding merge bases to assist very soon, I think this is a step in right direction.

I've also fixed the `fixupAndVerify` tests to match above.

src/internal/kopia/base_finder_test.go
```diff
@@ -701,9 +873,167 @@ func (suite *BaseFinderUnitSuite) TestGetBases() {
...
+					nil),
+				newBackupModel(
+					testBackup1,
+					true,
+					true,
+					false,
+					map[string]string{model.BackupTypeTag: model.AssistBackup},
+					nil),
 			},
 		},
```
ashmrtn, line 1038: I love the test coverage here since it tries a lot of different combinations. However, could you please add one more test case (mostly just a sanity check) to ensure we still get the expected merge base when the merge base is tagged with the new merge base tags (i.e. `map[string]string{model.BackupTypeTag: model.MergeBase}`)? The precise set of inputs/outputs don't matter so much so long as the base is selected

ashmrtn (commented):
LGTM, just minor nits

src/internal/kopia/base_finder_test.go
pandeyabs: Good suggestion! Done.

pandeyabs:
Please ignore sonarcloud failure(duplicate lines in `fixupAndVerify`). Duplication is temporary & will be removed as part of #3943.

@ashmrtn This has been rebased now. Could you please approve, if things look good? :)

ashmrtn (approved):
just nits mostly. There is one possible gap in the logic that may need patched, but it should be pretty quick. In theory it could be patched in a separate PR after this merges

src/internal/kopia/backup_bases_test.go
```diff
@@ -249,192 +238,223 @@ func (suite *BackupBasesUnitSuite) TestMergeBackupBases() {
...
+		for _, i := range assistInputs {
+			baseID := fmt.Sprintf("id%d", i.id)
+
+			reasons := make([]Reasoner, 0, len(i.cat))
+
+			for _, c := range i.cat {
+				reasons = append(reasons, NewReason("", ro, path.ExchangeService, c))
+			}
+
+			m := makeManifest(baseID, "", "b"+baseID, reasons...)
```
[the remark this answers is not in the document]
pandeyabs: wow! thank you!

pandeyabs:
@ashmrtn Were you referring to this[ comment thread ](https://github.com/alcionai/corso/pull/3907#discussion_r1288900785) wrt gap in logic? If yes, i think we are good. I also verified that the incomplete assist base isn't returned by `fixupAndVerify`

ashmrtn:
>Were you referring to (this)[https://github.com/alcionai/corso/pull/3907#discussion_r1288900785] wrt gap in logic? If yes, i think we are good. I also verified that the incomplete assist base isn't returned by fixupAndVerify.

Yea, that's what I was worried about. As long as `fixupAndVerify` doesn't return it it should be fine! Thanks for double checking

PR #3908, opened 2023-07-26
Process assist backups first while merging details

pandeyabs:
This PR introduces a few changes to `mergeDetails`.
1. We now merge newly produced deets with assist backups first.
2. We then source the existing deets from merge backups(`DetailsMergeInfoer`) as before. The only new change here is to update items which may have been modified during assist backup. 

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -653,13 +653,81 @@ func mergeDetails(
 		return nil
 	}
 
-	var addedEntries int
+	var (
+		addedEntries int
+		excludeList  = make(map[path.Path]struct{})
+	)
+
+	// Process partial backups first, as they are more recent
+	for _, partialBackup := range partialBackups {
+		err := mergeDetailsFromBackup(
+			ctx,
+			detailsStore,
+			partialBackup,
+			dataFromBackup,
+			deets,
+			&addedEntries,
+			errs,
+			nil)
+		if err != nil {
+			return clues.Wrap(err, "merging details from partial backup")
+		}
+
+		// TODO(pandeyabs): This doesn't handle folders yet.
+		for _, entry := range deets.Details().Entries {
+			rr, err := path.FromDataLayerPath(entry.RepoRef, true)
+			if err != nil {
+				return clues.New("parsing deets item info path").
+					WithClues(ctx).
+					With("repo_ref", path.NewElements(entry.RepoRef))
+			}
+
+			excludeList[rr] = struct{}{}
+		}
+	}
 
 	for _, baseBackup := range backups {
-		var (
-			mctx                 = clues.Add(ctx, "base_backup_id", baseBackup.ID)
-			manifestAddedEntries int
-		)
+		err := mergeDetailsFromBackup(
+			ctx,
+			detailsStore,
+			baseBackup,
+			dataFromBackup,
+			deets,
+			&addedEntries,
+			errs,
+			excludeList)
+		if err != nil {
+			return clues.Wrap(err, "merging details from base backup")
+		}
+	}
+
+	checkCount := dataFromBackup.ItemsToMerge()
+
+	if addedEntries != checkCount {
+		return clues.New("incomplete migration of backup details").
+			WithClues(ctx).
+			With(
+				"item_count", addedEntries,
+				"expected_item_count", checkCount)
+	}
+
+	return nil
+}
+
+func mergeDetailsFromBackup(
+	ctx context.Context,
+	detailsStore streamstore.Streamer,
+	baseBackup kopia.BackupEntry,
+	dataFromBackup kopia.DetailsMergeInfoer,
+	deets *details.Builder,
+	addedEntries *int,
+	errs *fault.Bus,
+	excludeList map[path.Path]struct{},
+) error {
+	var (
+		mctx                 = clues.Add(ctx, "base_backup_id", baseBackup.ID)
+		manifestAddedEntries int
+	)
 
 		baseDeets, err := getDetailsFromBackup(
 			mctx,
@@ -678,15 +746,20 @@ func mergeDetails(
 					With("repo_ref", path.NewElements(entry.RepoRef))
 			}
 
-			// Although this base has an entry it may not be the most recent. Check
-			// the reasons a snapshot was returned to ensure we only choose the recent
-			// entries.
-			//
-			// TODO(ashmrtn): This logic will need expanded to cover entries from
-			// checkpoints if we start doing kopia-assisted incrementals for those.
-			if !matchesReason(baseBackup.Reasons, rr) {
-				continue
-			}
+		// Although this base has an entry it may not be the most recent. Check
+		// the reasons a snapshot was returned to ensure we only choose the recent
+		// entries.
+		//
+		// TODO(ashmrtn): This logic will need expanded to cover entries from
+		// checkpoints if we start doing kopia-assisted incrementals for those.
+		if !matchesReason(baseBackup.Reasons, rr) {
+			continue
+		}
+
+		// If we've already added this entry from a more recent backup, skip it.
+		if _, ok := excludeList[rr]; ok {
+			continue
+		}
 
 			mctx = clues.Add(mctx, "repo_ref", rr)
 
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -402,7 +402,6 @@ func (op *BackupOperation) do(
 		toMerge,
 		deets,
 		writeStats,
-		op.Selectors.PathService(),
 		op.Errors)
 	if err != nil {
 		return nil, clues.Wrap(err, "merging details")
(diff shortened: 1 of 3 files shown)


src/internal/operations/backup.go
```diff
@@ -632,23 +633,98 @@ func getNewPathRefs(
...
+			if err != nil {
+				return nil, clues.New("parsing base item info path").
+					WithClues(mctx).
+					With("repo_ref", path.NewElements(entry.RepoRef))
+			}
+
+			// Although this assist base has an entry it may not be the most recent.
+			// Check the reasons a snapshot was returned to ensure we only choose the
+			// recent entries.
+			if !matchesReason(assistBackup.Reasons, rr) {
```
pandeyabs, line 681: @ashmrtn I want to address one case here that you had brought up.

> assist bases `F1` (oldest, original categories: [email, contacts], selected category: `[contacts]`) & `F2` (original category: [email], selected category: `[email]`), have different versions of the item.

I think we are okay here. Repo ref for an email item from `F1` will not match with the selected `F1` reason(`assistBackup.Reasons = contacts`) here. So this item will be discarded. 

This email item will be picked up from `F2`.

src/internal/operations/backup.go
```diff
@@ -632,23 +633,98 @@ func getNewPathRefs(
...
+		if err != nil {
+			return nil, clues.New("fetching details for assist backup")
+		}
+
+		for _, entry := range baseDeets.Items() {
+			rr, err := path.FromDataLayerPath(entry.RepoRef, true)
+			if err != nil {
+				return nil, clues.New("parsing base item info path").
+					WithClues(mctx).
+					With("repo_ref", path.NewElements(entry.RepoRef))
```
ashmrtn, line 675: nit: I think LoggableDir may be a little clearer here. It'll also properly handle trailing slashes (I guess I should go make a PR for that for `NewElements`)

src/internal/operations/backup.go
ashmrtn: ooo good call. I want to take some time to think about this more in the context of concurrent backups, but that logic definitely makes sense if none of the assist bases overlapped

src/internal/operations/backup.go
```diff
@@ -632,23 +633,98 @@ func getNewPathRefs(
...
+			repoRefToAssistDeets[entry.RepoRef] = *entry
+		}
+	}
+
+	// Walk through new deets and update items from assist deets if needed.
+	// There are 2 possibilities here.
+	// 1. The item didn't exist in assist deets. Do nothing.
+	// 2. The item existed in the assist deets. Source it from assist deets
+	// if mod time hasn't changed. Otherwise current deets have the most
+	// recent info, do nothing.
```
ashmrtn, line 694: > Source it from assist deets
> if mod time hasn't changed. Otherwise current deets have the most
> recent info, do nothing.

I think this is where stuff starts to get a little fuzzy and I'm not sure all the logic will play out how we'd like. Again, this goes back to whether we assume backups can be concurrent or not

If we assume that backups aren't concurrent then this logic will work out just fine

However, if we assume that backups can be concurrent then it's possible the older backup (didn't have the Reason for this item) has the newer version of the item but we didn't select it because of the check on 681

src/internal/operations/backup.go
```diff
@@ -709,6 +785,20 @@ func mergeDetails(
...
+				logger.Ctx(ctx).Infow("sourcing item info from assist backup",
+					"repo_ref", entry.RepoRef)
+
+				// If something is considered cached then we assume the mod time
+				// hasn’t changed. If the mod time hasn’t changed then we
+				// assume neither the item data nor metadata changed, hence it’s
+				// safe to source the item’s detail entry from assist deets.
+				item = assistDeets.ItemInfo
+			}
+
```
ashmrtn, line 801: I don't think this addition is needed. With how assist base details merging is implemented, the items it's merging with are already in deets but not in `dataFromBackup`. This function and the helper function(s) it calls should only interact with `dataFromBackup` to see if something should be added to deets. That should guarantee disjoint sets in such a way that we can't match on an item in both this function and the function to merge assist base details

src/internal/operations/backup.go
```diff
@@ -632,23 +633,98 @@ func getNewPathRefs(
...
+// the current deets.
+func mergeDetailsFromAssistBackups(
 	ctx context.Context,
 	detailsStore streamstore.Streamer,
-	backups []kopia.BackupEntry,
-	dataFromBackup kopia.DetailsMergeInfoer,
 	deets *details.Builder,
-	writeStats *kopia.BackupStats,
+	assistBackups []kopia.BackupEntry,
 	serviceType path.ServiceType,
```
ashmrtn, line 643: nit: could you remove this please? It appears to be unused (shame on me for not noticing that when I wrote this)
```suggestion
```

src/internal/operations/backup.go
```diff
@@ -632,23 +633,98 @@ func getNewPathRefs(
...
+}
+
+func mergeDetailsFromBaseBackups(
+	ctx context.Context,
+	detailsStore streamstore.Streamer,
+	deets *details.Builder,
+	dataFromBackup kopia.DetailsMergeInfoer,
+	backups []kopia.BackupEntry,
+	repoRefToAssistDeets map[string]details.Entry,
+	serviceType path.ServiceType,
```
ashmrtn, line 725: nit: also unused

```suggestion
```

src/pkg/backup/details/details.go
```diff
@@ -470,11 +483,35 @@ func (d *Details) add(
 		entry.ItemRef = withoutMetadataSuffix(entry.ItemRef)
 	}
 
+	if d.repoRefToEntry == nil {
+		d.repoRefToEntry = make(map[string]*Entry)
+	}
+
+	d.repoRefToEntry[entry.RepoRef] = &entry
 	d.Entries = append(d.Entries, entry)
```
ashmrtn, line 491: I don't think this will behave the way we want it to. `repoRefToEntry` has a pointer to the location of the passed in details but `Entries` has a copy of the passed in details. Furthermore, `Entries` can be reallocated if the number of entries outgrows the capacity of the slice. That means a pointer to an entry obtained by taking the address of what's in the slice (even if it's the entry we just appended) may not remain valid outside this function

ashmrtn (changes requested):
logic to update an entry needs changed a bit since it's most likely going to break if the underlying slice is resized

src/internal/operations/backup.go
pandeyabs: This is to support below case. 

**4. D1 is updated between C1 & F2. D1 does not move between F2 & C3**
    1. Since D1 didn’t move, it won’t be redownloaded as F2 has it and kopia will think it’s cached. But the input collections will have D1 because it changed between C1 (where we got the delta token from) and C3
    2. need to add logic to determine that D1 was indeed cached and pull the extension output for it from F2 during details merging
    3. incorrect to use D1 from C1 because it was changed in F2 → need to source from F2

src/internal/operations/backup.go
pandeyabs: Okay, let me think a bit more about concurrent backups with overlapping reasons.

src/pkg/backup/details/details.go
pandeyabs: Oh! My understanding was that the reference to Entry is appended to slice. Just confirmed thats not true.
Thanks! I'll fix this.

src/internal/operations/backup.go
pandeyabs: Thought about concurrent backups with overlapping reasons a bit more. Even if we address that, it doesn't give us a foolproof solution(see scenario below). I think we should proceed with this PR with a followup to potentially support concurrent backup cases. 
I'll write up an internal doc to discuss this topic. 

**Scenario:**

* Extending on the example in this comment thread. Say `C3[contacts]` follows `F1` & `F2` and finishes.
* Now lets say `F4[email, contact]` follows `C3`. During base selection, `F2` will be selected for `email` as an assist base. But there will be no assist base for `contacts` because `C3[contacts]` is most recent. 
* This means the `email` item which was more recent in `F1`(compared to `F2`) will not be available for merge here.

src/pkg/backup/details/details.go
```diff
@@ -470,11 +483,35 @@ func (d *Details) add(
...
+	repoRef string,
+	newInfo ItemInfo,
+) error {
+	if d.repoRefToEntry == nil {
+		d.repoRefToEntry = make(map[string]*Entry)
+	}
+
+	entry, ok := d.repoRefToEntry[repoRef]
+	if !ok {
+		return clues.New("entry not found").With("repo_ref", repoRef)
```
ashmrtn, line 506: nit: use PII safe logging for RepoRef please
```suggestion
		return clues.New("entry not found").With("repo_ref", path.LoggableDir(repoRef))
```

src/internal/operations/backup.go
pandeyabs: Sounds good! 

> Part of what makes it (a tad) easier to deal with the existing code is that chunks of it are agnostic to assumptions like that.

100% with you on that.

PR #3912, opened 2023-07-26
use proper resource ids in factory

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/cmd/factory/impl/common.go
+++ b/src/cmd/factory/impl/common.go
@@ -224,7 +224,8 @@ var (
 
 func generateAndRestoreDriveItems(
 	ctrl *m365.Controller,
-	resourceOwner, secondaryUserID, secondaryUserName string,
+	protectedResource idname.Provider,
+	secondaryUserID, secondaryUserName string,
 	acct account.Account,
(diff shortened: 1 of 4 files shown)


src/cmd/factory/impl/common.go
```diff
@@ -224,7 +224,8 @@ var (
 
 func generateAndRestoreDriveItems(
 	ctrl *m365.Controller,
-	resourceOwner, secondaryUserID, secondaryUserName string,
+	protectedResource idname.Provider,
+	secondaryUserID, secondaryUserName string,
```
ashmrtn, line 228: would it be possible to remove one of these? It seems like we're passing in the same value but cased differently

ashmrtn (approved):

src/cmd/factory/impl/common.go
ryanfkeepers: It's technically usable for onedrive, in case we want the difference, so I'm going to keep it in for now.  We can revisit as needed.

PR #3916, opened 2023-07-26
rename export interface

ryanfkeepers:
to better match current standards

#### Type of change


meain (approved):
Ahh, this is a much better name.

PR #3919, opened 2023-07-27
Export data from OD/SP in logevity tests

meain:
~This tries to export all the backup that is available in the logevity test repo. This should warn us in case we are not able to export any older versions.~ Had a discussion and decided it would be better to export just two items and thought we could start with trying to export the first and the last item.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/3920

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/longevity_test.yml
+++ b/.github/workflows/longevity_test.yml
@@ -230,6 +230,62 @@ jobs:
 
 ##########################################################################################################################################
 
+# Export OneDrive Test
+      - name: OneDrive Export test
+        run: |
+          set -euo pipefail
+          echo -e "\Export OneDrive test\n" >> ${CORSO_LOG_FILE}
+
+          echo -e "\Export OneDrive test - first entry\n" >> ${CORSO_LOG_FILE}
+          ./corso backup list onedrive 2>/dev/null | tail -n+2 | head -n1 | awk '{print $1}' |
+              while read -r line; do
+                  ./corso export onedrive \
+                    "/tmp/corso-export--$line" \
+                    --no-stats \
+                    --backup "$line" \
+                    2>&1 | tee ${{ env.CORSO_LOG_DIR }}/export_onedrive.txt
+              done
+
+          echo -e "\Export OneDrive test - last entry\n" >> ${CORSO_LOG_FILE}
+          ./corso backup list onedrive 2>/dev/null | tail -n1 | awk '{print $1}' |
+              while read -r line; do
+                  ./corso export onedrive \
+                    "/tmp/corso-export--$line" \
+                    --no-stats \
+                    --backup "$line" \
+                    2>&1 | tee ${{ env.CORSO_LOG_DIR }}/export_onedrive.txt
+              done
+
+##########################################################################################################################################
+
+# Export SharePoint Test
+      - name: SharePoint Export test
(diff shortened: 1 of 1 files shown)


meain:
Just got started with exports in logevity tests. This PR currently has code to export every backup that is available in the logevity test repo. It does not test for the data that is exported. What is the direction we should be taking this in or is this all we need?

/cc @vkamra

.github/workflows/longevity_test.yml
```diff
@@ -230,6 +230,62 @@ jobs:
...
+              done
+
+          echo -e "\Export OneDrive test - last entry\n" >> ${CORSO_LOG_FILE}
+          ./corso backup list onedrive 2>/dev/null | tail -n1 | awk '{print $1}' |
+              while read -r line; do
+                  ./corso export onedrive \
+                    "/tmp/corso-export--$line" \
+                    --no-stats \
+                    --backup "$line" \
+                    2>&1 | tee ${{ env.CORSO_LOG_DIR }}/export_onedrive.txt
```
ashmrtn, line 256: nit: this should be using a different stdout output file if we don't want to clobber the previous one

.github/workflows/longevity_test.yml
```diff
@@ -230,6 +230,62 @@ jobs:
...
 ##########################################################################################################################################
 
+# Export OneDrive Test
+      - name: OneDrive Export test
+        run: |
+          set -euo pipefail
+          echo -e "\Export OneDrive test\n" >> ${CORSO_LOG_FILE}
+
+          echo -e "\Export OneDrive test - first entry\n" >> ${CORSO_LOG_FILE}
+          ./corso backup list onedrive 2>/dev/null | tail -n+2 | head -n1 | awk '{print $1}' |
```
ashmrtn, line 240: does this and the last backup export command assume there's at least one backup? The other alternative is to write some wrapper code for the corso SDK. Not sure if that would be easier when it comes to logging/output, but may allow for greater control of backup selection

I'm alright with suggesting an SDK wrapper if that makes more sense for this use-case because sanity testing already gives us coverage ensuring the command works

meain: > does this and the last backup export command assume there's at least one backup?

Yup, I'm assuming that is an OK assumption to make as we are doing a backup right before this.

> but may allow for greater control of backup selection

Any specific requirement that you have in mind for selecting backups?

.github/workflows/longevity_test.yml
meain: Ahh, good call. Let me fix that.

.github/workflows/longevity_test.yml
ashmrtn: ah good point about the backup before this. I guess I should pay more attention to the unchanged lines sometimes haha. I think mostly just edge case handling when it comes to backup selection, but probably not worth worry over too much. I think worst case we'll just end up exporting the same backup 2x if it's the only backup at the time

ashmrtn (approved):

Issue #3799, opened 2023-07-13
Enable immutable backups in longevity tests


ashmrtn:
For better test coverage we should enable immutable backups in the longevity tests. This will require the following:
* finalizing/merging pending PRs for immutable backups
* creating a bucket with
  * versioning
  * object locking
  * lifecycle policy
* changing the bucket longevity tests use
* enabling immutable backups in corso once bucket configuration is complete

related to #3519

ashmrtn:
a good chunk of this is done including getting a repo setup with immutable backups

Remaining work
- [ ] enable point in time checking for deleted backup
- [ ] enable S3 object retention setting checking

PR #3923, opened 2023-07-27
Create S3 retention checker

ashmrtn:
Create a CLI utility that checks retention config
information for a small set of objects. Utility
works as follows:
* given a set of prefixes find one current version
  of each object
* for each object selected above, retrieve the
  retention config
* compare retention mode and expiry with the given
  mode and an expiry calculated from the current
  time

To attempt to reduce costs, the utility uses the
list API to fetch object names/versions and match
prefixes. It will cancel listing objects (i.e.
stop fetching new pages) once all prefixes are
matched

Also includes a flag to select non-current versions
of objects with the same prefix. However, setting
this flag may lead to more API calls because some
objects may not have non-current versions. In that
case all objects will be listed

sample usage to check current and non-current versions
of an object with prefix 'x' and current and
non-current versions of an object with prefix 'q'.
The current versions should expire in >= 8h and the
non-current versions should expire <= 7h
```
s3checker --bucket <> --bucket-prefix <> \
  --prefix x --prefix q \
  --retention-mode GOVERNANCE --live-retention-duration 8h \
  --with-non-current --dead-retention-duration 7h
```

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3799

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/cmd/s3checker/pkg/s3/s3.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return &Client{
+		client:     cli,
+		bucketName: opt.BucketName,
+		prefix:     opt.Prefix,
+	}, nil
+}
+
+func maybeAddObj(prefix string, obj ObjInfo, m map[string]ObjInfo) {
```
meain, line 103: Does this make any difference even if we add?

src/cmd/s3checker/pkg/s3/s3.go
```diff
@@ -0,0 +1,210 @@
...
+		var e minio.ErrorResponse
+
+		if errors.As(err, &e) {
+			if e.Code == noLockCode {
+				return nil, nil, clues.Stack(ErrNoRetention, err)
+			}
+		}
+	}
+
+	return mode, retainUntil, clues.Wrap(err, fmt.Sprintf(
```
meain, line 205: Not sure if we should do `clues.Wrap(...).With(..., ...)` here and expand at the other side.

meain (approved):
My exposure to retention is very minimal(I was not able to test the code with retention), but the code LGTM. Approving to unblock as this is a completely separate module just for CI.

src/cmd/s3checker/pkg/s3/s3.go
ashmrtn: hmmm well I suppose as long as we don't start keeping a list of items instead of a single item it doesn't really matter. It'll just change which item in the set we pick (first seen vs. random element). The main thing we want is to just select a single item for each prefix so we don't make a bunch of get retention calls later

Issue #3929, opened 2023-07-28
Increase logging during hierarchy merging for better debugging


ashmrtn:
To make things easier to debug it would be good to increase the amount of logging we have there. Some things to consider logging:
* if the subtree root was marked as deleted (effectively disabling merging)
* number of directories merged in total
* number of directories not moved, moved, and deleted

ashmrtn:
Going to close for now, we can have another issue to track prefix subtree deletion

PR #3930, opened 2023-07-28
Track and print stats about merging

ashmrtn:
Track and print information about the number
of directories that were moved, deleted,
recursively moved or deleted, and that
didn't move at all

This can help shed light on what's happening
and generally help with debugging issues
that may crop up

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3929

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -852,6 +852,7 @@ func traverseBaseDir(
 	expectedDirPath *path.Builder,
 	dir fs.Directory,
 	roots map[string]*treeMap,
+	stats *mergeStats,
 ) error {
 	ctx = clues.Add(ctx,
 		"old_dir_path", oldDirPath,
@@ -889,9 +890,26 @@ func traverseBaseDir(
 		// This directory was deleted.
 		if upb == nil {
 			currentPath = nil
+			stats.del++
 		} else {
-			// This directory was moved/renamed and the new location is in upb.
+			// This directory was explicitly mentioned and the new (possibly
+			// unchanged) location is in upb.
 			currentPath = upb.ToBuilder()
+
+			if oldDirPath.String() == currentPath.String() {
+				stats.noMove++
+			} else {
+				stats.move++
+			}
+		}
+	} else {
+		// Just stats tracking stuff.
+		if currentPath == nil {
+			stats.recursiveDel++
+		} else if oldDirPath.String() == currentPath.String() {
+			stats.noMove++
+		} else {
+			stats.recursiveMove++
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/internal/kopia/upload.go
```diff
@@ -993,6 +1012,21 @@ func logBaseInfo(ctx context.Context, m ManifestEntry) {
 		"base_backup_id", mbID)
 }
 
+type mergeStats struct {
```
ryanfkeepers, line 1015: What's the possibility of having this replaced with a [counter](https://github.com/alcionai/corso/blob/main/src/pkg/count/count.go) instead of a new struct?  I'm hoping to centralize stats aggregation around that package in the long term.

ryanfkeepers (approved):
Would love to see a counter here instead of a new struct.  But nothing blocking, so ✅ either way.

src/internal/kopia/upload.go
ashmrtn: a few things:

I don't have a preference one way or the other as to making a new `counter` here vs. a new struct

More broadly, I think the question really turns into do these stats need to be reported at the top-level? From my perspective, I see this mostly as a sanity check/debug logging that's good to have all the time and isn't that noisy. If we start talking about passing in a `counter` to the hierarchy merging code so that things can be aggregated at a higher level, I don't find that useful unless we want to start reporting these metrics in some way. The counters do feel a bit more heavy weight than necessary given that they're thread-safe and this code is single threaded, but that's likely me trying to over-optimize things that don't matter since kopia may be making calls to S3 for directory listings anyway

ryanfkeepers: > I think the question really turns into do these stats need to be reported at the top-level?

Personally, I don't believe the method of reporting is an important part of the consideration.  Counters are useful for when we want to feed stats up the chain, true.  But more simply it's a way to aggregate behavior without needing to worry about the smaller details.  Such as...

> The counters do feel a bit more heavy weight than necessary given that they're thread-safe and this code is single threaded

The goal is for the interface to be lightweight enough that implementers don't need to worry about heavy or lightweightness.  If we have to go through a decision process of "do I really need the extra utility?" every time we could introduce a counter, then the package is already dead.  What I want to establish is deadpan boilerplate around, "I need to count things, here's a thing that counts things, I call Inc() when I up the value."  Then the implementer is insulated against future changes like moving towards routines, without ever needing to worry about it.  Things should just work.

Issue #3895, opened 2023-07-24
[Bug]: some data still unmasked when `--mask-sensitive-data` passed


ashmrtn:
### What happened?

when using the `--mask-sensitive-data` flag some information like the user display name (for OneDrive/Exchange) is still visible. For full and incremental backup shows the user display name still appears when connecting to the repo (progress output) and at the end of the run (unclear what's printing it)

Additionally, the tenant ID and user ID is found in some places in the logs when printing collection info/info about items being backed up

### Corso Version?

Corso v0.11.1 dev

### Where are you running Corso?

N/A

PR #3935, opened 2023-07-28
Better handling of clues, error, and log data if it's PII

ashmrtn:
Add better handling for
* hiding possibly sensitive data when logging
* adding possibly sensitive data to context clues
* adding context clues to returned errors

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3895

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/upload.go
```diff
@@ -463,7 +465,10 @@ func streamBaseEntries(
 
 		entName, err := decodeElement(entry.Name())
 		if err != nil {
-			return clues.Wrap(err, "decoding entry name: "+entry.Name())
+			return clues.Wrap(
+				err,
+				"decoding entry name: "+clues.Hide(entry.Name()).Conceal(),
```
ashmrtn, line 470: Not sure if this is how we want to handle this. I think there's two possible strategies, but I don't know if we have the infrastructure for them right now
1. hide sensitive data in logged error messages (unclear how)
2. have clues info printed to CLI if it adds useful context

ryanfkeepers: > hide sensitive data in logged error messages (unclear how)

My preference is this, by using structured data.
```
return clues.Wrap(err, "decoding entry name").
  With("entry_name", clues.Hide(entry.Name())
```
Clues will automatically call `Conceal()` as part of marshalling the entry name to the structured data.

src/internal/kopia/upload.go
```diff
@@ -508,13 +513,15 @@ func streamBaseEntries(
 		}
 
 		if err := ctr(ctx, entry); err != nil {
-			return clues.Wrap(err, "executing callback on item").With("item_path", itemPath)
+			return clues.Wrap(err, "executing callback on item").
+				WithClues(ctx).
```
ryanfkeepers, line 517: double checking because I can't remember what ctr does.  It won't call clues.wrap().WithClues(ctx)?

ryanfkeepers (approved):

src/internal/kopia/upload.go
ashmrtn: yea, usually that's what I do. However, I guess the catch is if this is an error that will be printed on the CLI will the added clues in the error be printed to the CLI too?

I think we're kind of in a situation where we want
* obfuscated data if we're logging the error
* plaintext if it's printed to CLI

Since we often log the error itself injecting unhidden data into it could leak PII into the logs even with `--mask-sensitive-data`. On the other hand, using `With(key, value)` could leave out useful info from CLI messages if we don't  print clues on the CLI

src/internal/kopia/upload.go
ashmrtn: no, we should assume it's an opaque function. It's a function that kopia provides which actually uploads data

src/internal/kopia/upload.go
ryanfkeepers: > I think we're kind of in a situation where we want

Personally, I think this that pivot is more of a dance than is worth chasing.  If the user is printout out to the terminal, they're already in a position where they can run things with plaintext.  On that note, I'm not even sure if it's possible for us to expose a log in both places? 

Per the error display, this is why it should be a hard rule to keep extra data in the structured set and out of the log messages.  It allows us to dump the error message (just the message) to the terminal, not as a logging statement, without worry of exposing pii.

ashmrtn: I think maybe we're talking past each other a little here. I'd say there are places where we log errors so that we can debug things. In those cases we want to obfuscate PII so that it's easy to send off logs for debugging

However, the most common situation where we'll talk about maybe sending logs for debugging is a user will run corso and an error will occur. In that case it would be good to print enough error information to the CLI that the user can see if it was a simple error (e.x. invalid user passed to exchange backup because there's a typo in the user name)

This is what leads to split expectations about error output when it comes to CLI (plaintext) and logs (maybe obfuscated due to `--mask-sensitive-data`). Ideally, the workflow for debugging even simple errors in corso is not asking people to find and search through log files, it's just on the CLI for them

PR #3939, opened 2023-07-31
fix loop in restore config conceal

ryanfkeepers:
#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/control/restore.go
+++ b/src/pkg/control/restore.go
@@ -121,9 +121,9 @@ func (rc RestoreConfig) marshal() string {
 func (rc RestoreConfig) concealed() RestoreConfig {
 	return RestoreConfig{
 		OnCollision:        rc.OnCollision,
-		ProtectedResource:  clues.Hide(rc.ProtectedResource).Conceal(),
+		ProtectedResource:  clues.Conceal(rc.ProtectedResource),
 		Location:           path.LoggableDir(rc.Location),
-		Drive:              clues.Hide(rc.Drive).Conceal(),
+		Drive:              clues.Conceal(rc.Drive),
 		IncludePermissions: rc.IncludePermissions,
 	}
 }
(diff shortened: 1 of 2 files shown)


src/pkg/control/restore.go
```diff
@@ -121,9 +121,9 @@ func (rc RestoreConfig) marshal() string {
 func (rc RestoreConfig) concealed() RestoreConfig {
 	return RestoreConfig{
 		OnCollision:        rc.OnCollision,
-		ProtectedResource:  clues.Hide(rc.ProtectedResource).Conceal(),
+		ProtectedResource:  clues.Conceal(rc.ProtectedResource),
 		Location:           path.LoggableDir(rc.Location),
-		Drive:              clues.Hide(rc.Drive).Conceal(),
```
zackrossman, line 126: Missing context on the relevance of this `Hide` function. Can you explain where the 'loop' in the description is?

vkamra: The loop is below in `Format` I believe.

ryanfkeepers: Ah, this change is incidental.  The loop is on line 141.  `rc.concealed()` returns another restore config.  `fmt.Fprint`, when handed a struct, internally calls `v.Format(fs, r)`, which hits the Format func again, and thus loops.  So calling marshal before Fprint ensures we pass in a string instead of a struct with a Format method.

src/pkg/control/restore.go
```diff
@@ -138,7 +138,7 @@ func (rc RestoreConfig) Conceal() string {
 // used within a PrintF, suitable for logging, storing in errors,
 // and other output.
 func (rc RestoreConfig) Format(fs fmt.State, _ rune) {
-	fmt.Fprint(fs, rc.concealed())
+	fmt.Fprint(fs, rc.concealed().marshal())
```
vkamra, line 141: To confirm - the issue was that `concealed()` returned a `RestoreConfig` which `Fprintf` would then call `Format` on?

zackrossman (approved):

src/pkg/control/restore.go
vkamra: @ryanfkeepers - do we need the change on line 124/125 above? Are we guaranteed that the items we are trying to conceal have already been hidden i.e. `hashText` has been generated?

ryanfkeepers: the 124/126 changes are a more direct version of the same call.  Same result either way.

vkamra: Got it - I re-read the code. Makes sense.

PR #3963, opened 2023-08-04
Update backup details merge logic

ashmrtn:
Update backup details merge logic to use assist
backup bases. As the modTime check is already in
DetailsMergeInfoer there's not much else to do
here besides wiring things up

Overall, this solution is an alternative to the
previous one. It works by placing all cached
items in the DetailsMergeInfoer instead of adding
them to details (assuming they had a details
entry)

During details merging, we can cycle through
all bases once and track only the items we've
added to details (so we don't duplicate things).
This works because we know precisely which items
we should be looking for

ModTime comparisons in the DetailsMergeInfoer
ensure we get the proper version of each item
details

**Note:** This requires a minor patch to how
we determine if it's safe to persist a backup
model because now backups won't produce details
entries for cached items until `mergeDetails`
runs

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -361,7 +361,7 @@ func (op *BackupOperation) do(
 	err = mergeDetails(
 		ctx,
 		detailsStore,
-		mans.Backups(),
+		mans,
 		toMerge,
 		deets,
 		writeStats,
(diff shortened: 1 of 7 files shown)


src/internal/operations/backup.go
```diff
@@ -655,88 +763,68 @@ func mergeDetails(
...
+		// accidentally merge them again. This could happen if, for example, there's
+		// an assist backup and a merge backup that both have the same version of an
+		// item at the same path.
+		alreadySeenEntries = map[string]struct{}{}
+	)
 
-		baseDeets, err := getDetailsFromBackup(
-			mctx,
-			baseBackup.Backup,
+	// Just to be on the safe side merge details from assist bases first. It
```
ryanfkeepers, line 779: You say that twice in the same paragraph. (:
```suggestion
	// Merge details from assist bases first. It
```

ryanfkeepers (approved):

PR #3983, opened 2023-08-07
SDK option and logic to drop assist bases

ashmrtn:
Add a way for SDK users to drop
kopia-assisted incremental bases
thus forcing item data redownload
if the item wasn't sourced from
a merge base

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2360

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -57,9 +57,9 @@ type BackupOperation struct {
 
 	// when true, this allows for incremental backups instead of full data pulls
 	incremental bool
-	// When true, allows for kopia-assisted incremental backups instead of
-	// downloading and hashing all item data.
-	assistBackup bool
+	// When true, disables kopia-assisted incremental backups. This forces
+	// downloading and hashing all item data for items not in the merge base(s).
+	disableAssistBackup bool
 }
 
 // BackupResults aggregate the details of the result of the operation.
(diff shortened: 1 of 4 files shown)


src/pkg/control/options.go
```diff
@@ -62,6 +62,10 @@ type Toggles struct {
 	// DisableIncrementals prevents backups from using incremental lookups,
 	// forcing a new, complete backup of all data regardless of prior state.
 	DisableIncrementals bool `json:"exchangeIncrementals,omitempty"`
+	// DisableAssistCaching disables finding cached items in previous failed
+	// backups (i.e. kopia-assisted incrementals). Data dedupe will still occur
+	// since that is based on content hashes.
+	DisableAssistCaching bool `json:"disableAssistCaching,omitempty"`
```
ashmrtn, line 68: happy to take suggestions on better names if folks have ideas

ryanfkeepers: I mentioned this in the CLI PR, but I think `caching` is overused and difficult to correctly interpret.  Granted, having the space to call it `assistCaching` helps.  But I still think it's better avoided.  `DisableStorageAssist`, maybe.  Or, if you want to keep with the suggestion in the other PR, `ForceFullBackup`.

src/internal/operations/manifests.go
```diff
@@ -15,11 +15,44 @@ import (
...
+	bb, meta, useMergeBases, err := getManifestsAndMetadata(
+		ctx,
+		bf,
+		rp,
+		reasons,
+		fallbackReasons,
+		tenantID,
+		getMetadata)
+	if err != nil {
+		return nil, nil, false, clues.Stack(err).OrNil()
```
ryanfkeepers, line 35: we already know the err isn't nil.
```suggestion
		return nil, nil, false, clues.Stack(err)
```

ryanfkeepers (approved):

src/internal/operations/backup.go
```diff
@@ -312,6 +318,10 @@ func (op *BackupOperation) do(
 		lastBackupVersion = mans.MinBackupVersion()
 	}
 
+	// TODO(ashmrtn): This should probably just return a collection that deletes
+	// the entire subtree instead of returning an additional bool. That way base
+	// selection is controlled completely by flags and merging is controlled
+	// completely by collections.
```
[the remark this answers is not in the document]
ashmrtn: no, it should be safe due to how we select bases and persist data. The "deletion" happens during hierarchy merging so it's really just excluding directories from the base when building the in-memory state

If corso crashes partway through the item data backup with the "deleted" state in-memory then we won't select that as a base with the new corso extension base selection logic (it won't have a details snapshot)

If corso hits some error persisting item data while making the snapshot but does still persist the details then the base will be eligible as an assist base on the next backup. That's safe though because the assist base contains all the details we'll need and the original merge base still has all the undeleted items

PR #3984, opened 2023-08-07
Add CLI flag for disabling assist-incrementals

### LINKED ISSUE #2360: Flag to disable kopia-assisted incrementals

ashmrtn:
Adds hidden flag for disabling
kopia-assisted incrementals

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #2360

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/flags/options.go
```diff
@@ -110,6 +112,18 @@ func AddDisableIncrementalsFlag(cmd *cobra.Command) {
 	cobra.CheckErr(fs.MarkHidden(DisableIncrementalsFN))
 }
 
+// Adds the hidden '--disable-caching' cli flag which, when set, disables
+// kopia-assisted incremental backups.
+func AddDisableAssistCachingFlag(cmd *cobra.Command) {
```
ashmrtn, line 117: happy to take naming suggestions if folks have them

ryanfkeepers: `Greg`, preferably.

src/cli/flags/options.go
```diff
@@ -9,6 +9,7 @@ const (
 	DisableConcurrencyLimiterFN = "disable-concurrency-limiter"
 	DisableDeltaFN              = "disable-delta"
 	DisableIncrementalsFN       = "disable-incrementals"
+	DisableAssistCachingFN      = "disable-caching"
```
ryanfkeepers, line 12: I worry about this is overly generic, which makes it difficult to grok and potentially collides with other (maybe future) forms of caching.  Not sure there's some foo for `disable-foo` that won't run into similar problems in this case.  Perhaps some version of `force-foo` would communicate the intent more clearly.  `force-full` or `force-clean` or `force-new`.

ryanfkeepers (approved):

Issue #3998, opened 2023-08-09
[Bug]: Sanity tests sporadically fail on OD item permissions fetch


pandeyabs:
### What happened?

GET permissions fails if the item doesn't exist. This is due to a race condition between test run & purge scripts.
`getting item metadata: getting item metadata: error status code received from the API: Item not found:`

Fix here is to ignore GET permissions failure if it returns not found. This is safe to do here, since the graph call for permissions [looks up the item by id](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/drive.go#L254C17-L254C17), and not the permissions by id. Hence, we can safely assume that the item that this permission is tied to has been deleted.

### Corso Version?

Corso v0.11.1

### Where are you running Corso?

NA

PR #4001, opened 2023-08-09
Skip item if metadata fetch fails with 404

pandeyabs:
GET permissions fails if the item doesn't exist. We are hitting sanity test failures due to a race condition between test run & purge scripts.
`getting item metadata: getting item metadata: error status code received from the API: Item not found:`

Fix here is to ignore GET permissions failure if it returns not found. This is safe to do here, since the graph call for permissions [looks up the item by id](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/drive.go#L254C17-L254C17), and not the permissions by id. Hence, we can safely assume that the item that this permission is tied to has been deleted.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3998

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


meain (approved):

src/internal/m365/onedrive/collection.go
```diff
@@ -515,7 +515,14 @@ func (oc *Collection) populateDriveItem(
...
 	itemMeta, itemMetaSize, err = downloadItemMeta(ctx, oc.handler, oc.driveID, item)
 	if err != nil {
-		errs.AddRecoverable(ctx, clues.Wrap(err, "getting item metadata").Label(fault.LabelForceNoBackupCreation))
+		// Skip deleted items
+		if clues.HasLabel(err, graph.LabelStatus(http.StatusNotFound)) || graph.IsErrDeletedInFlight(err) {
+			logger.CtxErr(ctx, err).With("skipped_reason", fault.SkipNotFound).Info("item not found")
+			errs.AddSkip(ctx, fault.FileSkip(fault.SkipNotFound, oc.driveID, itemID, itemName, graph.ItemInfo(item)))
+		} else {
+			errs.AddRecoverable(ctx, clues.Wrap(err, "getting item metadata").Label(fault.LabelForceNoBackupCreation))
+		}
```
ryanfkeepers, line 524: Don't add skips for deleted items, please.  Skips should be reserved for non-transient cases where we deliberately ignore an existing item.  Race conditions with item deletions are transient, and does not pass over an existing item.

```suggestion
		if !clues.HasLabel(err, graph.LabelStatus(http.StatusNotFound)) && !graph.IsErrDeletedInFlight(err) {
			errs.AddRecoverable(ctx, clues.Wrap(err, "getting item metadata").Label(fault.LabelForceNoBackupCreation))
		}
```

ryanfkeepers (changes requested):

Issue #3993, opened 2023-08-08
Path extension to support multi-service expression


ryanfkeepers:
In order to support Teams design, the path design will need to undergo some revision to support service relationships.

Currently, the storage path assumes every entry contains only a single service.  IE: exchange, onedrive, or sharepoint- and with each of those, a single protected resource.  With teams we're introducing the idea of service and resource relationships.  Like how a Team contains a Site, or a Team chat is owned by a User.

Tracking these relationships requires us to change the way we handle and utilize paths, as well as some associated structures like selectors and reasons.

PR #4010, opened 2023-08-10
introduce service_resource tuple

ryanfkeepers:
the service resource tuple will form the basis for having paths contain multiple nested service
declarations, by allowing it to swap out the
Service() and ResourceOwner() funcs in favor of
a func that returns the ordered list of service-resource tuples found in the path.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3993

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/path/service_resource.go
+++ b/src/pkg/path/service_resource.go
@@ -1,0 +1,114 @@
+package path
+
+import (
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/common/str"
+	"github.com/alcionai/corso/src/internal/common/tform"
+)
+
+// ---------------------------------------------------------------------------
+// Tuple
+// ---------------------------------------------------------------------------
+
+// ServiceResource holds a service + resource tuple.  The tuple implies
+// that the resource owns some data in the given service.
+type ServiceResource struct {
+	Service           ServiceType
+	ProtectedResource string
+}
+
+func MakeServiceResource(
+	st ServiceType,
+	protectedResource string,
+) ServiceResource {
+	return ServiceResource{
+		Service:           st,
+		ProtectedResource: protectedResource,
+	}
+}
+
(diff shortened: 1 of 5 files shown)


src/pkg/path/service_resource.go
```diff
@@ -0,0 +1,114 @@
...
+// ---------------------------------------------------------------------------
+// Tuple
+// ---------------------------------------------------------------------------
+
+// ServiceResource holds a service + resource tuple.  The tuple implies
+// that the resource owns some data in the given service.
+type ServiceResource struct {
+	Service           ServiceType
+	ProtectedResource string
+}
```
ashmrtn, line 19: I'm a bit on the fence about this struct definition and some of the validation that occurs for `[]ServiceResource`s. The below are the main concerns I have around this (numbered for ease of reference, not importance):
1. all fields in the struct are public, is there a time when callers would ever choose to use `MakeServiceResource` instead of declaring it themselves?
2. what, if any, guarantees are we making around validity of a `ServiceResource` instance, especially since all fields are public? Does this mean it'll be possible for code to generate and pass around invalid instances? Is all code that interacts with a `ServiceResource` now responsible for ensuring that instance is valid in some way?
3. same as (2) but this time for `[]ServiceResource` and around checking that subservice(s) are valid for the ordering of the `ServiceResource`s in the slice

To draw some parallels, the strictness of the `Path` initializer provides guarantees about individual `Path` instances. If some part of the code is handed a `Path` instance then that code can assume that at least the `Path` has some basic characteristics (i.e. non-empty tenant and resource owner, valid `ServiceType`/`CategoryType`, at least one non-prefix segment)

I'm happy to discuss this offline if that would be simpler

ryanfkeepers:

PR #4015, opened 2023-08-10
distrubutes ServiceResource throughout path

ryanfkeepers:
Adds utilization of the ServiceResource struct to all of
the path package functionality.  In general, this takes
most instances where a service and resource
are requested, and replaces those values with a
slice of ServiceResource tuples.

To keep the review smaller, this change does not update
any packages outside of path.  Therefore the tests and
build are guaranteed to fail.  The next PR, which
updates all other packages with the changes, is
necessary for both PRs to move forward.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3993

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:

PR #4020, opened 2023-08-11
Add 'halves()' to path interface

ryanfkeepers:
Adds a func to the path interface which breaks up the path into its two distinct halves: the prefix builder, and the folder and item elements.  Likely to be useful in the future where prefixes will be of arbitrary length, and downstream consumers are primarly focused on the non-prefix portion of the path.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3993

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn:
have you run into a situation in the code where you've needed this? It seems like using functions to explicitly get some of the prefix elements and a call to `Folders()` would get the same result. I'm hesitant to lose the explicit accessors we have for the prefix elements (this returns a generic Builder instead)

ryanfkeepers:
> have you run into a situation in the code where you've needed this? It seems like using functions to explicitly get some of the prefix elements and a call to Folders() would get the same result.

@ashmrtn Only the one bit of test code that changed.  Though it _feels_ like something we'll use more of.  I could be wrong about that, though, and am happy to take it or leave it.

> I'm hesitant to lose the explicit accessors we have for the prefix elements (this returns a generic Builder instead)

Ah, good call.  Return a prefix-only path.Path instead?  (if we keep it)

ashmrtn:
>> have you run into a situation in the code where you've needed this? It seems like using functions to explicitly get some of the prefix elements and a call to Folders() would get the same result.
>
>Only the one bit of test code that changed. Though it feels like something we'll use more of. I could be wrong about that, though, and am happy to take it or leave it.

I'd prefer to leave it out unless we find ourselves needing it a lot in non-test code

>> I'm hesitant to lose the explicit accessors we have for the prefix elements (this returns a generic Builder instead)
>
>Ah, good call. Return a prefix-only path.Path instead? (if we keep it)

The above suggestion seems like it would be mostly useful if we need to print (or get the string of) just the prefix of the path? I'm not sure what else it really gets us apart from that since we can always just use individual accessor functions the the different prefix parts

ryanfkeepers:

Issue #4031, opened 2023-08-14
Point-in-time mode fails to list backups due to manifest compaction


ashmrtn:
Longevity tests are currently failing due to attempting to compact manifest blobs when opening the repo at a point in time. Setting the repo to read-only will likely solve this. The error occurs because kopia doesn't allow changes to repo state when using the point-in-time functionality

ashmrtn:
Further investigation reveals this could actually be an upstream issue as setting the repo to ReadOnly in the local config options still returns the same error

ashmrtn:
PRs with fixes pushed upstream. this issue is waiting for review and merge of upstream PRs and rebase of current kopia fork

Upstream PRs:
* kopia/kopia#3225
* kopia/kopia#3226

PR #4032, opened 2023-08-14
Print the point in time being used

ashmrtn:
When looking for recently deleted backups, print the point in time that kopia will be opened at so that if we need to manually debug something the value is right there.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4031

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cmd/longevity_test/longevity.go
+++ b/src/cmd/longevity_test/longevity.go
@@ -73,6 +73,12 @@ func pitrListBackups(
 	pitr time.Time,
 	backupIDs []string,
 ) error {
+	logAndPrint(
+		ctx,
+		"looking for %d backup(s) using timestamp %v\n",
+		len(backupIDs),
+		pitr)
+
 	if len(backupIDs) == 0 {
 		return nil
 	}


src/cmd/longevity_test/longevity.go
```diff
@@ -73,6 +73,12 @@ func pitrListBackups(
 	pitr time.Time,
 	backupIDs []string,
 ) error {
+	logAndPrint(
+		ctx,
+		"looking for %d backup(s) using timestamp %v\n",
```
ryanfkeepers, line 78: using/at/since?  backup/latest backup?
```suggestion
		"looking for %d backup(s) using timestamp %v\n",
```

ryanfkeepers (approved):

src/cmd/longevity_test/longevity.go
ashmrtn: timestamp is relative to when deletion happened not the time of the backup(s)

Issue #4019, opened 2023-08-11
Add batch delete function for backups


ashmrtn:
Right now backup deletion in corso is split between deleting snapshots (kopia wrapper) and backup models (model store). Internally, all of these objects are stored in kopia's manifest manager and can be deleted together. Doing so will help reduce the number of manifests objects that kopia writes

This idea can also be extended to delete multiple corso backups in a single go, further reducing the number of manifest blobs kopia writes

PR #4038, opened 2023-08-15
Don't fail deletion on missing backups

ashmrtn:
Allows ignoring missing backups during deletion by
passing in an addition bool param.

Since this changes the API, we can remove the final
commit to make ignoring missing backups the
default behavior. This will change CLI behavior
though

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4019

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/repository/repository.go
+++ b/src/pkg/repository/repository.go
@@ -633,13 +633,15 @@ func getBackupErrors(
 // DeleteBackups removes the backups from both the model store and the backup
 // storage.
 //
+// Missing models or snapshots during deletion do not cause errors.
+//
 // All backups are delete as an atomic unit so any failures will result in no
 // deletions.
 func (r repository) DeleteBackups(ctx context.Context, ids ...string) error {
 	return deleteBackups(ctx, store.NewWrapper(r.modelStore), ids...)
 }
 
-// deleteBackup handles the processing for Backup.
+// deleteBackup handles the processing for backup deletion.
 func deleteBackups(
 	ctx context.Context,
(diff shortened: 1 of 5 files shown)


vkamra (approved):

src/pkg/repository/repository.go
```diff
@@ -633,16 +633,26 @@ func getBackupErrors(
...
+) error {
+	return deleteBackups(ctx, store.NewWrapper(r.modelStore), failOnMissing, ids...)
 }
 
-// deleteBackup handles the processing for Backup.
+// deleteBackup handles the processing for backup deletion.
 func deleteBackups(
 	ctx context.Context,
 	sw store.BackupGetterModelDeleter,
+	failOnMissing bool,
```
ryanfkeepers, line 655: While I'm fine with this implementation, it does make me wonder if it's the most idiomatic.  Another way of handling batch processes like this that I've liked is to return an `[]error` of `len=len(ids)`.  Every entry in the returned slice relates to the same-index of the id that it processed, holding nil if no error occurred.  So for 3 ids, the response may be `[]error{nil, ErrNotFound, nil}`.  This way the caller can: 1/ easily associate errors with IDs. 2/ respond to each error holistically (what if we want to continue on some other error case as well?).

Just spitballing here.  Changes aren't necessary for this PR unless you want the refactor.

ryanfkeepers (approved):

PR #4046, opened 2023-08-16
remove all uses of iota

ryanfkeepers:
I've needed to catch gotchas that arise from contributors adding a value in the middle of an iota list, not to mention have dealt with prior bugs that happened the same way, now too many times to feel safe about its usage.

This PR removes the use of iota from all const declarations. The intent is to not allow the use of iota within the codebase.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3993

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn (commented):
Do we have a good way to do regression tests for these changes? Unfortunately I don't think our existing tests will catch these changes because they don't operate on backups from a previous version of corso

If we want to avoid iota in the future it's probably also a good idea to see if we can make a lint rule for it

ryanfkeepers:
> Do we have a good way to do regression tests for these changes?

What kind of regression tests are you looking for?  Just something to make sure all values are consistent across the change?  If so, I don't think I have any ideas.  I could add a set of unit tests.  But that'd be testing a const value which, in context of this change, makes sense, but is plain weird outside of this context.

ashmrtn:
looking at this PR a bit more I think there's two sorts of changes we're making:
* removal of in-memory-only iota values
* removal of persisted iota values

The in-memory-only ones are safe to merge. There's no previous state we need to worry about not matching against anymore.

>Just something to make sure all values are consistent across the change?

Yes, I'm looking for a way to gain confidence that the removal of persisted iota values doesn't break anything when attempting to make a new backup with a base created prior to this change or restore data from a backup made prior to this change. There's diffs for both the ServiceType and CategoryType string generators so it would be good to make sure nothing's gone awry there since we rely on the persisted string values for those according to the comment

ryanfkeepers:
> There's diffs for both the ServiceType and CategoryType string generators so it would be good to make sure nothing's gone awry there since we rely on the persisted string values for those according to the comment

Ah, you know what?  Those are unnecessary.  I'll just revert that part of the change.

src/pkg/backup/details/iteminfo.go
```diff
@@ -20,16 +20,17 @@ type ItemType int
...
 	// Exchange (00x)
-	ExchangeContact
-	ExchangeEvent
-	ExchangeMail
+	ExchangeContact ItemType = 1
+	ExchangeEvent   ItemType = 2
+	ExchangeMail    ItemType = 3
+
 	// SharePoint (10x)
-	SharePointLibrary ItemType = iota + 97 // 100
```
meain, line 30: I believe `SharePointLibrary` will be 101

ryanfkeepers: Arg, the comments are off!?

ryanfkeepers: yup, you were right.  Good catch!

src/internal/m365/collection/site/collection.go
```diff
@@ -26,14 +26,21 @@ import (
...
-	List
-	Drive
-	Pages
+	// Unknown is 2 due to a mishandling of iota.
+	// It should be 0, by standard, but now we're
+	// stuck with this for backwards compatibility.
+	Unknown DataCategory = 2
+	List    DataCategory = 3
+	Drive   DataCategory = 4
+	Pages   DataCategory = 5
```
ashmrtn, line 43: These are never persisted. It doesn't matter what their values are since there's no backwards compatibility issues we could encounter. They [appear](https://github.com/alcionai/corso/blob/8d3fdeeb8dbd8dc68284c1d33b0b431d309b4f5b/src/internal/m365/collection/site/collection.go#L200) to only be used to determine what data to retrieve (and `Drive` doesn't seem to be used at all?)

ashmrtn (approved):
would still really like to see some tests for this since there was an issue brought up

src/internal/m365/collection/site/collection.go
ryanfkeepers: mmkay, I removed them entirely in favor of path.CategoryTypes.

src/pkg/backup/details/iteminfo.go
ryanfkeepers: I've compared them betwen main and the branch.  Plus wrote a little unit test on each end that I deleted. They are, with absolute confidence, the same values.

PR #4068, opened 2023-08-18
channels and messages API

neha-Gupta1:
Message handler implementation and other APIs to fetch channels and messages data

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+// Channels is an interface-compliant provider of the client.
+type Channels struct {
+	Client
+}
+
+// ---------------------------------------------------------------------------
+// containers
+// ---------------------------------------------------------------------------
+
+// CreateContainer makes an channels with the name in the team
```
ryanfkeepers, line 34: Is this something we need right away?  Since we're going to start off with Export instead of Restore for channels, I think we can skip channel creation.

src/pkg/backup/details/iteminfo.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// Folder Management(30x)
 	FolderItem ItemType = 306
+
+	// GroupChannel(40x)
+	GroupChannel ItemType = 407
```
ryanfkeepers, line 41: Sanity check: if we go the route of treating channels as discrete items, then we should not plan to use them in the same way that we use Drives in onedrive and sharepoint.  If you'd prefer to handle channels the same way we handle drives, then we probably don't want a GroupChannel item type, in the same way that we don't have a Drive type.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+		Post(ctx, body, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "creating channel")
+	}
+
+	return ChannelsDisplayable{Channelable: container}, nil
+}
+
+// DeleteChannel removes a channel from user's M365 account
+func (c Channels) DeleteChannel(
```
ryanfkeepers, line 56: Similarly, do we expect to delete channels?  I can't think of any reason we'd want to do this, whether as part of backup or restore.  We might be better off without this.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+		ByChannelId(containerID).
+		Delete(ctx, nil)
+	if err != nil {
+		return graph.Stack(ctx, err)
+	}
+
+	return nil
+}
+
+// prefer GetChannelByID where possible.
```
ryanfkeepers, line 80: In the other cases where we've commented `prefer GetFooByID`, Foo is either Item or Container.  The comment effectively means "use the generic interface if you can".  (for that matter, we could probably get rid of one func by use of generics).  But since `GetChannelByID` is named for `Channel` and not Item or Container, there is no generic interface compliance, and so this function is unnecessary duplication.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+	container := ChannelsDisplayable{Channelable: cal}
+
+	if err := graph.CheckIDAndName(container); err != nil {
+		return nil, clues.Stack(err).WithClues(ctx)
+	}
+
+	return container, nil
+}
+
+func (c Channels) PatchChannel(
```
ryanfkeepers, line 159: What use cases do we have for patching a channel?  This might also be an endpoint we can remove.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+	}
+
+	return nil
+}
+
+// ---------------------------------------------------------------------------
+// message
+// ---------------------------------------------------------------------------
+
+// GetItem retrieves a Messageable item.
```
ryanfkeepers, line 183: Comments here are out of sync.  Please go through the file and make sure they match the data being used.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+func (c Channels) GetMessage(
+	ctx context.Context,
+	teamID, channelID, itemID string,
+	immutableIDs bool,
```
ryanfkeepers, line 187: Channels don't have immutable IDs.  This property can be removed.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+	}
+
+	if itm == nil {
+		return nil, clues.New("nil response mail message creation").WithClues(ctx)
+	}
+
+	return itm, nil
+}
+
+func (c Channels) DeleteMessage(
```
ryanfkeepers, line 235: We can probably skip delete for now as well, since we won't be restoring messages for v0/v1, and we won't need to handle restore collisions.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+
+	return nil
+}
+
+// ---------------------------------------------------------------------------
+// replies
+// ---------------------------------------------------------------------------
+
+// GetReplies retrieves a Messageable item.
+func (c Channels) GetReplies(
```
ryanfkeepers, line 267: Will this need a pager implementation?

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+	}
+
+	if itm == nil {
+		return nil, clues.New("nil response reply to message creation").WithClues(ctx)
+	}
+
+	return itm, nil
+}
+
+func (c Channels) DeleteReply(
```
ryanfkeepers, line 314: Similar to messages, I don't think we'll need to delete replies at this time.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+	}
+
+	return nil
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+func MessageInfo(msg models.ChatMessageable, size int64) *details.GroupsInfo {
```
ryanfkeepers, line 347: suggestion: namespacing this func helps reduce collision between this and other types of "messages".  The other funcs are methods on the Channel client, so they're already namespaced.
```suggestion
func ChannelMessageInfo(msg models.ChatMessageable, size int64) *details.GroupsInfo {
```

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// helper funcs
+// ---------------------------------------------------------------------------
+
+// ChannelsDisplayable is a wrapper that complies with the
+// models.Channelable interface with the graph.Container
+// interfaces.
+type ChannelsDisplayable struct {
```
ryanfkeepers, line 367: Something of a conflict of interests here.  Displayable is a concept we use for containers, not for items or drives.  This is related to the other comment checking whether we want to treat the container as an item or as a drive.  If we're going to match it to a displayable, it sounds like we want it to be a `Container`.  If we want it to be an Item, or similar to a drive, I don't think we want this compliance.

And, for that matter, I'm not sure this interface is valuable outside of the Exchange collection processor anyway.  I'd say see if we can cut it out, and only bring it back if necessary.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,376 @@
...
+		ByChatMessageId(itemID).
+		Get(ctx, nil)
+	if err != nil {
+		return nil, nil, graph.Stack(ctx, err)
+	}
+
+	return message, MessageInfo(message, size), nil
+}
+
+func (c Channels) PostMessage(
```
ryanfkeepers, line 211: Hmm, do we need `PostMessage` and `PostReply`?  We won't be restoring these values yet, only exporting, which is the primary purpose of having these funcs.  I suppose the question is whether we need them for testing purposes (and delete, too)?  Maybe.  Maybe not.

I'd suggest that for this PR specifically you can drop both funcs.  Focus only on what's necessary for backup.  We can come back and re-add them again if necessary.

ryanfkeepers (changes requested):
Lots of YAGNI to cut down on.

src/internal/tester/tconfig/config.go
```diff
@@ -45,6 +46,7 @@ const (
 	EnvCorsoM365TestSiteURL          = "CORSO_M365_TEST_SITE_URL"
 	EnvCorsoM365TestTeamID           = "CORSO_M365_TEST_TEAM_ID"
 	EnvCorsoM365TestGroupID          = "CORSO_M365_TEST_GROUP_ID"
+	EnvCorsoM365TestChannelID        = "CORSO_M365_TEST_CHANNEL_ID"
```
ryanfkeepers, line 49: I don't think it's in our interest to use these envs for any entity more granular than a resource (user, site, etc).  This is basically like using these envs to specify containers.  We already have selectors as a system to handle that granularity, and prior work has shown that looking for expected containers by name (ex: a site drive named "documents") is an effective management strategy.  Adding that identification here could both couple test expectations together unnecessarily, and increase the complexity and maintenance of baseline testing.

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -2,38 +2,143 @@ package api
 
 import (
 	"context"
+
```
ryanfkeepers, line 5: This PR is already getting a little to large to fully review.  Can you please move the pagers implementation into a follow-up PR?  In fact, any addition of new code on top of the original PR can be moved to a follow-up PR, for ease of review.  We can flatten them all before merging to main if extra assurance is needed.

src/pkg/backup/details/iteminfo.go
neha-Gupta1: Agreed. It should be of `chatMessage`  type in that case. Changing it. Let me know if that fits best.

src/internal/tester/tconfig/config.go
neha-Gupta1: okay. In that case we can safely remove the test setup for channels. And use the name to find channels if needed in test cases.

src/pkg/services/m365/api/channels.go
neha-Gupta1: We can add this implementation if required in the PR with pager.

neha-Gupta1:
@ryanfkeepers please let me know other reviews on this PR. I will move forward with other PRs once this code is stabilised. it will save re-basing and confusions due to changes here.

src/internal/m365/collection/groups/handler.go
```diff
@@ -0,0 +1,17 @@
+package groups
```
ryanfkeepers, line 1: use the `handlers.go` (with the 's') file please.

src/pkg/services/m365/api/channels.go
```diff
@@ -1 +1,194 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// replies
+// ---------------------------------------------------------------------------
+
+// GetReplies retrieves a Messageable item.
+func (c Channels) GetReplies(
+	ctx context.Context,
+	teamID, channelID, itemID string,
```
ryanfkeepers, line 138:
```suggestion
	teamID, channelID, messageID string,
```

ryanfkeepers (approved):
Test coverage is insufficient.  Going to approve this to allow it to merge, but all of these funcs need test coverage as a fast follow.

neha-Gupta1:
Yes. Still missing test-cases for messages and replies. I will add that in coming PRs.

Issue #3217, opened 2023-04-25
Cleanup incomplete/failed Corso backup state


vkamra:
Incomplete/Failed Corso backups can leave behind Kopia snapshots in the repositories. 

These are left behind because follow-up backup operations can leverage the existing data in Kopia snapshots.

We need to implement a mechanism to discover this state and clean it up.

PR #4081, opened 2023-08-21
Cleanup old assist bases

ashmrtn:
Add helper functions and tests that find
old assist bases and add them to the set
of backups to garbage collect. This
allows us to remove backups that aren't
displayed to the user because they had
some errors during them

The more recent assist backups can still
be used for incremental backups so we
don't want to garbage collect all of
them

This still does not wire this code into
any existing corso function, just adds
additional cleanup logic

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3217

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/cleanup_backups_test.go
```diff
@@ -529,6 +625,173 @@ func (suite *BackupCleanupUnitSuite) TestCleanupOrphanedData() {
 			buffer:    24 * time.Hour,
 			expectErr: assert.NoError,
 		},
+		// Tests dealing with assist base cleanup.
+		{
+			name: "AssistBase NotYoungest InBufferTime Noops",
```
pandeyabs, line 630: Given these tests are testing some complex code behavior, could you please add some comments for each test for more context?

pandeyabs: Also, can we add a test which has both assist & merge bases? Just to establish the cleanup behavior that we are always going to keep the most recent assist base, even if it's superseded by a merge base.

pandeyabs (approved):
LGTM! This is very well thought out!

src/internal/kopia/cleanup_backups.go
```diff
@@ -114,8 +126,11 @@ func cleanupOrphanedData(
 		return clues.Wrap(err, "getting all backup models")
 	}
 
-	toDelete := maps.Clone(deets)
-	maps.Copy(toDelete, dataSnaps)
+	// assistBackups is the set of backups that have a
+	//   * a label denoting their an assist backup
```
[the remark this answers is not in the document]
ashmrtn: oh whoops, guess I should proof read things more often

src/internal/kopia/cleanup_backups_test.go
ashmrtn: >can we add a test which has both assist & merge bases

Good callout! I've added a new test for that and added comments to all the assist base tests

PR #4084, opened 2023-08-22
Simplify group_handler by leveraging library_handler

meain:
https://github.com/alcionai/corso/pull/4030#discussion_r1296246853

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/collection/drive/library_handler.go
```diff
@@ -128,6 +133,7 @@ func (h libraryBackupHandler) NewLocationIDer(
 	driveID string,
 	elems ...string,
 ) details.LocationIDer {
+	// TODO(meain): path related changes for groups
```
ryanfkeepers, line 136: Can you tell me about how you imagine the groups' path changes might affect the location path here?

Separately, does the TODO make sense as something internal to this func, or should it go into `group_handler` with a todo to override this func?

src/internal/m365/collection/drive/library_handler.go
```diff
@@ -313,18 +321,37 @@ func augmentLibraryItemInfo(
...
-		DriveName:  driveName,
-		ItemName:   ptr.Val(item.GetName()),
-		ItemType:   details.SharePointLibrary,
-		Modified:   ptr.Val(item.GetLastModifiedDateTime()),
-		Owner:      creatorEmail,
-		ParentPath: pps,
-		SiteID:     siteID,
-		Size:       size,
-		WebURL:     weburl,
+	switch path.SharePointService {
```
ryanfkeepers, line 324: Interesting switch conditional you have there.

Kinda feels like this func should move into a different file, since it's servicing two different handlers now?  Maybe in that third file you could hook in the standard item_handler augment as well.  The only reason there were separate functions previously was because the augment funcs existed in different packages.

src/internal/m365/service/sharepoint/backup.go
```diff
@@ -80,7 +80,7 @@ func ProduceBackupCollections(
 			spcs, canUsePreviousBackup, err = site.CollectLibraries(
 				ctx,
 				bpc,
-				drive.NewLibraryBackupHandler(ac.Drives(), scope),
+				drive.NewLibraryBackupHandler(ac.Drives(), scope, path.SharePointService),
```
ryanfkeepers, line 83: Nit:  and could probably do this other places, too.
```suggestion
				drive.NewLibraryBackupHandler(ac.Drives(), scope, bpc.Selector.PathService()),
```

ryanfkeepers (approved):

src/internal/m365/collection/drive/library_handler.go
meain: Hmm, I guess this might not require a change according to the current path change design. The change would have been to create a `NewSharePointLocationIDer` for groups and not necessarily this function. I mostly added in the TODO there as I was kinda unsure.

PR #4087, opened 2023-08-22
Shuffle around logic for details merging

### LINKED ISSUE #3716: [Bug]: Exchange backup fails if we can't enumerate service folders
### What happened?

The case I investigated we got a 404 when attempting to get the user's calendars. This eventually caused the backup in CI to [fail](https://github.com/alcionai/corso/actions/runs/5406562647/jobs/9823595921#step:7:268) when merging backup details with the following error
```text
tester.go:39: TestBackupOpIntegrationSuite/TestBackup_Run_incrementalExchange/delete_a_folder run at 2023-06-28T23:04:14.19626645Z
      backup_integration_test.go:1273: 
          	Error Trace:	/home/runner/work/corso/corso/src/internal/operations/backup_integration_test.go:1273
          	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:112
          	Error:      	Received unexpected error:
          	            	empty location key
          	            		/home/runner/work/corso/corso/src/internal/kopia/merge_details.go:81
          	            	getting new paths
          	            		/home/runner/work/corso/corso/src/internal/operations/backup.go:637
          	            	getting updated info for entry
          	            		/home/runner/work/corso/corso/src/internal/operations/backup.go:747
          	            	merging details
          	            		/home/runner/work/corso/corso/src/internal/operations/backup.go:357
          	            	running backup
          	            		/home/runner/work/corso/corso/src/internal/operations/backup.go:226
          	Test:       	TestBackupOpIntegrationSuite/TestBackup_Run_incrementalExchange/delete_a_folder
```

### Fix
Always attempt to parse the location info from the old entry prior to lookup. Even for older backups we should have the invariant that either the LocationRef is non-empty or the RepoRef contains enough information to generate the old LocationRef

Priority of this is a bit unclear. Technically the backup should still fail overall because we couldn't backup something (calendars in this case), but properly handling this case can lead to better error messages in the long run and better best-effort behavior

### Analysis
Connecting to the repo and dumping the details for the backup immediately before the failing one shows that the details entry has all the expected information. However, we appear to be failing on getting the information when backing up info. Not being able to get the info causes us to fail to merge later because we don’t have enough to do the lookup

Inspecting the logs for the failing backup shows a 404 when getting calendars (info trimmed and formatted for legibility)
```
2023-06-28T23:04:24.537Z  ERROR graph/middleware.go:176 graph api error: 404 Not Found  {
	"service": "Exchange",
	"category": "events",
	"resp_status_code": "404",
	"resp_content_len": "-1",
	"resp_status": "404 Not Found",
	"method": "GET",
	"url": "https://graph.microsoft.com/v1.0/users/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee/calendars?$select=id,name&$skip=10&$top=10",
	"request_content_len": "0",
	"response": "HTTP/1.1 404 Not Found
{
	\"error\":{
		\"code\":\"ErrorItemNotFound\",
		\"message\":\"The specified object was not found in the store.\",
		\"innerError\":{
			\"date\":\"2023-06-28T23:04:24\",
			\"request-id\":\"37278584-159f-4ce3-b91c-dc0db4b55ae6\",
			\"client-request-id\":\"04ab5101-d45e-4d7d-99a4-b3249058436c\"
		}
	}
}"
}
```

The rest of the backup appears to proceed normally from there, but the backup still fails overall during details merging. The root cause is the assumption that all exchange folders will generate collections even if they aren’t changed. Having a collection allows us to get the [location](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/kopia/upload.go#L543) of the folder even if it’s merged. The location is then [passed](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/kopia/upload.go#L557) to `streamBaseEntries` and added as part of the [information](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/kopia/upload.go#L499) that’s eventually [propagated](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/kopia/upload.go#L204) to the merge handler when `FinishedFile` is called.

When looking up items for merging, we don’t attempt to extract location from Exchange right now. This leads to us calling `GetNewPathRef` with an empty old location. Overall, it causes the [if-block](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/kopia/merge_details.go#L78-L82) in the lookup code to return an error because the information that we found has no location and no location was passed in for the lookup.

The [merge code](https://github.com/alcionai/corso/blob/7795f58c30c8b04eb8ad3cf5c73896f53d297820/src/internal/operations/backup.go#L627) is currently setup to not attempt parsing a LocationRef in all cases. I don’t believe there’s really a reason not to attempt it. It should be possible to define the set of backups for which there will exist a LocationRef or from which we can parse the location from the RepoRef, much like we’re doing for OneDrive and SharePoint. That alone should solve the issue.

I believe the parsing checks would go something like the following:

- calendar
    - < backup version 2 -> parse from RepoRef
    - >= backup version 2 -> source from LocationRef
- email and contacts
    - attempt to parse from LocationRef
    - if LocationRef is empty attempt to parse from RepoRef

The above is based on the fact that for email and contacts the following changes were made:

- support for populating LocationRef was added in #2829 (2023-04-03)
- switch to folder IDs was added in #3373 (2023-05-11)

This means that at all times either the LocationRef or the RepoRef for email and contacts should contain what we want for the LocationRef for unchanged items. Furthermore, since what will be used as the LocationRef is properly formatted (i.e. proper directory structure) when found in the RepoRef we don’t have to worry about additional code there. Also, no contacts are stored directly in the `contacts` prefix folder of the kopia hierarchy, so we shouldn’t have to worry about the LocationRef ever being empty unexpectedly

We could probably just treat calendars the same as well as either LocationRef will be empty or populated and I don't believe there was a time in-between where we had IDs in the RepoRef but empty LocationRef. Nor will events ever appear outside a calendar in the folder hierarchy

### Corso Version?

any after ~May

N/A

### Relevant log output

_No response_

ashmrtn:
Shuffle around some logic for details merging so that
we always attempt to extract a LocationRef from the
backup base entry that's currently being examined.

A LocationRef should always be available from either
the LocationRef field in the details entry (newer
backups) or by extracting it from the RepoRef (older
backups)

Manually tested incremental backups for exchange in
the following scenarios:
1. v0.3.0 backup (calendars use IDs in RepoRef) ->
   incremental with this patch -> incremental with
   this patch
1. v0.2.0 backup (exchange uses folder names in
   RepoRef) -> incremental with this patch ->
   incremental with this patch

The above tests should cover the cases where:
* base backup details don't have LocationRef for
  exchange items
* base backup details have LocationRef for exchange
  items

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3716

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/operations/backup_test.go
```diff
@@ -606,6 +606,25 @@ func (suite *BackupOpUnitSuite) TestBackupOperation_MergeBackupDetails_AddsItems
...
+			[]string{
+				tenant,
+				path.ExchangeService.String(),
+				ro,
+				path.EmailCategory.String(),
+				"work",
+				"item1",
+			},
+			true,
+		)
```
ryanfkeepers, line 621:
```suggestion
			true)
```

src/pkg/backup/details/entry.go
```diff
@@ -68,32 +71,48 @@ func (de Entry) ToLocationIDer(backupVersion int) (LocationIDer, error) {
...
+	case ExchangeEvent:
+		if backupVersion >= 2 {
+			return nil, clues.New("no previous location for calendar entry").
+				With("repo_ref", rr)
+		}
+
+		fallthrough
+	case ExchangeMail:
+		fallthrough
+	case ExchangeContact:
```
ryanfkeepers, line 92:
```suggestion
	case ExchangeMail, ExchangeContact:
```

src/pkg/backup/details/entry.go
```diff
@@ -68,32 +71,48 @@ func (de Entry) ToLocationIDer(backupVersion int) (LocationIDer, error) {
...
+		baseLoc = path.Builder{}.Append(rr.Folders()...)
+
+	case OneDriveItem:
+		fallthrough
+	case SharePointLibrary:
```
ryanfkeepers, line 97:
```suggestion
	case OneDriveItem, SharePointLibrary:
```

ryanfkeepers (approved):

src/pkg/backup/details/entry.go
ashmrtn: ahhh I always forget you can do that for some reason haha

Issue #3989, opened 2023-08-08
Teams Channels Backup


ryanfkeepers:
# Work
## Packages

- [ ]  `/src/pkg/<provider>/collections/<processor>.go` to handle category-specific processing.

## Interface & Data Compliance

- [x]  `/src/pkg/path/resource_path.go` - canonical enums for representing each category.
- [x]  `/src/pkg/selectors/<service>.go` - scope and const additions for the data category.
- [x]  `/src/pkg/backup/details/<service>.go` - details iteminfo representation for the category.
- [x] RepoRef and LocationRef layout is completed with examples.

## Enumeration

- [x]  All containers.
    - [ ]  ~~May want a container cacher implementation.~~
- [x]  Items-in-container.
- [x]  Items-in-container by delta.
- [x]  All-items (if per-container delta is not supported).

## Backup v0

- [x]  The service can backup all data in a given category.
- [x]  The backup process can reduce the data it selects for backup by using selectors (necessary for testing).
- [x]  The backup path is complete and stable.
    - [x]  `RepoRef` describes the location of the data according to the storage provider.
    - [x]  `Path Locations` describes the human representation of each item’s location, as opposed to the storage path (repoRef).
- [x]  `Backup Details` entries contains a representation for that service, and includes the minimum necessary metadata for the data category.

## Backup v1

- [x]  Incremental Support
    - [x]  Does the category support `delta tokens`?  If not, does an alternative method exist to get the “latest additions and removals”
    - [x]  `Metadata` production around delta tokens and current kopia storage paths.
    - [x]  `Global Excludes` map to detail the deletion of items which are not tracked by container.

PR #4093, opened 2023-08-23
add channels info filters to selectors

ryanfkeepers:
Adds groups channel message filters to group selectors. Includes scopes for message creation time, last response time, and message created by.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/selectors/groups.go
```diff
@@ -334,9 +404,16 @@ const (
...
+	// channel and drive selection
 	GroupsInfoSiteLibraryDrive groupsCategory = "GroupsInfoSiteLibraryDrive"
 	GroupsInfoChannel          groupsCategory = "GroupsInfoChannel"
+
+	// data contained within details.ItemInfo
+	GroupsInfoChannelMessageCreatedAfter    groupsCategory = "GroupsInfoChannelMessageCreatedAfter"
+	GroupsInfoChannelMessageCreatedBefore   groupsCategory = "GroupsInfoChannelMessageCreatedBefore"
+	GroupsInfoChannelMessageCreator         groupsCategory = "GroupsInfoChannelMessageCreator"
+	GroupsInfoChannelMessageLastReplyAfter  groupsCategory = "GroupsInfoChannelMessageLastReplyAfter"
+	GroupsInfoChannelMessageLastReplyBefore groupsCategory = "GroupsInfoChannelMessageLastReplyBefore"
```
neha-Gupta1, line 416: In future we may be able to add more selectors like - filter on basis of 
- messages with events
- members in channel

src/pkg/selectors/groups.go
```diff
@@ -309,7 +310,76 @@ func (s *groups) LibraryItems(libraries, items []string, opts ...option) []Group
 // -------------------
 // ItemInfo Factories
 
-// TODO
+// MessageCreator produces one or more groups channelMessage info scopes.
+// Matches any channel message created by the specified user.
+// If any slice contains selectors.Any, that slice is reduced to [selectors.Any]
+// If any slice contains selectors.None, that slice is reduced to [selectors.None]
+// If any slice is empty, it defaults to [selectors.None]
+func (s *GroupsRestore) MessageCreator(creator string) []GroupsScope {
```
meain, line 318: Following from https://github.com/alcionai/corso/pull/4092#discussion_r1302921568 , do we need separate creator and owner?

src/pkg/selectors/groups.go
```diff
@@ -309,7 +310,76 @@ func (s *groups) LibraryItems(libraries, items []string, opts ...option) []Group
...
+			[]string{creator},
+			filters.In),
+	}
+}
+
+// MessageCreatedAfter produces a channel message created-after info scope.
+// Matches any message where the creation time is after the timestring.
+// If the input equals selectors.Any, the scope will match all times.
+// If the input is empty or selectors.None, the scope will always fail comparisons.
+func (s *GroupsRestore) MessageCreatedAfter(timeStrings string) []GroupsScope {
```
meain, line 332: I haven't added `CreatedAfter` and other funcs for Libraries, but was wondering if we need this filter separately for different services.

src/pkg/selectors/groups.go
ryanfkeepers: True, but no need to get ahead of ourselves.  We haven't identified a use case for these properties yet.  When we figure out the UX needed, we can revisit their addition.  The properties I've added are table stakes, as identified by Georgi and I in conversation.

src/pkg/selectors/groups.go
ryanfkeepers: See reply to that comment.  I think I'm going to keep with the separation, unless there's strong feelings otherwise.

src/pkg/selectors/groups.go
ryanfkeepers: You're welcome to try and consolidate the interanls if you want, but I don't think there's much left to reduce/reuse.

The primary reason for duplicating the funcs themeselves (ie: CreatedAfter) is that we want the selector to provide a curated list of scopes that are only applicable to that specific resource.  And, of course, that we need to provide selector-type-specific consts to the underlying scope.  Providing a generic "CreatedAfter" scope wouldn't allow for either of those.

ryanfkeepers:
@neha-Gupta1 @meain I'm not seeing any comments that indicate blocking issues.  Is this PR ready for approval?

src/pkg/selectors/groups.go
meain: Hmm, my concern was mostly around the UX. When restoring messages(which I know is not possible as of now), should we have `MessageCreatedAfter` override `CreatedAfter` as we will have to restore the attachments referenced in the message?

meain (approved):

src/pkg/selectors/groups.go
ryanfkeepers: Ah, I grok.  No, let's not do that at this time.  Per prior sync with Georgi, we're not going to restore attachments at the same time as messages.  It's fine to keep them separated.

Issue #3990, opened 2023-08-08
Teams Libraries Backup


ryanfkeepers:
# Work
## Packages

- [ ]  `/src/pkg/<provider>/<service>/<collection-processor>.go` to handle category-specific processing.

## Interface & Data Compliance

- [ ]  `/src/pkg/path/resource_path.go` - canonical enums for representing each category.
- [ ]  `/src/pkg/selectors/<service>.go` - scope and const additions for the data category.
- [ ]  `/src/pkg/backup/details/<service>.go` - details iteminfo representation for the category.
- [ ] RepoRef and LocationRef layout is completed with examples.

## Enumeration

- [ ]  All containers.
    - [ ]  May want a container cacher implementation.
- [ ]  Items-in-container.
- [ ]  Items-in-container by delta.
- [ ]  All-items (if per-container delta is not supported).

## Backup v0

- [ ]  The service can backup all data in a given category.
- [ ]  The backup process can reduce the data it selects for backup by using selectors (necessary for testing).
- [ ]  The backup path is complete and stable.
    - [ ]  `RepoRef` describes the location of the data according to the storage provider.
    - [ ]  `Path Locations` describes the human representation of each item’s location, as opposed to the storage path (repoRef).
- [ ]  `Backup Details` entries contains a representation for that service, and includes the minimum necessary metadata for the data category.

## Backup v1

- [ ]  Incremental Support
    - [ ]  Does the category support `delta tokens`?  If not, does an alternative method exist to get the “latest additions and removals”
    - [ ]  `Metadata` production around delta tokens and current kopia storage paths.
    - [ ]  `Global Excludes` map to detail the deletion of items which are not tracked by container.

PR #4095, opened 2023-08-23
Handle groups without SharePoint sites 

meain:
Security groups will not have an associated site. This avoids trying
to backup the site if we get an error that the site is not available.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/groups/backup.go
+++ b/src/internal/m365/service/groups/backup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
+	"golang.org/x/exp/slices"
 
 	"github.com/alcionai/corso/src/internal/common/idname"
 	"github.com/alcionai/corso/src/internal/common/prefixmatcher"
@@ -20,6 +21,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/backup/metadata"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
@@ -50,6 +52,27 @@ func ProduceBackupCollections(
 		ctx,
 		"group_id", clues.Hide(bpc.ProtectedResource.ID()),
 		"group_name", clues.Hide(bpc.ProtectedResource.Name()))
+
+	resp, err := ac.Groups().GetByID(ctx, bpc.ProtectedResource.ID())
+	if err != nil {
+		return nil, nil, false, clues.Wrap(err, "getting group").WithClues(ctx)
+	}
+
+	// Not all groups will have associated SharePoint
+	// sites. Distribution channels and Security groups will not
+	// have one. This check is to skip those groups.
+	groupTypes := resp.GetGroupTypes()
+	hasSharePoint := slices.Contains(groupTypes, "Unified")
+
+	// If we don't have SharePoint site, there is nothing here to
+	// backup as of now.
+	if !hasSharePoint {
+		logger.Ctx(ctx).
+			With("group_id", bpc.ProtectedResource.ID()).
+			Infof("No SharePoint site found for group")
+
+		return nil, nil, false, clues.Stack(graph.ErrServiceNotEnabled, err).WithClues(ctx)
+	}
 
 	for _, scope := range b.Scopes() {
 		if el.Failure() != nil {


src/internal/m365/service/groups/backup.go
```diff
@@ -66,6 +67,15 @@ func ProduceBackupCollections(
 			// default one.
 			resp, err := ac.Groups().GetRootSite(ctx, bpc.ProtectedResource.ID())
 			if err != nil {
+				if clues.HasLabel(err, graph.LabelsNoProvisionResource) {
+					logger.Ctx(ctx).With("group_id", bpc.ProtectedResource.ID()).Infof("No SharePoint site found for group")
+
+					// TODO(meain): how should we handle a subservice missing?
```
meain, line 73: What exactly should we do here?

src/internal/m365/service/groups/backup.go
```diff
@@ -66,6 +67,15 @@ func ProduceBackupCollections(
 			// default one.
 			resp, err := ac.Groups().GetRootSite(ctx, bpc.ProtectedResource.ID())
 			if err != nil {
+				if clues.HasLabel(err, graph.LabelsNoProvisionResource) {
+					logger.Ctx(ctx).With("group_id", bpc.ProtectedResource.ID()).Infof("No SharePoint site found for group")
```
ryanfkeepers, line 71:
```suggestion
					logger.Ctx(ctx).
					    With("group_id", bpc.ProtectedResource.ID()).
					    Infof("No SharePoint site found for group")
```

src/internal/m365/service/groups/backup.go
ryanfkeepers: Continue is fine.  A switch or set of conditions will be needed in the future, since different groups may have access to multiple different systems.  But the current design process you're working towards (error now, continue later, more after that) is a great approach.

meain: Cool, I've just updated the comment for now.

src/internal/m365/graph/errors.go
```diff
@@ -62,16 +62,18 @@ const (
...
 const (
-	IOErrDuringRead   errorMessage = "IO error during request payload read"
-	MysiteURLNotFound errorMessage = "unable to retrieve user's mysite url"
-	MysiteNotFound    errorMessage = "user's mysite not found"
-	NoSPLicense       errorMessage = "Tenant does not have a SPO license"
+	IOErrDuringRead         errorMessage = "IO error during request payload read"
+	MysiteURLNotFound       errorMessage = "unable to retrieve user's mysite url"
+	MysiteNotFound          errorMessage = "user's mysite not found"
+	NoSPLicense             errorMessage = "Tenant does not have a SPO license"
+	UnableProvisionResource errorMessage = "unable to provision resource" // happens if group sharepoint site not available
```
ryanfkeepers, line 69:
```suggestion
	UnableToProvisionResource errorMessage = "unable to provision resource" // happens if group sharepoint site not available
```

src/internal/m365/graph/errors.go
```diff
@@ -62,16 +62,18 @@ const (
...
 )
 
 const (
 	LabelsMalware             = "malware_detected"
 	LabelsMysiteNotFound      = "mysite_not_found"
 	LabelsNoSharePointLicense = "no_sharepoint_license"
+	LabelsNoProvisionResource = "no_provision_resource"
```
ryanfkeepers, line 76: nit: Wording here is a bit clunky.
```suggestion
	LabelsCannotProvisionResource = "cannot_provision_resource"
```

ryanfkeepers (commented):
I feel uneasy about this approach.  We're looking for a clear identifier about whether a group contains a site or not, and this error isn't a very clear identifier.  Given that we're reacting to an error instead of looking for declarative state, we can't guarantee that we won't see this issue crop up in transient ways due to graph server issues.  Further, since the only usable part of the error is its message, we have weak guarantees about the consistency of responses, and thus the consistency of identification.

Before we go ahead with this design, I'd like you to do a review of the data that's available to site-capable and non-site-capable group resources to ensure we don't have any other way to detect the state.  One place to start with is the `unified` tag in the `groupType` property.  Thatt seems to already coincide with whether a group has access to conversations or not, and is the identifier we plan to use for said data collection set.  It may indicate the same for sites.  Or, if not, there may be some other non-error indicator to use.

meain:
@ryanfkeepers I could use a re-review on this whenever you have time.

src/internal/m365/service/groups/backup.go
```diff
@@ -47,6 +48,35 @@ func ProduceBackupCollections(
...
+		if gt == "Unified" {
+			hasSharePoint = true
+			break
+		}
+	}
+
+	// If we don't have SharePoint site, there is nothing here to
+	// backup as of now.
+	// TODO: Update when we have more than just SharePoint sites
+	if !hasSharePoint {
```
ryanfkeepers, line 72:
```suggestion
	if slices.Contains(groupTypes, "Unified") {
	// or, if you want to not worry about case sensitivity
	if filters.Equal(groupTypes).Compare("Unified") {
```

ryanfkeepers (approved):

Issue #3844, opened 2023-07-18
Add service level isolation


pandeyabs:
### What happened?

While getting user info during backup, we try to [discover](https://github.com/alcionai/corso/blob/22f990a709b996a1f1783ec81ebb909430a79111/src/pkg/services/m365/api/users.go#L174C18-L174C18) all services which are enabled for the user. 

This approach has potential drawbacks. For example, assume that the user intends to an exchange backup. The backup may fail during service discovery if the user does not have a onedrive or if corso doesn't have file permissions. While we do handle such scenarios gracefully, we have noticed that graph api may change error messages on us, leading to backup failures. Additionally, it's also the right thing to do, to be transparent to the user.

Proposed changes:
1. Corso should only discover & enable requested services. If the user is asking to do exchange backups, we should not attempt onedrive/sharepoint discovery.
2. This can be taken to another level by forcing isolation within a service (e.g. calendar/mail/contacts in exchange). Although this may not always be possible. For e.g. corso may need to verify that the user has an existing mailbox, before backing calendars.

### Corso Version?

Corso v0.10.0

### Where are you running Corso?

Linux

ryanfkeepers:
Additional request: as we're investigating the isolation of per-service access checks, can we include checks for appropriate isolation in the cli layer as well?

pandeyabs:
Thanks @ryanfkeepers. Sounds good!

PR #4096, opened 2023-08-23
Add service isolation

pandeyabs:
1. Renamed `IsBackupRunnable` to `IsServiceEnabled` & extended to restore operation. Removed `checkServiceEnabled`. Reasons:
- The above 2 functions were doing the same thing. Removed `checkServiceEnabled` in favor of `IsBackupRunnable` since we want this check to be as soon as possible during backup/restore op initialization
- Renamed `IsBackupRunnable` to `IsServiceEnabled`, because a) we are only doing service enabled checks right now, b) common code that can be used for both restore & backup.

2. Wire corso code to use new helpers in internal/m365/common.go
3. Note: SDK wiring and related integ tests will be added in a follow up PR

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3844

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/backup.go
```diff
@@ -204,35 +203,81 @@ func verifyBackupInputs(sels selectors.Selector, siteIDs []string) error {
...
 }
 
-type getInfoer interface {
-	GetInfo(context.Context, string) (*api.UserInfo, error)
-}
-
 func checkServiceEnabled(
 	ctx context.Context,
-	gi getInfoer,
+	ctrl *Controller,
```
ryanfkeepers, line 208: This is a bit overbearing, since y7ou're still only using `ctrl.AC.foo()`.  I'd suggest passing along the api client instead of the full controller.  That way you can still use `ac.Users()` and `ac.Sites()` as needed.
```suggestion
	ac api.Client,
```

src/pkg/services/m365/api/users.go
```diff
@@ -283,6 +284,22 @@ func (c Users) getMailboxSettings(
 	return settings, nil
 }
 
+func (c Users) IsExchangeServiceEnabled(
```
ryanfkeepers, line 287: This is effectively a duplicate of the service-layer [UserHasMailbox()](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/users.go#L63) func.  Can we use that func instead of adding a new one?

As a design philosophy: more interperative functionality like this (ie, calling some api and transforming the results according to a given interpretation) are better located in the service layer, not the api layer.  The API layer is an owner for each specific calls itself.  Not wrappers on those calls.  We're not perfect on this, but that's the ideal.

That would ultimately mean we'd want to move GetMailInbox to the service layer.  Granted, that can feel a little weird when corso itself wants to consume the func, since the service layer is SDK oriented.  Hmm.  Maybe the best middle ground is to remove this func, keep with UserHasMailbox in the service layer, have corso call GetMailInbox, and better integrate EvaluateMailboxError into GetMailInbox.  That would set us up to remove this intermediate func.

src/pkg/services/m365/api/users.go
```diff
@@ -301,6 +318,62 @@ func (c Users) GetMailInbox(
...
+) (MailboxInfo, error) {
+	mailFolderFound := true
+	// check whether the user is able to access their inbox.
+	// if they cannot, we can assume they are ineligible for exchange backups.
+	inbx, err := c.GetMailInbox(ctx, userID)
+	if err != nil {
+		if err := EvaluateMailboxError(graph.Stack(ctx, err)); err != nil {
+			logger.CtxErr(ctx, err).Error("getting user's mail folder")
+			return MailboxInfo{}, err
+		}
```
ryanfkeepers, line 333: These conditionals can be combined for cleaner handling.
```suggestion
	if err := EvaluateMailboxError(err); err != nil {
			logger.CtxErr(ctx, err).Error("getting user's mail folder")
			return MailboxInfo{}, graph.Stack(ctx, err)
		}
```

src/pkg/services/m365/api/users.go
```diff
@@ -301,6 +318,62 @@ func (c Users) GetMailInbox(
 	return inbox, nil
 }
 
+func (c Users) GetMailboxInfo(
```
ryanfkeepers, line 321: This is a copy of the mailbox code from `GetInfo`, right?  The other GetInfo funcs are marked as deprecated, but not the one in this file.  Does this allow us to get rid of that func as well?

ryanfkeepers (approved):
Some refinement wanted around functions and package ownership, but generally lgtm.

src/pkg/services/m365/api/users.go
pandeyabs: Hey! Changes here are mostly a copy paste from existing code, it's not cleaned up yet. Also just FYI, most of the times I put out draft PRs to let CI run through all the tests & catch any glaring issues. That being said, I appreciate the early review comments :)

src/pkg/services/m365/api/users.go
pandeyabs: Yep, correct

src/pkg/services/m365/api/users.go
pandeyabs: > . Granted, that can feel a little weird when corso itself wants to consume the func, since the service layer is SDK oriented.

Yeah, this was my thinking behind adding this func. I want common code which checks if exchange is enabled. Currently we do this separately for SDK & CLI which can lead to bugs.

> and better integrate EvaluateMailboxError into GetMailInbox.

Let me look into it.

src/internal/m365/backup.go
pandeyabs: Yeah, that's fair.

src/pkg/services/m365/api/users.go
pandeyabs: The latter.

pandeyabs:
@ryanfkeepers Could you please take another look at this PR? :) It has changed quite a bit since the last review, as i refactored & moved a lot of code into an earlier PR.

src/internal/operations/restore.go
```diff
@@ -228,6 +229,25 @@ func (op *RestoreOperation) do(
 		"restore_protected_resource_id", restoreToProtectedResource.ID(),
 		"restore_protected_resource_name", clues.Hide(restoreToProtectedResource.Name()))
 
+	// Check if the resource has the service enabled to be able to restore.
+	runnable, err := op.rc.IsServiceEnabled(
+		ctx,
+		op.Selectors.PathService(),
+		restoreToProtectedResource.ID())
+	if err != nil {
+		logger.CtxErr(ctx, err).Error("verifying restore is runnable")
```
ryanfkeepers, line 238: This is going to double-log the error when it gets caught again on line 171.  You can drop this log and let things bubble up.

src/internal/operations/restore.go
```diff
@@ -228,6 +229,25 @@ func (op *RestoreOperation) do(
...
+		op.Errors.Fail(clues.Wrap(err, "verifying restore is runnable"))
```
ryanfkeepers, line 239: The `run()` func that calls this already establishes `errors.Fail` when an error is returned.  This line should be removed.

src/internal/operations/restore.go
```diff
@@ -228,6 +229,25 @@ func (op *RestoreOperation) do(
...
+
+		return nil, clues.Stack(err).WithClues(ctx)
```
ryanfkeepers, line 241:
```suggestion
		return nil, clues.Wrap(err, "verifying restore is runnable").WithClues(ctx)
```

src/internal/operations/restore.go
```diff
@@ -228,6 +229,25 @@ func (op *RestoreOperation) do(
...
+	}
+
+	if !runnable {
+		logger.CtxErr(ctx, graph.ErrServiceNotEnabled).Error("checking if restore is enabled")
+		op.Errors.Fail(clues.Wrap(err, "checking if restore is enabled"))
+
+		return nil, clues.Stack(graph.ErrServiceNotEnabled).WithClues(ctx)
```
ryanfkeepers, line 248: same duplication with these: they can be dropped, since the calling method will handle it.  Then the returned error just needs to get augmented.
```suggestion
		return nil, clues.Wrap(graph.ErrServiceNotEnabled, "verifying service is enabled for restore").WithClues(ctx)
```

src/internal/operations/inject/inject.go
```diff
@@ -37,6 +42,12 @@ type (
 			ctr *count.Bus,
 		) (*details.Details, error)
 
+		IsServiceEnabled(
```
ryanfkeepers, line 45: Since this is declared twice, it should get moved into its own interface, then you can embed that one within these other two.

src/internal/operations/backup.go
```diff
@@ -202,11 +202,11 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	ctx, flushMetrics := events.NewMetrics(ctx, logger.Writer{Ctx: ctx})
 	defer flushMetrics()
 
-	var runnable bool
-
-	// IsBackupRunnable checks if the user has services enabled to run a backup.
-	// it also checks for conditions like mailbox full.
-	runnable, err = op.bp.IsBackupRunnable(ctx, op.Selectors.PathService(), op.ResourceOwner.ID())
+	// Check if the user has the service enabled in order for us to run a backup.
```
ryanfkeepers, line 205:
```suggestion
	// Check if the protected resource has the service enabled in order for us to run a backup.
```

src/internal/m365/service/groups/enabled.go
```diff
@@ -0,0 +1,26 @@
...
+	resource string,
+) (bool, error) {
+	_, err := gbi.GetByID(ctx, resource)
+	if err != nil {
+		// TODO(meain): check for error message in case groups are
+		// not enabled at all similar to sharepoint
+		return false, err
+	}
+
+	return true, nil
```
ryanfkeepers, line 25: We should already have called "getByID" at this point to get the resource's ID and Name, so I don't think we're gaining anything here by calling it again.
```suggestion
	return true, nil
```

pandeyabs: Sounds good. @meain is this okay with you?

src/internal/operations/restore.go
pandeyabs: thanks! These are existing logs that I carried over during refactor. Will remove.

PR #4097, opened 2023-08-23
Cleanup duplicate code

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/backup.go
+++ b/src/internal/m365/backup.go
@@ -147,39 +147,6 @@ func (ctrl *Controller) ProduceBackupCollections(
 	return colls, ssmb, canUsePreviousBackup, nil
 }
 
-// IsBackupRunnable verifies that the users provided has the services enabled and
-// data can be backed up. The canMakeDeltaQueries provides info if the mailbox is
-// full and delta queries can be made on it.
-func (ctrl *Controller) IsBackupRunnable(
-	ctx context.Context,
-	service path.ServiceType,
-	resourceOwner string,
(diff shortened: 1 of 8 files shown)


ryanfkeepers (changes requested):
Problem: this code is not errant duplication.  It's an intentional promotion of backup enablement checks upstream from the provider/service implementation.  We want to shortcut the process of initializing backups as soon as possible in the operation, and the best way to do that is to check support at the top of the operation.

If anything, the deeper checks within the provider/service are the duplication to this implementation, and should be removed in favor of this.  And, iirc, we also need to extend this pattern to restores.

pandeyabs:
>  We want to shortcut the process of initializing backups as soon as possible in the operation, and the best way to do that is to check support at the top of the operation. If anything, the deeper checks within the provider/service are the duplication to this implementation, and should be removed in favor of this. And, iirc, we also need to extend this pattern to restores.
 
Thanks! I chose to remove this func instead of deeper checks, so that we can have parity between restore & backup code.
But your reasoning makes sense. I'll make the changes.

pandeyabs:
Closing this in favor of earlier checks (1. in [this](https://github.com/alcionai/corso/pull/4096) PR)

PR #4099, opened 2023-08-23
add sanity test boilerplate for groups

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3989 

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/.github/actions/backup-restore-test/action.yml
+++ b/.github/actions/backup-restore-test/action.yml
@@ -12,7 +12,7 @@ inputs:
     required: false
     default: ""
   restore-args:
-    description: Arguments to pass for restore
+    description: Arguments to pass for restore; restore is skipped when missing.
     required: false
     default: ""
   test-folder:
@@ -28,6 +28,10 @@ inputs:
     description: Value for the --collisions flag
     requried: false
     default: "replace"
+  with-export:
+    description: Runs export tests when true
+    required: false
+    default: false
 
 outputs:
   backup-id:
@@ -52,6 +56,7 @@ runs:
             tee $GITHUB_OUTPUT
 
     - name: Restore ${{ inputs.service }} ${{ inputs.kind }}
+      if: inputs.restore != ""
       id: restore
       shell: bash
       working-directory: src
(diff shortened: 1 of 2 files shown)


.github/workflows/sanity-test.yaml
```diff
@@ -318,6 +321,53 @@ jobs:
...
           log-dir: ${{ env.CORSO_LOG_DIR }}
+          with-export: true
+
+##########################################################################################################################################
+
+# Groups and Teams
+
+      # generate new entries for test
+      - name: Groups - Create new data
+        if: never() # TODO: enable when ready
```
meain, line 332: I don't think `never()` is available. I think we can however just use `false` though.

.github/actions/backup-restore-test/action.yml
```diff
@@ -52,6 +56,7 @@ runs:
             tee $GITHUB_OUTPUT
 
     - name: Restore ${{ inputs.service }} ${{ inputs.kind }}
+      if: inputs.restore != ""
```
meain, line 59: I'm assuming you meant `restore-args`.

.github/actions/backup-restore-test/action.yml
```diff
@@ -86,10 +92,10 @@ runs:
         ./sanity-test
 
     - name: Export ${{ inputs.service }} ${{ inputs.kind }}
+      if: ${{ inputs.with-export == true }}
```
meain, line 95: Nit: do we need the `== true` check?

meain (approved):
Need a few changes,  but approving to unblock.

.github/actions/backup-restore-test/action.yml
ryanfkeepers: whoops!  Thank you, yes.

.github/workflows/sanity-test.yaml
ryanfkeepers: so it seems.  Good catch.

.github/actions/backup-restore-test/action.yml
ryanfkeepers: probably not.  But it's not really a bug.  Just excess.

Issue #3988, opened 2023-08-08
Teams & Groups Implementation


ryanfkeepers:
High level tracking for Implementation of Teams and Groups support.

# Subtasks
- [x] [Channels Backup](https://github.com/alcionai/corso/issues/3989)
- [x] [Channels Export](https://github.com/alcionai/corso/issues/3991)
- [x] [Libraries Backup](https://github.com/alcionai/corso/issues/3990)
- [x] [Libraries Restore](https://github.com/alcionai/corso/issues/3992)
- [x] #4259
- [x] [Path Extension Support](https://github.com/alcionai/corso/issues/3993)
- [x] #4308
- [x] #4309
- [x] #4310
- [x] https://github.com/alcionai/corso/issues/4462
- [x] https://github.com/alcionai/corso/issues/4431
- [x] https://github.com/alcionai/corso/issues/4430

# Future tech-debt
- https://github.com/alcionai/corso/issues/4312
- https://github.com/alcionai/corso/issues/4405

# Boilerplate Implementation
## Packages / Files

- [x]  `/src/internal/<provider>/<service>` owns all backup/restore/export logic for that provider.
- [x]  `CLI`
    - [x]  `/cli/backup/<service>` to own all cli commands for backups
    - [x]  `/cli/restore/<service>` to own all cli commands for restores
    - [x]  `/cli/export/<service>` to own all cli commands for exports
    - [x]  `/cli/flags/<service>` to own all cli flags for the service
    - [x]  `/cli/utils/<service>` to own all cli utils/flag parsers for the service

## Interface & Data Compliance

- [x]  `/src/pkg/path/resource_path.go` -  canonical enums for representing each service.
- [x]  `/src/pkg/backup/details/<service>.go` - details iteminfo representation for the service.
- [x]  `/src/pkg/selectors/<service>.go` - selectors for the service.

## Enumeration

- [x]  Protected Resources must have an enumeration api available in the `/src/pkg/services/<provider>/_.go` file, to be consumed by SDK users.
- [x]  Protected Resources must have an enumeration api available in the `/src/pkg/services/<provider>/api/_.go` file, to be consumed internally.

vkamra:
Closing this issue now. Future work will be tracked by separate tickets

PR #4101, opened 2023-08-23
add groups, teams e2e tests

ryanfkeepers:
add e2e test suites for groups and teams backups.
These tests are currently skipped and waiting for
completion of e2e functionality.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/groups_e2e_test.go
+++ b/src/cli/backup/groups_e2e_test.go
@@ -33,6 +33,63 @@ var (
 )
 
 // ---------------------------------------------------------------------------
+// tests that require no existing backups
+// ---------------------------------------------------------------------------
+
+type NoBackupGroupsE2ESuite struct {
+	tester.Suite
+	dpnd dependencies
+	its  intgTesterSetup
+}
+
+func TestNoBackupGroupsE2ESuite(t *testing.T) {
+	t.Skip("enable when e2e is complete for groups")
+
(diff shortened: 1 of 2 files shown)


src/cli/backup/groups_e2e_test.go
```diff
@@ -0,0 +1,622 @@
...
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	suite.its = newIntegrationTesterSetup(t)
+	suite.dpnd = prepM365Test(t, ctx)
+}
+
+func (suite *BackupGroupsE2ESuite) TestGroupsBackupCmd_channelMessages() {
+	runGroupsBackupCategoryTest(suite, channelMessages)
```
meain, line 122: Do we have to check for channel messages in groups as we will be letting users only backup messages for teams explicitly. Not that it makes any difference, but just thought I should ask.

meain (approved):

src/cli/backup/groups_e2e_test.go
ryanfkeepers: I guess I don't fully understand the question.  What sorts of checks are you thinking of?

meain: My bad, I was wondering if we had to have tests for messages in groups or should we limit that only to teams.

PR #4106, opened 2023-08-24
check for zero len string with 0

ryanfkeepers:
according to https://docs.github.com/en/actions/learn-github-actions/expressions

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/.github/actions/backup-restore-test/action.yml
+++ b/.github/actions/backup-restore-test/action.yml
@@ -56,7 +56,7 @@ runs:
             tee $GITHUB_OUTPUT
 
     - name: Restore ${{ inputs.service }} ${{ inputs.kind }}
-      if: ${{ inputs.restore-args != 0 }}
+      if: ${{ inputs.restore-args }}
       id: restore
       shell: bash
       working-directory: src
@@ -78,7 +78,7 @@ runs:
             cat /tmp/corsologs
 
     - name: Check restore ${{ inputs.service }} ${{ inputs.kind }}
-      if: ${{ inputs.restore-args != 0 }}
+      if: ${{ inputs.restore-args }}
       shell: bash
       working-directory: src
       env:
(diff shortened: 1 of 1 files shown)


.github/actions/backup-restore-test/action.yml
```diff
@@ -56,7 +56,7 @@ runs:
             tee $GITHUB_OUTPUT
 
     - name: Restore ${{ inputs.service }} ${{ inputs.kind }}
-      if: ${{ inputs.restore-args != "" }}
+      if: ${{ inputs.restore-args != 0 }}
```
ashmrtn, line 59: I think maybe we just want `inputs.restore-args`? Or maybe drop the `${{ }}`? From the GitHub actions [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif)

>When you use expressions in an if conditional, you may omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. Using the ${{ }} expression syntax turns the contents into a string, and strings are truthy. For example, if: true && ${{ false }} will evaluate to true.

The part of the docs on [expressions](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) also has the below buried in it (can't link directly so scroll down to the table)

String | Parsed from any legal JSON number format, otherwise NaN.Note: empty string returns 0.
-- | --

ashmrtn: you can also manually launch a run of sanity tests by going to the sanity tests workflow in the Actions page and choosing the branch and user to run with. That may help with debugging

ryanfkeepers: changed to a truthiness check.

ashmrtn (approved):

PR #4109, opened 2023-08-25
Channel pager implementation

neha-Gupta1:
Pager implementation for channel and message pager.

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual


src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
 // ---------------------------------------------------------------------------
 
 type ChannelDeltaEnumerator interface {
-	DeltaGetPager
+	PageLinker
 	ValuesInPageLinker[models.Channelable]
 	SetNextLinker
 }
 
 // TODO: implement
```
ryanfkeepers, line 91:
```suggestion
```

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
-// var _ ChannelDeltaEnumerator = &channelsPageCtrl{}
+var _ ChannelDeltaEnumerator = &channelPageCtrl{}
```
ryanfkeepers, line 92: Channels does not use a delta query query, and thus should not be named `delta` in any location.
```suggestion
var _ ChannelEnumerator = &channelPageCtrl{}
```

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
+}
+
+func (c Channels) NewChannelPager(
+	teamID,
+	channelID string,
+	fields []string,
+) *channelPageCtrl {
+	requestConfig := &teams.ItemChannelsRequestBuilderGetRequestConfiguration{
+		QueryParameters: &teams.ItemChannelsRequestBuilderGetQueryParameters{
+			Select: fields,
```
ryanfkeepers, line 107: Same questions as for messages: do we have any reason to select fields, and do we need to prefer the page size?

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
+
+// ---------------------------------------------------------------------------
+// non delta item pager
+// ---------------------------------------------------------------------------
+
+type MessageItemIDType struct {
+	ItemID string
+}
+
+type channelItemPageCtrl struct {
```
ryanfkeepers, line 160:
```suggestion
type channelMessagePageCtrl struct {
```

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
+}
+
+func (p *channelPageCtrl) GetOdataNextLink() *string {
+	// No next link preent in the API result
+	emptyString := ""
+	return &emptyString
+}
+
+// ---------------------------------------------------------------------------
+// non delta item pager
```
ryanfkeepers, line 153: This should get moved up to co-locate with the other message pager.  The current page layout goes `messages > channels > messages`, which doesn't flow very well.
```suggestion
// non delta channel message pager
```

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
+	ItemID string
+}
+
+type channelItemPageCtrl struct {
+	gs      graph.Servicer
+	builder *teams.ItemChannelsItemMessagesRequestBuilder
+	options *teams.ItemChannelsItemMessagesRequestBuilderGetRequestConfiguration
+}
+
+func (c Channels) GetItemIDsInContainer(
```
ryanfkeepers, line 166: This pattern is an artifact of the drive pager from which you took the boilerplate.  The drive backup system has a basis for enumerating only the IDs.  But I'm not sure there's any point to this when enumerating channel messages.  A simpler implementation is to have the non-delta enumeration return the complete chanMessageable, much like the delta enumeration does.

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -14,30 +21,205 @@ type ChannelMessageDeltaEnumerator interface {
...
+		Teams().
+		ByTeamId(teamID).
+		Channels().
+		ByChannelId(containerID).
+		Messages()
+
+	return &channelItemPageCtrl{c.Stable, builder, options}
+}
+
+//lint:ignore U1000 False Positive
```
ryanfkeepers, line 212: is this necessary, or can it be removed?

ryanfkeepers (changes requested):
All functions in this change need test coverage.

src/pkg/services/m365/api/channels_pager.go
neha-Gupta1: Implemented

src/pkg/services/m365/api/channels_pager.go
neha-Gupta1: Yes. Its required. Linter fails because of it.

ryanfkeepers:
closing: changes were ported to other PRs

PR #4112, opened 2023-08-25
Fix path collisions for SharePoint export

meain:
Previously if we had multiple SharePoint document libs, we would have merged the contents of both in the export. This separates it by document lib.

After:
- DocumentLibA/FileA
- DocumentLibB/FileB

Before:
- FileA
- FileB

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/export.go
+++ b/src/internal/m365/collection/drive/export.go
@@ -1,0 +1,140 @@
+package drive
+
+import (
+	"context"
+	"strings"
+
+	"github.com/alcionai/clues"
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
+	"github.com/alcionai/corso/src/internal/version"
+	"github.com/alcionai/corso/src/pkg/export"
+	"github.com/alcionai/corso/src/pkg/fault"
+)
+
+var _ export.Collection = &ExportCollection{}
+
+// ExportCollection is the implementation of export.ExportCollection for OneDrive
+type ExportCollection struct {
+	// baseDir contains the path of the collection
+	baseDir string
+
+	// backingCollection is the restore collection from which we will
+	// create the export collection.
+	backingCollection data.RestoreCollection
+
+	// backupVersion is the backupVersion of the backup this collection was part
+	// of. This is required to figure out how to get the name of the
+	// item.
+	backupVersion int
+}
+
+func NewExportCollection(
+	baseDir string,
+	backingCollection data.RestoreCollection,
+	backupVersion int,
+) ExportCollection {
+	return ExportCollection{
+		baseDir:           baseDir,
+		backingCollection: backingCollection,
+		backupVersion:     backupVersion,
+	}
+}
+
+func (ec ExportCollection) BasePath() string {
+	return ec.baseDir
+}
+
+func (ec ExportCollection) Items(ctx context.Context) <-chan export.Item {
+	ch := make(chan export.Item)
+	go items(ctx, ec, ch)
+
+	return ch
+}
+
+// items converts items in backing collection to export items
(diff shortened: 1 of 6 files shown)


src/internal/m365/service/sharepoint/export.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+// ProduceExportCollections will create the export collections for the
+// given restore collections.
+func ProduceExportCollections(
+	ctx context.Context,
+	backupVersion int,
```
ryanfkeepers, line 23: Any chance of getting an `ExportProducerConfig` similar to the [backup and restore versions](https://github.com/alcionai/corso/blob/main/src/internal/operations/inject/containers.go#L24) we have?  As a separate PR, of course.

src/internal/m365/service/sharepoint/export.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, dc := range dcs {
+		drivePath, err := path.ToDrivePath(dc.FullPath())
+		if err != nil {
+			return nil, clues.Wrap(err, "transforming path to drive path").WithClues(ctx)
+		}
+
+		driveName, ok := backupDriveIDNames.NameOf(drivePath.DriveID)
+		if !ok {
+			// This should not happen, but just in case
+			logger.Ctx(ctx).With("drive_id", drivePath.DriveID).Warn("drive name not found, using drive id")
```
ryanfkeepers, line 45: Avoid `Warn` level logging, please.  (also, the linter should have caught this...)
```suggestion
			logger.Ctx(ctx).With("drive_id", drivePath.DriveID).Info("drive name not found, using drive id")
```

src/internal/m365/collection/drive/export.go
```diff
@@ -0,0 +1,141 @@
...
+				Name: name,
+				Body: item.ToReader(),
+			},
+			Error: err,
+		}
+	}
+
+	eitems, erecovereable := errs.ItemsAndRecovered()
+
+	// Return all the items that we failed to get from kopia at the end
```
ryanfkeepers, line 84:
```suggestion
	// Return all the items that we failed to source from the persistence layer
```

src/internal/m365/collection/drive/export.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// items converts items in backing collection to export items
+func items(ctx context.Context, ec ExportCollection, ch chan<- export.Item) {
+	defer close(ch)
+
+	errs := fault.New(false)
+
+	// There will only be a single item in the backingCollections
+	// for OneDrive
```
ryanfkeepers, line 63: Is this comment accurate?  It feels weird to say "there's only one" and then run a range instead of just reading from the channel once.  One or the other isn't fully correct.  Are there errors if we see more than one item?  Do we have to behave differently?  Or would we be safe to remove this comment and not worry about needing to know the details?

ryanfkeepers (approved):
Some nits, but lgtm.

src/internal/m365/collection/drive/export.go
meain: Ahh, good catch. This comment was from an earlier implementation where we had `backingCollection` as `[]data.RestoreCollection`. We had moved away from that format, but I forgot to remove this comment.

src/internal/m365/service/sharepoint/export.go
meain: Hmm, that could be useful to unify the interface. I'll add it my backlog.

PR #4115, opened 2023-08-25
Use a larger data chunk to test compression

### LINKED ISSUE #3156: Add a regression test to ensure compression is enabled
We added a regression test to the Kopia repo but should add one to Corso as well to ensure that if Kopia behavior changes - we catch it.

ashmrtn:
Test that compression is on by default by having one file be large and easily compressible. Compression is indirectly checked by looking at how many bytes were uploaded.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #3156

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/kopia/wrapper_test.go
```diff
@@ -765,18 +766,24 @@ func (suite *KopiaIntegrationSuite) TearDownTest() {
...
 
 func (suite *KopiaIntegrationSuite) TestBackupCollections() {
+	c1 := exchMock.NewCollection(
+		suite.storePath1,
+		suite.locPath1,
+		5)
+	// Add a 4k chunk of data that should be compressible. This helps check
+	// compression is enabled because we do some testing on the number of bytes
+	// uploaded during the first backup.
+	c1.Data[0] = []byte(strings.Repeat("abcdefgh", 512))
```
meain, line 776: Just for my own understanding, where exactly are checking for compression?

ashmrtn: good question! Unfortunately we can't do a direct check on the uploaded data blobs to check for compression. Instead, what we do is check the number of bytes uploaded ([here](https://github.com/alcionai/corso/blob/3e43028a883bb2a38b3fe96eaeab745f31193445/src/internal/kopia/wrapper_test.go#L899) and the assertion below). From manual testing, these assertions pass when compression is enabled but fail if it's not

meain (approved):

PR #4116, opened 2023-08-25
implementation for path ChannelMessagesCategory

neha-Gupta1:
- enumerating channels
- Implementing BackupHandler
- Implementaion of path.ChannelMessagesCategory backup

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual


src/internal/m365/collection/groups/handlers.go
```diff
@@ -16,6 +16,7 @@ type BackupHandler interface {
 	) (models.Channelable, error)
 	NewChannelsPager(
 		teamID string,
+		fields []string,
```
ryanfkeepers, line 19: Field selection shouldn't be exposed to any layer above the api, as that conflicts with producing standardized data.  Unless, of course, there's a need to create multiple instances of pagers, each with different fields.  That doesn't seem to me like a pattern we want to take on.  It's much safer and simpler to assume that pagers can be treated like singletons (ie, we'd only ever need to construct one) than to let them be stateful.  If we really need per-enumeration field selection, that's better of handled by some func that accepts a standard pager as part of its properties.

src/internal/m365/collection/groups/groups_handler.go
ryanfkeepers, line 1: This should probably be named `channels_handler.go`.  That's the scope being processed, not the group as a whole.

src/internal/m365/collection/groups/groups_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/microsoft/kiota-abstractions-go/serialization"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+)
+
+var _ BackupHandler = &groupBackupHandler{}
+
+type groupBackupHandler struct {
+	ac      api.Channels
+	groupID string
+	scope   selectors.GroupsScope
```
ryanfkeepers, line 17: is `scope` used at all here?  If not, it can be removed.  We can always add it back in later if necessary.

src/internal/m365/collection/groups/backup.go
```diff
@@ -306,7 +334,7 @@ func includeContainer(
 	directory := ptr.Val(gd.GetDisplayName())
 
 	// TODO(keepers): awaiting parent branch to update to main
```
ryanfkeepers, line 336:
```suggestion
```

src/internal/m365/collection/groups/backup.go
```diff
@@ -153,11 +179,13 @@ func populateCollections(
 		// }
 
 		// ictx = clues.Add(ictx, "previous_path", prevPath)
-
+		// TODO: Neha check this
+		var fields []string
 		// TODO: the handler should provide this implementation.
+		// TODO: if we doing this messages are items for us.
```
ryanfkeepers, line 185: I'm not sure what this comment is supposed to mean.  Can you please explain?

src/internal/m365/collection/groups/backup.go
```diff
@@ -153,11 +179,13 @@ func populateCollections(
 		// }
 
 		// ictx = clues.Add(ictx, "previous_path", prevPath)
-
+		// TODO: Neha check this
+		var fields []string
```
ryanfkeepers, line 183: see other comments on the topic, please.

ryanfkeepers (changes requested):

src/internal/m365/collection/groups/handlers.go
neha-Gupta1: Make sense. Updated the code.

src/internal/m365/collection/groups/backup.go
neha-Gupta1: I was figuring out some changes. This comment is no more needed.

src/pkg/services/m365/api/channels.go
```diff
@@ -128,6 +126,27 @@ func (c Channels) GetMessage(
 	return message, ChannelMessageInfo(message, size), nil
 }
 
+// GetMessage retrieves a ChannelMessage item.
+func (c Channels) GetMessageByID(
```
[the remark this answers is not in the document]
neha-Gupta1: yes. Updated it.

ryanfkeepers:
closing: work has been ported to other PRs

PR #4125, opened 2023-08-28
Fix path for group metadata

meain:
Previously we were using sharepoint site id.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -47,8 +47,15 @@ const restrictedDirectory = "Site Pages"
 type Collections struct {
 	handler BackupHandler
 
-	tenantID      string
+	tenantID string
+
+	// resourceOwner is the owner of the top level resource
 	resourceOwner string
+
+	// serviceOwner is the owner of the service we are backing up.
+	// serviceOwner and resourceOwner will be only when backing up a
+	// SharePoint site associated with a Group.
+	serviceOwner string
 
 	statusUpdater support.StatusUpdater
 
@@ -69,6 +76,7 @@ func NewCollections(
 	bh BackupHandler,
 	tenantID string,
 	resourceOwner string,
+	serviceOwner string,
 	statusUpdater support.StatusUpdater,
 	ctrlOpts control.Options,
 ) *Collections {
@@ -76,6 +84,7 @@ func NewCollections(
 		handler:       bh,
 		tenantID:      tenantID,
 		resourceOwner: resourceOwner,
+		serviceOwner:  serviceOwner,
 		CollectionMap: map[string]map[string]*Collection{},
 		statusUpdater: statusUpdater,
 		ctrl:          ctrlOpts,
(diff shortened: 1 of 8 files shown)


src/internal/m365/collection/drive/collections.go
```diff
@@ -47,9 +47,16 @@ const restrictedDirectory = "Site Pages"
 type Collections struct {
 	handler BackupHandler
 
-	tenantID      string
+	tenantID string
+
+	// resourceOwner is the owner of the top level resource
```
ryanfkeepers, line 52:
```suggestion
	// resourceOwner is the resource (eg: user, site, etc) whose drive data is
	// stored by the backup.
	// TODO: rename to protectedResource
```

src/internal/m365/collection/drive/collections.go
```diff
@@ -47,9 +47,16 @@ const restrictedDirectory = "Site Pages"
...
 	resourceOwner string
 
+	// serviceOwner is the owner of the service we are backing up.
+	// serviceOwner and resourceOwner will be only when backing up a
+	// SharePoint site associated with a Group.
+	serviceOwner string
```
ryanfkeepers, line 58: I'd like to challenge this implementation a bit.  The resource and service owners are so often the same that I worry thier usage will be perpetually unclear.  That confusion could lead to accidental and difficult-to-pinpoint bugs.  For example, I think the choice here to store the metadata under a path that matches sharepoint formatting instead of matching the standard groups formatting both feels accidental and could lead to unexpected side effects.

A more holistic implementation would be to extend the handler with a func like `metadataDir() path.Elements`, and modify `graph.NewMetadataEntry` to accept an Elements slice or Path instead of only a string.  This way we:
* allow any backup producer to arbitrarily nest metadata locations (reusable)
* maintain a single owner within the collection (simple)
* maintain the same prefix pathing for all data in the collection (consistent)

ryanfkeepers (changes requested):

src/internal/m365/collection/drive/collections.go
meain: Yup, that makes sense and was the plan long term. To be frank this was more of a short term fix, but I guess it does not make much sense to go this route. I think I will just implement passing partial paths downstream as I am starting with it anyways.

ryanfkeepers: > I think I will just implement passing partial paths downstream as I am starting with it anyways.

Can you clarify that for me?  Does that mean you're going to close this PR and shift the work elsewhere?  Amend this PR in a follow-up PR?  Amend this PR directly?

meain: Closing in favor of https://github.com/alcionai/corso/pull/4156

PR #4126, opened 2023-08-28
Update test helpers for groups backup

meain:
This still is not complete, but this fixes a few places in the test suite helpers where we did not accommodate for groups.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/test/helper_test.go
+++ b/src/internal/operations/test/helper_test.go
@@ -133,8 +133,12 @@ func prepNewTestBackupOp(
 	bod.sw = store.NewWrapper(bod.kms)
 
 	connectorResource := resource.Users
-	if sel.Service == selectors.ServiceSharePoint {
+
+	switch sel.Service {
+	case selectors.ServiceSharePoint:
 		connectorResource = resource.Sites
+	case selectors.ServiceGroups:
+		connectorResource = resource.Groups
 	}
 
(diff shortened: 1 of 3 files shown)


src/internal/operations/test/helper_test.go
```diff
@@ -575,14 +581,21 @@ type ids struct {
 	DriveRootFolderID string
 }
 
+type gids struct {
+	ID                        string
+	RootSiteID                string
+	RootSiteDriveID           string
+	RootSiteDriveRootFolderID string
+}
+
```
ryanfkeepers, line 590: Kinda wish I could recursively nest a struct in itself, because ids containing `RootSite ids` would work great here.  But I think this is a good enough halfway stop.

Mostly suggesting the change as a Nit to ensure we have separate declarations for "what ids we want" vers contextual siloing/subownership of ids.
```suggestion
type gids struct {
        ids
	RootSite ids
}

```

meain: I've changed it to the following. Didn't quite understand the use of using ids in the root level.

``` go
type gids struct {
	ID       string
	RootSite ids
}
```

PR #4135, opened 2023-08-28
add channel and messages enumerators

ryanfkeepers:
adds enumerators for channels, and a delta
of channel messages.
Also updates the its struct to include group ids, and reformats some of the struct schema.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/channels_pager.go
+++ b/src/pkg/services/m365/api/channels_pager.go
@@ -158,10 +158,6 @@ func (p *channelPageCtrl) GetPage(
 	return resp, graph.Stack(ctx, err).OrNil()
 }
 
-func (p *channelPageCtrl) GetOdataNextLink() *string {
-	return p.GetOdataNextLink()
-}
-
 func (p *channelPageCtrl) ValuesIn(l PageLinker) ([]models.Channelable, error) {
 	return getValues[models.Channelable](l)
 }
--- a/src/pkg/services/m365/api/channels_pager.go
+++ b/src/pkg/services/m365/api/channels_pager.go
@@ -1,43 +1,223 @@
 package api
 
 import (
+	"context"
+
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"github.com/microsoftgraph/msgraph-sdk-go/teams"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/m365/graph"
+	"github.com/alcionai/corso/src/pkg/logger"
 )
 
 // ---------------------------------------------------------------------------
-// item pager
+// channel message pager
 // ---------------------------------------------------------------------------
 
-type ChannelMessageDeltaEnumerator interface {
-	DeltaGetPager
-	ValuesInPageLinker[models.ChatMessageable]
-	SetNextLinker
+var _ DeltaPager[models.ChatMessageable] = &channelMessageDeltaPageCtrl{}
+
+type channelMessageDeltaPageCtrl struct {
+	resourceID, channelID string
+	gs                    graph.Servicer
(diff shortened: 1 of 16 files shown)


src/pkg/services/m365/api/channels_pager.go
```diff
@@ -1,43 +1,223 @@
...
+			continue
+		}
+
+		if err != nil {
+			return nil, DeltaUpdate{}, graph.Wrap(ctx, err, "retrieving page of channels")
+		}
+
+		vals, err := pager.ValuesIn(page)
+		if err != nil {
+			return nil, DeltaUpdate{}, graph.Wrap(ctx, err, "extracting channels from response")
```
ashmrtn, line 109: nit: should this be `extracting channel messages`?

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -1,43 +1,223 @@
...
+		}
+
+		vs = append(vs, vals...)
```
ashmrtn, line 112: do we need to handle deleted messages in delta results?

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -1,43 +1,223 @@
...
+		vs               = []models.ChatMessageable{}
+		pager            = c.NewChannelMessageDeltaPager(teamID, channelID)
+		invalidPrevDelta = len(prevDelta) == 0
+		newDeltaLink     string
+	)
+
+	// Loop through all pages returned by Graph API.
+	for {
+		page, err := pager.GetPage(graph.ConsumeNTokens(ctx, graph.SingleGetOrDeltaLC))
+		if graph.IsErrInvalidDelta(err) {
```
ashmrtn, line 93: hmmm I know this probably follows similar code we have, but looking at it there's the assumption that we get the invalid delta token error on the first call to the API. It's probably fine to just check for `IsErrInvalidDelta`, but we should probably clear `vs` when we see it

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -1,43 +1,223 @@
...
+
+func (p *channelPageCtrl) GetPage(
+	ctx context.Context,
+) (PageLinker, error) {
+	resp, err := p.builder.Get(ctx, p.options)
+	return resp, graph.Stack(ctx, err).OrNil()
+}
+
+func (p *channelPageCtrl) GetOdataNextLink() *string {
+	return ptr.To("")
```
ashmrtn, line 161: is this intentionally returning "no next page"? If it is could you please add a comment that states why it's alright to assume there's no next page

ashmrtn (approved):
mostly nits and maybe one or two things that were missed

src/pkg/services/m365/api/channels_pager.go
ryanfkeepers: We aren't persisting metadata yet, so it's all non-incremental through the incremental pager.  Proper incremental handling will arrive in v1.

Also, I'm planning to do a refactor between v0 and v1 that centralizes the delta page runners (we're halfway there already) which should resolve this concern on its own.

src/pkg/services/m365/api/channels_pager.go
ryanfkeepers: Hah, no, not intended.  This is copied code which slipped through my attention.  Good catch, ty!

src/pkg/services/m365/api/channels_pager.go
ryanfkeepers: Ah, good call, thank you.

PR #4137, opened 2023-08-28
Tighten checks for assist base garbage collection

ashmrtn:
Do a bit better job checking assist bases for
garbage collection by considering whether
there's a newer merge base already

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3217

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/cleanup_backups.go
+++ b/src/internal/kopia/cleanup_backups.go
@@ -266,7 +266,7 @@ func cleanupOrphanedData(
 	// This sort of edge case will ideally happen only for a few assist bases at
 	// a time. Assuming this function is run somewhat periodically, missing these
 	// edge cases is alright because they'll get picked up on a subsequent run.
-	assistItems := collectOldAssistBases(ctx, assistBackups)
+	assistItems := collectOldAssistBases(ctx, mostRecentMergeBase, assistBackups)
 
 	logger.Ctx(ctx).Debugw(
 		"garbage collecting old assist bases",
@@ -293,6 +293,10 @@ func transferTags(snap *manifest.EntryMetadata, bup *backup.Backup) error {
 	tenant, err := decodeElement(snap.Labels[kopiaPathLabel])
 	if err != nil {
 		return clues.Wrap(err, "decoding tenant from label")
+	}
+
+	if bup.Tags == nil {
+		bup.Tags = map[string]string{}
 	}
 
 	bup.Tags[tenantTag] = tenant
@@ -364,6 +368,7 @@ func keysForBackup(bup *backup.Backup) ([]string, error) {
 
 func collectOldAssistBases(
 	ctx context.Context,
+	mostRecentMergeBase map[string]time.Time,
 	bups []*backup.Backup,
 ) []manifest.ID {
 	// maybeDelete is the set of backups that could be deleted. It starts out as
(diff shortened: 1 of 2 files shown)


src/internal/kopia/cleanup_backups.go
```diff
@@ -128,11 +128,17 @@ func cleanupOrphanedData(
...
 	}
 
-	// assistBackups is the set of backups that have a
-	//   * a label denoting they're an assist backup
-	//   * item data snapshot
-	//   * details snapshot
-	var assistBackups []*backup.Backup
+	var (
+		// assistBackups is the set of backups that have a
+		//   * a label denoting they're an assist backup
```
ryanfkeepers, line 133:
```suggestion
		// assistBackups is the set of backups that have a
		//   * label denoting they're an assist backup
```

src/internal/kopia/cleanup_backups.go
```diff
@@ -128,11 +128,17 @@ func cleanupOrphanedData(
...
+		//   * item data snapshot
+		//   * details snapshot
+		assistBackups []*backup.Backup
+		// mostRecentMergeBase holds the creation time of the most recent merge base
+		// for the given Reason. The Reason -> key translation is done by
```
ryanfkeepers, line 138: Nit: took me a second to grok the key design from that explanation.
```suggestion
		// mostRecentMergeBase maps the reason to its most recent creation time.
		// The key is created using keysForBackup.
```

src/internal/kopia/cleanup_backups.go
```diff
@@ -188,31 +194,57 @@ func cleanupOrphanedData(
...
 
+			// This is a little messy to have, but can simplify the logic below.
+			// The state of tagging in corso isn't all that great right now and we'd
+			// really like to consolidate tags and clean them up. For now, we're
+			// going to copy tags that are related to Reasons for a backup from the
+			// item data snapshot to the backup model. This makes the function
+			// checking if assist backups should be garbage collected a bit easier
+			// because now they only have to source data from backup models.
+			if err := transferTags(d, &bm); err != nil {
+				logger.CtxErr(ctx, err).Errorw(
```
ryanfkeepers, line 205: I feel minorly uneasy about logging Error level when we don't treat the situation as an error.  Worst case scenario it can lead to debugging confusion where future maintainers will have to decide if an error log was supposed to result in an erroneous run.  The comments help a lot.  But I think an info log would translate a little more nicely here.

src/internal/kopia/cleanup_backups.go
```diff
@@ -188,31 +194,57 @@ func cleanupOrphanedData(
...
+				assistBackups = append(assistBackups, &bm)
+				continue
+			}
+
+			// If it's a merge base track the time it was created so we can check
+			// later if we should remove all assist bases or not.
+			tags, err := keysForBackup(&bm)
+			if err != nil {
+				logger.CtxErr(ctx, err).Error(
+					"getting Reason keys for merge base. May keep an additional assist base")
```
ryanfkeepers, line 236:
```suggestion
				logger.CtxErr(ctx, err).
					Error("getting Reason keys for merge base. May keep an additional assist base")
```

src/internal/kopia/cleanup_backups.go
```diff
@@ -188,31 +194,57 @@ func cleanupOrphanedData(
...
+			}
+
+			for _, tag := range tags {
+				t, ok := mostRecentMergeBase[tag]
+				if ok && t.After(bm.CreationTime) {
```
ryanfkeepers, line 241: technically you don't need the OK check, right?  Zero valued time is the epoch.

ryanfkeepers (approved):

src/internal/kopia/cleanup_backups.go
ashmrtn: yea, wasn't really sure what level to put it under. Info makes sense

src/internal/kopia/cleanup_backups.go
ashmrtn: yea, that should work

PR #4145, opened 2023-08-29
enumerate ids, get messages in Items()

ryanfkeepers:
delays getting the full channel message data until we iterate over each message in the collection's Items() stream. The initial enumeration has had its footprint reduced to only ids. Also fixes up the failing operations backup test.

Follow up PRs still need to:
* populate each message with all its replies on the get-later
* turn on other integration testing at the operations and ci layers

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/groups/backup.go
+++ b/src/internal/m365/collection/groups/backup.go
@@ -176,9 +176,7 @@ func populateCollections(
 			path.Builder{}.Append(cName),
 			qp.Category,
 			statusUpdater,
-			ctrlOpts,
-			cID,
-			cName)
+			ctrlOpts)
 
 		channelCollections[cID] = &edc
 
--- a/src/internal/m365/collection/groups/backup.go
+++ b/src/internal/m365/collection/groups/backup.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"golang.org/x/exp/maps"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/data"
@@ -147,7 +148,7 @@ func populateCollections(
 
 		// ictx = clues.Add(ictx, "previous_path", prevPath)
 
-		items, _, err := bh.getChannelMessagesDelta(ctx, cID, "")
+		items, _, err := bh.getChannelMessageIDsDelta(ctx, cID, "")
 		if err != nil {
 			el.AddRecoverable(ctx, clues.Stack(err))
 			continue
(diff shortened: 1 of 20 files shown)


src/internal/m365/collection/groups/backup.go
```diff
@@ -147,7 +148,7 @@ func populateCollections(
 
 		// ictx = clues.Add(ictx, "previous_path", prevPath)
 
-		items, _, err := bh.getChannelMessagesDelta(ctx, cID, "")
+		items, _, err := bh.getChannelMessageIDsDelta(ctx, cID, "")
```
ashmrtn, line 151: Ideally we should get IDs and modtimes so that we can send enough data to trigger kopia assisted incremental backups without having to fetch the full item. The modtime returned for that should be the same modtime we eventually return for backup details. To simplify the logic a bit, it may be best to fetch all fields that are required in backup details here (maybe with the exception of size)

I'm not sure if I'm looking at the correct code, but it does seem like this call is getting all fields during the delta query under the hood

ashmrtn (approved):

src/internal/m365/collection/groups/collection.go
```diff
@@ -280,7 +291,8 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 
 			// TODO: the item getter should provide this func when we switch
 			// to a lookup-id-then-item pattern.
-			info := api.ChannelMessageInfo(item, int64(len(data)), col.chanID, col.chanName)
+			info.ChannelID = col.chanID
+			info.ChannelName = col.chanName
```
neha-Gupta1, line 295: Are we going to store teams info on meta data level too? That would help us in creating teams it that was also deleted.

src/internal/m365/collection/groups/backup.go
ryanfkeepers: > Ideally we should get IDs and modtimes so that we can send enough data to trigger kopia assisted incremental backups without having to fetch the full item

Sounds good.  I'll take care of this in a follow-up PR.

> I'm not sure if I'm looking at the correct code, but it does seem like this call is getting all fields during the delta query under the hood

You're looking in the right place.  Property selection isn't supported for items at this time.  We could optimize around passing the entire item, if wanted.  I opted not to at this time.

src/internal/m365/collection/groups/collection.go
ryanfkeepers: we have a couple different types of metadata.  Did you mean alongside previous paths and delta tokens? Or did you mean as a `foo.meta` item?  If the latter (.meta), I'd like to avoid that unless absolutely necessary.  And since we're not planning to restore teams or channels, we might not need that info.

PR #4148, opened 2023-08-30
gather replies when streaming message items

ryanfkeepers:
when streaming channel message items into a backup, collect all replies in a message at the same time that the item gets looked up.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/groups/collection.go
+++ b/src/internal/m365/collection/groups/collection.go
@@ -35,10 +35,6 @@ const (
 )
 
 type Collection struct {
-	// TODO(keepers): figure out a better way of handing these down
-	chanID   string
-	chanName string
-
 	protectedResource string
 	stream            chan data.Item
 
@@ -86,11 +82,8 @@ func NewCollection(
 	statusUpdater support.StatusUpdater,
 	ctrlOpts control.Options,
 	// doNotMergeItems bool,
-	chanID, chanName string,
 ) Collection {
 	collection := Collection{
-		chanID:   chanID,
-		chanName: chanName,
 		added:    map[string]struct{}{},
 		category: category,
 		ctrl:     ctrlOpts,
@@ -268,10 +261,13 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 			writer := kjson.NewJsonSerializationWriter()
 			defer writer.Close()
 
+			flds := col.fullPath.Folders()
+			parentFolderID := flds[len(flds)-1]
+
 			item, info, err := col.getter.getChannelMessage(
 				ctx,
 				col.protectedResource,
-				col.chanID,
+				parentFolderID,
 				id)
 			if err != nil {
 				logger.CtxErr(ctx, err).Info("writing channel message to serializer")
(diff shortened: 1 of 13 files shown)


src/internal/m365/collection/groups/collection.go
```diff
@@ -268,10 +261,13 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
 			writer := kjson.NewJsonSerializationWriter()
 			defer writer.Close()
 
+			flds := col.fullPath.Folders()
+			parentFolderID := flds[len(flds)-1]
+
 			item, info, err := col.getter.getChannelMessage(
 				ctx,
 				col.protectedResource,
-				col.chanID,
```
ashmrtn, line 274: I'd actually prefer a separate variable for the channel ID rather than extracting it from the storage path. Taking it from the storage path implies we'll never change the storage path format (I hope we don't, but best to avoid the coupling if we can)

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -141,6 +152,109 @@ func (c Channels) GetChannelMessageIDsDelta(
...
+
+		nextLink := ptr.Val(page.GetOdataNextLink())
+		if len(nextLink) == 0 {
+			break
+		}
+
+		pager.SetNext(nextLink)
+	}
+
+	logger.Ctx(ctx).Debugf("retrieved %d channel message replies", len(vs))
```
ashmrtn, line 253: nit: stray debug log?
```suggestion
```

ashmrtn (approved):

src/internal/m365/collection/groups/collection.go
```diff
@@ -86,11 +82,8 @@ func NewCollection(
 	statusUpdater support.StatusUpdater,
 	ctrlOpts control.Options,
 	// doNotMergeItems bool,
-	chanID, chanName string,
 ) Collection {
 	collection := Collection{
-		chanID:   chanID,
-		chanName: chanName,
```
neha-Gupta1, line 93: Out of curiosity- where will we get channel ID and Name from?

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -67,7 +68,12 @@ func (c Channels) NewChannelMessageDeltaPager(
 	}
 
 	options := &teams.ItemChannelsItemMessagesDeltaRequestBuilderGetRequestConfiguration{
-		Headers: newPreferHeaders(preferPageSize(maxNonDeltaPageSize)),
+		QueryParameters: &teams.ItemChannelsItemMessagesDeltaRequestBuilderGetQueryParameters{},
```
neha-Gupta1, line 71: Is query param required here?

src/pkg/services/m365/api/channels_pager.go
ryanfkeepers: Not stray.  Could be an info, i suppose.

src/pkg/services/m365/api/channels_pager.go
ryanfkeepers: Yes, so that we don't hit a nil-pointer on line 76.

src/internal/m365/collection/groups/collection.go
ryanfkeepers: From the repoRef path.

src/internal/m365/collection/groups/collection.go
ryanfkeepers: This doesn't prevent us from changing the storage path.  It just means that if/when we do, we'll need to field the channelID some other way to support that refactor.  While I understand the want to avoid copuling, I think this is fairly harmless in the scheme of things.

PR #4153, opened 2023-08-30
Use Libraries prefix when exporting SharePoint files

meain:
As of now we only export SP libraries, but in future we will have to export lists and pages. This ensure that we can accommodate them later.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/sharepoint/export.go
+++ b/src/internal/m365/service/sharepoint/export.go
@@ -47,6 +47,7 @@ func ProduceExportCollections(
 		}
 
 		baseDir := path.Builder{}.
+			Append("Libraries").
 			Append(driveName).
 			Append(drivePath.Folders...)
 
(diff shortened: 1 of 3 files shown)


src/internal/m365/service/sharepoint/export.go
```diff
@@ -47,6 +47,7 @@ func ProduceExportCollections(
 		}
 
 		baseDir := path.Builder{}.
+			Append("Libraries").
```
ryanfkeepers, line 50: Can we use the path category string here instead?

ryanfkeepers (approved):

src/internal/m365/service/sharepoint/export.go
meain: `path.LibrariesCategory` works out to lower case libraries. I personally feel like `Libraries` is better than `libraries` for the folder name. That said, we can maybe create another constant. Let me know if you have any thoughts. I'm merging it for now as there will be no logical changes.

Issue #4154, opened 2023-08-30
Path changes for storing groups data


meain:
Related: https://github.com/alcionai/corso/issues/3990

PR #4155, opened 2023-08-30
Use path prefixes for backup paths

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/4154

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/group_handler.go
+++ b/src/internal/m365/collection/drive/group_handler.go
@@ -1,11 +1,10 @@
 package drive
 
 import (
+	odConsts "github.com/alcionai/corso/src/internal/m365/service/onedrive/consts"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/selectors"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
-
-	odConsts "github.com/alcionai/corso/src/internal/m365/service/onedrive/consts"
 )
 
 var _ BackupHandler = &groupBackupHandler{}
@@ -53,9 +52,8 @@ func (h groupBackupHandler) PathPrefix(
 
 func (h groupBackupHandler) CanonicalPath(
 	folders *path.Builder,
-	tenantID, resourceOwner string,
+	tenantID string,
 ) (path.Path, error) {
-	// TODO: drop resourceOwner
 	return folders.ToDataLayerPath(
 		tenantID,
 		h.groupID,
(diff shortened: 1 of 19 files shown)


src/internal/m365/collection/drive/group_handler.go
```diff
@@ -27,16 +33,36 @@ func NewGroupBackupHandler(groupID string, ac api.Drives, scope selectors.Groups
...
+		odConsts.RootPathDir)
 }
 
-func (h groupBackupHandler) ServiceCat() (path.ServiceType, path.CategoryType) {
-	return path.GroupsService, path.LibrariesCategory
+func (h groupBackupHandler) CanonicalPath(
+	folders *path.Builder,
+	tenantID string,
+) (path.Path, error) {
+	return folders.ToDataLayerPath(
```
ashmrtn, line 57: nit: I'd prefer manipulating parameters here instead of adding a vararg parameter to `ToDataLayerPath`. The surface area of the path package is already fairly large :/

src/pkg/path/builder.go
```diff
@@ -306,6 +306,7 @@ func (pb Builder) ToDataLayerPath(
 	service ServiceType,
 	category CategoryType,
 	isItem bool,
+	elems ...string,
```
ashmrtn, line 309: if we keep this will you please add tests for it?

src/internal/m365/collection/drive/group_handler.go
meain: I'm not entirely sure I get the change you are suggesting. If you talking about just appending to the path here, the problem is that `folders.ToDataLayerPath` adds these as prefixes. Let me know if I am missing something obvious.

src/internal/m365/collection/drive/group_handler.go
```diff
@@ -27,16 +33,36 @@ func NewGroupBackupHandler(groupID string, ac api.Drives, scope selectors.Groups
...
+		odConsts.SitesPathDir,
+		h.siteID,
+		odConsts.DrivesPathDir,
+		driveID,
+		odConsts.RootPathDir)
 }
 
-func (h groupBackupHandler) ServiceCat() (path.ServiceType, path.CategoryType) {
-	return path.GroupsService, path.LibrariesCategory
+func (h groupBackupHandler) CanonicalPath(
```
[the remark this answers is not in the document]
meain: Hmm, makes sense. Let me take that up in a follow up PR.

PR #4161, opened 2023-08-30
enable groups integration tests

ryanfkeepers:
enables nightly cli e2e tests, operations layer integration tests, and sanity tests for groups and teams.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/groups_e2e_test.go
+++ b/src/cli/backup/groups_e2e_test.go
@@ -1,0 +1,615 @@
+package backup_test
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/alcionai/clues"
+	"github.com/google/uuid"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/alcionai/corso/src/cli"
+	"github.com/alcionai/corso/src/cli/config"
+	"github.com/alcionai/corso/src/cli/flags"
+	"github.com/alcionai/corso/src/cli/print"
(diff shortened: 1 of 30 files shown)


src/cli/backup/groups_e2e_test.go
```diff
@@ -0,0 +1,615 @@
+package backup_test
```
ryanfkeepers, line 1: These files were added in https://github.com/alcionai/corso/pull/4101/files.  I'm not sure how they got sliced out.

src/cli/backup/groups_e2e_test.go
```diff
@@ -0,0 +1,615 @@
...
+
+	cmd := cliTD.StubRootCmd(
+		"backup", "delete", "groups",
+		"--config-file", suite.dpnd.configFilePath,
+		"--"+flags.BackupFN, uuid.NewString())
+	cli.BuildCommandTree(cmd)
+
+	// unknown backupIDs should error since the modelStore can't find the backup
+	err := cmd.ExecuteContext(ctx)
+	require.Error(t, err, clues.ToCore(err))
```
neha-Gupta1, line 594: Should the actual error be also compared instead of just checking that an error is received. The error could be on any level and we might not be catching it.

neha-Gupta1, line 594: Should the actual error also get compared here? Because currently we are just checking if err is not nil. And error could be at any level.

ryanfkeepers: We can rely on lower level tests for that granularity.

neha-Gupta1 (approved):

PR #4162, opened 2023-08-30
split message ids into added, deleted

ryanfkeepers:
splits the retrieval of channel message ids into two maps: one for new (added) messages, another for deleted messages.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/channels_pager_test.go
+++ b/src/pkg/services/m365/api/channels_pager_test.go
@@ -56,27 +56,28 @@ func (suite *ChannelsPagerIntgSuite) TestEnumerateChannelMessages() {
 	ctx, flush := tester.NewContext(t)
 	defer flush()
 
-	msgIDs, du, err := ac.GetChannelMessageIDsDelta(
+	addedIDs, _, du, err := ac.GetChannelMessageIDsDelta(
 		ctx,
 		suite.its.group.id,
 		suite.its.group.testContainerID,
 		"")
 	require.NoError(t, err, clues.ToCore(err))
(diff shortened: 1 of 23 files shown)


src/pkg/services/m365/api/channels_pager_test.go
```diff
@@ -56,27 +56,28 @@ func (suite *ChannelsPagerIntgSuite) TestEnumerateChannelMessages() {
...
-	msgIDs, du, err = ac.GetChannelMessageIDsDelta(
+	addedIDs, deletedIDs, du, err := ac.GetChannelMessageIDsDelta(
 		ctx,
 		suite.its.group.id,
 		suite.its.group.testContainerID,
 		du.URL)
 	require.NoError(t, err, clues.ToCore(err))
-	require.Empty(t, msgIDs, "should have no new messages from delta")
+	require.Empty(t, addedIDs, "should have no new messages from delta")
+	require.Empty(t, deletedIDs, "should have no deleted messages from delta")
```
neha-Gupta1, line 76: A test case with few deleted messages can be added.

ryanfkeepers: This is an integration test that is run with a minimum time between incrementals to ensure that we receive zero data when no changes are made.  Setting up to test deleted items would require either mocking the api (somewhat pointless at this layer) or actually deleting items (which we don't need to engage in).  Therefore: no tests for deletions at this time.

PR #4173, opened 2023-08-31
fixing bugs found during extensive manual tests

ryanfkeepers:
fix up the following bugs:
* reinstate previousPath in the metadata retrieval
* implement DoNotMergeItems for real
* nil pointer protection for the 'from' property
* some additional logging

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


website/docs/support/known-issues.md
```diff
@@ -25,3 +25,5 @@ Below is a list of known Corso issues and limitations:
   not be inheritable by children.
 
 * Link shares with password protection can't be restored.
+
+* All replies in a Teams Conversation are removed from backup when the root message gets deleted.
```
ryanfkeepers, line 29: Not sure if this explanation is sufficient.  Thoughts?  @gmatev, pinging for scrutiny here.

vkamra: @ryanfkeepers - can you clarify (here in the PR comment ) how the user gets into the situation and what the impact is? Based on that - can review what should be added to known-issues.

ryanfkeepers: Sure.  Easiest way to explain is with this image.
image

What's going on here is that I created a conversation, replied to it multiple times, and then deleted the initial message.  In the Teams app, the conversation survives with a "deleted" root message and all replies intact.

However, when utilizing the API, these properties are unavailable.  Replies are only retrievable with a message ID, and the deleted message ID now returns 404 for both GetMessage and GetMessageReplies endpoints.

This could lead users to expect that all replies are backed up and restorable, while in Corso I'm unable to get any reply in the conversation.

meain (approved):

website/docs/support/known-issues.md
gmatev: @ryanfkeepers as phrased, I think this is pretty confusing as stated. 

 I would probably remove this all together or replace it with something along the lines of "Message replies are only backed up when their corresponding parent message is still available at the time of backup".

ryanfkeepers: will do, thanks.

Issue #4175, opened 2023-09-01
[Bug]: Backup of german exchange contacts doesn’t work


lts-developer:
### What happened?

A new backup of the exchange contacts on an empty storage doesn’t work, when the language setting on the website https://outlook.office.com/mail/ (> Settings > General > Language and time > Language) is any other than “English” (tested and doesn’t work with: “German” and “Italian”).
The backup job status says “Completed”, but when you try to restore the contacts with Corso CLI you get “no items match the provided selectors”.
It maybe has something to do with the Outlook-folder-names. In English “Inbox”, “Contacts”, etc. and in German “Posteingang”, „Kontakte“, etc.?
Tested with a new full backup and with incremental backups.

Another language-related problem is: 
When you change the language and “Rename default folders so their names match the specified language” on the website https://outlook.office.com/mail/ the incremental (and maybe even the new full backup, if there is an existing backup of the mailbox on the storage) backup completely fail with the error:
           running backup: merging details: getting updated info for entry: getting new paths: empty location key
(As soon as you switch back to the previously set language the backup is working again as usual.)
Tested with “German” and “Italian”.

./corso backup create exchange --mailbox [address@office365.com](mailto:address@office365.com)
Connecting to repository:      1s done
Connecting to M365:         done
Connecting to M365:         done
Backing Up ∙ [address@office365.com](mailto:address@office365.com)
Discovering items to backup:      4s done
∙ contacts:      0s done
∙ email:      3s done
∙ events:      0s done
Backing up data:      2s done
Error:
        running backup: merging details: getting updated info for entry: getting new paths: empty location key
Completed Backups:
No backups available
Error: 1 of 1 backups failed

### Corso Version?

Corso v0.12.0

### Where are you running Corso?

Linux x86_64 - Debian 11

ryanfkeepers:
Thanks for the report, @lts-developer!  We're digging into the root cause now.

ryanfkeepers:
Alrighty!  Quick follow-up.  We are able to repro both issues.  For the first one (new backup) we also have a root cause: one line of our code is comparing the conatacts folder display name with the graph api's english [well-known folder name](https://learn.microsoft.com/en-us/graph/api/resources/mailfolder?view=graph-rest-1.0) const.  For the second issue (incremental) we don't have a root cause yet, but with a functional repro we should have that figured out in short order.

[...]

PR #4180, opened 2023-09-01
centralize contact location in cache resolver

ryanfkeepers:
moves all handling of the "contacts" root folder presence into the contacts container cache.  This ensures we can remove any one-off code specifically targeting contacts' peculiar root folder handling out of generic code paths and into a data-specific owner.

The addition of a "defaultRootLocation" api is necessary otherwise the cache lookup on restore always fails (or worse, gets confused by the well known folder and an identically named subfolder), and we end up with 409 responses and other issues.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4175 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/backup_test.go
+++ b/src/internal/m365/collection/exchange/backup_test.go
@@ -91,12 +91,10 @@ func (mg mockGetter) GetAddedAndRemovedItemIDs(
 
 var _ graph.ContainerResolver = &mockResolver{}
 
-type (
-	mockResolver struct {
-		items []graph.CachedContainer
-		added map[string]string
-	}
-)
+type mockResolver struct {
+	items []graph.CachedContainer
+	added map[string]string
+}
 
 func newMockResolver(items ...mockContainer) mockResolver {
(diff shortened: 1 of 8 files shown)


ashmrtn:
this looks good though there's one thing I'm unsure of. Could you please check the backup details for a backup with another contact folder besides the default? I'm wondering if now they'll be nested under the default folder

If we want to have LocationRef not nest them we'd probably need to tweak how we create folders (now we have a parent ID but don't want to recursively build the LocationRef with it). Calendars works around that by explicitly adding all fields in the call similar to [this](https://github.com/alcionai/corso/blob/8d2070409f7cdd96e0e4f96b52a2d48bd658ca6c/src/pkg/services/m365/api/contacts_pager.go#L71). That may keep us from properly handling subfolders of a child of the default folder though :/

It's fine to back-burner this if you don't want to try to work around all this

ryanfkeepers:
@ashmrtn going to close this out, due to staleness and higher priority from other projects

PR #4184, opened 2023-09-04
Add sites metadata in Groups backup

meain:
Add sites metadata (previous paths) under `groupID/libraries/sites/previouspath`.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/group_handler.go
+++ b/src/internal/m365/collection/drive/group_handler.go
@@ -65,7 +65,7 @@ func (h groupBackupHandler) MetadataPathPrefix(tenantID string) (path.Path, erro
 
 	p, err = p.Append(false, odConsts.SitesPathDir, h.siteID)
 	if err != nil {
-		return nil, clues.Wrap(err, "appending sites to metadata path")
+		return nil, clues.Wrap(err, "appending site id to metadata path")
 	}
 
 	return p, nil
--- a/src/internal/m365/collection/drive/group_handler.go
+++ b/src/internal/m365/collection/drive/group_handler.go
@@ -86,6 +86,17 @@ func (h groupBackupHandler) CanonicalPath(
 	)
 }
 
+func (h groupBackupHandler) SitePathPrefix(tenantID string) (path.Path, error) {
+	return path.Build(
+		tenantID,
+		h.groupID,
+		h.service,
+		path.LibrariesCategory,
+		false,
+		odConsts.SitesPathDir,
+		h.siteID)
+}
+
 func (h groupBackupHandler) IsAllPass() bool {
 	return h.scope.IsAny(selectors.GroupsLibraryFolder)
 }
--- a/src/internal/m365/service/groups/backup.go
+++ b/src/internal/m365/service/groups/backup.go
@@ -42,7 +42,7 @@ func ProduceBackupCollections(
 		categories           = map[path.CategoryType]struct{}{}
 		ssmb                 = prefixmatcher.NewStringSetBuilder()
 		canUsePreviousBackup bool
-		sites                = map[string]string{}
+		sitesPreviousPaths   = map[string]string{}
 	)
 
 	ctx = clues.Add(
@@ -72,8 +72,6 @@ func ProduceBackupCollections(
 				return nil, nil, false, err
 			}
 
-			sites[ptr.Val(resp.GetId())] = ptr.Val(resp.GetName())
-
 			pr := idname.NewProvider(ptr.Val(resp.GetId()), ptr.Val(resp.GetName()))
 			sbpc := inject.BackupProducerConfig{
 				LastBackupVersion: bpc.LastBackupVersion,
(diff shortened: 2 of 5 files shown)


src/internal/m365/service/groups/backup.go
```diff
@@ -118,5 +129,49 @@ func ProduceBackupCollections(
...
+
 	return collections, ssmb.ToReader(), canUsePreviousBackup, el.Failure()
 }
+
+func getSitesMetadataCollection(
+	tenantID, groupID string,
+	sites map[string]string,
+	su support.StatusUpdater,
+) (data.BackupCollection, error) {
+	// TODO(meain): Should we store this one level above? If so, how would we
```
meain, line 152: Let me know if anyone has any thoughts on the location/content for this metadata.

ryanfkeepers: I'm interpreting the question as, `"should we store the map of sites at the categoryType dir, instead of the sitePath-const dir?"`  If I read that correctly then, eh, it's sixes.  There's no difference in group resource (since one group holds many sites), and the siteIDs already compose unique keys.  But there's no harm in the namespacing.

src/internal/m365/service/groups/backup.go
```diff
@@ -77,15 +79,24 @@ func ProduceBackupCollections(
 				Selector:          bpc.Selector,
 			}
 
+			bh := drive.NewGroupBackupHandler(
+				bpc.ProtectedResource.ID(),
+				ptr.Val(resp.GetId()),
+				ac.Drives(),
+				scope,
+			)
```
ryanfkeepers, line 87:
```suggestion
				scope)
```

src/internal/m365/collection/drive/group_handler.go
```diff
@@ -63,6 +63,11 @@ func (h groupBackupHandler) MetadataPathPrefix(tenantID string) (path.Path, erro
 		return nil, clues.Wrap(err, "making metadata path")
 	}
 
+	p, err = p.Append(false, odConsts.SitesPathDir, h.siteID)
+	if err != nil {
+		return nil, clues.Wrap(err, "appending sites to metadata path")
```
ryanfkeepers, line 68:
```suggestion
		return nil, clues.Wrap(err, "appending site id to metadata path")
```

src/internal/m365/service/groups/backup.go
```diff
@@ -77,15 +79,24 @@ func ProduceBackupCollections(
...
+
+			cp, err := bh.SitePathPrefix(creds.AzureTenantID)
+			if err != nil {
+				return nil, nil, false, clues.Wrap(err, "getting canonical path")
+			}
+
+			sitesPreviousPaths[ptr.Val(resp.GetId())] = cp.String()
```
ryanfkeepers, line 94: This could use a comment for clarity.  It looks like siteID -> path, but the value in the map seems to lack utility.  Unless we're pointing to different prefixes, this should be easily (and consistently) reproducable just from the site ID.  So is the value there as a default, or is it expected to get used somewhere?  Could it instead be a `map[string]struct{}`?

src/internal/m365/service/groups/backup.go
```diff
@@ -118,5 +129,49 @@ func ProduceBackupCollections(
...
+		path.LibrariesCategory,
+		false)
+	if err != nil {
+		return nil, clues.Wrap(err, "making metadata path")
+	}
+
+	p, err = p.Append(false, odConsts.SitesPathDir)
+	if err != nil {
+		return nil, clues.Wrap(err, "appending sites to metadata path")
+	}
```
ryanfkeepers, line 167: Since this pattern is coming up a lot, and feels a bit clunky (imo), thoughts on cleaning it up with a `BuildMetadata` constructor similar to the current [path.Build()](https://github.com/alcionai/corso/blob/main/src/pkg/path/path.go#L134) func?

ryanfkeepers (commented):
I'd like to see a test that ensures correct production of metadata collections before approval.  Otherwise, LGTM!

src/internal/m365/service/groups/backup.go
ryanfkeepers: Ah, it came up in my own day.  So here's the change! https://github.com/alcionai/corso/pull/4192

src/internal/m365/service/groups/backup.go
meain: I don't see a use for it, but just thought I would retain a common format and name(`previouspath`) here. I think if we are changing the format, we should ideally also change the name to something new so that we don't break assumptions of the format within the file with the name `previouspath`. I am open to any thoughts on the name.

src/internal/m365/service/groups/backup.go
meain: Let me add a todo here for now, I'll update it once #4192 is merged.

meain:
Added a test as well @ryanfkeepers. This should now be good to go.

ryanfkeepers (approved):

PR #4186, opened 2023-09-04
Wire service isolation changes into SDK code

pandeyabs:
Wires `IsServiceEnabled` code into `UserHasMailbox`, `UserHasDrive`, and `UserGetMailboxInfo` SDK APIs.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

*  https://github.com/alcionai/corso/issues/3844

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -9,6 +9,7 @@ import (
 	msgraphgocore "github.com/microsoftgraph/msgraph-sdk-go-core"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/microsoftgraph/msgraph-sdk-go/users"
+	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/internal/common/idname"
 	"github.com/alcionai/corso/src/internal/common/ptr"
@@ -266,6 +267,18 @@ func EvaluateMailboxError(err error) error {
 	}
 
 	return err
+}
+
+// IsErrMailboxNotFound inspects the secondary errors inside MailboxInfo and
+// determines whether the resource has a mailbox.
+func IsErrMailboxNotFound(errs []error) bool {
+	for _, err := range errs {
+		if errors.Is(err, ErrMailBoxNotFound) {
+			return true
+		}
+	}
+
+	return false
 }
 
 func (c Users) GetMailboxSettings(
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -20,7 +20,7 @@ import (
 
 // Variables
 var (
-	ErrMailBoxSettingsNotFound     = clues.New("mailbox settings not found")
+	ErrMailBoxNotFound             = clues.New("mailbox not found")
 	ErrMailBoxSettingsAccessDenied = clues.New("mailbox settings access denied")
 )
 
@@ -215,7 +215,7 @@ func (c Users) GetInfo(ctx context.Context, userID string) (*UserInfo, error) {
 	}
 
 	if !mailFolderFound {
-		mi.ErrGetMailBoxSetting = append(mi.ErrGetMailBoxSetting, ErrMailBoxSettingsNotFound)
+		mi.ErrGetMailBoxSetting = append(mi.ErrGetMailBoxSetting, ErrMailBoxNotFound)
 		userInfo.Mailbox = mi
 
 		return userInfo, nil
--- a/src/pkg/services/m365/api/users.go
+++ b/src/pkg/services/m365/api/users.go
@@ -269,9 +269,9 @@ func EvaluateMailboxError(err error) error {
 	return err
 }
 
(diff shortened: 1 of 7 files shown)


src/pkg/services/m365/users.go
```diff
@@ -66,16 +67,20 @@ func UserHasMailbox(ctx context.Context, acct account.Account, userID string) (b
...
-	_, err = ac.Users().GetMailInbox(ctx, userID)
-	if err != nil {
-		if err := api.EvaluateMailboxError(err); err != nil {
-			return false, clues.Stack(err)
-		}
+	return exchange.IsServiceEnabled(ctx, ac.Users(), userID)
+}
 
-		return false, nil
+func UserGetMailboxInfo(
```
juleslasarte, line 73: Based on your PR in ark, it looks like with this change i need to do two calls to get this info for the user. I.e

```
if (userHasMailbox) {
   // check mailbox purpose
   }
```

otherwise the errors are not caught when the mailbox doesn't exist. I believe its all a call to the same api, which feels wasteful. Could GetMailbox info catch the errors and return a nil mailbox + a corso error such as "MailboxDoesntExist" ?

ryanfkeepers: That sounds like a reasonable request.

pandeyabs: > Could GetMailbox info catch the errors and return a nil mailbox + a corso error such as "MailboxDoesntExist" ?

Thanks. We do [catch this error](https://github.com/alcionai/corso/blob/main/src/internal/m365/service/exchange/enabled.go#L67) and return an embed MailboxInfo error `api.ErrMailBoxSettingsNotFound`, along with `error = nil`. 

Ark can inspect the embed error to figure out if a user has a mailbox, and then look at the purpose. This way we can eliminate the `UserHasMailbox` call and just rely on `UserGetMailboxInfo`.

Since `api.ErrMailBoxSettingsNotFound` is slightly misleading, I'll also fix it to say `api.ErrMailboxNotFound`.

Also, I am against returning a nil mailbox + corso error here, since it breaks the error reporting contract we have throughout the SDK layer.

pandeyabs: @juleslasarte @ryanfkeepers  this is ready for a final review

Changes:
1. I've renamed `api.ErrMailBoxSettingsNotFound` to `api.ErrMailboxNotFound`.
2. Introduced `IsErrMailboxNotFound` which can be used by ark (via its m365 connector) to inspect secondary `MailboxInfo` errors & check if a resource has a mailbox. 
3. The corresponding ark PR is now updated with this logic. 

There isn't a better way to do this without disturbing a lot of code, which I feel is an overkill. Also, even if we return `nil, corso error` instead of secondary errors, we would still need ark components to call into the m365 connector to interpret the corso error. Otherwise we risk bleeding corso dependencies into ark components.

src/pkg/services/m365/api/users.go
```diff
@@ -268,6 +269,18 @@ func EvaluateMailboxError(err error) error {
 	return err
 }
 
+// IsErrMailboxNotFound inspects the secondary errors inside MailboxInfo and
+// determines whether the resource has a mailbox.
+func IsErrMailboxNotFound(errs []error) bool {
```
ryanfkeepers, line 274: Can you please move this to the graph package, so that it can be co-located with the other errors comparators?

ryanfkeepers, line 274: Nit: `isAny` to imply that it will return true if any error matches.
```suggestion
func IsAnyErrMailboxNotFound(errs []error) bool {
```

src/pkg/services/m365/users.go
ryanfkeepers: > we would still need ark components to call into the m365 connector to interpret the corso error

We have a [package for that](https://github.com/alcionai/corso/blob/main/src/pkg/errs/errs.go), actually.  The idea there is that any granularity of low-level error in corso can be compred as a more generic, easily scrutinized sentinel.  That way, SDK users don't go digging into low level packages like m365.  They just check the error to see if it's a, say, "ServiceNotEnabled" err.

ryanfkeepers (approved):
handful of nits, but nothing blocking.  @juleslasarte should take another look as well.

src/pkg/services/m365/api/users.go
pandeyabs: This is not a graph error. It's defined in api.users

src/pkg/services/m365/users.go
pandeyabs: TIL, thanks. I'll keep this code as is to avoid further refactoring of shared code. It's an extra redirection or two to inspect secondary error code, which is fine imo.

juleslasarte (approved):

PR #4188, opened 2023-09-05
Fix import cycles in groups backup

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper_test.go
+++ b/src/internal/kopia/wrapper_test.go
@@ -26,7 +26,7 @@ import (
 	"github.com/alcionai/corso/src/internal/data"
 	dataMock "github.com/alcionai/corso/src/internal/data/mock"
 	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
-	m365Mock "github.com/alcionai/corso/src/internal/m365/mock"
+	mockColl "github.com/alcionai/corso/src/internal/m365/mock/collection"
 	exchMock "github.com/alcionai/corso/src/internal/m365/service/exchange/mock"
 	"github.com/alcionai/corso/src/internal/tester"
 	"github.com/alcionai/corso/src/pkg/backup/details"
@@ -1136,7 +1136,7 @@ func (suite *KopiaIntegrationSuite) TestBackupCollections_NoDetailsForMeta() {
 					streams = append(streams, ms)
 				}
 
-				mc := &m365Mock.BackupCollection{
+				mc := &mockColl.BackupCollection{
 					Path:    storePath,
 					Loc:     locPath,
 					Streams: streams,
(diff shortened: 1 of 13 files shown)


src/internal/kopia/wrapper_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/alcionai/corso/src/internal/data"
 	dataMock "github.com/alcionai/corso/src/internal/data/mock"
 	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
-	m365Mock "github.com/alcionai/corso/src/internal/m365/mock"
+	mockColl "github.com/alcionai/corso/src/internal/m365/mock/collection"
```
ryanfkeepers, line 29: nit: I would have stuck with the curent alias.  It doesn't match the package name as well, but `m365Mock.BackupCollection` is easier to grok than `mockColl.BackupCollection` (especially if/when we need to handle different providers).

src/internal/m365/mock/collection/collection.go
```diff
@@ -1,4 +1,4 @@
-package mock
+package collection
```
ryanfkeepers, line 1: Is this even specific to m365 at this point?  Could it get moved into the `data` package as a generic mock instead?  That feels like a much more appropriate owner.

src/internal/m365/service/groups/backup.go
```diff
@@ -93,7 +95,8 @@ func ProduceBackupCollections(
 					fb)
 
 				// Backup is always best effort and we don't error out even if this fails
-				operations.LogFaultErrors(ctx, fb.Errors(), "collecting metadata")
+				helpers.LogFaultErrors(ctx, fb.Errors(), "collecting site metadata")
+				logger.CtxErr(ctx, err).Info("collecting site metadata")
```
ryanfkeepers, line 99: very nitpicky: but this way the info log kinda appears like a header to the error list.
```suggestion
				logger.CtxErr(ctx, err).Info("collecting site metadata")
				helpers.LogFaultErrors(ctx, fb.Errors(), "collecting site metadata")
```

src/internal/operations/helpers/helpers.go
```diff
@@ -1,4 +1,4 @@
-package operations
+package helpers
```
ryanfkeepers, line 1: Generic package names like `"helpers"` are generally considered a smell by the golang community, as they indicate that the code either doesn't have a distinct enough identity to warrant separate ownership or that the identity is misunderstood, and the result often turns into a sort of dumping ground for code that could often be refined towards an identity

In this case, I think it's the latter: that the identity is misunderstood.  We don't want a blob of "helper" funcs.  So far, this package only contains errors-oriented functionality, so that sounds like the best choice of names: `/operations/errors/errors.go`.

That said, I'm also dubious about the need to move these files out at all.  The only imports are pkg/ control, fault, and logger, none of which are going to create an import cycle.  Seems to me that you could keep this in the original package as `operations/errors.go`.  That would help reduce additional imports, additional subpackages, and reduce the number of packages with a common name (ie: errors).  Unless I'm missing something about the imports, is it possible to keep this in the original package?

ryanfkeepers (changes requested):

meain:
Closing as we closed https://github.com/alcionai/corso/pull/4187

PR #4196, opened 2023-09-06
refactor api pagers to deduplicate spaghetti

ryanfkeepers:
the page-querying design in the api layer has grown organically from the roots of its implementation prior to centralizing calls in the api layer.  As a result, there's a lot of duplicate, or near- duplicate, code handling.  The duplication will, as we grow the layer, eventually lead to maintenance issues and "I forgot that spot" gotchas.

This PR is a first step towards wrangling that code into a clean and centralized pattern.
* Page controller structs are consistent for each type (one for deltas, one for non-deltas).
* All page controllers call into either enumerateItems or deltaEnumerateItems as their centralized runner.
* All variations on data control (only ids, collision keys, etc) are handled as a post-process to the enumeration calls, instead of duplicating the runner process itself in every iteration.

Future PRs can further reduce issues by:
* conforming container enumeration to the same pattern.
* moving pager loop processing out of higher layers and into the api layer.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -229,7 +229,9 @@ func runBackups(
 		err = bo.Run(ictx)
 		if err != nil {
 			if errors.Is(err, graph.ErrServiceNotEnabled) {
-				logger.Ctx(ctx).Infow("service not enabled", "resource_owner_name", bo.ResourceOwner.Name())
+				logger.Ctx(ctx).Infow("service not enabled",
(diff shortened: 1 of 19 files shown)


ryanfkeepers:
messed up the git commit, closing and remaking

PR #4200, opened 2023-09-06
ensure groups resource is set

ryanfkeepers:
ensure that the repository creation correctly sets the resource category to match the
service.

This change contains a lot of deletion of code where we had previously pushed ownership of identifying the resource category outward to upper layers of code.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3990

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/controller.go
+++ b/src/internal/m365/controller.go
@@ -59,7 +59,6 @@ type Controller struct {
 func NewController(
 	ctx context.Context,
 	acct account.Account,
-	rc resource.Category,
 	pst path.ServiceType,
 	co control.Options,
 ) (*Controller, error) {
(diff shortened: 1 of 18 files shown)


src/internal/m365/controller.go
```diff
@@ -75,6 +74,17 @@ func NewController(
 		return nil, clues.Wrap(err, "creating api client").WithClues(ctx)
 	}
 
+	rc := resource.UnknownResource
```
ryanfkeepers, line 77: This is the heart of the fix.  All other changes are deletions to comply with moving the rc identification into this func.

juleslasarte (approved):
this is working for me in ark!

Issue #3991, opened 2023-08-08
Teams Channels Export


ryanfkeepers:
# Work
## Packages

- [x]  `/src/pkg/<provider>/<service>/<collection-processor>.go` to handle category-specific processing.

## Interface & Data Compliance

- [x]  `/src/pkg/path/resource_path.go` - canonical enums for representing each category.
- [x]  `/src/pkg/selectors/<service>.go` - scope and const additions for the data category.
- [x]  `/src/pkg/backup/details/<service>.go` - details iteminfo representation for the category.
- [x] RepoRef and LocationRef layout is completed with examples.

## Enumeration

- [x]  All containers.
    - [ ]  May want a container cacher implementation.
- [x]  Items-in-container.
- [x]  Items-in-container by delta.
- [x]  All-items (if per-container delta is not supported).

## Export v0

- [x]  The service can export all data in a given category.
- [x]  Selectors can reduce the exported data set by specifying owners, containers, and items.
- [x]  Directory creation occurs if the export location does not currently exist.
- [x]  Convert individual items to the expected format from backup data
- [x]  Combine multiple items into a individual file for file formats that can hold multiple items (eg: pst exports for exchange)

ryanfkeepers:
@neha-Gupta1 heads up, I modified this ticket to cover `Export` instead of `Restore`, since we'll be implementing data exports first.

PR #4202, opened 2023-09-07
some code cleanup before groups export

ryanfkeepers:
various tidbits of data cleanup before moving forward with adding export behavior to groups.
* move duplicate collections mocks into data/mock
* move the export collection struct into pkg/export (to prevent future duplicates in the next PR)
* rename export.Collection to Collectioner, because it's an interface.
* some other non-logic rearrangement

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3991 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/export/export.go
```diff
@@ -16,28 +22,50 @@ type Collection interface {
...
+type BaseCollection struct {
+	// BaseDir contains the destination path of the collection.
+	BaseDir string
 
-	// Body is the body of the item. This is an io.ReadCloser and the
-	// SDK consumer is responsible for closing it.
-	Body io.ReadCloser
+	// BackingCollection is the restore collection from which we will
+	// create the export collection.
+	BackingCollection data.RestoreCollection
```
meain, line 32: `BackingCollection` sometimes have to be `[]data.RestoreCollection` for cases like pst export.

src/pkg/export/export.go
```diff
@@ -16,28 +22,50 @@ type Collection interface {
...
-	// the name of the item and an io.ReadCloser which contains the
-	// body of the item.
-	Data ItemData
+	// Name is the name of the item. This is the name that the item
+	// would have had in the service.
+	Name string
+
+	// Body is the body of the item. This is an io.ReadCloser and the
+	// SDK consumer is responsible for closing it.
+	Body io.ReadCloser
```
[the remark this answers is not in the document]
ryanfkeepers: I'm happy to make a code bump in ark to comply.  Should be a minimial change.

src/pkg/export/export.go
ryanfkeepers: Good to know!  I'll get that fixed up.

src/internal/m365/collection/drive/export.go
```diff
@@ -13,83 +13,61 @@ import (
...
-		name, err := getItemName(ctx, itemUUID, ec.backupVersion, ec.backingCollection)
-
-		ch <- export.Item{
-			ID: itemUUID,
-			Data: export.ItemData{
-				Name: name,
-				Body: item.ToReader(),
-			},
-			Error: err,
+	for _, rc := range drc {
```
meain, line 40: Although it can be multiple collections, as of now I don't see a case where anything drive/file related having multiple collections. It would be good to have a comment here that we are only expecting one collection in this.

meain (approved):

src/internal/m365/collection/drive/export.go
ryanfkeepers: Eh, that seems like a conflict of interest.  Either we need the slice or we don't.  But keeping it in and saying we don't want it feels weird.  I'm going to keep it in for now, and we can revisit if needed.

PR #4205, opened 2023-09-07
Incrementals for Groups SharePoint backup

meain:
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/data/mock/collection.go
```diff
@@ -100,3 +100,61 @@ func (c Collection) DoNotMergeItems() bool {
 func (c Collection) FetchItemByName(ctx context.Context, name string) (data.Item, error) {
 	return &Item{}, clues.New("not implemented")
 }
+
+type RestoreCollection struct {
```
meain, line 104: `BackupCollection` and `RestoreCollection` moved from `src/internal/m365/mock/collection.go`

src/internal/m365/backup.go
```diff
@@ -186,3 +188,56 @@ func verifyBackupInputs(sels selectors.Selector, cachedIDs []string) error {
 
 	return nil
 }
+
+func (ctrl *Controller) CollectMetadata(
```
meain, line 192: This does not really have to be a method on `Controller`.

meain:
@ryanfkeepers This is rewrite of https://github.com/alcionai/corso/pull/4187

src/internal/m365/backup.go
ryanfkeepers: The method design is appropriate in this case, actually, since it allows us to pass the controller into the metadata generation as an interface.  However, the larger function design has issues (see my other comment).

The only code we need here are lines 207-211 (and whatever else to transform them to a restore collection).

src/internal/m365/backup.go
```diff
@@ -186,3 +188,56 @@ func verifyBackupInputs(sels selectors.Selector, cachedIDs []string) error {
...
+		}
+
+		for _, fp := range filePaths {
+			rp, err := path.BuildRestorePaths(
+				reason.Tenant(),
+				reason.ProtectedResource(),
+				reason.Service(),
+				reason.Category(),
+				fp,
+			)
```
ryanfkeepers, line 225: We'll get the habit formed one of these days 😉 
```suggestion
				fp)
```

ryanfkeepers (changes requested):

src/internal/m365/backup.go
meain: I'm getting a linter for this. 🙈

src/pkg/path/path.go
```diff
@@ -127,6 +127,36 @@ type RestorePaths struct {
 	RestorePath Path
 }
 
+func BuildRestorePaths(
```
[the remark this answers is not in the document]
meain: I've changed it to use `BuildMetadata` with call to `Dir` at the call site.

src/internal/data/mock/collection.go
meain: I personally don't think we should fold them together. I think having two separate structs for backup and restore would be useful to keep the tests more readable. That said, if you have a strong opinion on this, I'll merge them.

src/internal/operations/manifests.go
```diff
@@ -127,52 +128,3 @@ func getManifestsAndMetadata(
 
 	return bb, collections, true, nil
 }
-
-// collectMetadata retrieves all metadata files associated with the manifest.
-func collectMetadata(
```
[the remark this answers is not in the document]
meain: I am not sure I fully agree with the argument. Here is what I think.

> Instead, this func shold remain where it is (we can assume we'll always need the basic set of metadata), and the provider/service-specific implementation should call a second, additional function that optionally extends the set of restoreCollections.

I don't think we should have the code first call this function to get the "basic" metadata and then call the method on `Controller` to get "additional metadata". This is introduces an unnecessary distinction between metadata belonging to services with only one underlying service and the rest(for example, in case of SP under group, there is no "basic" metadata).

> we have to duplicate those provider-agnostic parts

I don't think we are necessarily duplicating anything as the only implementation is in Controller and not in any of the specific services. It only calls out to providers that don't use the defaults like groups and asks it to return the set of metadata files we should use instead of the default set.

src/internal/data/mock/collection.go
ryanfkeepers: Outside of tests I agree with you.  But inside of tests, the behavior is so similar (and there's no overlaps) I don't see any value in it.  Personal opinion: production code should be as obvious as possible (in terms of ownership, behavior, etc), test code should be as streamlined/centralized as possible (to minimize variations, maintenance, one-off issues, gotchas, etc).  This addresses the most fundamental difficulties with each space: prod code can be difficult to grok (high rate of change, lots of one-off fixes, naming is difficult), test code can be difficult to maintain cleanly (it's easy to let test behavior slip behind prod behavior).

That said, this isn't a hard line.  Follow your gut and we'll be fine.

src/internal/operations/manifests.go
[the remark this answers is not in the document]
ryanfkeepers: Though the introduction of teams I'm increasingly bothered by the amount of code we've had to touch to enable or correct functionality purely by introducing the concept of the service.  The cognitive load is overwhelming, as is the amount of maintenance and the potential for bugs.  It's still early in the life of the repo and this problem will not scale well into the future.  "How do we do more with less work" is a philosophy we need to start adopting.

The fundamental question here is whether every provider should be _required_ to provide a metadata factory for the code to run successfully.  If we go that path, it's one more place where things don't work for the next addition until we add the support.  We don't often add providers, true.  But the direction still pushes us towards "do less with more".  Especially since this change removes code that is already doing more with less.

Anywhere that we can ensure the code works across changes without additional attention is good for our future maintenance and contribution, not to mention stability.

src/internal/data/mock/collection.go
meain: I have folded `BackupCollection` and `RestoreCollection` into Collection. We have to however have a separate `RestoreCollection` with the current implementation as this one takes a data.Collection instead of what the `BackupCollection` was taking.

src/internal/operations/manifests.go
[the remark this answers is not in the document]
meain: > Anywhere that we can ensure the code works across changes without additional attention is good for our future maintenance and contribution, not to mention stability.

I might be missing something, but I feel like the current code retains the old functionality in that any new additional services that gets added do not have to do anything if they use the default metadata location. This is done by making use of [`groups.MetadataFiles`](https://github.com/alcionai/corso/blob/dd7617385855ffc5bd33c38cbad94028c1a397bf/src/internal/m365/backup.go#L209). This changes gives us a place where we can hook into and use non standard metadata in case the service deviates from it.

This changes also has minimal impact compared to introducing a custom struct and difference between "standard/basic" metadata and "additional" metadata.

---

I might also be misunderstanding your ask here. Are you suggesting that we could use the `collectMetadata` function to get the default set and add others on top by appending to the same array? One problem that I can see with it is that `sites` in `groups` do not have the standard set of metadata files. So it is useless (and possibly error prone) to try to get the default set of metadata first.

src/internal/m365/collection/drive/collections.go
```diff
@@ -191,11 +191,11 @@ func deserializeMetadata(
 
 var errExistingMapping = clues.New("mapping already exists for same drive ID")
 
-// deserializeMap takes an reader and a map of already deserialized items and
+// DeserializeMap takes an reader and a map of already deserialized items and
```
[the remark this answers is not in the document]
ryanfkeepers: imo `/collection/drive` is appropariate owner for shared resources like this, so long as they're drive-ish specific.  `/service/drive` would not be appropriate, however.

src/internal/m365/backup.go
ryanfkeepers: I like the idea of returning paths.  That way we can centralize all metadata path -> collection generation within the operations layer, further removing it from provider/service ownership.  Better control over how we generate those collections that way.

src/internal/operations/manifests.go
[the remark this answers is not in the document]
ryanfkeepers: > any new additional services

additional services, yes.  But it doesn't set us up well for the next provider.

Am I in the weeds a bit here?  Yes, absolutely, but even in the weeds the issue stands.  Perhaps a better way to frame my concern is from this perspective: if this code were its very first introduction- ie, the first time we started reading metadata before a backup- I'd have no problem moving forward with what you've written.  Without any other implementation to compare against, this is a perfectly adequate solution.  It'd get a pass.  Then, in the future, I'd be even happier to see a PR that refines this implementation by moving it out of the m365 package to treat it as a service-agnostic process.  Which would put things, more or less, at our current implementation.

It's not that your implementation has issues or significant impact.  It's that you're walking backwards from a goal post we've already reached.  This is why I like Ash's suggestion to have the provider return paths instead of collections.  It keeps us walking along the vector that got us past this goal post in the first place.  My prior suggestion involved redistributing metadata collection generation, which is also a step back from where we're at.  Is it also an acceptable solution?  Sure.  But not redistributing that code again is a better solution.

> Are you suggesting that we could use the collectMetadata function to get the default set and add others on top by appending to the same array? One problem that I can see with it is that sites in groups do not have the standard set of metadata files. So it is useless (and possibly error prone) to try to get the default set of metadata first.

This was my suggestion, yes.  Standard and non-standard files shouldn't pose any problems here because we're not passing files or file content around, we're using collections, and collections as a structure are resilient to differences in both the files and the directories they contain.  It is up to the service to describe the paths to those non-standard files, and it is up to that same service to consume those collections to retrieve and parse the files.  Those details are opaque to all other processes and consumers.

src/internal/m365/backup.go
meain: I like the idea of returning a list of paths from here. I have to however point out that in case of Groups SharePoint libraries, we first read the `.../libraries/previouspaths` file to figure out what sites are available in `groups.MetadataFiles` which will construct and read the metadata file for which we will need kopia to be passed into here.

src/internal/operations/manifests.go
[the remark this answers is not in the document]
meain: >  Standard and non-standard files shouldn't pose any problems here because we're not passing files or file content around, we're using collections, and collections as a structure are resilient to differences in both the files and the directories they contain. It is up to the service to describe the paths to those non-standard files, and it is up to that same service to consume those collections to retrieve and parse the files. Those details are opaque to all other processes and consumers.

I think you misunderstood my concern here. The "standard" set of metadata includes `.../libraries/{previouspaths,delta}` for libraries, which would be returned in the collections, but we do not write a `delta` file at the top level for libraries.

ryanfkeepers (approved):
reviewed in sync, everything should be ready to go!

src/internal/m365/backup.go
ashmrtn: that kind of makes it feel like having additional nesting to get the metadata is probably not the best choice in the long run. Having to pass kopia directly into the graph connector is going to further complicate already complex logic and lead to more possible failure scenarios

Are there viable alternatives to the layout that's in use at the moment? How much effort would be required to implement one of the alternative?

ashmrtn: alright, sorry, paged back in some context for this. To make sure I have the state of the world right, what we have for metadata is something like
```
- tenantID
  - GroupsMetadata
    - siteIDList
    - libraries
      - siteID1
        - previouspaths
        - previousdeltas
      - siteID2
        - previouspaths
        - previousdeltas
```

And it's the part where we need to know the siteID that's causing us problems right? We've essentially made a circular dependency in that in order to know the path for the metadata files for a site we need to know the ID of the site which is itself stored in a metadata file. We're also not currently resolving this circular dependency in an alternate way by, for example, enumerating the set of sites prior to asking for metadata (in which case the set of sites that we need to pull metadata for would be known and the siteIDList file would give us the set of deleted sites. Sites that have been deleted shouldn't require metadata from corso because we can just send a single collection at the site level marking the entire subtree deleted)

So given the above, the long-term options appear to be
1. keep what we have laid out re: layout and base selection
2. change layout so we don't need to know additional state to get all required metadata (i.e. collapse metadata into top-level files, store in more generic location, etc. The key is to pick some location that doesn't require additional information about the data we are/have backed up previously in order to find)
3. add more advanced logic in kopia to do wildcard and/or recursive matches so we just do something like "return me all collections with data under this prefix" though it'd get tricker to define what the restore path for those collections should be in the generic case
4. lookup the sites being backed up with graph connector prior to fetching metadata. This would remove the circular dependency because we'd know the paths we need to fetch metadata for. We should ignore errors if things aren't found for a site (i.e. a new site was added since the last backup). Other logic should handle if we're only missing one of the metadata files (i.e. have previous deltas but not previous paths)

meain: > We're also not currently resolving this circular dependency in an alternate way by, for example, enumerating the set of sites prior to asking for metadata (in which case the set of sites that we need to pull metadata for would be known and the siteIDList file would give us the set of deleted sites. Sites that have been deleted shouldn't require metadata from corso because we can just send a single collection at the site level marking the entire subtree deleted)

This sounds like a viable solution to me. We could implement this without having to make any major changes IIUC. The only trouble if anything would be is how we manage errors and making sure we are erroring out because it is a new site and not because we did not have it in kopia.

ashmrtn: >The only trouble if anything would be is how we manage errors and making sure we are erroring out because it is a new site and not because we did not have it in kopia.

IIRC, we already use a separate fault bus when fetching metadata so the errors from there shouldn't cause the backup to fail overall. Not finding metadata for a new site should be something we can ignore. It's new, we'll just do a full enumeration of the data in the site. Not finding metadata for a site we previously backed up should be manageable. We should mark all collections in it as new/do not merge and delete the old hierarchy that existed for the site (so we don't accidentally pull forward folders that were deleted)

Issue #3992, opened 2023-08-08
Teams Libraries Restore


ryanfkeepers:
# Work
## Packages

- [ ]  `/src/pkg/<provider>/<service>/<collection-processor>.go` to handle category-specific processing.

## Interface & Data Compliance

- [ ]  `/src/pkg/path/resource_path.go` - canonical enums for representing each category.
- [ ]  `/src/pkg/selectors/<service>.go` - scope and const additions for the data category.
- [ ]  `/src/pkg/backup/details/<service>.go` - details iteminfo representation for the category.
- [ ] RepoRef and LocationRef layout is completed with examples.

## Enumeration

- [ ]  All containers.
    - [ ]  May want a container cacher implementation.
- [ ]  Items-in-container.
- [ ]  Items-in-container by delta.
- [ ]  All-items (if per-container delta is not supported).

## Restore v0

- [ ]  The service can restore all data in a given category.
- [ ]  Selectors can reduce the restored data set by specifying owners, containers, and items.
- [ ]  Container creation occurs if the restore location does not currently exist.

## Restore v1

- [ ]  Existing items within a container are enumerated and produce a map of `collisionKeys` for use in advanced restore comparisons.
- [ ]  advanced restore location options
    - [ ]  restore in place
    - [ ]  restore to different resource
    - [ ]  restore to different location
- [ ]  advanced restore collision options `(replace, skip, copy)`
- [ ]  restore-to-different-resource (ie: different user, site, team, etc).

PR #4209, opened 2023-09-08
Restore for Groups SharePoint Sites

meain:
Still a few pending items like CLI handling and restoring specific items. Will be handled in a follow up.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3992

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/path/drive.go
```diff
@@ -27,6 +29,11 @@ func ToDrivePath(p Path) (*DrivePath, error) {
 			With("path_folders", p.Folder(false))
 	}
 
+	if p.Service() == GroupsService {
+		// Groups have an extra /sites/<siteID> in the path
+		return &DrivePath{DriveID: folders[3], Root: folders[4], Folders: folders[5:]}, nil
+	}
```
ryanfkeepers, line 35: Hmm, this brings me back to the concern about one-off service handling, and that we're doing too much of that.  As a preface: this is a more managable concern than other palces, and we're already doing a _lot_ of this, so I won't block the PR on it, but I do want to voice said concern.

Ideally, I think this func should accept path.Elements instead of a path, where that builder only holds the path tail containing the drive ID and onward.  And hat we hold the assumption that the caller knows how to shorten a path to that tail.  That way we can skip all "if service foo then bar" checks in favor of expecting a well-formed list of elements from packages which should already own the specific format of their path.

src/cli/restore/groups.go
```diff
@@ -77,5 +77,25 @@ func restoreGroupsCmd(cmd *cobra.Command, args []string) error {
...
+
+	if flags.RunModeFV == flags.RunModeFlagTest {
+		return nil
+	}
+
+	if err := utils.ValidateGroupsRestoreFlags(flags.BackupIDFV, opts); err != nil {
+		return err
+	}
+
+	sel := utils.IncludeGroupsRestoreDataSelectors(ctx, opts)
```
ryanfkeepers, line 91: What happens here if the user specifies channel data for restore?

src/cli/restore/groups.go
```diff
@@ -77,5 +77,25 @@ func restoreGroupsCmd(cmd *cobra.Command, args []string) error {
...
+	utils.FilterGroupsRestoreInfoSelectors(sel, opts)
```
ryanfkeepers, line 92: None of the info selectors have been added to the groups restore command.  Is that coming in a follow-up?

src/cli/restore/groups.go
ryanfkeepers, line 1: The teams file needs these same additions.

ryanfkeepers (commented):
I feel like most of these files should have tests, and I'm not seeing them.  This is a lot of addition of behavior for the service.

Also, please make sure you're duplicating cli design between both the groups and teams files for the time being.

src/cli/restore/groups.go
meain: Since we are dropping teams, I'm assuming this is not necessary anymore.

src/cli/restore/groups.go
meain: Yup, this is available in https://github.com/alcionai/corso/pull/4218

src/cli/restore/groups.go
meain: It is not something I have not added for now. Let me take this as a follow up.

src/pkg/path/drive.go
meain: Hmm, I understand your concern. I am not sure if passing in a trimmed path would be easy enough. We could however consider moving this to a service specific path func. In any case let me drop a TODO for now to be picked up later.

meain:
I've added tests except for e2e tests. Let me add those in a follow up PR. I'm not adding anything specific for teams as we are dropping them.

ryanfkeepers (approved):

PR #4225, opened 2023-09-11
use alias for teams cli

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
@meain you can ignore my other requests to duplicate behavior for teams files after this PR.

src/cli/restore/groups.go
```diff
@@ -41,6 +41,7 @@ func addGroupsCommands(cmd *cobra.Command) *cobra.Command {
 // TODO: correct examples
 const (
 	groupsServiceCommand          = "groups"
+	teamsServiceCommand           = "teams"
 	groupsServiceCommandUseSuffix = "--backup <backupId>"
 
 	groupsServiceCommandRestoreExamples = `# Restore file with ID 98765abcdef in Bob's last backup (1234abcd...)
```
meain, line 47: Should we mention something in the docs/examples that teams and groups are the same so as to avoid any confusion for the user?

meain (approved):
LGTM, we should probably consider documenting that teams is an alias, but approving to unblock.

PR #4228, opened 2023-09-12
Minor refactor of storage provider stringer

pandeyabs:
One of the setup PRs before introducing local storage support for corso.

Changes
1. Export `storage.storageProvider` and rename it to `storage.Provider`. Without the rename, linter errors out with `exported: type name will be used as storage.StorageProvider by other packages, and that stutters; consider calling this Provider`
2. Rerun stringer - `stringer -type=Provider -linecomment`
3. Introduce a new Provider - `FS` for filesystem. This is in line with kopia naming for local/network attached storage.
4. Fix hardcoded string values for storage providers in test code.
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/storage/storage.go
```diff
@@ -8,12 +8,13 @@ import (
...
+type Provider int
 
-//go:generate stringer -type=storageProvider -linecomment
+//go:generate stringer -type=Provider -linecomment
 const (
-	ProviderUnknown storageProvider = 0 // Unknown Provider
-	ProviderS3      storageProvider = 1 // S3
+	ProviderUnknown Provider = 0 // Unknown Provider
+	ProviderS3      Provider = 1 // S3
+	ProviderFS      Provider = 2 // FS
```
ryanfkeepers, line 17: FS is a bit difficult to parse if you don't already know what it means.  Can we use the phrase `LocalFileStorage` instead?  Clarity is preferred over brevity for not-often referenced consts like this.

ryanfkeepers (approved):
lgtm, just nits

src/pkg/storage/storage.go
```diff
@@ -24,7 +25,7 @@ var (
 // Storage defines a storage provider, along with any configuration
 // required to set up or communicate with that provider.
 type Storage struct {
-	Provider storageProvider
+	Provider Provider
```
pandeyabs, line 28: Need to fix this. Converting diff to draft for now.

src/pkg/storage/storage.go
pandeyabs: `LocalFileStorage` is misleading too. Since it can be network attached as well. Maybe `Filesystem` would work better.

src/pkg/storage/storage.go
pandeyabs: Re: `Provider Provider` field.

Looks like naming a struct field the same name as its type is idiomatic as per golang conventions. Also, `Type` suffix is discouraged. @ryanfkeepers Need your thoughts on this! Should I keep this as is, i.e. `Provider Provider`?

PR #4229, opened 2023-09-12
add non-delta message paging path

ryanfkeepers:
similar to exchange, there are conditions under
which channel messages are unable to use the
delta api.  In particular, when the channel has no valid email property.  This change ensures that
we enumerate the entire channel under those
conditions, instead of failing out.

Technically, the only situation we know of where
the email property is missing, and thus the channel cannot make delta queries, is when there are
zero messages in the channel already.  However,
there may be unknown cases that we haven't caught, so instead of skipping the channel we're going to
handle non-delta enumeration for future safeguarding.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3989 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/item_pager.go
+++ b/src/pkg/services/m365/api/item_pager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/alcionai/clues"
 
@@ -172,12 +173,15 @@ func deltaEnumerateItems[T any](
 // shared enumeration runner funcs
 // ---------------------------------------------------------------------------
 
+type addedAndRemovedHandler[T any] func(items []T) ([]string, []string, error)
+
 func getAddedAndRemovedItemIDs[T any](
 	ctx context.Context,
 	pager Pager[T],
 	deltaPager DeltaPager[T],
 	prevDeltaLink string,
 	canMakeDeltaQueries bool,
+	aarh addedAndRemovedHandler[T],
 ) ([]string, []string, DeltaUpdate, error) {
 	if canMakeDeltaQueries {
 		ts, du, err := deltaEnumerateItems[T](ctx, deltaPager, prevDeltaLink)
@@ -186,7 +190,7 @@ func getAddedAndRemovedItemIDs[T any](
 		}
 
 		if err == nil {
-			a, r, err := addedAndRemovedByAddtlData(ts)
+			a, r, err := aarh(ts)
 			return a, r, du, graph.Stack(ctx, err).OrNil()
 		}
 	}
@@ -198,13 +202,19 @@ func getAddedAndRemovedItemIDs[T any](
 		return nil, nil, DeltaUpdate{}, graph.Stack(ctx, err)
 	}
 
-	a, r, err := addedAndRemovedByAddtlData[T](ts)
+	a, r, err := aarh(ts)
 
 	return a, r, du, graph.Stack(ctx, err).OrNil()
 }
 
-type getIDAndAddtler interface {
+type getIDer interface {
 	GetId() *string
+}
+
+// for added and removed by additionalData[@removed]
+
+type getIDAndAddtler interface {
+	getIDer
 	GetAdditionalData() map[string]any
 }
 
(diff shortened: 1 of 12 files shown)


src/pkg/services/m365/api/item_pager.go
```diff
@@ -230,3 +240,33 @@ func addedAndRemovedByAddtlData[T any](items []T) ([]string, []string, error) {
...
+	for _, item := range items {
+		giaddt, ok := any(item).(getIDAndDeletedDateTimer)
+		if !ok {
+			return nil, nil, clues.New("item does not provide id and deleted date time getters").
+				With("item_type", fmt.Sprintf("%T", item))
+		}
+
+		// if the additional data contains a `@removed` key, the value will either
+		// be 'changed' or 'deleted'.  We don't really care about the cause: both
+		// cases are handled the same way in storage.
```
ashmrtn, line 263: nit: could you please update this comment block to match the code below it?

src/pkg/services/m365/api/item_pager.go
```diff
@@ -172,12 +173,15 @@ func deltaEnumerateItems[T any](
 // shared enumeration runner funcs
 // ---------------------------------------------------------------------------
 
+type addedAndRemovedHandler[T any] func(items []T) ([]string, []string, error)
+
 func getAddedAndRemovedItemIDs[T any](
```
ashmrtn, line 178: I'm a little concerned by the growing number of parameters we're gathering here. We may need to refactor this at some point in part to help that issue but also because I've been trying to get some stuff up and running for Exchange kopia assisted incrementals and I've found events doesn't fit the expected pattern. It doesn't return mod times in delta tokens which throws off some of the other logic in this function

ashmrtn (approved):

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: whoops, can do.

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: I actually just put this through a fairly heavy refactor and this is, I think, as tight as we're going to make the func.  Then again, this is unexported, and should only be getting called through funcs [like this](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/mail_pager.go#L282), which should provide a nicer api to interface with.  Additional parameters like this shouldn't upset anything that higher levels are trying to do.

What it sounds like you're bringing up isn't a concern with the function handle, so much as how we're natively reducing the data to ids, and that's causing you to slice out modified times.  Is that right?  If so, it should be a fairly light lift to extend that dataset.  And since I've already got myself elbow-deep in the code, I'd be happy to make the change.

ashmrtn: >What it sounds like you're bringing up isn't a concern with the function handle, so much as how we're natively reducing the data to ids, and that's causing you to slice out modified times

Unfortunately it's not quite that simple :( Graph's delta results don't have a mod time returned in them which leaves us up a creek a bit when it comes to following existing code paths we have. I was hoping that I could just switch this function to return `map[string]time.Time` for added items (and other functions that call this) so that we get that into the collection. We could then leverage that to delay fetching item data until we know we need it since the file path and mod time is sufficient to get kopia's cached item checks working

However, since we don't have mod times for some items we either need to split events off to use it's own stuff or figure out a more flexible way to deal with things here. I'll keep thinking on it tomrrow

ryanfkeepers: > However, since we don't have mod times for some items we either need to split events off to use it's own stuff or figure out a more flexible way to deal with things here. I'll keep thinking on it tomrrow

Incidentally, this shares some overlap with channel messages.  The messages themselves return their mod time.  However, since we're packing the replies into the message, that mod time won't reflect whether the entire message thread has had a more recent addition.  And at a surface glance, there's no easy way to get that info without enumerating all replies.

We might need to revisit the approach at a broad scale, beyond just the one or two services using it.  Maybe we can sync tomorrow and noodle on it together.

PR #4232, opened 2023-09-12
get replies when getting message

ryanfkeepers:
This was somehow sliced out of changes persisted
in prior branch merges.  It re-adds persisting replies as part of message content retrieval.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3989

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/pkg/services/m365/api/channels.go
+++ b/src/pkg/services/m365/api/channels.go
@@ -108,8 +108,6 @@ func (c Channels) GetChannelMessage(
 	ctx context.Context,
 	teamID, channelID, messageID string,
 ) (models.ChatMessageable, *details.GroupsInfo, error) {
-	var size int64
-
 	message, err := c.Stable.
 		Client().
 		Teams().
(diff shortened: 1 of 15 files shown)


src/pkg/services/m365/api/channels.go
```diff
@@ -169,15 +174,19 @@ func ChannelMessageInfo(
 		msgCreator = ptr.Val(from.GetUser().GetDisplayName())
 	}
 
+	if msg.GetBody() != nil {
+		content = ptr.Val(msg.GetBody().GetContent())
+	}
```
neha-Gupta1, line 179: Are we handling the replies of type other than `text` message. For eg, an attachment reply?

neha-Gupta1 (approved):
Looks good. No blocking changes required.

src/pkg/services/m365/api/channels.go
ryanfkeepers: All replies are ChatMessageable types.  An attachment-only reply may have zero length content, but it will otherwise function as normal.

PR #4233, opened 2023-09-13
uncomment code to remove deleted messages

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3989

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/groups/collection.go
+++ b/src/internal/m365/collection/groups/collection.go
@@ -142,12 +142,9 @@ func (col Collection) DoNotMergeItems() bool {
 
 // Item represents a single item retrieved from exchange
 type Item struct {
-	id string
-	// TODO: We may need this to be a "oneOf" of `message`, `contact`, etc.
-	// going forward. Using []byte for now but I assume we'll have
-	// some structured type in here (serialization to []byte can be done in `Read`)
+	id      string
 	message []byte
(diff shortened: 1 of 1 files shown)


src/internal/m365/collection/groups/collection.go
```diff
@@ -220,31 +217,30 @@ func (col *Collection) streamItems(ctx context.Context, errs *fault.Bus) {
...
+			defer func() { <-semaphoreCh }()
 
-	// 		col.stream <- &Item{
-	// 			id:      id,
-	// 			modTime: time.Now().UTC(), // removed items have no modTime entry.
-	// 			deleted: true,
-	// 		}
+			col.stream <- &Item{
+				id:      id,
+				modTime: time.Now().UTC(), // removed items have no modTime entry.
```
neha-Gupta1, line 232: can deleted Time be used here? Because it would be the last time item was modified.

neha-Gupta1 (approved):
Looks good other than a small comment which can be discussed.

src/internal/m365/collection/groups/collection.go
ryanfkeepers: It could, but we're better off using Now to ensure this item always gets processed.

Issue #1416, opened 2022-11-02
Support for local storage


gmatev:
for even easier evaluation allow using local storage for a test repo:

```
mkdir -p ~/.corso/repo

corso repo init filesystem --path ~/.corso/repo
corso repo connect filesystem --path ~/.corso/repo
```

Should be straightforward passthrough to Kopia https://kopia.io/docs/repositories/#local-or-network-attached-storage

vkamra:
This is done. https://corsobackup.io/docs/setup/repos/#filesystem-storage

PR #4237, opened 2023-09-13
Remove tight CLI coupling with S3 storage provider

pandeyabs:
One of the 2 remaining setup PRs before we can introduce local storage repos.

**Changes:**
1. Read storage provider(`provider`) from config file except for `repo init * ` or `repo connect *` commands.
2. Apply flag overrides based on provider type( e.g. `S3FlagOverrides` if provider is `S3`)
3. Propagate storage provider type to functions which read/write config. These functions arbitrate on config hierarchy - flags, env, config file, in that order. 

**Reasons**
* Reason 1 is needed is because config file is the source of truth for storage provider for all commands except `repo init` or `repo connect`. In the exception cases, we pick the provider in command (e.g. `s3`) as the source of truth. e.g. consider a `repo init s3`, followed by `repo init filesystem`.During `repo init filesystem`, config file would indicate `S3` provider, but the correct behavior here is to select `filesystem` provider.
* One alternative was to push provider from the init/connect cmds into an override flag, and let the config code decide on hierarchy. However, this felt hacky. provider here is not a flag to begin with. It's part of init/connect commands.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/1416

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -290,7 +290,16 @@ func genericDeleteCommand(
 
 	ctx := clues.Add(cmd.Context(), "delete_backup_id", bID)
 
-	r, _, _, _, err := utils.GetAccountAndConnect(ctx, pst, repo.S3Overrides(cmd))
+	provider, overrides, err := repo.GetStorageProviderAndOverrides(ctx, cmd)
+	if err != nil {
+		return Only(ctx, err)
+	}
+
+	r, _, _, _, err := utils.GetAccountAndConnect(
+		ctx,
+		pst,
+		provider,
+		overrides)
 	if err != nil {
 		return Only(ctx, err)
 	}
@@ -316,7 +325,16 @@ func genericListCommand(
 ) error {
 	ctx := cmd.Context()
 
-	r, _, _, _, err := utils.GetAccountAndConnect(ctx, service, repo.S3Overrides(cmd))
+	provider, overrides, err := repo.GetStorageProviderAndOverrides(ctx, cmd)
+	if err != nil {
+		return Only(ctx, err)
+	}
+
+	r, _, _, _, err := utils.GetAccountAndConnect(
+		ctx,
+		service,
+		provider,
+		overrides)
 	if err != nil {
 		return Only(ctx, err)
 	}
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -289,16 +289,10 @@ func genericDeleteCommand(
 
 	ctx := clues.Add(cmd.Context(), "delete_backup_id", bID)
 
-	provider, overrides, err := utils.GetStorageProviderAndOverrides(ctx, cmd)
-	if err != nil {
-		return Only(ctx, err)
-	}
-
-	r, _, _, _, err := utils.GetAccountAndConnect(
+	r, _, _, _, err := utils.GetAccountAndConnectWithOverrides(
 		ctx,
-		pst,
-		provider,
-		overrides)
+		cmd,
+		pst)
 	if err != nil {
 		return Only(ctx, err)
 	}
@@ -324,16 +318,10 @@ func genericListCommand(
 ) error {
 	ctx := cmd.Context()
 
-	provider, overrides, err := utils.GetStorageProviderAndOverrides(ctx, cmd)
-	if err != nil {
-		return Only(ctx, err)
-	}
-
-	r, _, _, _, err := utils.GetAccountAndConnect(
+	r, _, _, _, err := utils.GetAccountAndConnectWithOverrides(
 		ctx,
-		service,
-		provider,
-		overrides)
+		cmd,
+		service)
 	if err != nil {
 		return Only(ctx, err)
 	}
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/alcionai/corso/src/cli/flags"
 	. "github.com/alcionai/corso/src/cli/print"
-	"github.com/alcionai/corso/src/cli/repo"
 	"github.com/alcionai/corso/src/cli/utils"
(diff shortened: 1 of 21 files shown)


src/cmd/longevity_test/longevity.go
```diff
@@ -16,20 +16,25 @@ import (
...
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
+	"github.com/alcionai/corso/src/pkg/storage"
 	"github.com/alcionai/corso/src/pkg/store"
 )
 
 // deleteBackups connects to the repository and deletes all backups for
 // service that are at least deletionDays old. Returns the IDs of all backups
 // that were deleted.
+// Only supported for S3 repos currently.
```
pandeyabs, line 26: To extend support to non-s3 storage providers, we'd need to pass in new cli flags to longevity tests to indicate provider type.

src/cli/config/storage.go
```diff
@@ -92,7 +94,6 @@ func configureStorage(
 		return store, clues.Wrap(err, "reading s3 configs from corso config file")
 	}
 
-	s3Overrides(overrides)
```
pandeyabs, line 95: dead code removed.

src/cli/cli.go
```diff
@@ -64,15 +64,24 @@ func preRun(cc *cobra.Command, args []string) error {
 	avoidTheseDescription := []string{
 		"Initialize a repository.",
 		"Initialize a S3 repository",
+		"Connect to a S3 repository",
```
pandeyabs, line 67: Explanation:
* Consider a case where we do a `repo connect` but the config file is not present. In that case, `preRun` will fail with this PR because `GetStorageProviderAndOverrides` will return an error(config file not found).
* We could special case this scenario for connect and source provider from the command, but it's hacky and will make it worse.
* Other alternative is to skip telemetry for connect, like we do for init. Telemetry for following backup/restore/* cmds will work as expected.

src/cmd/longevity_test/longevity.go
ryanfkeepers: Will there be a PR coming for this support?  Seems like the change will be incomplete without it.

ryanfkeepers (approved):
nits about code location and some consolidation, but generally lgtm.  Approving to unblock.

src/cmd/longevity_test/longevity.go
ashmrtn: this is the longevity test command that is only used in longevity testing. I think for the moment at least we can punt adding support for this until a bit later. The commands like this are already a bit hacky because they hook into some parts of the CLI config parsing but not all of it. TBH I think refactoring the config parsing code so it's more portable would provide the most benefit towards bringing CI commands like this into a better state

src/cmd/longevity_test/longevity.go
```diff
@@ -16,20 +16,25 @@ import (
...
 func deleteBackups(
 	ctx context.Context,
 	service path.ServiceType,
 	deletionDays int,
 ) ([]string, error) {
 	ctx = clues.Add(ctx, "cutoff_days", deletionDays)
 
-	r, _, _, _, err := utils.GetAccountAndConnect(ctx, service, nil)
+	r, _, _, _, err := utils.GetAccountAndConnect(ctx,
```
ashmrtn, line 34: nit: could you please put the `ctx` parameter on a newline as well?
```suggestion
	r, _, _, _, err := utils.GetAccountAndConnect(
		ctx,
```

src/cli/backup/backup.go
```diff
@@ -290,7 +290,16 @@ func genericDeleteCommand(
...
+	provider, overrides, err := repo.GetStorageProviderAndOverrides(ctx, cmd)
+	if err != nil {
+		return Only(ctx, err)
+	}
+
+	r, _, _, _, err := utils.GetAccountAndConnect(
+		ctx,
+		pst,
+		provider,
+		overrides)
```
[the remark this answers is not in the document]
pandeyabs: Can't move `utils.GetStorageProviderAndOverrides` into `utils.GetAccountAndConnect` unfortunately. 

Main reason is that it'll break longevity tool which also uses `utils.GetAccountAndConnect`. This tool doesn't utilize corso config constructs like flags, config file. So reading from config file & applying flag overrides are not applicable here.

I'd prefer to leave this as is and address it with a proper config refactor later.

src/cli/repo/s3.go
```diff
@@ -228,7 +238,7 @@ func connectS3Cmd(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func S3Overrides(cmd *cobra.Command) map[string]string {
+func S3FlagOverrides(cmd *cobra.Command) map[string]string {
```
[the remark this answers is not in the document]
pandeyabs: Thanks! Done.

src/cli/cli.go
ryanfkeepers: Ah, check. Thanks for the clarity.

src/cli/backup/backup.go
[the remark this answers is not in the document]
ryanfkeepers: Well, one more pitch before we leave it.  What if you created a new func to combine the two, left `utils.GetAccountAndConnect` so that longevity can call it.  Then have everything else call the new func?  It's a minor touch to the code, and hits all the places you're already updating, making it fairly within scope of this change.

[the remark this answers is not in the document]
pandeyabs: Thanks! That's a good idea. Although it'll not completely eliminate calling `utils.GetStorageProviderAndOverrides` from cli. We will need it for `GetConfigFromRepo` which is directly used by `init` & `connect` commands.

I'll see what works best. Should be quick.

PR #4242, opened 2023-09-13
use display name to look up groups

ryanfkeepers:
Allows lookup of groups using their display name in addition to their ID.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3988 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/groups.go
```diff
@@ -106,14 +110,44 @@ func (c Groups) GetByID(
...
 
-	resp, err := service.Client().Groups().ByGroupIdString(identifier).Get(ctx, nil)
-	if err != nil {
-		err := graph.Wrap(ctx, err, "getting group by id")
+	ctx = clues.Add(ctx, "resource_identifier", identifier)
 
-		return nil, err
+	var group models.Groupable
+
+	if uuidRE.MatchString(identifier) {
```
neha-Gupta1, line 117: A very corner case- if `uuid` is itself used a display name, the code will think it's a group ID and fail finding the correct group.

meain: Hmm, I guess it is unlikely. I guess we could look for names if the uuid look up fails in case we are really concerned about it.

ryanfkeepers: No harm in attempting the fallback, I suppose.

neha-Gupta1 (approved):
Apart from a corner case everything looks good.

PR #4243, opened 2023-09-13
Add CLI for local storage

pandeyabs:
New commands to initialize & connect to a repo on local or network attached storage.
* `repo init filesystem --path /tmp/repo`
* `repo connect filesystem --path /tmp/repo`

Includes basic unit & e2e tests. More coverage to be added in a following PR to keep the size contained.

**Updates:**
* Added Repo path sanitization i.e. handle relative paths, make paths cross platform compatible, etc. 
* Removed retention artifacts, not supported for filesystem storage.
* cli docs - auto updated.
* Manually tested with all corso backup/restore/export commands.

**Doesn't include** 
1. Symlinks
2. User ids wiring into repo.
3. Repos documentation update - in an upcoming PR.
4. Prefix support -> kopia doesn't support prefixes for `filesystem` storage
5. More E2E tests. 

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #1416 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/storage/filesystem.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	c.Path = str.First(overrides[FilesystemPath], c.Path)
+
+	return c.validate()
+}
+
+// TODO(pandeyabs): We need to sanitize paths e.g. handle relative paths,
+// make paths cross platform compatible, etc.
```
ashmrtn, line 88: I think a call to `filepath.Abs(<input path>)` in the appropriate place will probably solve the issue. It should be done prior to writing values to the config file though (i.e. config file should contain absolute paths) otherwise things will get a little weird depending on the current working directory

I'm not sure what the symlink resolution is for `filepath.Abs`, but you can just leave a TODO for it probably

the `filepath` package is also already platform-specific, so it should handle all the checks for paths that don't adhere to the platform's expectations (i.e. wrong slash type) as well as give back the proper format

ashmrtn (commented):
generally lgtm with a few minor nits. Not sure what the broader expectations are around structure given other reviews for the PR before this one

src/cli/repo/filesystem_e2e_test.go
```diff
@@ -0,0 +1,166 @@
...
+		t,
+		[][]string{storeTD.AWSStorageCredEnvs, tconfig.M365AcctCredEnvs})})
+}
+
+func (suite *FilesystemE2ESuite) cleanup(t *testing.T, repoPath string) {
+	// sanity check that we are not deleting wrong directory by accident
+	assert.Contains(t, repoPath, os.TempDir())
+	err := os.RemoveAll(repoPath)
+	require.NoError(t, err, clues.ToCore(err))
+}
```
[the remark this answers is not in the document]
pandeyabs: Thanks! TIL.

src/pkg/storage/filesystem.go
pandeyabs: Yeah! I wanted to address all of these in a later PR. I think it makes sense to have `filepath.Abs` atleast. Thanks!

src/pkg/storage/testdata/storage.go
```diff
@@ -51,6 +52,30 @@ func NewPrefixedS3Storage(t tester.TestT) storage.Storage {
 	return st
 }
 
+func NewFilesystemStorage(t tester.TestT) storage.Storage {
+	now := tester.LogTimeOfTest(t)
+	dir := "filesystem_test" + now
+	repoPath := filepath.Join(t.TempDir(), dir)
```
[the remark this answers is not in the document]
pandeyabs: nope, we dont need it.

src/pkg/storage/filesystem.go
pandeyabs: I've added support to make absolute paths & handle special cases. See `MakeAbsoluteFilePath`. It's currently being invoked by `filesystem` CLI code, which is fine since the `path` is sourced from `init` and `connect` CLI flags ( and eventually written to config file).

Ideally it should live in `FilesystemConfig` scope but that'd need some error propagation changes in `Configurer` interfaces. I'll handle that in a later PR, or as part of bigger config refactor.

src/cli/cli.go
```diff
@@ -65,6 +65,8 @@ func preRun(cc *cobra.Command, args []string) error {
 		"Initialize a repository.",
 		"Initialize a S3 repository",
 		"Connect to a S3 repository",
+		"Initialize a repository on local or network storage.",
```
ryanfkeepers, line 68: we should probably make a const out of these one day.  Not an issue in this PR, just thinking out loud.

src/cli/utils/flags_test.go
```diff
@@ -122,3 +122,26 @@ func (suite *FlagUnitSuite) TestAddS3BucketFlags() {
...
+			assert.Equal(t, true, flags.SucceedIfExistsFV, flags.SucceedIfExistsFN)
+		},
+	}
+
+	flags.AddFilesystemFlags(cmd)
+
+	cmd.SetArgs([]string{
+		"test",
+		"--" + flags.FilesystemPathFN, "/tmp/test",
+		"--" + flags.SucceedIfExistsFN,
```
ryanfkeepers, line 142: This is missing some of the azure creds flag checks.  Would you mind adding those as well?  We'd like to have comprehensive flag coverage in these tests.

src/cli/utils/utils.go
```diff
@@ -243,12 +243,36 @@ func GetStorageProviderAndOverrides(
...
-		overrides = flags.S3FlagOverrides(cmd)
+		return provider, flags.S3FlagOverrides(cmd), nil
+	case storage.ProviderFilesystem:
+		return provider, flags.FilesystemFlagOverrides(cmd), nil
+	}
+
+	return provider, nil, clues.New("unknown storage provider: " + provider.String())
+}
+
+func MakeAbsoluteFilePath(p string) (string, error) {
```
ryanfkeepers, line 256: Seems like this has some opinions about what `p` should be.  Could you please add a func header describing expected i/o for different variations of p, so that future maintainers don't need to read the func exhaustively to understand its behavior?

src/pkg/storage/testdata/storage.go
```diff
@@ -51,6 +52,28 @@ func NewPrefixedS3Storage(t tester.TestT) storage.Storage {
...
+
+	st, err := storage.NewStorage(
+		storage.ProviderFilesystem,
+		&storage.FilesystemConfig{
+			Path: repoPath,
+		},
+		storage.CommonConfig{
+			Corso: GetAndInsertCorso(""),
+		})
+	require.NoError(t, err, "creating storage", clues.ToCore(err))
```
ryanfkeepers, line 72: Formatting is required if the assertion contains a message.
```suggestion
	require.NoErrorf(t, err, "creating storage: %+v", clues.ToCore(err))
```

src/pkg/storage/testdata/storage.go
```diff
@@ -51,6 +52,28 @@ func NewPrefixedS3Storage(t tester.TestT) storage.Storage {
 	return st
 }
 
+func NewFilesystemStorage(t tester.TestT) storage.Storage {
+	now := tester.LogTimeOfTest(t)
+	repoPath := filepath.Join(t.TempDir(), now)
+
+	err := os.MkdirAll(repoPath, 0700)
+	require.NoError(t, err, "creating filesystem repo", clues.ToCore(err))
```
ryanfkeepers, line 60: formatting also needed here.

src/cli/utils/flags_test.go
```diff
@@ -122,3 +122,26 @@ func (suite *FlagUnitSuite) TestAddS3BucketFlags() {
...
 }
+
+func (suite *FlagUnitSuite) TestAddFilesystemFlags() {
+	t := suite.T()
+
+	cmd := &cobra.Command{
+		Use: "test",
+		Run: func(cmd *cobra.Command, args []string) {
+			assert.Equal(t, "/tmp/test", flags.FilesystemPathFV, flags.FilesystemPathFN)
+			assert.Equal(t, true, flags.SucceedIfExistsFV, flags.SucceedIfExistsFN)
```
ashmrtn, line 133: nit: this can be shortened by using a more specific assertion
```suggestion
			assert.True(t, flags.SucceedIfExistsFV, flags.SucceedIfExistsFN)
```

src/internal/kopia/filesystem.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	fsCfg := cfg.(*storage.FilesystemConfig)
+	opts := filesystem.Options{
+		Path: fsCfg.Path,
+	}
+
+	store, err := filesystem.New(ctx, &opts, false)
+	if err != nil {
+		return nil, clues.Stack(err).WithClues(ctx)
```
ryanfkeepers, line 31: filesystem.New should be expected to attach clues from the ctx already.

Is a `clues.Wrap` more appropropriate than a Stack here? 
```suggestion
		return nil, clues.Stack(err)
```

src/internal/tester/tconfig/config.go
```diff
@@ -20,6 +20,7 @@ const (
 	TestCfgEndpoint        = "endpoint"
 	TestCfgPrefix          = "prefix"
 	TestCfgStorageProvider = "provider"
+	TestCfgFilesystemPath  = "path"
```
ryanfkeepers, line 23: `path` feels really generic.  `dir`, maybe?  or `filepath`?

src/cli/repo/filesystem.go
```diff
@@ -0,0 +1,215 @@
...
+	. "github.com/alcionai/corso/src/cli/print"
+	"github.com/alcionai/corso/src/cli/utils"
+	"github.com/alcionai/corso/src/internal/events"
+	ctrlRepo "github.com/alcionai/corso/src/pkg/control/repository"
+	"github.com/alcionai/corso/src/pkg/repository"
+	"github.com/alcionai/corso/src/pkg/storage"
+)
+
+const (
+	fsProviderCmd          = "filesystem"
```
ryanfkeepers, line 19: Nit: long form `Command` for consts, short for `Cmd` for funcs, please.  It's not much, but it keeps things consistent.
```suggestion
	fsProviderCommand          = "filesystem"
```

src/cli/repo/filesystem.go
```diff
@@ -0,0 +1,215 @@
...
+		Short:   "Connect to a repository on local or network storage.",
+		Long:    `Ensures a connection to an existing repository on local or network storage.`,
+		RunE:    connectFilesystemCmd,
+		Args:    cobra.NoArgs,
+		Example: fsProviderCmdConnectExamples,
+	}
+}
+
+// connects to an existing filesystem repo.
+func connectFilesystemCmd(cmd *cobra.Command, args []string) error {
```
ryanfkeepers, line 155: For another PR: could I ask you to create generic `connectRepo` and `initRepo` handlers, much like we do with other cli commands?  Since azure should be a fast-follow here, it'll cut down on the boilerplate duplication for that implementation.

src/internal/kopia/filesystem.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	fsCfg := cfg.(*storage.FilesystemConfig)
+	opts := filesystem.Options{
+		Path: fsCfg.Path,
+	}
+
+	store, err := filesystem.New(ctx, &opts, false)
```
ashmrtn, line 29: The final parameter seems to be whether kopia should attempt to create the parent directories for the repo. I get that always setting it to false makes tests easier, but can it lead to failures during CLI usage? We don't seem to have code at the moment to handle intermediate directory creation

This can be addressed later, but at the very least we should add a note to the CLI help explaining what directories need to exist prior to running corso so that users aren't surprised

ashmrtn (approved):

ryanfkeepers (commented):
Handful of nits, but generally lgtm.  Will let Ash take the final approval.

As a side note: this PR is larger than we normally like to see, and should be split between the CLI part and kopia part.  That'd make for a much easier review on our end.  Since the review is already effectively complete, that's something to keep in mind for future changes.

PR #4245, opened 2023-09-14
minimize channel messages exports

ryanfkeepers:
reduces channel message export data to the minimal set of valuable info: message content, creator,
creation and modification time, and replies (each
reply has the same data, sans other replies).

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3991  

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/export/groups_test.go
+++ b/src/cli/export/groups_test.go
@@ -71,6 +71,7 @@ func (suite *GroupsUnitSuite) TestAddGroupsCommands() {
 
 				// bool flags
 				"--" + flags.ArchiveFN,
+				"--" + flags.RawFN,
 			})
 
 			cmd.SetOut(new(bytes.Buffer)) // drop output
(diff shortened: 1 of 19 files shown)


src/cli/export/groups_test.go
```diff
@@ -71,6 +71,7 @@ func (suite *GroupsUnitSuite) TestAddGroupsCommands() {
 
 				// bool flags
 				"--" + flags.ArchiveFN,
+				"--" + flags.RawFN,
```
meain, line 74: I was thinking we could add something like `--format raw`. This could be later expanded to `--format html` or something for an html export of the chats. These would be similar for exchange where the formats would be `pst`, `mbox` etc.

Related: https://github.com/alcionai/corso/blob/edf753382ea99cb255f0ac1747adb97dcd5b8069/src/pkg/control/export.go#L15

ryanfkeepers: sure, sounds good to me

meain (approved):
LGTM, unless @gmatev has any additional thoughts on the flags.

PR #4248, opened 2023-09-14
Add incremental backup test for groups

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/3990

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/metadata.go
+++ b/src/internal/m365/collection/exchange/metadata.go
@@ -15,13 +15,13 @@ import (
 
 // MetadataFileNames produces the category-specific set of filenames used to
 // store graph metadata such as delta tokens and folderID->path references.
-func MetadataFileNames(cat path.CategoryType) []string {
+func MetadataFileNames(cat path.CategoryType) [][]string {
 	switch cat {
 	// TODO: should this include events?
 	case path.EmailCategory, path.ContactsCategory:
-		return []string{metadata.DeltaURLsFileName, metadata.PreviousPathFileName}
+		return [][]string{{metadata.DeltaURLsFileName}, {metadata.PreviousPathFileName}}
 	default:
-		return []string{metadata.PreviousPathFileName}
+		return [][]string{{metadata.PreviousPathFileName}}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


src/internal/m365/collection/exchange/metadata.go
```diff
@@ -15,13 +15,13 @@ import (
 
 // MetadataFileNames produces the category-specific set of filenames used to
 // store graph metadata such as delta tokens and folderID->path references.
-func MetadataFileNames(cat path.CategoryType) []string {
+func MetadataFileNames(cat path.CategoryType) [][]string {
```
ryanfkeepers, line 18: Okay, so, I think I grok _why_ this change is wanted.  But I'm not seeing where the changes are consumed.  I'd expect some caller of this func would need to change its behavior based on the return value, or that the func it passes that return value into would need to change, but I'm not seeing either of those.  Am I missing it?  Where is this double slice utilized?

meain: This change is consumed by [`checkMetadataFileExists`](https://github.com/alcionai/corso/blob/0748e138bc6c234b8ee9462875369721af486f96/src/internal/operations/test/helper_test.go#L273). The signature for that has changed in this PR. This change is also reflected [here](https://github.com/alcionai/corso/blob/0748e138bc6c234b8ee9462875369721af486f96/src/internal/operations/test/exchange_test.go#L246).

ryanfkeepers: This is _only_ used in testing?  That feels... weird.  Going to approve to unblock, but I think that func should be moved into the test package if its not used in live code.

ryanfkeepers (approved):

src/internal/m365/collection/exchange/metadata.go
meain: Let me take it up separately.

PR #4251, opened 2023-09-14
Introduce new interfaces for storage configuration

pandeyabs:
Introducing a new `Configurer` interface to abstract out storage config information(for s3, filesystem etc) from caller code. I consider this as a short term solution. We need to consolidate overall config handling in a better way, but that's out of scope for this PR chain.

Testing
* Most of the changes here are code movement under the hood. So relying on existing tests.
* I'll address any test gaps in a later PR.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/1416

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/backup/helpers_test.go
+++ b/src/cli/backup/helpers_test.go
@@ -140,8 +140,10 @@ func prepM365Test(
 		recorder = strings.Builder{}
 	)
 
-	cfg, err := st.S3Config()
+	sc, err := st.StorageConfig()
 	require.NoError(t, err, clues.ToCore(err))
+
+	cfg := sc.(*storage.S3Config)
 
 	force := map[string]string{
 		tconfig.TestCfgAccountProvider: account.ProviderM365.String(),
--- a/src/cli/flags/s3.go
+++ b/src/cli/flags/s3.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/credentials"
 	"github.com/alcionai/corso/src/pkg/storage"
 )
@@ -55,8 +54,7 @@ func S3FlagOverrides(cmd *cobra.Command) map[string]string {
 
 func PopulateS3Flags(flagset PopulatedFlags) map[string]string {
 	s3Overrides := make(map[string]string)
-	// TODO(pandeyabs): Move account overrides out of s3 flags
-	s3Overrides[account.AccountProviderTypeKey] = account.ProviderM365.String()
+
 	s3Overrides[storage.StorageProviderTypeKey] = storage.ProviderS3.String()
 
 	if _, ok := flagset[AWSAccessKeyFN]; ok {
--- a/src/cli/flags/s3.go
+++ b/src/cli/flags/s3.go
@@ -53,9 +53,9 @@ func S3FlagOverrides(cmd *cobra.Command) map[string]string {
 }
 
 func PopulateS3Flags(flagset PopulatedFlags) map[string]string {
-	s3Overrides := make(map[string]string)
-
-	s3Overrides[storage.StorageProviderTypeKey] = storage.ProviderS3.String()
+	s3Overrides := map[string]string{
+		storage.StorageProviderTypeKey: storage.ProviderS3.String(),
+	}
 
 	if _, ok := flagset[AWSAccessKeyFN]; ok {
 		s3Overrides[credentials.AWSAccessKeyID] = AWSAccessKeyFV
--- a/src/pkg/storage/s3.go
+++ b/src/pkg/storage/s3.go
@@ -109,7 +109,7 @@ func buildS3ConfigFromMap(config map[string]string) (*S3Config, error) {
 	return c, c.validate()
 }
 
-func (c *S3Config) validate() error {
+func (c S3Config) validate() error {
 	check := map[string]string{
 		Bucket: c.Bucket,
 	}
@@ -133,7 +133,7 @@ func s3Overrides(in map[string]string) map[string]string {
 	}
 }
 
-func (c *S3Config) s3ConfigsFromStore(kvg KVStoreGetter) {
+func (c *S3Config) s3ConfigsFromStore(kvg Getter) {
 	c.Bucket = cast.ToString(kvg.Get(BucketNameKey))
 	c.Endpoint = cast.ToString(kvg.Get(EndpointKey))
 	c.Prefix = cast.ToString(kvg.Get(PrefixKey))
@@ -141,7 +141,7 @@ func (c *S3Config) s3ConfigsFromStore(kvg KVStoreGetter) {
 	c.DoNotVerifyTLS = cast.ToBool(kvg.Get(DisableTLSVerificationKey))
 }
 
-func (c *S3Config) s3CredsFromStore(kvg KVStoreGetter) {
+func (c *S3Config) s3CredsFromStore(kvg Getter) {
 	c.AccessKey = cast.ToString(kvg.Get(AccessKey))
 	c.SecretKey = cast.ToString(kvg.Get(SecretAccessKey))
 	c.SessionToken = cast.ToString(kvg.Get(SessionToken))
@@ -150,7 +150,7 @@ func (c *S3Config) s3CredsFromStore(kvg KVStoreGetter) {
 var _ Configurer = &S3Config{}
 
 func (c *S3Config) ApplyConfigOverrides(
-	kvg KVStoreGetter,
+	kvg Getter,
 	readConfigFromStore bool,
 	matchFromConfig bool,
 	overrides map[string]string,
@@ -158,13 +158,8 @@ func (c *S3Config) ApplyConfigOverrides(
 	if readConfigFromStore {
 		c.s3ConfigsFromStore(kvg)
 
-		if b, ok := overrides[Bucket]; ok {
-			overrides[Bucket] = common.NormalizeBucket(b)
-		}
-
-		if p, ok := overrides[Prefix]; ok {
-			overrides[Prefix] = common.NormalizePrefix(p)
-		}
+		overrides[Bucket] = common.NormalizeBucket(overrides[Bucket])
+		overrides[Prefix] = common.NormalizePrefix(overrides[Prefix])
 
 		if matchFromConfig {
 			providerType := cast.ToString(kvg.Get(StorageProviderTypeKey))
@@ -214,7 +209,7 @@ func (c *S3Config) ApplyConfigOverrides(
 var _ WriteConfigToStorer = &S3Config{}
 
 func (c *S3Config) WriteConfigToStore(
-	kvs KVStoreSetter,
+	kvs Setter,
 ) {
 	s3Config := c.normalize()
 
--- a/src/pkg/storage/s3.go
+++ b/src/pkg/storage/s3.go
@@ -4,6 +4,8 @@ import (
 	"strconv"
 
 	"github.com/alcionai/clues"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/spf13/cast"
 
 	"github.com/alcionai/corso/src/internal/common"
 	"github.com/alcionai/corso/src/internal/common/str"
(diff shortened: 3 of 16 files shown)


src/pkg/storage/s3.go
```diff
@@ -100,3 +121,107 @@ func (c S3Config) validate() error {
...
+
+			if err := mustMatchConfig(kvg, s3constToTomlKeyMap, s3Overrides(overrides)); err != nil {
+				return clues.Wrap(err, "verifying s3 configs in corso config file")
+			}
+		}
+	}
+
+	c.s3CredsFromStore(kvg)
+
+	aws := credentials.AWS{
```
pandeyabs, line 183: Note: I've made some changes to this code here. [Original code](https://github.com/alcionai/corso/blob/edf753382ea99cb255f0ac1747adb97dcd5b8069/src/cli/config/storage.go#L99) for reference.

**Changes:**
1. Original code was not populating `S3Config.AWS` if the creds are sourced from env vars. Fixed it. Note that this wasn't causing any issues because kopia calls into minio later which patches[ empty creds from env vars](https://github.com/minio/minio-go/blob/5aaca534bbcfdfcecbfd2126c494b154fde16564/pkg/credentials/env_aws.go#L41).
This allows us to test for config correctness.

2. We were validating aws credentials. This is unnecessary since kopia/minio will do that for us later anyway. I've removed this.

src/cli/flags/s3.go
```diff
@@ -55,8 +54,7 @@ func S3FlagOverrides(cmd *cobra.Command) map[string]string {
 
 func PopulateS3Flags(flagset PopulatedFlags) map[string]string {
 	s3Overrides := make(map[string]string)
-	// TODO(pandeyabs): Move account overrides out of s3 flags
-	s3Overrides[account.AccountProviderTypeKey] = account.ProviderM365.String()
+
 	s3Overrides[storage.StorageProviderTypeKey] = storage.ProviderS3.String()
```
ryanfkeepers, line 58: I can't include line 56 in this, but this can be shortened ot a single line:
```suggestion
	s3Overrides = map[string]string{storage.StorageProviderTypeKey: storage.ProviderS3.String()}
```

src/cli/backup/helpers_test.go
```diff
@@ -140,9 +140,11 @@ func prepM365Test(
 		recorder = strings.Builder{}
 	)
 
-	cfg, err := st.S3Config()
+	sc, err := st.StorageConfig()
 	require.NoError(t, err, clues.ToCore(err))
 
+	cfg := sc.(*storage.S3Config)
```
ryanfkeepers, line 146: Struct interpretation from interfaces should be avoided whenever possible.  The `S3Config` func is doing all the same behavior here, with better guarantees and clarity for end users.  I'd prefer we keep with that pattern.  "Accept interfaces, return structs", as the idiom goes.

src/pkg/storage/s3.go
```diff
@@ -66,29 +88,28 @@ func (c S3Config) StringConfig() (map[string]string, error) {
...
+		c.Prefix = orEmptyString(config[keyS3Prefix])
+		c.DoNotUseTLS = str.ParseBool(config[keyS3DoNotUseTLS])
+		c.DoNotVerifyTLS = str.ParseBool(config[keyS3DoNotVerifyTLS])
 	}
 
 	return c, c.validate()
 }
 
-func (c S3Config) validate() error {
+func (c *S3Config) validate() error {
```
ryanfkeepers, line 112: pointer receiver not needed unless we're modifying `c`, which it doesn't look like is happening.

src/pkg/storage/s3.go
ryanfkeepers: per 2/ can we get some examples of failed validation?  We want to ensure that the response to the user is able to be clearly understood as a credentials issue.

src/pkg/storage/s3.go
```diff
@@ -100,3 +121,107 @@ func (c S3Config) validate() error {
...
+	if readConfigFromStore {
+		c.s3ConfigsFromStore(kvg)
+
+		if b, ok := overrides[Bucket]; ok {
+			overrides[Bucket] = common.NormalizeBucket(b)
+		}
+
+		if p, ok := overrides[Prefix]; ok {
+			overrides[Prefix] = common.NormalizePrefix(p)
+		}
```
ryanfkeepers, line 167: so long as these funcs return an empty string when given an empty string, the conditionals are unnecessary.  Golang will automatically return an empty string if the key is not populated.
```suggestion
overrides[Bucket] = common.NormalizeBucket(overrides[Bucket])
overrides[Prefix] = common.NormalizePrefix(overrides[Prefix])
```

src/pkg/storage/storage.go
```diff
@@ -92,3 +93,79 @@ func orEmptyString(v any) string {
...
+func NewStorageConfig(provider ProviderType) (Configurer, error) {
+	switch provider {
+	case ProviderS3:
+		return &S3Config{}, nil
+	}
+
+	return nil, clues.New("unsupported storage provider: " + provider.String())
+}
+
+type KVStoreGetter interface {
```
ryanfkeepers, line 115: Nit: Simpler interface names are better.
```suggestion
type Getter interface {
```

src/pkg/storage/storage.go
```diff
@@ -92,3 +93,79 @@ func orEmptyString(v any) string {
...
+	Get(key string) any
+}
+
+type KVStoreSetter interface {
```
ryanfkeepers, line 119: Same nit
```suggestion
type Setter interface {
```

src/pkg/storage/storage.go
```diff
@@ -92,3 +93,79 @@ func orEmptyString(v any) string {
 
 	return v.(string)
 }
+
+func (s Storage) StorageConfig() (Configurer, error) {
```
ryanfkeepers, line 97: I'm dubious about the value of this function.  Noting how you're required to cast back to the service config struct in all locations of its usage, I am left with the expectation that:
1. All callers will need to have _a priori_ knowledge of which provider is in use before calling this func anyway.
2. Returning a `Configurer` interface isn't helping the callers, anyway, because the data within the config is their end goal, not the behavior of the configurer.

I think a different approach is in order.  I think sticking with the original `func (s Storage) S3Config() (S3Config, error)` will have more mileage here, purely due to 2/.  It will require callers to know which provider to request at the time of usage.  But that's already the pattern in place, a pattern that isn't changing as part of this PR.

Othewise you could go really big and ditch the provider-specific structs altogether in favor of a map (or map wrapper).  That'll let you more directly focus on what _data_ is being passed around, instead of the behavior built in to the data structure.

ryanfkeepers (changes requested):

src/pkg/storage/storage.go
ashmrtn: I think we should look for an acceptable middle-ground solution for the moment. The current goal is not to refactor all the config stuff, though there's already been a significant amount of that to enable local storage. Do we need to set time aside to do the refactor? Absolutely, but for the moment let's limit the scope of changes so we can finish up this feature

Looking at this PR and the next one, the reason we have this is so that callers can get a handle to the specific config for the provider without us having to create `ConfigForX` for every provider we have. Eventually we'll need to determine what solution we want to implement, but for now I think this is an alright tradeoff to make

If we want to make larger changes to how all this is handled I think we should come up with a plan that we agree on. This code can be either easy to handle in the future or very painful depending on what solution we pick so we should think carefully before changing things

src/cli/backup/helpers_test.go
ashmrtn: tbh I'm a bit on the fence about that one. It makes sense, until some code decides to call some non-interface function (or field) on the struct you returned when they really shouldn't be and there's coupling where you don't expect it

```go

type Bar interface {
  Print() string
}

type foo struct {}

func getFoo() *foo { return &foo{} }

func (f *foo) Print() string {
  return "hello world"
}

func (f *foo) PrettyPrint() string {
  return "hello world!"
}

func printer(b Baz) {
  fmt.Println(b.Print())
}

func main() {
  f := getFoo()
  // Now main is coupled to this specific implementation.
  fmt.Printf("got struct that says: %s\n", f.PrettyPrint())

  // printer still uses the interface though so we can change implementations.
  printer(f)
}
```

ryanfkeepers: Well, the coupling is effectively a guarantee because different service providers supply discrete data sets, and what the consumer is really after here is that data, not standardized behavior.  There is a requirement, at some level, of a priori knowledge about what data should be available or not, and no interface adequately satisfies that expectation.

Well, techincally there is an interface that satisfies it.  `GetData() map[string]string`.  Then we expect the caller to know what keys are associated with the expected provider.  But if they know the provider's keys, then they also know the provider, which means they'd also know to call `GetSpecificProviderConfig() SpecificProviderCfg`.

src/pkg/storage/storage.go
ryanfkeepers: > Looking at this PR and the next one, the reason we have this is so that callers can get a handle to the specific config for the provider without us having to create ConfigForX for every provider we have.

The difference between the two is reduced to:
```
// ConfigForX
func buildXConfig(base map[string]string) (XConfig, error) {...}

func XConfig() (Xconfig, error) {
  return buildXConfig(m)
} 

myCfg, err := XConfig()

// proposed
func buildXConfig(base map[string]string) (XConfig, error) {...}

func StorageConfig() (configurer, error) {
  return buildXConfig(m)
} 

iCfg, err := StorageConfig()

myConfig, ok := iCfg.(*Xconfig)
```

My point in writing that out is to show that the difference here is moot.  Interfaces cannot solve all problems and, unfortunately, this is one of those cases.  What we need here is not described by shared behavior.  And, sincerely, I'm not trying to make a soapbox out of it.  If I could come up with an interface which got around it (within a reasonable amount of work), I would go for that path.  But especially as we're looking for a more direct approach to the problem with less refactoring, I think the best course here is to accept that an interface is more abstraction than is beneficial, and to keep with the straightforward data production.  However inelegant that solution may seem.

src/cli/backup/helpers_test.go
ashmrtn: >the coupling is effectively a guarantee because different service providers supply discrete data sets, and what the consumer is really after here is that data, not standardized behavior

In this particular case that's true. However, it's not true in the general case, especially for code that could be refactored to be generic code that leverages interfaces/implementations

ryanfkeepers: Generally agreed, yes.  I still think that struct extraction from an interface is generally an anti-pattern in golang.  It smells like an attempt to recreate OOP covariance, which golang explicitly avoids supporting.  That suggests the design issue lies elsewhere (ex: if the interface doesn't provide the needed behavior, then we have the wrong interface), and the solution by casting to structure is a symptom of that other issue.

ryanfkeepers (approved):
approving to unblock.  Per sync, we can migrate ongoing design discussion elsewhere.

src/pkg/storage/storage.go
pandeyabs: > Othewise you could go really big and ditch the provider-specific structs altogether in favor of a map (or map wrapper)

That was my first instinct while doing this mini config refactor. However, I found that it may end up disturbing a lot of code. I am going to pass on this for now. 

> Returning a Configurer interface isn't helping the callers, anyway, because the data within the config is their end goal, not the behavior of the configurer.

This is partially correct. It's true that callers are interested in data within the config. However, in several cases the callers pass this `Configurer` interface to `WriteRepoConfig`. See [example](https://github.com/alcionai/corso/blob/03ed65a0c00c159a91897b3dc8aa365199bbaaec/src/cli/repo/s3.go#L152). You could argue here that we could ditch this in favor of a config map that is fed to viper (and eventually written to config). 
But `WriteRepoConfig` is called from several places and it's a tangled mess. Unraveling this will push back the local storage work out even further. 

So can we keep the code in this PR as is, unless you have strong reservations against it @ryanfkeepers ? :)

pandeyabs: I didn't do a great job explaining last friday, so let me retry. Consider this [example](https://github.com/alcionai/corso/blob/e50fe44d8a96b86e27f03998455c87e35aa76651/src/cli/utils/utils.go#L108) from the next PR in this chain. 

Here, I am no longer casting interface to a struct i.e. `AccountConnectAndWriteRepoConfig` and `WriteRepoConfig` are both agnostic to the storage provider. Here I need `StorageConfig()`, with the current setup, unless i 

> I'd like to follow up on the design discussion outside of this PR. But it's fine for now.

Just to be on the same page, since we've had a ton of discussion here, are you suggesting we do below for every storage provider,
```
func XConfig() (Xconfig, error) {
  return buildXConfig(m)
} 

```
instead of 
```
iCfg, err := StorageConfig()
myConfig, ok := iCfg.(*Xconfig)
``` ? 

Is this the only suggestion here, or are there more?

ryanfkeepers: > I didn't do a great job explaining last friday, so let me retry. Consider this [example](https://github.com/alcionai/corso/blob/e50fe44d8a96b86e27f03998455c87e35aa76651/src/cli/utils/utils.go#L108) from the next PR in this chain.

You did a fine job explaining.  The linked code is great works exactly like we'd want, given that it's acting within in a provider-agnostic space.

> Just to be on the same page, since we've had a ton of discussion here, are you suggesting we do below for every storage provider,

Yes, that is exaclty my request.  At least, for cases where we would otherwise cast from the interface to the struct.  We seem to have a generic use case for `StorageConfig`, which is a great case for it, and we should keep that around.  But in non-generic cases I don't believe we should be using it.

PR #4255, opened 2023-09-14
match item type in groups selectors info filter

ryanfkeepers:
adds item type comparisons to the info filter during groups selector reduction.  This ensures cross-
contamination of item types on shared info properties does not occur.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3988 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/selectors/groups.go
+++ b/src/pkg/selectors/groups.go
@@ -773,9 +773,9 @@ func (s GroupsScope) matchesInfo(dii details.ItemInfo) bool {
 	acceptableItemType := -1
 
 	switch infoCat.leafCat() {
-	case GroupsChannelMessage:
-		acceptableItemType = int(details.SharePointLibrary)
 	case GroupsLibraryItem:
+		acceptableItemType = int(details.SharePointLibrary)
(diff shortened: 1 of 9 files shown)


src/pkg/selectors/groups.go
```diff
@@ -770,6 +770,19 @@ func (s GroupsScope) matchesInfo(dii details.ItemInfo) bool {
 		return false
 	}
 
+	acceptableItemType := -1
+
+	switch infoCat {
```
meain, line 775: This seems a lot like [`leafCat`](https://github.com/alcionai/corso/blob/ce422f0b1d1d92756731b396290198a100fc4b25/src/pkg/selectors/groups.go#L556). Maybe we can reuse that?

ryanfkeepers: Hey!  I completely forgot about that 😅.  Thanks for the reminder, this is exactly what it's here for.

ashmrtn (approved):

PR #4257, opened 2023-09-15
handle path for channel

neha-Gupta1:
Handle the path creation for Onedrive and Channel.

- [ ] :no_entry: No

#### Type of change

- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual

### DIFF
--- a/src/internal/m365/service/onedrive/restore.go
+++ b/src/internal/m365/service/onedrive/restore.go
@@ -110,6 +110,10 @@ func AugmentRestorePaths(
 	colPaths := map[string]path.RestorePaths{}
 
 	for _, p := range paths {
+		if p.StoragePath.Category() == path.ChannelMessagesCategory {
+			continue
+		}
+
 		first := true
 
 		for {
--- a/src/internal/m365/service/onedrive/restore.go
+++ b/src/internal/m365/service/onedrive/restore.go
@@ -109,8 +109,13 @@ func AugmentRestorePaths(
 	// of the directory.
 	colPaths := map[string]path.RestorePaths{}
 
+	allowedCategories := map[path.CategoryType]struct{}{
+		path.LibrariesCategory: {},
+		path.FilesCategory:     {},
+	}
+
 	for _, p := range paths {
-		if p.StoragePath.Category() == path.ChannelMessagesCategory {
+		if _, ok := allowedCategories[p.StoragePath.Category()]; !ok {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


src/internal/operations/restore.go
```diff
@@ -430,7 +430,9 @@ func formatDetailsForRestoration(
 
 	if sel.Service == selectors.ServiceOneDrive ||
 		sel.Service == selectors.ServiceSharePoint ||
-		sel.Service == selectors.ServiceGroups {
+		(sel.Service == selectors.ServiceGroups &&
+			len(paths) > 0 &&
+			(fds.Items()[0].Groups.ItemType == details.SharePointLibrary)) {
```
meain, line 435: I don't think this is the right check. I think we should probably have this check inside `AugmentRestorePath` on individual path entries.

neha-Gupta1: Yes. Doing the changes

src/internal/m365/service/onedrive/restore.go
```diff
@@ -110,6 +110,10 @@ func AugmentRestorePaths(
 	colPaths := map[string]path.RestorePaths{}
 
 	for _, p := range paths {
+		if p.StoragePath.Category() == path.ChannelMessagesCategory {
```
meain, line 113: Looks good. We should probably enhance the test for `AugmentRestorePaths` though.

neha-Gupta1: Working on creating dataset for the testcase.

ryanfkeepers: This doesn't feel like the right ownership.  `/service/onedrive` should never see a ChannelMessage category path.  Are we sure this is in the right place?  Alternatively, are we sure there's not a more severe bug, of which this is only a symptom?

My assumption is that either: 1/ `AugmentRestorePaths` is owned by the wrong package and needs to be moved to a more generic location or 2/ this should be an inclusive check (ex: `if not category files/folders then continue`) instead of an exclusive one (ie: `allow everything except channel messages`).  2/ is probably the more correct solution.

ryanfkeepers (changes requested):

neha-Gupta1:
@ryanfkeepers this seems to fix the issue(sanity run gave expected results). But we can't completely ignore that it might be a bigger bug. 
And I will meanwhile incorporate the changes you suggested.

ryanfkeepers (approved):

PR #4260, opened 2023-09-15
Add export support for Group SharePoint

meain:
Ideally goes in after https://github.com/alcionai/corso/pull/4257 .
We could merge this instead of https://github.com/alcionai/corso/pull/4258 but I thought we could skip for now as this as been tested much less. But outside of me being paranoid, I think this should be in a good shape.

If this do go in, we can update the CHANGELOG entry to say that we do support export for SP libs.

This will also need more e2e(sanity) tests which I'll add in a follow up PR.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/4259

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/groups/export.go
+++ b/src/internal/m365/service/groups/export.go
@@ -25,7 +25,7 @@ func ProduceExportCollections(
 	exportCfg control.ExportConfig,
 	opts control.Options,
 	dcs []data.RestoreCollection,
-	backupDriveIDNames idname.CacheBuilder,
+	backupDriveIDNames idname.Cacher,
 	deets *details.Builder,
 	errs *fault.Bus,
 ) ([]export.Collectioner, error) {
@@ -51,7 +51,7 @@ func ProduceExportCollections(
 				[]data.RestoreCollection{restoreColl},
 				backupVersion,
 				exportCfg)
-		case path.FilesCategory:
+		case path.LibrariesCategory:
 			drivePath, err := path.ToDrivePath(restoreColl.FullPath())
 			if err != nil {
 				return nil, clues.Wrap(err, "transforming path to drive path").WithClues(ctx)
@@ -77,11 +77,11 @@ func ProduceExportCollections(
 			el.AddRecoverable(
 				ctx,
 				clues.New("unsupported category for export").With("category", cat))
+
+			continue
 		}
 
-		if coll != nil {
-			ec = append(ec, coll)
-		}
+		ec = append(ec, coll)
 	}
 
 	return ec, el.Failure()
--- a/src/internal/m365/service/groups/export.go
+++ b/src/internal/m365/service/groups/export.go
@@ -2,7 +2,7 @@ package groups
 
 import (
 	"context"
-	ospath "path"
+	stdlibpath "path"
 
 	"github.com/alcionai/clues"
 
@@ -67,7 +67,7 @@ func ProduceExportCollections(
 			}
 
 			folders := restoreColl.FullPath().Folders()
-			siteName := folders[1] // use siteID by fault
+			siteName := folders[1] // use siteID by default
 
 			webURL, ok := backupSiteIDWebURL.NameOf(siteName)
 			if !ok {
@@ -79,7 +79,7 @@ func ProduceExportCollections(
 				// We can't use the actual name anyways as it might
 				// contain invalid characters. This should also avoid
 				// possibility of name collisions.
-				siteName = ospath.Base(webURL)
+				siteName = stdlibpath.Base(webURL)
 			}
 
 			baseDir := path.Builder{}.
--- a/src/internal/m365/service/groups/export.go
+++ b/src/internal/m365/service/groups/export.go
@@ -3,12 +3,17 @@ package groups
 import (
 	"context"
 
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/common/idname"
 	"github.com/alcionai/corso/src/internal/data"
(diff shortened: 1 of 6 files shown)


src/internal/m365/service/groups/export.go
```diff
@@ -20,6 +25,7 @@ func ProduceExportCollections(
 	exportCfg control.ExportConfig,
 	opts control.Options,
 	dcs []data.RestoreCollection,
+	backupDriveIDNames idname.CacheBuilder,
```
ryanfkeepers, line 28: Write methods aren't required here.
```suggestion
	backupDriveIDNames idname.Cache,
```

src/internal/m365/service/groups/export.go
```diff
@@ -33,19 +39,48 @@ func ProduceExportCollections(
...
 		case path.ChannelMessagesCategory:
 			folders = append(folders, fp.Folders()...)
-		}
 
-		coll := groups.NewExportCollection(
-			path.Builder{}.Append(folders...).String(),
-			[]data.RestoreCollection{restoreColl},
-			backupVersion)
+			coll = groups.NewExportCollection(
+				path.Builder{}.Append(folders...).String(),
```
ryanfkeepers, line 50: I know this wasn't your addition, but it suddenly bothers me.
```suggestion
			coll = groups.
			    NewExportCollection(path.Builder{}.Append(folders...).
			    String(),
```

src/internal/m365/service/groups/export.go
```diff
@@ -33,19 +39,48 @@ func ProduceExportCollections(
...
-		ec = append(ec, coll)
+			coll = drive.NewExportCollection(
+				baseDir.String(),
+				[]data.RestoreCollection{restoreColl},
+				backupVersion)
+		default:
+			el.AddRecoverable(
+				ctx,
+				clues.New("unsupported category for export").With("category", cat))
+		}
```
ryanfkeepers, line 79:
```suggestion
				clues.New("unsupported category for export").With("category", cat))
				
				continue
		}
```

ryanfkeepers (approved):

src/internal/m365/service/groups/export.go
meain: LOL, not sure why I added a `nil` check instead of this. Good catch.

src/internal/m365/service/groups/export.go
meain: This looks like an odd way to split it. Any reason for keeping `String()` in a separate line? Just wanted to make sure you didn't mistype anything.

src/internal/m365/service/groups/export.go
```diff
@@ -33,21 +41,65 @@ func ProduceExportCollections(
...
+			if !ok {
+				// This should not happen, but just in case
+				logger.Ctx(ctx).With("site_id", folders[1]).Info("site weburl not found, using site id")
+			}
+
+			if len(webURL) != 0 {
+				// We can't use the actual name anyways as it might
+				// contain invalid characters. This should also avoid
+				// possibility of name collisions.
+				siteName = ospath.Base(webURL)
```
meain, line 82: The name of the site in restore is the basename from URL. This might not always be a 1:1 match.
Also, as of now, we are not storing the name of the site. Should we consider storing that along with `siteID` and `webURL`.
/cc @ryanfkeepers

meain:
@ryanfkeepers I was missing adding site name in path. It is required here as in future we will have multiple sites. I could use one more round of review here.

src/internal/m365/service/groups/export.go
ryanfkeepers: splitting on function chains within builder patterns is pretty standard.  This is effectively:
```go
err := do.
  foo().
  bar().
  baz()
```

Personally, when there's multiple funcs in a chain, I prefer this style over "split on the longest func param.  ie:
```go
// my favorite is above
// second favorite
err := do.
  foo(long, list, of, params).
  bar().
  
// second favorite pt 2
err := do.
  foo(
    even,
    longer,
    param,
    set).
  bar()
  
// not my favorite
err := do.foo(
    params, get, their, own, line,
  ).bar()
```

ryanfkeepers: wait a minute, I just realized I read the code wrong.  Yeah, what I suggested is gonzo.  Disregard this suggestion. 🤦

src/internal/m365/service/groups/export.go
ryanfkeepers: I think we're fine to leave it.  Since users can extrapolate the dir name from the webURL, and its already in a human readable format, this should be sufficient.

ryanfkeepers (approved):
some nits; lgtm

src/internal/m365/service/groups/export.go
```diff
@@ -33,21 +41,65 @@ func ProduceExportCollections(
...
+
+			driveName, ok := backupDriveIDNames.NameOf(drivePath.DriveID)
+			if !ok {
+				// This should not happen, but just in case
+				logger.Ctx(ctx).With("drive_id", drivePath.DriveID).Info("drive name not found, using drive id")
+				driveName = drivePath.DriveID
+			}
+
+			folders := restoreColl.FullPath().Folders()
+			siteName := folders[1] // use siteID by fault
```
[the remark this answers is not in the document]
meain: Woops!

PR #4265, opened 2023-09-15
minor touchups from manual testing

ryanfkeepers:

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/groups/backup.go
+++ b/src/internal/m365/service/groups/backup.go
@@ -67,9 +67,14 @@ func ProduceBackupCollections(
 			break
 		}
 
+		catStr := scope.Category().PathType().String()
+		if scope.Category().PathType() == path.ChannelMessagesCategory {
+			catStr = "messages"
+		}
+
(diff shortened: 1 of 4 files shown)


src/internal/m365/service/groups/backup.go
```diff
@@ -67,9 +67,14 @@ func ProduceBackupCollections(
 			break
 		}
 
+		catStr := scope.Category().PathType().String()
+		if scope.Category().PathType() == path.ChannelMessagesCategory {
+			catStr = "messages"
```
meain, line 72: Should we maybe update the stringer value? I don't see why that has to be `channelMessages` to begin with.

meain (approved):
Approving to unblock.

src/internal/m365/service/groups/backup.go
ryanfkeepers: I was thinking of doing a follow-up for this.

Issue #2023, opened 2023-01-03
Delay fetching Exchange item data from Graph API until it's clear kopia will be uploading the item


ashmrtn:
Kopia-assisted incrementals allows us to skip uploading items that kopia already has to kopia again. However, the logic for GraphConnector still fetches all item data. This data fetch eats into the number of requests that graph API allows us

Instead of always fetching item data, we should only fetch the data if kopia requests it

ashmrtn:
also related to
* #1702 
as it requires knowing when kopia actually needs data

ashmrtn:
thinking about this more, it also has implications for how backup details for these items are handled. A plan for that should be pinned down prior to starting work on this

vkamra:
Related also to #3007

vkamra:
This (along with #3007) will help with scenarios such as:
- Delta token expired
- Delta-based backups cannot be leveraged because the mailbox is full

ashmrtn:
Now that we added support for sourcing details entries from assist bases, the requirements for enabling kopia assisted incrementals have become a bit easier to meet. Currently the requirements are:
* complete ItemInfo is materialized if the item data is read by kopia (i.e. not required otherwise)
* item mod time is available
* item mod time matches mod time in ItemInfo if ItemInfo is materialized
* item name is available

Since we merge assist base details, we need only know the mod time and item name when first telling kopia about the item. We can put off materializing the ItemInfo until we know kopia read the file data and it wasn't cached

However, this leaves us with a couple new questions (numbered for ease of reference):
1. how do we handle errors when trying to fetch item data?
1. how do we handle items that are deleted between the time we enumerated them and the time we try to fetch their data?

For the first case, we can just return an error when opening the lazy reader. This will be caught by kopia and handled similar to an error fetching OneDrive/SharePoint data

For the second case, we need to be a little more careful (and this needs validated through manual testing). We should probably do three things:
* return a reader that has no data (i.e. create an empty file in kopia). We need to return some non-error because we don't want to fail the backup (by returning an error) but we've also already told kopia about the item (so it could see if it was cached) so we need to give it something
* return a sentinel error from Info() (call to get ItemInfo)
* check for the sentinel error when [adding the item to the details builder](https://github.com/alcionai/corso/blob/8d2070409f7cdd96e0e4f96b52a2d48bd658ca6c/src/internal/kopia/upload.go#L229). If we have the sentinel then don't add the details entry. This will make it so the item isn't searchable by users in the final backup

[...]

PR #4266, opened 2023-09-15
Fetch mod time when getting added and removed items

ashmrtn:
Also return mod time when available if
getting the set of added and removed
items. This will be leveraged in later
PRs to implement kopia assisted
incrementals for exchange

Does not change any logic in
collections right now, just adds the
fields to be returned

Also adds an additional return value
denoting if the mod times are expected
to be valid. This is required because
events delta cannot return mod time

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #2023

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/services/m365/api/config.go
```diff
@@ -26,6 +26,7 @@ const (
 	isCancelled       = "isCancelled"
 	isDraft           = "isDraft"
 	mobilePhone       = "mobilePhone"
+	modifiedTime      = "lastModifiedDateTime"
```
ryanfkeepers, line 29: Nit: I'm trying to keep these var names parallel to their values to minimize cognitive load.
```suggestion
	lastModifiedDateTime      = "lastModifiedDateTime"
```

src/pkg/services/m365/api/channels_pager.go
```diff
@@ -142,15 +151,15 @@ func (c Channels) GetChannelMessageIDs(
...
 		ctx,
 		c.NewChannelMessagePager(teamID, channelID),
 		c.NewChannelMessageDeltaPager(teamID, channelID, prevDeltaLink),
 		prevDeltaLink,
 		canMakeDeltaQueries,
-		addedAndRemovedByDeletedDateTime)
+		addedAndRemovedByDeletedDateTime[models.ChatMessageable])
 
-	return added, removed, du, clues.Stack(err).OrNil()
+	return maps.Keys(added), removed, du, clues.Stack(err).OrNil()
```
ryanfkeepers, line 162: To maintain consistency across implementations, can this please switch to a `map[string]time.Time` return value?

src/pkg/services/m365/api/item_pager.go
```diff
@@ -182,29 +188,29 @@ func getAddedAndRemovedItemIDs[T any](
 	prevDeltaLink string,
 	canMakeDeltaQueries bool,
 	aarh addedAndRemovedHandler[T],
-) ([]string, []string, DeltaUpdate, error) {
+) (map[string]time.Time, bool, []string, DeltaUpdate, error) {
```
ryanfkeepers, line 191: This return is getting a bit choky.  Thoughts on condensing it into something like below on a follow-up?

```go
type addedAndRemoved struct {
  Added map[string]time.Time
  Removed []string
  CanUseModTime bool
  Delta DeltaUpdate
}

func getAddedAndRemovedItetmIDs[T any](...) (addedAndRemoved, error) {...}
```

PR #4282, opened 2023-09-18
add humanstring to path category type

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/pkg/path/category_type.go
+++ b/src/pkg/path/category_type.go
@@ -29,31 +29,47 @@ const (
 	ChannelMessagesCategory CategoryType = 9 // channelMessages
 )
 
-func ToCategoryType(category string) CategoryType {
-	cat := strings.ToLower(category)
+var strToCat = map[string]CategoryType{
+	strings.ToLower(EmailCategory.String()):           EmailCategory,
+	strings.ToLower(ContactsCategory.String()):        ContactsCategory,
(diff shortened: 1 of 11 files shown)


src/pkg/path/category_type.go
```diff
@@ -29,31 +29,47 @@ const (
...
+	FilesCategory:           "Files",
+	LibrariesCategory:       "Libraries",
+	ListsCategory:           "Lists",
+	PagesCategory:           "Pages",
+	DetailsCategory:         "Details",
+	ChannelMessagesCategory: "Messages",
+}
+
+// HumanString produces a more human-readable string version of the category.
+func (cat CategoryType) HumanString() string {
```
ashmrtn, line 66: this is starting to feel more like we should be defining a struct for it rather than creating more and more maps

ashmrtn (approved):

src/pkg/path/category_type.go
ryanfkeepers: I've tossed around the same idea in my head for a while.  Ultimately, unless we expand the behavior even further (I don't think that feels very likely, personally) the lift outweighs the value at this moment.  It's a two-way door, and always an option for the future.

PR #4294, opened 2023-09-19
add standardized tests for flags in backup cmds

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test


src/cli/backup/backup.go
```diff
@@ -279,6 +279,10 @@ func genericDeleteCommand(
 		return nil
 	}
 
+	if flags.RunModeFV == flags.RunModeFlagTest {
```
meain, line 282: I'm not sure if this is the best approach. Maybe we should split adding flags and the use of `utils.AddCommand`?

ryanfkeepers: It's not the best approach, agreed.  But it's also the pattern we've established in other areas of code.  Let's move this PR forward and we can address the pattern separately.

ryanfkeepers:
@meain could this get a final review, please?

pandeyabs:
I'll also review it shortly today. Sorry about the delay!

meain (approved):
Woops, I somehow missed the notification on this one.

src/cli/backup/onedrive_test.go
```diff
@@ -103,14 +84,229 @@ func (suite *OneDriveUnitSuite) TestAddOneDriveCommands() {
...
+	// required to ensure a dry run.
+	flags.AddRunModeFlag(cmd, true)
+
+	c := addOneDriveCommands(cmd)
+	require.NotNil(t, c)
+
+	// Test arg parsing for few args
+	cmd.SetArgs([]string{
+		oneDriveServiceCommand,
+		"--" + flags.RunModeFN, flags.RunModeFlagTest,
```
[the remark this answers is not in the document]
pandeyabs: Same goes for assertions

src/cli/backup/backup.go
pandeyabs: Can we create an issue to track this please? As an alternative, I feel we could add some lightweight integration tests instead, on an empty mailbox/drive/sp/groups.

pandeyabs (approved):
Approving to unblock. Not sure if we want to proliferate use of `isTestMode()` pattern in code.

src/cli/backup/backup.go
meain: I don't know if we should be adding integrations test for these, as that feels like adding unnecessary complexity just to test availability of flags.

ryanfkeepers: Agreed, we should avoid integration tests for this purpose.

src/cli/backup/onedrive_test.go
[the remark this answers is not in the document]
ryanfkeepers: I have follow-up PRs which already move a large set of common assertions to shared code.

PR #4297, opened 2023-09-19
aggregate flags for consistency and clarity

ryanfkeepers:
aggregates common flags, such as storage and provider credentials, in places that provide better guarantees that these flags get attached to their expected
commands.

Also aggregates testing setup and assertions around common flags to reduce boilerplate and chaff test
code, allowing flag-checking unit tests to better
identify bespoke flags in each command.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/flags/repo.go
+++ b/src/cli/flags/repo.go
@@ -33,6 +33,16 @@ func AddBackupIDFlag(cmd *cobra.Command, require bool) {
 	}
 }
 
+// ---------------------------------------------------------------------------
+// storage
+// ---------------------------------------------------------------------------
+
+func AddAllStorageFlags(cmd *cobra.Command) {
+	AddCorsoPassphaseFlags(cmd)
+	AddS3BucketFlags(cmd)
+	AddAWSCredsFlags(cmd)
+}
+
 func AddAWSCredsFlags(cmd *cobra.Command) {
 	fs := cmd.Flags()
 	fs.StringVar(&AWSAccessKeyFV, AWSAccessKeyFN, "", "S3 access key")
@@ -47,4 +57,12 @@ func AddCorsoPassphaseFlags(cmd *cobra.Command) {
 		CorsoPassphraseFN,
 		"",
 		"Passphrase to protect encrypted repository contents")
+}
+
+// ---------------------------------------------------------------------------
+// Provider
+// ---------------------------------------------------------------------------
+
+func AddAllProviderFlags(cmd *cobra.Command) {
+	AddAzureCredsFlags(cmd)
 }
--- a/src/cli/flags/repo.go
+++ b/src/cli/flags/repo.go
@@ -39,7 +39,6 @@ func AddBackupIDFlag(cmd *cobra.Command, require bool) {
 
 func AddAllStorageFlags(cmd *cobra.Command) {
 	AddCorsoPassphaseFlags(cmd)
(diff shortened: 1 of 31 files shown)


src/cli/flags/testdata/repo.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"gotest.tools/v3/assert"
+
+	"github.com/alcionai/corso/src/cli/flags"
+)
+
+func PreparedStorageFlags() []string {
```
meain, line 12: Ahh, this is so much better :D

meain (approved):

src/cli/flags/repo.go
```diff
@@ -33,6 +33,16 @@ func AddBackupIDFlag(cmd *cobra.Command, require bool) {
 	}
 }
 
+// ---------------------------------------------------------------------------
+// storage
+// ---------------------------------------------------------------------------
+
+func AddAllStorageFlags(cmd *cobra.Command) {
+	AddCorsoPassphaseFlags(cmd)
+	AddS3BucketFlags(cmd)
```
pandeyabs, line 42: S3 bucket and AWS creds flags are valid for AWS S3 storage only. Calling this function from `corso backup * *` or `corso restore * *` will expose these flags for other storage providers. Additionally, this won't expose any flags for other storage providers like filesystem. We should fix this. 

Also, quick sanity check: We are adding `AddS3BucketFlags` here. We currently only expose these flags for `repo * * `, and config file serves as the source of truth for bucket params for other commands. Given we `matchFromConfig`, any discrepancies between flags & config file should lead to an error, but we should validate this.

pandeyabs (changes requested):
`AddAllStorageFlags` looks problematic to me. Have left a comment.

src/cli/flags/repo.go
ryanfkeepers: > S3 bucket and AWS creds flags are valid for AWS S3 storage only. 

Correct, but every operation command needs to expose the total set of storage provider flags, because those commands are not scoped to a specific provider.  IE: it's `corso restore exchange` not `corso restore s3 exchange`.

> Additionally, this won't expose any flags which are meant for providers like filesystem

The main branch at the time of this PR didn't contain those flags.  I'll update the code in future PRs as needed.

> We are adding AddS3BucketFlags here. We currently only expose these flags for repo * * ,

Ah, good catch, will fix.

ryanfkeepers: @pandeyabs `AddS3BucketFlags()` has been removed from the func.  Can you give this PR another review, please?

Also, since you jumped in, would you kindly also review the parent PR?  https://github.com/alcionai/corso/pull/4294  Thanks!

ryanfkeepers: > Not saying we shouldn't expose bucket flags. It shouldn't break anything IMO. Just asking to verify functionality

I think it's better to keep flags out unelss the command explicitly supports them.  If we don't want the user overriding the s3 bucket during these commands, we shouldn't expose them.  Likewise, no-op flags will confuse end users, and should also be avoided.

pandeyabs (approved):

PR #4298, opened 2023-09-19
move container enum to pager pattern

ryanfkeepers:
moves exchange container enumeration funcs in
the api to the pager pattern established with item enumeration.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/contacts_pager.go
+++ b/src/pkg/services/m365/api/contacts_pager.go
@@ -94,12 +94,9 @@ func (c Contacts) EnumerateContainers(
 			break
 		}
 
-		gcc, ok := c.(graph.CachedContainer)
-		if !ok {
-			el.AddRecoverable(ctx, clues.New("container does not implement graph.CachedContainer"))
-		}
+		gncf := graph.NewCacheFolder(c, nil, nil)
 
-		if err := fn(gcc); err != nil {
+		if err := fn(&gncf); err != nil {
 			errs.AddRecoverable(ctx, graph.Stack(ctx, err).Label(fault.LabelForceNoBackupCreation))
 			continue
 		}
--- a/src/pkg/services/m365/api/contacts_pager.go
+++ b/src/pkg/services/m365/api/contacts_pager.go
@@ -18,25 +18,29 @@ import (
 // container pager
 // ---------------------------------------------------------------------------
 
-// EnumerateContainers iterates through all of the users current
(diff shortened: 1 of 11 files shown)


ashmrtn:
this seems to have some non-trivial errors during CI that should probably be resolved prior to review

src/pkg/services/m365/api/contacts_pager.go
```diff
@@ -45,44 +49,57 @@ func (c Contacts) EnumerateContainers(
...
+func (p *contactsFoldersPageCtrl) ValidModTimes() bool {
+	return true
+}
+
+// EnumerateContainers iterates through all of the users current
+// contacts folders, transforming each to a graph.CacheFolder, and calling
+// fn(cf).
+// Contact folders are represented in their current state, and do
+// not contain historical data.
+func (c Contacts) EnumerateContainers(
```
ashmrtn, line 75: I feel like this is getting close to pushing too much functionality into what should ideally be a fairly thin wrapper around the Graph API. The original intent behind pulling out the Graph calls and placing them behind an interface was so we could mock them, allowing us to test a broader range of situations without having to manufacture state in the remote service

Granted, the original design is not particularly conducive to pulling the Graph calls out into a separate package, but we should be considerate of the original intent behind the refactoring effort

ryanfkeepers: Generally agreed.  Do I assume correctly that you've got your eye on the `fn` processor, in particular?  The rest of the func is more or less benign, imo.  If so, I'd be happy to follow up by splitting out the two parts (enumeration and transformation) to their respective layers.  But for this PR in particular, I'd prefer to keep the scope of changes where they are.

ashmrtn: yea, that's mostly the concern. The more complexity we push into functions like this, the more difficulty we'll have testing that these functions work as expected since they either require complex networking mock setups or special setups in the external service

As long as we prioritize splitting this to reduce the logic in functions like this is seems good

ashmrtn (approved):

src/pkg/services/m365/api/contacts_pager.go
ryanfkeepers: PR up: https://github.com/alcionai/corso/pull/4346

Issue #4299, opened 2023-09-19
Remove casts to concrete exchange collection implementation in tests


ashmrtn:
Some tests in `internal/m365/collection/exchange/backup_test.go` like [this](https://github.com/alcionai/corso/blob/b00ac292c7a68f000530b9f07e38d9f0e538e6af/src/internal/m365/collection/exchange/backup_test.go#L1179) rely on being able to cast a collection to `prefetchCollection`. While the tests pass for the moment because the mock returns data in such a way that it ends up with a `prefetchCollection` these tests should really be updated to just use the `data.BackupCollection` API

PR #4300, opened 2023-09-19
Lazily fetch Exchange item data when possible

### LINKED ISSUE #2023: Delay fetching Exchange item data from Graph API until it's clear kopia will be uploading the item

ashmrtn:
Implement lazy data fetch for Exchange items.
Use a new collection type to clearly denote
when items can be lazily fetched vs. requiring
eager fetch

**This PR changes how the read bytes stat is
updated.** Lazily fetched items will not
update the read bytes stat. This stat doesn't
appear to be used anywhere at the moment

For items that are deleted between the time
enumeration takes place and the time the data
for them needs fetched, the corso will:
* return an empty reader for the item
* not add the item to backup details
* delete the (empty) item from kopia on the
  next backup

Manually tested deleting an item between
enumeration and data fetch

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes #2023

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/collection.go
+++ b/src/internal/m365/collection/exchange/collection.go
@@ -413,8 +413,6 @@ func (col *lazyFetchCollection) streamItems(
 ) {
 	var (
 		success     int64
-		totalBytes  int64
-		wg          sync.WaitGroup
 		colProgress chan<- struct{}
 
 		user = col.user
@@ -427,7 +425,7 @@ func (col *lazyFetchCollection) streamItems(
 			col.statusUpdater,
 			len(col.added)+len(col.removed),
 			int(success),
-			totalBytes,
+			0,
 			col.FullPath().Folder(false),
 			errs.Failure())
 	}()
@@ -487,8 +485,6 @@ func (col *lazyFetchCollection) streamItems(
 			colProgress <- struct{}{}
 		}
 	}
-
-	wg.Wait()
 }
 
 // Item represents a single item retrieved from exchange
--- a/src/internal/m365/collection/exchange/collection.go
+++ b/src/internal/m365/collection/exchange/collection.go
@@ -178,16 +178,11 @@ func NewCollection(
 	items itemGetterSerializer,
 	added map[string]time.Time,
 	removed []string,
+	validModTimes bool,
 	statusUpdater support.StatusUpdater,
-) prefetchCollection {
-	collection := prefetchCollection{
-		baseCollection: bc,
-		user:           user,
-		added:          maps.Clone(added),
-		removed:        map[string]struct{}{},
-		getter:         items,
-		statusUpdater:  statusUpdater,
-	}
+) data.BackupCollection {
+	add := maps.Clone(added)
+	remove := make(map[string]struct{}, len(removed))
 
 	// Remove any deleted IDs from the set of added IDs because items that are
 	// deleted and then restored will have a different ID than they did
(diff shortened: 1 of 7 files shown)


ashmrtn:
added #4299 as follow up but not sure when it will be addressed

src/internal/m365/collection/exchange/collection.go
```diff
@@ -174,18 +176,46 @@ func NewCollection(
...
+	add := maps.Clone(added)
+	remove := make(map[string]struct{}, len(removed))
+
+	// Remove any deleted IDs from the set of added IDs because items that are
+	// deleted and then restored will have a different ID than they did
+	// originally.
+	//
+	// TODO(ashmrtn): If we switch to immutable IDs then we'll need to handle this
+	// sort of operation in the pager since this would become order-dependent
+	// unless Graph started consolidating the changes into a single delta result.
```
meain, line 193: It is possible for an deletion to happen midway through us fetching pages from graph and thus we can never be certain that one item will not have multiple entries. I'm not sure how this plays into immutable IDs, but just wanted to mention it.

src/internal/m365/collection/exchange/collection.go
```diff
@@ -174,18 +176,46 @@ func NewCollection(
...
 	items itemGetterSerializer,
+	added map[string]time.Time,
+	removed []string,
+	validModTimes bool,
 	statusUpdater support.StatusUpdater,
-) prefetchCollection {
-	collection := prefetchCollection{
+) data.BackupCollection {
+	add := maps.Clone(added)
+	remove := make(map[string]struct{}, len(removed))
```
meain, line 185: nit: Since they are only used once, why not rename them to a bigger ones and use normal names for the vars used within the func.

```suggestion
	added := maps.Clone(origAdded)
	removed := make(map[string]struct{}, len(origRemoved))
```

src/internal/m365/collection/exchange/collection.go
```diff
@@ -12,6 +12,8 @@ import (
 	"time"
 
 	"github.com/alcionai/clues"
+	"github.com/spatialcurrent/go-lazy/pkg/lazy"
```
meain, line 15: I know we did not add this package in this PR, but this is a repo with 2 stars last updated 2 years ago. Should we use something else or maybe just implement a simple version of this within corso?

meain (approved):
LGTM

src/internal/m365/collection/exchange/collection.go
ashmrtn: heh I wondered about that too when we added it to the OneDrive code previously. I guess since we haven't had issues with it so far we may as well keep using it? We can always switch if we do find some problem

To play devils advocate a bit, it's possible that there hasn't been updates to the repo simply because it doesn't do much from a logical standpoint and was made well enough that there haven't been issues reported

src/internal/m365/collection/exchange/collection.go
ashmrtn: I think we can catch duplicate IDs if changes happen mid-paging. The problem is if we have immutable IDs the handling of the duplicates becomes order dependent

For background, if an item is changed/moved/deleted while paging through results then either 1. the item will appear again in the result set or 2. the item will appear in the next delta result set in the next backup

The code below handles the first case by removing IDs in the added set if they also appear in the removed set. If we have multiple adds for the same item the map automatically consolidates them into a single item fetch

This would need to change if we switched to immutable IDs because in that case it would be possible to see a series of results like `(add item A, remove item A, add item A)`. The "correct" outcome in that case would be to fetch data for item A. Without immutable IDs the delta results look like `(add item A, remove item A, add item A')` so we avoid the ordering issue altogether

PR #4318, opened 2023-09-20
add throttling, token counters to operation

ryanfkeepers:
Adds context embedding to pkg/count, and embeds
the backup, restore, and export operation ctx with the operations count.bus.  The bus is extracted from the ctx in api middleware to track throttling response counts, as well as token consumption counts.  Counts are globally tracked, not on a per-time or per resource basis.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/count/keys.go
```diff
@@ -3,10 +3,12 @@ package count
 type key string
 
 const (
+	APICallTokensConsumed key = "api-call-tokens-consumed"
```
pandeyabs, line 6: Could you add a small description for each of these?

src/pkg/count/count.go
```diff
@@ -106,3 +108,33 @@ func (b *Bus) TotalValues() map[string]int64 {
 
 	return b.Values()
 }
+
+// ---------------------------------------------------------------------------
+// context embedding
+// ---------------------------------------------------------------------------
+
+type countKey string
```
pandeyabs, line 116: `countBusKey` instead of `countKey`?

pandeyabs (approved):
LGTM. Great pattern!

src/pkg/count/count.go
ryanfkeepers: Eh, it's unexported.  Could just be named `key`.  But I don't think it's critical either way.

src/pkg/count/keys.go
ryanfkeepers: Sure, why not.

PR #4323, opened 2023-09-20
Fix extracting category from BaseCollection

ashmrtn:

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/collection.go
+++ b/src/internal/m365/collection/exchange/collection.go
@@ -252,16 +252,24 @@ func (col *prefetchCollection) streamItems(
 	errs *fault.Bus,
 ) {
 	var (
+		category    path.CategoryType
 		success     int64
 		totalBytes  int64
 		wg          sync.WaitGroup
 		colProgress chan<- struct{}
 
 		user = col.user
-		log  = logger.Ctx(ctx).With(
-			"service", path.ExchangeService.String(),
-			"category", col.FullPath().Category().String())
 	)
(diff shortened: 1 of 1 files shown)


src/internal/m365/collection/exchange/collection.go
```diff
@@ -252,17 +252,25 @@ func (col *prefetchCollection) streamItems(
...
 		colProgress chan<- struct{}
 
 		user = col.user
-		log  = logger.Ctx(ctx).With(
-			"service", path.ExchangeService.String(),
-			"category", col.FullPath().Category().String())
 	)
 
+	if col.FullPath() != nil {
+		category = col.FullPath().Category()
```
ryanfkeepers, line 265: no problems with this implementation, but do you think it'd be worth moving this to the base collection as a `col.Category()` func?  Right now it's up to every caller to check this same pattern.

ryanfkeepers (approved):

src/internal/m365/collection/exchange/collection.go
ashmrtn: that's an excellent idea. Since the issue is already in main though, let me queue this to avoid long-term issues (in case we forget about anything) then we can get a nicer solution added to BaseCollection

Issue #4119, opened 2023-08-26
Enhancement: pass more than one id to the delete command


zombiehoffa:
Would be nice to be able to pass more than one ID to the delete command on the cli so that when I'm going through the pain of pulling out ids based on date I can just pass the whole group of backups older than my desired date range to the delete command rather than coding a loop to pass each one individually.

vkamra:
Can leverage #4019

neha-Gupta1:
@zombiehoffa a new flag `backups` has been exposed to handle deletion of multiple IDs by providing `,`(comma) separated backupIDs. 
Currently the functionality is present in latest main. And will be available from release  `v0.14.0`

vkamra:
This is done.

PR #4335, opened 2023-09-21
allow delete of multiple backup IDs

neha-Gupta1:
allow deletion of multiple IDs in single command

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* https://github.com/alcionai/corso/issues/4119

#### Test Plan

- [ ] :muscle: Manual
- [ ] :green_heart: E2E


src/cli/flags/repo.go
```diff
@@ -17,13 +17,26 @@ const (
...
 	AWSSessionTokenFV    string
 	CorsoPassphraseFV    string
 	SucceedIfExistsFV    bool
 )
 
+// AddBackupIDFlag adds the --backup flag.
+func AddDeleteBackupIDFlag(cmd *cobra.Command, require bool) {
+	cmd.Flags().StringSliceVar(
+		&DeleteBackupIDFV,
+		BackupFN, nil,
```
ryanfkeepers, line 32: re-using flag names can cause confusion and gotchas.  Since the primary difference here is to allow a list of backups, instead of a single one, I suggest a rename from `DeleteBackupIDF*` to `BackupIDsF*`.  This will allow two wholly separate flag declarations with no collision gotchas.

neha-Gupta1: Yes. Make sense. Let me update it.

src/cli/flags/repo.go
```diff
@@ -17,13 +18,26 @@ const (
...
+	BackupIDsFV          []string
 	AWSAccessKeyFV       string
 	AWSSecretAccessKeyFV string
 	AWSSessionTokenFV    string
 	CorsoPassphraseFV    string
 	SucceedIfExistsFV    bool
 )
 
+// AddBackupIDFlag adds the --backup flag.
+func AddDeleteBackupIDFlag(cmd *cobra.Command, require bool) {
```
ryanfkeepers, line 30: Nit: since the flag was changed to not say "delete", and also to make sure it doesn't accidentally get confused with the singular backup.
```suggestion
func AddMultipleBackupIDsFlag(cmd *cobra.Command, require bool) {
```

ryanfkeepers (approved):
thanks for the changes!

vkamra (changes requested):
This is a breaking change. Adding a block to discuss the implications.

vkamra:
@neha-Gupta1 @ryanfkeepers - suggest we keep the `--backup` flag so that this doesn't break existing users (and scripts) silently.

If we feel strongly about moving to `--backups` then add support for both and we deprecate `--backup` by announcing in changelog, removing from docs and giving it a few releases

neha-Gupta1:
> @neha-Gupta1 @ryanfkeepers - suggest we keep the `--backup` flag so that this doesn't break existing users (and scripts) silently.
> 
> If we feel strongly about moving to `--backups` then add support for both and we deprecate `--backup` by announcing in changelog, removing from docs and giving it a few releases

@ryanfkeepers yours thoughts on this please.

vkamra (approved):

Issue #4191, opened 2023-09-05
Collections struct refactoring 


ryanfkeepers:
Our per-package collections haven't aged particularly well with their organic growth since starting as smaller, specialized packages.  Now that we're adding groups to the mix, we're seeing the potential for real gains when it comes to cleaning up the old and copied code.

Some places to investigate include:
* ensuring comments throughout the file are accurate
* much of the base functionality (value getters especially) could be ported to a shared "collection" that gets embeddeed in each local iteration.
  * this will also help ensure we use public funcs for things like getting handles to paths
* certain duplicate properties could be reduced (i.e. probably don't need separate mod time field)

PR #4339, opened 2023-09-21
Create and use generic tombstone collection

ashmrtn:
Incremental backups requires us to mark some
folders as deleted by creating a collection
with state `data.DeletedState`

This PR creates a simple, generic "tombstone"
collection that does just that

The PR additionally uses the tombstone
collection in place of more complicated
implementations where any easy switch is
possible. Deleted collections in OneDrive
require more work since tests attempt to
cast to a concrete type

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4191

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/data/implementations.go
+++ b/src/internal/data/implementations.go
@@ -6,6 +6,7 @@ import (
 	"github.com/alcionai/clues"
 
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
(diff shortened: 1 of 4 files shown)


src/internal/data/implementations.go
```diff
@@ -136,3 +137,30 @@ func (col BaseCollection) DoNotMergeItems() bool {
...
+}
+
+// tombstoneCollection is a collection that marks a folder (and folders under it
+// if they aren't explicitly noted in other collecteds) as deleted. It doesn't
+// contain any items.
+type tombstoneCollection struct {
+	BaseCollection
+}
+
+func (col *tombstoneCollection) Items(context.Context, *fault.Bus) <-chan Item {
```
ryanfkeepers, line 161: would it be easier to return Nil here?

Also, you might want to add a comment explaining that this is always expected to be a no-op, and why.

ryanfkeepers (approved):

src/internal/data/implementations.go
ashmrtn: ah yea, nil should work. I sort of forgot that range loops play nice with nil heh

There is a comment on the struct definition that states it never has items, but I'll add something to this function as well

Issue #2025, opened 2023-01-03
Verify flags before connecting to Kopia and checking M365 credentials


ashmrtn:
The current split for the CLI first creates and initializes a repository struct and then verifies flags the user passed in. Creating the repository object allows passing it to a lower-level function thus giving better opportunities for testing. However, it can sometimes take several seconds to connect to kopia. This leads to a poor user experience because the feedback loop is elongated (i.e. spend 10s connecting only to be told a flag was misspelled).

To fix this, two things should be done:
* repository struct creation should be separated from initialization, thus preserving our ability to write unit tests with mocks
* flag verification should be done after repository creation but before repository initialization

vkamra:
@ryanfkeepers - can you link the other ticket where M365 credentials are not being validated in repo connect

ryanfkeepers:
https://github.com/alcionai/corso/issues/1426

gmatev:
Related to #1426

PR #4352, opened 2023-09-22
add generic details command

ryanfkeepers:
centralizes details command processing in the cli

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2025

### DIFF
--- a/src/cli/backup/sharepoint_test.go
+++ b/src/cli/backup/sharepoint_test.go
@@ -1,7 +1,6 @@
 package backup
 
 import (
-	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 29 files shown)


src/cli/backup/sharepoint_test.go
```diff
@@ -391,51 +387,3 @@ func (suite *SharePointUnitSuite) TestSharePointBackupCreateSelectors() {
 		})
 	}
 }
-
-func (suite *SharePointUnitSuite) TestSharePointBackupDetailsSelectors() {
```
vkamra, line 395: Do we have coverage for sharepoint details now? The change in `backup_test` only includes exchange. Can we leave this test in here?

vkamra:
Change looks good. Question on test coverage for sharepoint/onedrive/groups.

vkamra (approved):
Change LGTM. Question on test coverage and whether we need to leave the existing tests or add a few more test cases in `backup_test.go`

pandeyabs (approved):

src/cli/backup/sharepoint_test.go
ryanfkeepers: This wasn't exercising integration behavior in details, it was just a roundabout way of testing the selectors behavior via mocks, all of which is already covered in other sections. The `backup_test` reference to exchange doesn't exercise exchange in any particular way either.  We primarily need a valid service to stand in place there.

As separate toil, I'd be happy to go through and re-evaluate test coverage after these cases for security, however.  They seem to have grow in a couple different directions across time and could use some re-alignment.

PR #4355, opened 2023-09-22
connect to data provider during init/connect

ryanfkeepers:
Takes two steps towards better data provider valiation on init and connect:
1. abstracts the data provider in a repository to replace hard-coded references to m365 with an interface
for connecting to any generic data provider specified by the account.Provider.
2. connects to the data provider during the init and connect steps so that the provider details are validated early in usage, presumably before we write said
connection credentials to the config file.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #2025

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/path/service_type.go
+++ b/src/pkg/path/service_type.go
@@ -14,6 +14,10 @@ var ErrorUnknownService = clues.New("unknown service string")
 //
 // Metadata services are not considered valid service types for resource paths
 // though they can be used for metadata paths.
+//
+// The string representaton of each enum _must remain the same_.  In case of
+// changes to those values, we'll need migration code to handle transitions
+// across states else we'll get marshalling/unmarshalling errors.
 type ServiceType int
 
 //go:generate stringer -type=ServiceType -linecomment
--- a/src/pkg/path/service_type.go
+++ b/src/pkg/path/service_type.go
@@ -14,10 +14,6 @@ var ErrorUnknownService = clues.New("unknown service string")
 //
 // Metadata services are not considered valid service types for resource paths
 // though they can be used for metadata paths.
-//
-// The order of the enums below can be changed, but the string representation of
-// each enum must remain the same or migration code needs to be added to handle
-// changes to the string format.
 type ServiceType int
 
 //go:generate stringer -type=ServiceType -linecomment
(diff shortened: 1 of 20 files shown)


src/pkg/path/service_type.go
```diff
@@ -14,10 +14,6 @@ var ErrorUnknownService = clues.New("unknown service string")
 //
 // Metadata services are not considered valid service types for resource paths
 // though they can be used for metadata paths.
-//
-// The order of the enums below can be changed, but the string representation of
```
ashmrtn, line 18: (just skimming cause I was curious) we should keep this comment. It documents the expectations for updating these enums (i.e. can't change string form of them)

ryanfkeepers: check, will do

src/pkg/repository/data_providers.go
```diff
@@ -7,48 +7,68 @@ import (
...
 	pst path.ServiceType,
-	acct account.Account,
-	co control.Options,
 ) (*m365.Controller, error) {
-	if !m365nonce {
-		m365nonce = true
+	if r.Provider != nil {
+		ctrl, ok := r.Provider.(*m365.Controller)
+		if !ok {
+			return nil, clues.New("Attempted to access multiple repository providers")
```
pandeyabs, line 62: Could you explain this error message? Not making sense to me.

src/pkg/repository/repository.go
```diff
@@ -184,10 +216,11 @@ func (r *repository) Connect(ctx context.Context) (err error) {
 		}
 	}()
 
-	// ctrl, err = connectToM365(ctx, sel.PathService(), r.Account, r.Opts)
-	// if err != nil {
-	// 	return operations.BackupOperation{}, clues.Wrap(err, "connecting to m365")
-	// }
+	if cfg.Service != path.UnknownService {
```
pandeyabs, line 219: This answers my question from the earlier PR in the chain.

pandeyabs (approved):

src/pkg/repository/data_providers.go
ryanfkeepers: worth adding a comment here to avoid future confusion, thanks for the highlight.

PR #4358, opened 2023-09-25
Supply context for http requests

pandeyabs:
We aggregate graph API tokens(XMRUs) by 
1. binding a count bus to `ctx` during backup initialization.
2. Incrementing counters after extracting the bus reference from `ctx` in metrics middleware. We create a new count bus if bus reference is not found.

While testing the count bus, I found that it wasn't producing correct total XMRU values. This is because for http requests created by graph requestor, we are passing `context.Background()` during request creation. This results in creation of multiple count busses during 2) above, leading to incorrect total values.

We don't have this problem for [http requests created](https://github.com/microsoft/kiota-http-go/blob/b162cb63903518fdf02bd4e281e7729bcad899fc/nethttp_request_adapter.go#L271) via graph servicer. Because kiota http adapter attaches supplied `ctx` to request.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/graph/http_wrapper.go
+++ b/src/internal/m365/graph/http_wrapper.go
@@ -82,7 +82,7 @@ func (hw httpWrapper) Request(
 	body io.Reader,
 	headers map[string]string,
 ) (*http.Response, error) {
-	req, err := http.NewRequest(method, url, body)
+	req, err := http.NewRequestWithContext(ctx, method, url, body)
 	if err != nil {
 		return nil, clues.Wrap(err, "new http request")
 	}


ryanfkeepers (approved):

pandeyabs:
@ryanfkeepers Looks like this change is causing tests to timeout(except last run). So I'm removing it from mergequeue for now. It's possible that rate limiter middleware might be impacted by this fix ( `QueueRequest` -> `ctxLimiter`)

I'm taking a look. Will likely continue investigation tomorrow. We can let this be for now.

pandeyabs:
And just when I wrote above comment, the integration tests passed :/

ryanfkeepers:
> I'm taking a look. Will likely continue investigation tomorrow. We can let this be for now.

@pandeyabs Actually, you just jogged my memory.  We purposefully avoid attaching the rate limiter here (by not attaching the context) specifically because item content downloads do not fall into the drive rate limiting system.  So this change _will_ have unwanted side effects of unnecessary over-limiting.

That said, skipping the limiter by skipping the context is an anti-pattern and a smell, and we need to find a way to work around that without dropping the other things we want in the context.

Would you mind reverting this PR in the short term until that larger control gets fixed?  I've created an issue on the topic and assigned you to it here: https://github.com/alcionai/corso/issues/4364.

pandeyabs:
> specifically because item content downloads do not fall into the drive rate limiting system.

They do. Graph throttling issues that we have seen with SDK consumers are with item content downloads going over the 1250 / min limit. It's undocumented by MSFT though. Will discuss more offline.

pandeyabs:
Closing this. Going to push this change as part of https://github.com/alcionai/corso/pull/4393/files so that we don't break main in the interim.

PR #4361, opened 2023-09-25
ensure account details on init/connect

ryanfkeepers:
use m365 connection details to retrieve an
application token on init and connect as a way to
validate that the user has provided correct and
usable connection details.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/controller.go
+++ b/src/internal/m365/controller.go
@@ -81,6 +81,11 @@ func NewController(
 
 	var rCli *resourceClient
 
+	// no failure for unknown service.
+	// In that case we create a controller that doesn't attempt to look up any resource
+	// data.  This case helps avoid unnecessary service calls when the end user is running
+	// repo init and connect commands via the CLI.  All other callers should be expected
+	// to pass in a known service, or else expect downstream failures.
 	if pst != path.UnknownService {
 		rc := resource.UnknownResource
 
--- a/src/internal/m365/controller.go
+++ b/src/internal/m365/controller.go
@@ -79,20 +79,24 @@ func NewController(
 		return nil, clues.Wrap(err, "creating api client").WithClues(ctx)
 	}
 
-	rc := resource.UnknownResource
+	var rCli *resourceClient
 
-	switch pst {
-	case path.ExchangeService, path.OneDriveService:
-		rc = resource.Users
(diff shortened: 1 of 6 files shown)


src/internal/m365/controller.go
```diff
@@ -79,20 +79,24 @@ func NewController(
...
-	case path.ExchangeService, path.OneDriveService:
-		rc = resource.Users
-	case path.GroupsService:
-		rc = resource.Groups
-	case path.SharePointService:
-		rc = resource.Sites
-	}
+	var rCli *resourceClient
+
+	if pst != path.UnknownService {
```
pandeyabs, line 84: Can't we fail fast for `path.UnknownService`? We are creating a `Controller` for unknown service later in this function. Is this by design?

src/pkg/services/m365/api/access.go
```diff
@@ -0,0 +1,62 @@
...
+func (c Client) Access() Access {
+	return Access{c}
+}
+
+// Access is an interface-compliant provider of the client.
+type Access struct {
+	Client
+}
+
+func (c Access) GetToken(
```
pandeyabs, line 27: Can we add a test for this? Also, is this documented by MSFT somewhere? Could you please leave the link as a comment?

pandeyabs (approved):
LGTM. Left some comments about missing test.

src/pkg/services/m365/api/access.go
ryanfkeepers: d'oh!  I did miss that, didn't I?  Thanks for the catch.

src/internal/m365/controller.go
ryanfkeepers: This is by design.  I added a comment for explanation.  Truthfully, it's not my favorite workaround, since it's an ambiguous usage of the UnknownService const.  I originally thought about adding a `NoService` const in the same set, which would allow code to clearly identify the difference between "no service is expected here" and "I should have a service but I don't".  Then I felt uncomfortable about the addition and withdrew it with the intent to bring it up in design discussion.

PR #4365, opened 2023-09-25
Create generic items without Info()

ashmrtn:
Create generic Item implementations that
don't implement the ItemInfo interface.
These implementations can be used for
things like metadata files.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4191

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/data/item.go
+++ b/src/internal/data/item.go
@@ -16,16 +16,23 @@ import (
 )
 
 var (
+	_ Item        = &unindexedPrefetchedItem{}
+	_ ItemModTime = &unindexedPrefetchedItem{}
+
 	_ Item        = &prefetchedItem{}
 	_ ItemInfo    = &prefetchedItem{}
 	_ ItemModTime = &prefetchedItem{}
+
+	_ Item        = &unindexedLazyItem{}
+	_ ItemModTime = &unindexedLazyItem{}
+
 	_ Item        = &lazyItem{}
 	_ ItemInfo    = &lazyItem{}
 	_ ItemModTime = &lazyItem{}
 )
 
 func NewDeletedItem(itemID string) Item {
-	return &prefetchedItem{
+	return &unindexedPrefetchedItem{
 		id:      itemID,
 		deleted: true,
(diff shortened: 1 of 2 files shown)


src/internal/data/item.go
```diff
@@ -34,24 +41,26 @@ func NewDeletedItem(itemID string) Item {
...
 }
 
-// prefetchedItem represents a single item retrieved from the remote service.
-type prefetchedItem struct {
+// unindexedPrefetchedItem represents a single item retrieved from the remote
+// service.
+//
+// This item doesn't implement ItemInfo so it's safe to use for items like
+// metadata that shouldn't appear in backup details.
+type unindexedPrefetchedItem struct {
```
ryanfkeepers, line 61: The name here is getting a little chonky.  `Prefetched` wasn't entirely easy to grok, but worked out as we had either prefetched or non-prefetched items, so things could be extrapolated.  `UnindexedPrefetched` implies the possibility of an `IndexedPrefetched` item, and I don't know where that would get us.  Further, without any reference to an index elsewhere in this code, and no explanation in the comment, I'm not even sure what the indexing here is referring to.

I feel like this decomposition could use a re-do.  Or, at least, a renaming.  some thoughts:
```go
// basically UnindexedPrefetched
type BaseItem struct {
  id, reader, modTime
}

type ItemAndInfo struct {
  BaseItem, ItemInfo
}

type LazyItem struct {
  ItemAndInfo, Getter, Errs
}
```

Low effort go at it.  Thoughts?

ryanfkeepers (approved):
Technically workable, so approving to unblock, but I think the names need a review if we want concepts to remain maintainable.

src/internal/data/item.go
ashmrtn: I'll come back and fix naming after I get the PR train merged

PR #4370, opened 2023-09-26
sanity test refactor

ryanfkeepers:
refactoring the sanity tests with three goals:
1. move from env vars to cli commands so that unsupported commands fail loudly.
2. set up support for groups restore and export testing.
3. introduce some code re-use throughout.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #3988

#### Test Plan

- [x] :green_heart: E2E


meain:
Would be useful to manually run the sanity test workflow on the changes in this branch. I've added https://github.com/alcionai/corso/pull/4382 to simplify this in the future.

meain:
We would also need a small change here. Left over from my PR. This should be changed to `new-data-creation-groups`.
https://github.com/alcionai/corso/blob/b7c02e6c7a701760d7267ef93e8b14e813d0ba28/.github/workflows/sanity-test.yaml#L367

src/cmd/sanity_test/export/onedrive.go
```diff
@@ -36,37 +29,19 @@ func CheckOneDriveExport(
...
-		}
-
-		relPath, err := filepath.Rel(folderName, path)
-		if err != nil {
-			return clues.Stack(err)
-		}
-
-		exportFileSizes[relPath] = info.Size()
-		if startTime.After(info.ModTime()) {
-			startTime = info.ModTime()
```
meain, line 55: This `startTime` was getting used later, but in the new code, this was changed to `envs.StartTime`. Same for other cases.

src/pkg/services/m365/api/drive.go
```diff
@@ -84,6 +84,26 @@ func (c Drives) GetRootFolder(
 	return root, nil
 }
 
+// TODO: pagination controller needed for completion.
```
meain, line 87: We should at least increase the number of items we retrieve. I believe the default is 10 which might be too low. Probably same for `GetContainerChildren`.

src/cmd/sanity_test/export/onedrive.go
ryanfkeepers: Why would we use `startTime` for this, anyway?  Seems like a conflict of purpose.

src/pkg/services/m365/api/drive.go
ryanfkeepers: tests are succeeding at the moment, but I do plan to follow up on the pagination.  It's currently pending another PR that'll make implementation super simple.

src/cmd/sanity_test/export/onedrive.go
meain: I meant that startTime is getting used in line 72 and it might be expecting the updated value from here. Maybe switch the one in line 72 to `env.StartTime`.

meain (approved):
Just a note about `startTime`.

PR #4375, opened 2023-09-26
separate delta not supported from invalid

ryanfkeepers:
groups backup has a failing case where delta
queries are not supported, but we aren't distinguishing
that case from the invalid delta case as expected. This
results in backup failures, instead of success with no data.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3988

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/graph/errors_test.go
+++ b/src/internal/m365/graph/errors_test.go
@@ -247,11 +247,6 @@ func (suite *GraphErrorsUnitSuite) TestIsErrInvalidDelta() {
 			err:    odErr(string(syncStateInvalid)),
 			expect: assert.True,
 		},
-		{
-			name:   "deltatoken not supported oDataErrMsg",
-			err:    odErrMsg("fnords", string(parameterDeltaTokenNotSupported)),
-			expect: assert.True,
-		},
 		// next two tests are to make sure the checks are case insensitive
 		{
 			name:   "resync-required oDataErr camelcase",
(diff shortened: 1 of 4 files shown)


src/internal/m365/graph/errors_test.go
```diff
@@ -271,6 +266,55 @@ func (suite *GraphErrorsUnitSuite) TestIsErrInvalidDelta() {
...
+			err:    nil,
+			expect: assert.False,
+		},
+		{
+			name:   "non-matching",
+			err:    assert.AnError,
+			expect: assert.False,
+		},
+		{
+			name:   "as",
```
ashmrtn, line 286: should we be checking if a wrapped error is handled correctly?

ashmrtn (approved):

src/internal/m365/graph/errors_test.go
ryanfkeepers: That feels a bit extraneous, given how well tested error wrapping systems are already.

ashmrtn: we did previously had a bug where some of code was expecting the error type itself (i.e. direct type-assertion to error type) and not a wrapped version of the error (i.e. obtain error via `errors.As` or `errors.Is`)

ryanfkeepers: I do recall that case.  I'm still going to punt here, because the topic is getting at the broader design of error handling here, for which we already have a ticket, and the work can be left to that.  In the meantime I'll see about a little bump on the side.

Issue #4328, opened 2023-09-20
Formalize way to set serialization version in service


ashmrtn:
As part of an easy fix for Exchange kopia assisted incrementals we bumped the serialization version for all services to show we were persisting deletion markers if the item was deleted in flight

We should formalize how services will configure the serialization version and check if they got a valid serialization version. Right now all of the serialization version handling is done in kopia wrapper by some custom reader/writers

Moving these to a generic package and having all services wrap their data readers in them and track their own disjoint sets of acceptable versions would likely be the first step in this process

PR #4378, opened 2023-09-26
Create exported readers for injecting and retrieving serialization version info

ashmrtn:
Previously kopia wrapper has been transparently injecting
and stripping out a "serialization version" for all items
persisted in kopia. The version thus far has been
hard-coded to `1` and has been stored in big endian format
in the first 4 bytes of every kopia file

This PR is a step towards getting serialization versions
on a per-service/per-item basis. It exposes serialization
readers that inject and strip out info to other packages

This PR also slightly changes the serialization version
format. The changes are two-fold:
* the MSB is now used to represent if the item was
  deleted between the time it was discovered and when
  its data was requested
* the serialization version number is set to uint16. This
  has no impact on existing persisted data since all
  versions are currently `1` (don't use more than
  16-bits). This size can be expanded in the future if
  needed, but is mostly to enforce the idea that higher
  order bits shouldn't be used for version numbers right
  now

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4328

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/common/readers/serialization_version.go
+++ b/src/internal/common/readers/serialization_version.go
@@ -1,0 +1,181 @@
+package readers
+
+import (
+	"bytes"
+	"encoding/binary"
+	"io"
+	"os"
+	"unsafe"
+
+	"github.com/alcionai/clues"
+)
+
+// persistedSerializationVersion is the size of the serialization version in
+// storage.
+//
+// The current on-disk format of this field is written in big endian. The
+// highest bit denotes if the item is empty because it was deleted between the
+// time we told the storage about it and when we needed to get data for it. The
+// lowest two bytes are the version number. All other bits are reserved for
+// future use.
+//
+//	MSB 31         30             16        8       0 LSB
+//	     +----------+----+---------+--------+-------+
+//	     | del flag |   reserved   | version number |
+//	     +----------+----+---------+--------+-------+
+type persistedSerializationVersion = uint32
+
+// SerializationVersion is the in-memory size of the version number that gets
+// added to the persisted serialization version.
+//
+// Right now it's only a uint16 but we can expand it to be larger so long as the
+// expanded size doesn't clash with the flags in the high-order bits.
+type SerializationVersion uint16
+
+// DefaultSerializationVersion is the current (default) version number for all
+// services. As services evolve their storage format they should begin tracking
+// their own version numbers separate from other services.
+const DefaultSerializationVersion SerializationVersion = 1
+
+var (
+	versionFormatSize                               = int(unsafe.Sizeof(persistedSerializationVersion(0)))
+	delInFlightMask   persistedSerializationVersion = 1 << ((versionFormatSize * 8) - 1)
+)
+
+// SerializationFormat is a struct describing serialization format versions and
+// flags to add for this item.
+type SerializationFormat struct {
+	Version     SerializationVersion
+	DelInFlight bool
+}
+
+func NewVersionedBackupReader(
+	format SerializationFormat,
+	baseReaders ...io.ReadCloser,
+) (io.ReadCloser, error) {
+	if format.DelInFlight && len(baseReaders) > 0 {
+		// This is a conservative check, but we can always loosen it later on if
+		// needed. At the moment we really don't expect any data if the item was
+		// deleted.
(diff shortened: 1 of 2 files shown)


src/internal/common/readers/serialization_version.go
```diff
@@ -0,0 +1,181 @@
...
+type SerializationVersion uint16
+
+// DefaultSerializationVersion is the current (default) version number for all
+// services. As services evolve their storage format they should begin tracking
+// their own version numbers separate from other services.
+const DefaultSerializationVersion SerializationVersion = 1
+
+var (
+	versionFormatSize                               = int(unsafe.Sizeof(persistedSerializationVersion(0)))
+	delInFlightMask   persistedSerializationVersion = 1 << ((versionFormatSize * 8) - 1)
```
ryanfkeepers, line 42: ❤️

src/internal/common/readers/serialization_version.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return res, nil
+}
+
+type versionedBackupReader struct {
+	// baseReaders is a reference to the original readers so we can close them.
+	baseReaders []io.ReadCloser
+	// combined is the reader that will return all data.
+	combined io.Reader
```
ryanfkeepers, line 93: You don't think it's worth keeping a reference to the version buffer or reader here, just in case?  I'm not sure what I'd use it for either.  Just paranoid, I guess.

src/internal/common/readers/serialization_version.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (vbr *versionedBackupReader) Close() error {
+	if vbr.combined == nil {
+		return nil
+	}
+
+	vbr.combined = nil
+
+	var errs *clues.Err
```
ryanfkeepers, line 118: hmm, this feels more like a case for fault recoverables than errs.  Create a fault bus here, add recoverable errors throughout.  If recoverable > 0, return a clean sentinel while logging each error case. 

My concern with using clues stacking here is that its format implies that deeper stacks occurred within context of earlier ones, not distinct from them (wtb `clues.Tree(err, errs...)`?  No, just kidding. ...maybe).  Which could make debugging a bit odd.  I think what you really mean to communicate is that 1/ at least one closer failed (return value), and 2/ one or more errors occurred (logs).  Easier with fault, harder with clues.

ryanfkeepers (approved):
lgtm, some nits for talking points, but nothing blocking

src/internal/common/readers/serialization_version.go
pandeyabs, line 42: Can we make these consts?

src/internal/common/readers/serialization_version.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+// SerializationFormat is a struct describing serialization format versions and
+// flags to add for this item.
+type SerializationFormat struct {
+	Version     SerializationVersion
+	DelInFlight bool
+}
+
+func NewVersionedBackupReader(
```
pandeyabs, line 52: Could you leave a small description for this func?

src/internal/common/readers/serialization_version.go
ashmrtn: technically we do have a reference to the version reader in `baseReaders`. If we find some reason we need it going forward we can add that. For the moment though I'm not sure what we'd use it for. Kopia wrapper doesn't do any validation/inspection of the written bytes so it doesn't seem like we'd need the version format at all

pandeyabs (approved):
:rocket: :ship:

src/internal/common/readers/serialization_version.go
ashmrtn: unfortunately using fault requires a context for logging. I don't really want to have to keep a reference to that here (and pass it in in all the places this will be used). The current option will work well enough even if the error output is a tad odd. I think having the `reader_index` clue in there will help readers figure out what's going on when viewed in conjunction with this code

PR #4391, opened 2023-09-27
correctly attach store/provider flags

ryanfkeepers:
corrects the storage and provider flag positioning to attach to the child command instead of the parent. Also corrects unit tests for flags to ensure flags are preset as expected.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/backup/groups_test.go
```diff
@@ -128,108 +128,64 @@ func (suite *GroupsUnitSuite) TestValidateGroupsBackupCreateFlags() {
...
+			groupsServiceCommand,
+			[]string{
+				"--" + flags.RunModeFN, flags.RunModeFlagTest,
+				"--" + flags.GroupFN, flagsTD.FlgInputs(flagsTD.GroupsInput),
+				"--" + flags.CategoryDataFN, flagsTD.FlgInputs(flagsTD.GroupsCategoryDataInput),
+				"--" + flags.FetchParallelismFN, flagsTD.FetchParallelism,
+				"--" + flags.FailFastFN,
+				"--" + flags.DisableIncrementalsFN,
+				"--" + flags.ForceItemDataDownloadFN,
+				"--" + flags.DisableDeltaFN,
```
meain, line 149: Should we move all the flags that are not specific to a single service for each group (backup, restore, list, details) into a package level variable, ie things like fail fast, disable incremental, disable delta etc.

src/cli/restore/onedrive_test.go
```diff
@@ -39,73 +37,56 @@ func (suite *OneDriveUnitSuite) TestAddOneDriveCommands() {
...
+						"--" + flags.FileFN, flagsTD.FlgInputs(flagsTD.FileNameInput),
+						"--" + flags.FolderFN, flagsTD.FlgInputs(flagsTD.FolderPathInput),
+						"--" + flags.FileCreatedAfterFN, flagsTD.FileCreatedAfterInput,
+						"--" + flags.FileCreatedBeforeFN, flagsTD.FileCreatedBeforeInput,
+						"--" + flags.FileModifiedAfterFN, flagsTD.FileModifiedAfterInput,
+						"--" + flags.FileModifiedBeforeFN, flagsTD.FileModifiedBeforeInput,
+						"--" + flags.CollisionsFN, flagsTD.Collisions,
+						"--" + flags.DestinationFN, flagsTD.Destination,
+						"--" + flags.ToResourceFN, flagsTD.ToResource,
+						"--" + flags.NoPermissionsFN,
```
meain, line 64: These set is common for onedrive, sp and groups. Maybe we can collect it to a global var.

meain (approved):
LGTM, few nits

src/cli/backup/groups_test.go
ryanfkeepers: Two notes: support and scope.

Per support: not all of these flags are equally supported.  Our application of features like ForceDataDownload, concurrency and paging controls, and etc are sometimes half-implemented, sometimes service-specific.  That makes it difficult to cleanly and easily aggregate all these flags as expected.  Here's an issue to track that standardization, we can follow up on it from there: https://github.com/alcionai/corso/issues/4400

Per scope: this request is out of scope for this PR.  I've made a tech-debt issue to track the movement.  Should be open to contribution when anyone has time to do so: https://github.com/alcionai/corso/issues/4399.

PR #4393, opened 2023-09-28
Push item download requests through drive rate limiter

pandeyabs:
**Changes:** 
1. Count item download requests under drive rate limit quota . Currently this is under exchange limiter.
2. Use 1 rate limit token instead of 2 for drive calls by default. 
3. Use 2 tokens instead of 1 for initial delta query ( which has no token). 

Sharing internal docs separately to go along with the review. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/graph/concurrency_middleware.go
+++ b/src/internal/m365/graph/concurrency_middleware.go
@@ -149,8 +149,7 @@ const limiterConsumptionCtxKey limiterConsumptionKey = "corsoGraphRateLimiterCon
 const (
 	// https://learn.microsoft.com/en-us/sharepoint/dev/general-development
 	// /how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling
-	defaultLC      = 1
-	driveDefaultLC = 1
+	defaultLC = 1
 	// limit consumption rate for single-item GETs requests,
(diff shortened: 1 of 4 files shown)


src/internal/m365/graph/concurrency_middleware.go
```diff
@@ -150,10 +150,12 @@ const (
 	// https://learn.microsoft.com/en-us/sharepoint/dev/general-development
 	// /how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling
 	defaultLC      = 1
-	driveDefaultLC = 2
+	driveDefaultLC = 1
```
ryanfkeepers, line 153: if 1 is the default for drive, then do we need two consts?  This can probably get removed in favor of the single default.

ryanfkeepers (approved):

src/internal/m365/graph/concurrency_middleware.go
pandeyabs: That works.

Issue #4371, opened 2023-09-26
Possible incorrect incremental backups for Groups


meain:
Backups make use of `canUsePreviousBackup` variable to decide if we should be using the previous backup as a base and do an incremental update or if we should be doing a full backup. This causes issue with the Groups incremental backups as messages and sites can have different value for the value and we might use just the delta update but with the variable set to true. This can be even more dangerous when we have multiple sites being backed up as we might just run into metadata parsing issue just for one site.

Here is a possible scenario where we would end up with an incorrect backup:
- Backing up sites
  - Metadata looks good, so we set `canUsePreviousBackup` to true
  - We only send back incremental set of collections
- Now we try to back up messages
  - Metadata is wrong
  - We set `canUsePreviousMetadata` to false and send back the full collection

In this case, even though `canUsePreviousMetadata` is false at the end of the function, we only have incremental set of collections for sites. This will cause our newer backup to only have the incremental data within in.

Ref:  https://github.com/alcionai/corso/blob/2bd7ee45ddf2eb6d0cf1a02af8931949d1e0a3b8/src/internal/m365/service/groups/backup.go#L31

ashmrtn:
I think ideally we should try to move away from having the additional `canUsePreviousBackup` return value and instead send back collections that delete the root of the hierarchy (and associated subdirectories) if needed. For example, if we were doing a drive backup and found out that the delta token expired we'd send back a collection for path `tenant/service/user/files/drive-id/root` marked with state `DeletedState`

This change can be done bit by bit in the code, but would allow individual control over different data types/instances of a data type

PR #4394, opened 2023-09-28
Return tombstone instead of canUsePreviousBackup

meain:
This is necessary to fix the correctness of the backup.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/4371

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -147,7 +147,11 @@ func deserializeMetadata(
 				// these cases. We can make the logic for deciding when to continue vs.
 				// when to fail less strict in the future if needed.
 				if err != nil {
-					return nil, nil, false, clues.Stack(err).WithClues(ictx)
+					errs.Fail(clues.Stack(err).WithClues(ictx))
+
+					breakLoop = true
+
+					break
 				}
 			}
 		}
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -135,11 +135,6 @@ func deserializeMetadata(
 					continue
 				}
 
-				if err == nil {
-					// Successful decode.
-					continue
-				}
-
 				// This is conservative, but report an error if either any of the items
 				// for any of the deserialized maps have duplicate drive IDs or there's
 				// some other problem deserializing things. This will cause the entire
@@ -149,9 +144,7 @@ func deserializeMetadata(
 				if err != nil {
 					errs.Fail(clues.Stack(err).WithClues(ictx))
 
-					breakLoop = true
-
-					break
+					return map[string]string{}, map[string]map[string]string{}, false, nil
 				}
 			}
 		}
--- a/src/internal/m365/collection/groups/channel_handler.go
+++ b/src/internal/m365/collection/groups/channel_handler.go
@@ -67,6 +67,15 @@ func (bh channelsBackupHandler) canonicalPath(
 			false)
 }
 
+func (bh channelsBackupHandler) MessagePathPrefix(tenantID string) (path.Path, error) {
+	return path.Build(
+		tenantID,
+		bh.protectedResource,
+		path.GroupsService,
+		path.ChannelMessagesCategory,
+		false)
+}
+
 func (bh channelsBackupHandler) GetChannelMessage(
 	ctx context.Context,
 	teamID, channelID, itemID string,
--- a/src/internal/m365/collection/groups/channel_handler.go
+++ b/src/internal/m365/collection/groups/channel_handler.go
@@ -67,7 +67,7 @@ func (bh channelsBackupHandler) canonicalPath(
 			false)
 }
 
-func (bh channelsBackupHandler) MessagePathPrefix(tenantID string) (path.Path, error) {
+func (bh channelsBackupHandler) PathPrefix(tenantID string) (path.Path, error) {
 	return path.Build(
 		tenantID,
 		bh.protectedResource,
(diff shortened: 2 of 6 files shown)


src/internal/m365/collection/drive/collections.go
```diff
@@ -147,7 +147,11 @@ func deserializeMetadata(
 				// these cases. We can make the logic for deciding when to continue vs.
 				// when to fail less strict in the future if needed.
 				if err != nil {
-					return nil, nil, false, clues.Stack(err).WithClues(ictx)
+					errs.Fail(clues.Stack(err).WithClues(ictx))
+
+					breakLoop = true
+
+					break
```
meain, line 154: I initially did this as this was the only way to get `deserializeMetadata` to return something other than true for `canUsePreviousBackup`. But this feels like we should ideally not be having a hard fail on this. What we there was some network issue midway through writing the metadata file? Any thoughts @ashmrtn @ryanfkeepers

ashmrtn: we should be able to assume that either the metadata file is complete (written fully but possibly malformed if there was a bug in the code) or not there at all due to how kopia works. We're ignoring things like bit rot and bit flips for the moment, but those are much more involved to handle

is there a reason we can't continue to just return false here? Since we don't know what we don't know if we see an error (i.e. we don't know what the previous drives were) we should just fetch everything again. I don't expect we'll have multiple metadata file sets, but I think the conservative approach of assuming we can't trust anything if one thing is bad is ok. I'm honestly not sure how to reason about potential overlap between different meatadata file sets

ashmrtn: also, I think the end goal should be creating/returning tombstone collections in code in files like this rather than the caller of this code. I realize that's a much larger change though so it's not reasonable to expect it to be done in this PR

src/internal/m365/collection/groups/channel_handler.go
```diff
@@ -67,6 +67,15 @@ func (bh channelsBackupHandler) canonicalPath(
 			false)
 }
 
+func (bh channelsBackupHandler) MessagePathPrefix(tenantID string) (path.Path, error) {
```
ryanfkeepers, line 70: `message` won't abstract well if/when we add other handlers.
```suggestion
func (bh channelsBackupHandler) PathPrefix(tenantID string) (path.Path, error) {
```

src/internal/m365/collection/drive/collections.go
```diff
@@ -147,7 +147,11 @@ func deserializeMetadata(
 				// these cases. We can make the logic for deciding when to continue vs.
 				// when to fail less strict in the future if needed.
 				if err != nil {
```
ryanfkeepers, line 149: this is the else part of an if/else, without the else.  Since you're touch this code, would you mind updating the code here and above?  Either to remove the `if err == nil` and kkeep this conditional, or to keep the continue-on-nil and remove this conditional instead.  We don't need both.

src/internal/m365/collection/drive/collections.go
ryanfkeepers: > is there a reason we can't continue to just return false here?

don't see why not.  This is doing the same eventual effect, just via a longer path.  Returning here will skip the block on 167, but that's redundant anyway since we'll end up inside 187 regardless.

> But this feels like we should ideally not be having a hard fail on this.

Agreed. Seems reasonable.  Since we're using the return value on 190 anyway (empty maps, false, nil), why not replicate that return here?

> I think the end goal should be creating/returning tombstone collections in code in files like this rather than the caller of this code

100% agreed.  Two questions: 1/ should we make a tech debt ticket to track the refactor, or include it as a v2 in this change?  2/ The more we democratize collection creation across multiple owners, does that increase the possibility of producing conflicting collection sets?  Is this something we should have preventative checks at some higher level to avoid?

ryanfkeepers (approved):
approving to unblock.  Conversation can continue as needed.

src/internal/m365/collection/drive/collections.go
ashmrtn: >Two questions: 1/ should we make a tech debt ticket to track the refactor, or include it as a v2 in this change?

Tech debt may be better. It's going to end up touching all service collection generators since they all have it right now. While I'd like to see the work done, I don't want to bog down other things by requiring it be done with this work

>2/ The more we democratize collection creation across multiple owners, does that increase the possibility of producing conflicting collection sets? Is this something we should have preventative checks at some higher level to avoid?

There's some basic checks on this in kopia wrapper, for checking that there's not multiple collections referring to the same FullPath(). I don't recall if there's a check that there's not multiple collections referring to the same PreviousPath(), there may be. Are there other checks you think we should have?

Also note that the checks above are _runtime_ checks. We can't tell before hand if there'll be collisions in an easy manner. It's one of the things we kind of have to reason through when writing the code

ryanfkeepers: > Are there other checks you think we should have?

Not off the top of my head.  But if memory serves, those checks will fail out the backup, right?  Ideally, we'd be able to "fix" the state before failing out (not sure off the top of my head).  Getting into a state of consistent failure (or, perhaps worse, inconsisstent failure) due to this scares me a bit, since replication of that state might be very difficult.

meain: > also, I think the end goal should be creating/returning tombstone collections in code in files like this rather than the caller of this code. I realize that's a much larger change though so it's not reasonable to expect it to be done in this PR

Agreed, I am planning to pick it up next. Just wanted to do the minimum necessary to ensure correctness of groups backup first. Lemme create a [tech-debt ticket](https://github.com/alcionai/corso/issues/4405)

Issue #3943, opened 2023-08-01
Add explicit 1:1 mapping between snapshots & associated backup models. 


pandeyabs:
### What happened?

Now that we require assist snapshots to have an associated backup model, we should make the 1:1 mapping between the backup model and the snapshot explicit for both merge & assist backups.

**Changes**
* Don't include merge snapshot in assist snapshots during base selection. This would also need some changes to kopia wrapper.
*  Return `{ManifestEntry, BackupEntry}` tuples for merge bases & assist bases during base selection.

### Corso Version?

Corso v0.11.1

### Where are you running Corso?

NA

### Relevant log output

```shell
NA
```

ashmrtn:
Basic work for separation is done. Some other cleanup can be done to consolidate data fields

PR #4402, opened 2023-09-28
Unify backup bases fields

ashmrtn:
Merge fields in the backup bases struct since
we assume they need to be tracked together
anyway

This PR attempts to keep the API as close to
what it currently is as possible. A future PR
will go through and update the
API/tests/mocks

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3943

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/backup_bases.go
+++ b/src/internal/kopia/backup_bases.go
@@ -406,10 +406,10 @@ func fixupReasons(
 			continue
 		}
 
-		// We've got at least on base for this Reason. The below finds which base to
-		// keep based on the creation time of the bases. If there's multiple bases
-		// in the input slice then we'll log information about the ones that we
-		// didn't add to the result set.
+		// We've got at least one base for this Reason. The below finds which base
+		// to keep based on the creation time of the bases. If there's multiple
+		// bases in the input slice then we'll log information about the ones that
+		// we didn't add to the result set.
 
 		// Sort in reverse chronological order so that it's easy to find the
 		// youngest base.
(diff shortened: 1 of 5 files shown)


src/internal/kopia/backup_bases.go
```diff
@@ -55,56 +58,67 @@ func (bb *backupBases) SnapshotAssistBases() []ManifestEntry {
 		return nil
 	}
 
+	res := []ManifestEntry{}
+
+	for _, ab := range bb.assistBases {
```
ryanfkeepers, line 63: it's not much, but I count six locations where we do this same iteration.  Is that worth it's own func?

src/internal/kopia/base_finder.go
```diff
@@ -206,7 +206,7 @@ func (b *baseFinder) findBasesInSet(
 	ctx context.Context,
 	reason identity.Reasoner,
 	metas []*manifest.EntryMetadata,
-) (*backupBase, *backupBase, error) {
+) (*BackupBase, *BackupBase, error) {
```
ryanfkeepers, line 209: Nit: I know it's not part of the changes here, per se, but return values like this make me just a little uneasy since it's difficult to distinguish which value is which without referencing either the func implementation or other code.  Could we pack these into a struct as a way of labeling them?

ryanfkeepers (commented):
Lgtm.  Going to let @pandeyabs have final approval, since he's more an owner of this code than I am.

pandeyabs:
Sorry, I didn't notice that @ryanfkeepers had tagged me a while back. I'll review this tomorrow.

src/internal/kopia/backup_bases.go
ashmrtn: this will be going away in the next PR. This is just to keep the interface intact so there's (relatively) less changes

src/internal/kopia/base_finder.go
ashmrtn: I'll pick this up in a later PR. I'd like to get this one done and there's already quite a bit going on here

pandeyabs (approved):

PR #4397, opened 2023-09-28
Update kopia password

neha-Gupta1:
Update Kopia repo password from Corso CLI

#### Does this PR need a docs update or release note?

- [ ] :clock1: Yes, but in a later PR

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* https://github.com/alcionai/corso/issues/4061

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Issue #4061, opened 2023-08-18
Provide CLI support to modify Corso passphrase


vkamra:
The Corso passphrase for a repository can currently only be modified by using the Kopia CLI.

This tracks providing Corso CLI support to do this.

ashmrtn:
[here's](https://github.com/kopia/kopia/blob/1d0f31f32e565ac4f652bc86ec5a22062cabcdf1/cli/command_repository_change_password.go#L25) the kopia CLI code that handles changing passwords. Part of it is custom (or at least partly custom) functions to make sure passwords match etc. We'll need to figure out some way to handle that ourselves perhaps.

The other part is actually changing the password. There's a kopia function that we can call for that

vkamra:
This is in-progress

vkamra:
Done with #4518

PR #4406, opened 2023-09-29
update Kopia password

neha-Gupta1:
---
This PR is raised as part of splitting https://github.com/alcionai/corso/pull/4397 into smaller parts
It cover connecting with Kopia and updating repos password

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* https://github.com/alcionai/corso/issues/4061

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test

### DIFF
--- a/src/internal/kopia/conn.go
+++ b/src/internal/kopia/conn.go
@@ -592,7 +592,9 @@ func (w *conn) UpdatePassword(ctx context.Context, password string, opts reposit
 	defer kopiaRef.Close(ctx)
 
 	kopiaRepo := kopiaRef.Repository.(repo.DirectRepository)
-	err := kopiaRepo.FormatManager().ChangePassword(ctx, password)
+	if err := kopiaRepo.FormatManager().ChangePassword(ctx, password); err != nil {
+		return clues.Wrap(err, "unable to update password")
+	}
 
-	return clues.Wrap(err, "unable to update password")
+	return nil
 }
--- a/src/internal/kopia/conn.go
+++ b/src/internal/kopia/conn.go
@@ -578,3 +578,17 @@ func (w *conn) LoadSnapshot(
 func (w *conn) SnapshotRoot(man *snapshot.Manifest) (fs.Entry, error) {
 	return snapshotfs.SnapshotRoot(w.Repository, man)
 }
+
+func (w *conn) UpdatePassword(ctx context.Context, password string, opts repository.Options) error {
+	kopiaRef := NewConn(w.storage)
+	if err := kopiaRef.Connect(ctx, opts); err != nil {
+		return clues.Wrap(err, "connecting kopia client")
+	}
+
+	defer kopiaRef.Close(ctx)
+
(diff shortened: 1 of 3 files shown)


src/pkg/repository/repository_test.go
```diff
@@ -236,6 +236,37 @@ func (suite *RepositoryIntegrationSuite) TestConnect() {
...
+
+	// now re-connect
+	err = r.Connect(ctx)
+	assert.NoError(t, err, clues.ToCore(err))
+
+	err = r.UpdatePassword(ctx, "newpass")
+	require.NoError(t, err, clues.ToCore(err))
+
+	// now reconnect with new pass
+	err = r.Connect(ctx)
```
ashmrtn, line 266: I'm not sure this is doing what we hope it is. I think it would be safer to use a separate repo instance, especially since the repo struct [caches](https://github.com/alcionai/corso/blob/2fe0e8643e39c59a880d68ad4c0045b52dd23f39/src/pkg/repository/repository.go#L244) the storage config which itself [contains](https://github.com/alcionai/corso/blob/2fe0e8643e39c59a880d68ad4c0045b52dd23f39/src/pkg/storage/testdata/storage.go#L47) the [repo passphrase](https://github.com/alcionai/corso/blob/2fe0e8643e39c59a880d68ad4c0045b52dd23f39/src/pkg/storage/testdata/storage.go#L78)

I'm not really a fan of the below, but I believe it should allow updating an existing storage struct with a new password
```go
st.Config[storeTD.CorsoPassphrase] = "newpass"
```

ashmrtn (approved):
approving to unblock though tests need a small adjustment to ensure they do what we think they do

src/internal/kopia/conn.go
```diff
@@ -578,3 +578,17 @@ func (w *conn) LoadSnapshot(
 func (w *conn) SnapshotRoot(man *snapshot.Manifest) (fs.Entry, error) {
 	return snapshotfs.SnapshotRoot(w.Repository, man)
 }
+
+func (w *conn) UpdatePassword(ctx context.Context, password string, opts repository.Options) error {
```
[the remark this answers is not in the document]
neha-Gupta1: Done.

src/pkg/repository/repository_test.go
neha-Gupta1: passphrase is saved in - `common_corsoPassphrase` and its not exposed mostly because of security reasons. And not sure if we should touch it. Also, the integration test in coming PRs will cover the other part of this test-case. As of know I will move with it. And if we are not satisfied with the testing, I will work on improving it.

ashmrtn: Given that this is just a test case it should be ok from a security perspective to mess with it. Not my favorite design paradigm, but at least it works

neha-Gupta1: Done

PR #4407, opened 2023-09-29
Fix restore backup tests

ashmrtn:
Exchange tests inadvertently got disabled since
it wasn't finding path matches for returned
BackupCollections. This switches to using
LocationPath which does allow for matching

Most contacts tests are disabled since restore
doesn't support nested folders

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/controller_test.go
+++ b/src/internal/m365/controller_test.go
@@ -887,6 +887,7 @@ func (suite *ControllerIntegrationSuite) TestRestoreAndBackup_core() {
 				},
 			},
 		},
+		// TODO(ashmrtn): Re-enable when we can restore contacts to nested folders.
 		//{
 		//	name:    "MultipleContactsSingleFolder",
 		//	service: path.ExchangeService,
@@ -1044,6 +1045,7 @@ func (suite *ControllerIntegrationSuite) TestRestoreAndBackup_core() {
 
 func (suite *ControllerIntegrationSuite) TestMultiFolderBackupDifferentNames() {
 	table := []restoreBackupInfo{
+		// TODO(ashmrtn): Re-enable when we can restore contacts to nested folders.
 		//{
 		//	name:    "Contacts",
 		//	service: path.ExchangeService,
--- a/src/internal/m365/controller_test.go
+++ b/src/internal/m365/controller_test.go
@@ -860,6 +860,33 @@ func (suite *ControllerIntegrationSuite) TestRestoreAndBackup_core() {
 				},
 			},
 		},
+		{
+			name:    "MultipleContactsInRestoreFolder",
+			service: path.ExchangeService,
+			collections: []stub.ColInfo{
+				{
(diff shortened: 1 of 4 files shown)


src/internal/m365/controller_test.go
```diff
@@ -887,49 +887,76 @@ func (suite *ControllerIntegrationSuite) TestRestoreAndBackup_core() {
...
-						{
-							Name:      "someencodeditemID5",
-							Data:      exchMock.ContactBytes("Bernard"),
-							LookupKey: "Bernard",
-						},
-					},
-				},
-			},
-		},
+		//{
```
ryanfkeepers, line 890: what's the plan for these tests?  Delete them?  Fix them later?

src/internal/m365/helper_test.go
```diff
@@ -923,30 +923,9 @@ func checkHasCollections(
...
-			fp.ProtectedResource(),
-			fp.Service(),
-			fp.Category(),
-			false)
-		if !assert.NoError(t, err, clues.ToCore(err)) {
-			continue
-		}
-
-		gotNames = append(gotNames, p.String())
+		gotNames = append(gotNames, loc.String())
```
ryanfkeepers, line 928: well that feels good

src/internal/m365/helper_test.go
```diff
@@ -967,14 +946,18 @@ func checkCollections(
 
 	for _, returned := range got {
 		var (
+			expectedColData map[string][]byte
```
ryanfkeepers, line 949:
```suggestion
			expectedColDataByLoc map[string][]byte
```

ryanfkeepers (approved):

src/internal/m365/controller_test.go
ashmrtn: ideally we'll (at some point) all restores to nested contacts folders. We can then re-enable these tests. I'll add a TODO about that to the code

ryanfkeepers: I'm a little worried about expectations there.  Msoft's hierarchy support for contacts, whether its increasing or decreasing, is difficult to track.  It _does_ feel slightly like tags are their long term goal? Same result either way, I suppose.  Something to make clear in the todo, maybe.

ashmrtn: well, if we're not testing it it will be extra difficult to track changes for (though may result in a bit more noise). At some point we'll need to figure out a long term plan for support for contacts and what that should entail

ryanfkeepers: Oh, oh, no, I'm not saying we shouldn't test it.  Just pondering what exactly we want to test.  It's weirdness either way. Ah well, this is a good topic for a sync design discussion sometime.

PR #4414, opened 2023-09-30
iterate pager results with channels

ryanfkeepers:
corrects the prior PRs full-scale batching of driveItems by adding a channel to the api layer's enumerateItems and driveEnumerateItems funcs.  To avoid changes
in most parts of code, also adds batch wrappers
that hide these changes from existing pager usage.

DriveItem collection processing now ranges over
a channel of pager results, allowing it to process pages in a stream rather than batching all items.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/drive_pager.go
+++ b/src/pkg/services/m365/api/drive_pager.go
@@ -77,7 +77,7 @@ func (c Drives) GetItemsInContainerByCollisionKey(
 	ctx = clues.Add(ctx, "container_id", containerID)
 	pager := c.NewDriveItemPager(driveID, containerID, idAnd("name")...)
 
-	items, err := enumerateItems(ctx, pager)
+	items, err := batchEnumerateItems(ctx, pager)
 	if err != nil {
 		return nil, graph.Wrap(ctx, err, "enumerating drive items")
 	}
@@ -101,7 +101,7 @@ func (c Drives) GetItemIDsInContainer(
 	ctx = clues.Add(ctx, "container_id", containerID)
 	pager := c.NewDriveItemPager(driveID, containerID, idAnd("file", "folder")...)
 
-	items, err := enumerateItems(ctx, pager)
+	items, err := batchEnumerateItems(ctx, pager)
 	if err != nil {
 		return nil, graph.Wrap(ctx, err, "enumerating contacts")
 	}
@@ -133,7 +133,7 @@ type DriveItemDeltaPageCtrl struct {
 
 func (c Drives) newDriveItemDeltaPager(
 	driveID, prevDeltaLink string,
-	selectProps ...string,
+	cc CallConfig,
 ) *DriveItemDeltaPageCtrl {
 	preferHeaderItems := []string{
 		"deltashowremovedasdeleted",
@@ -147,8 +147,8 @@ func (c Drives) newDriveItemDeltaPager(
 		QueryParameters: &drives.ItemItemsItemDeltaRequestBuilderGetQueryParameters{},
 	}
 
-	if len(selectProps) > 0 {
-		options.QueryParameters.Select = selectProps
+	if len(cc.Select) > 0 {
+		options.QueryParameters.Select = cc.Select
 	}
 
 	builder := c.Stable.
@@ -197,26 +197,33 @@ func (p *DriveItemDeltaPageCtrl) ValidModTimes() bool {
 	return true
 }
 
-// EnumerateDriveItems will enumerate all items in the specified drive and hand them to the
-// provided `collector` method
+// EnumerateDriveItems will enumerate all items in the specified drive and stream them page
+// by page, along with the delta update and any errors, to the provided channel.
 func (c Drives) EnumerateDriveItemsDelta(
 	ctx context.Context,
 	driveID string,
 	prevDeltaLink string,
-	selectProps []string,
-) (
-	[]models.DriveItemable,
-	DeltaUpdate,
-	error,
-) {
-	pager := c.newDriveItemDeltaPager(driveID, prevDeltaLink, selectProps...)
+	cc CallConfig,
+) NextPageResulter[models.DriveItemable] {
+	deltaPager := c.newDriveItemDeltaPager(
+		driveID,
+		prevDeltaLink,
+		cc)
 
-	items, du, err := deltaEnumerateItems[models.DriveItemable](ctx, pager, prevDeltaLink)
-	if err != nil {
-		return nil, du, clues.Stack(err)
+	npr := &nextPageResults[models.DriveItemable]{
+		pages: make(chan nextPage[models.DriveItemable]),
 	}
 
-	return items, du, nil
+	// asynchronously enumerate pages on the caller's behalf.
+	// they only need to consume the pager and call Results at
+	// the end.
+	go deltaEnumerateItems[models.DriveItemable](
+		ctx,
+		deltaPager,
+		npr,
+		prevDeltaLink)
+
+	return npr
 }
 
 // ---------------------------------------------------------------------------
@@ -354,11 +361,17 @@ func GetAllDrives(
 	ctx context.Context,
 	pager Pager[models.Driveable],
 ) ([]models.Driveable, error) {
-	ds, err := enumerateItems(ctx, pager)
-	if err != nil && (clues.HasLabel(err, graph.LabelsMysiteNotFound) ||
-		clues.HasLabel(err, graph.LabelsNoSharePointLicense)) {
+	ds, err := batchEnumerateItems(ctx, pager)
+
+	// no license or drives available.
+	// return a non-error and let the caller assume an empty result set.
+	// TODO: is this the best way to handle this?
+	// what about returning a ResourceNotFound error as is standard elsewhere?
+	if err != nil &&
+		(clues.HasLabel(err, graph.LabelsMysiteNotFound) || clues.HasLabel(err, graph.LabelsNoSharePointLicense)) {
 		logger.CtxErr(ctx, err).Infof("resource owner does not have a drive")
-		return make([]models.Driveable, 0), nil // no license or drives.
+
+		return make([]models.Driveable, 0), nil
 	}
 
 	return ds, graph.Stack(ctx, err).OrNil()
--- a/src/pkg/services/m365/api/item_pager.go
+++ b/src/pkg/services/m365/api/item_pager.go
@@ -28,6 +28,85 @@ type DeltaUpdate struct {
 	Reset bool
 }
 
+type NextPager[T any] interface {
+	// reset should only true on the iteration where the delta pager's Reset()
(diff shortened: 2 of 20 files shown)


ashmrtn:
if the PR before this one isn't going to reduce the total changes, can we please break up this PR into either multiple commits in this PR or multiple PRs? Right now it's ~2k additions + removals in a single commit which feels a bit intractable for review

src/pkg/services/m365/api/drive_pager.go
```diff
@@ -197,25 +197,26 @@ func (p *DriveItemDeltaPageCtrl) ValidModTimes() bool {
 	return true
 }
 
-// EnumerateDriveItems will enumerate all items in the specified drive and hand them to the
-// provided `collector` method
+// EnumerateDriveItems will enumerate all items in the specified drive and stream them page
+// by page, along with the delta update and any errors, to the provided channel.
 func (c Drives) EnumerateDriveItemsDelta(
 	ctx context.Context,
+	ch chan<- NextPage[models.DriveItemable],
```
ashmrtn, line 204: (commenting here because overall PR comments have no threading an become difficult to track)

I'm not comfortable with the design of this API. It feels a lot like we're trying to make something that fits both the iterator pattern and the "get all the things and return them" pattern but not doing a great job in the end because of that

The current code is laid out such that drives use something more akin to an iterator rather than the pager model we have for Exchange where everything is returned in a single result

This PR attempts to add channels to allow a more iterator-based approach, but it complicates several things in the overall design/usage of the system. The things that stand out the most are:
1. channel creation and ownership. I think the pager should be making a channel and returning it instead of taking in a channel. This is especially relevant because the code called by the pager function closes the channel at the end of iteration. The result is unclear ownership of the channel which will likely confuse people
1. handling of delta resets. The Exchange design can transparently handle delta resets because it only returns the results at the end. That means that internally it can clear the results if it needs to and other code doesn't have to worry about that detail at all. With the new design, it seems like it would be really easy to forget to reset the result in the caller when a delta reset occurs. Splitting the logic of switching pagers/resetting pagers and handling results sets just generally seems error prone and should be avoided. Doing so would require pushing that handling up a later and, quite likely, terminating iteration when the pager needs reset. That would force callers to handle that state explicitly
2. return value/expected usage in general. The return value from this function and expected usage is confusing in general. From what I can tell, this function assumes the caller will create a channel, create a goroutine to drain the channel, and then call this function to get results. It feels like a gotcha where it's too easy to deadlock. The usual iterator/channel pattern is to first create the iterator with a call and then traverse it with a second call. That separates the creation of the iterator (non-blocking) from the traversal of the iterator (potentially blocking). What we have here is a combined creation and iteration phase that could block (and even deadlock if used incorrectly).

ryanfkeepers: I can solve one of these issues: a better interface for having the caller iterate through pages.  Latest change swaps out the channel for a pager interface that's simpler to produce and consume.  The caller won't need to worry about either production of the channel, nor the control of routines.

Data management during reset, however, isn't up for change.  Any caller that wants to stream rather than batch their request (and at this time, only onedrive opts for streamed enumeration rather than batched) is necessarily opting to handle resets on behalf of the api layer.

src/pkg/services/m365/api/drive_pager.go
```diff
@@ -353,11 +360,17 @@ func GetAllDrives(
...
 	pager Pager[models.Driveable],
 ) ([]models.Driveable, error) {
-	ds, err := enumerateItems(ctx, pager)
-	if err != nil && (clues.HasLabel(err, graph.LabelsMysiteNotFound) ||
-		clues.HasLabel(err, graph.LabelsNoSharePointLicense)) {
+	ds, err := batchEnumerateItems(ctx, pager)
+
+	// no license or drives available.
+	// return a non-error and let the caller assume an empty result set.
+	// TODO: is this the best wayy to handle this?
```
ashmrtn, line 367: nit: spelling
```suggestion
	// TODO: is this the best way to handle this?
```

src/pkg/services/m365/api/item_pager.go
```diff
@@ -27,6 +27,57 @@ type DeltaUpdate struct {
...
+
+	return np.items, np.reset, false
+}
+
+func (npr *nextPageResults[T]) Results() (DeltaUpdate, error) {
+	// if the pager hasn't closed yet, drain out the pages iterator
+	// to avoid leaking routines, and to ensure we get results.
+	for npr.pages != nil {
+		<-npr.pages
+	}
```
ashmrtn, line 72: nit: could you please switch this to range directly over the channel? I think the code should just call `close(npr.pages)` when it's done instead of having a separate close method (unless there's a high probability of needing to do other work)

If it doesn't just call close on the channel then please somehow synchronize access with a mutex. With the current code it's possible to get an NPE if `close()` is called between the if-check and attempting to read from the channel

src/pkg/services/m365/api/item_pager.go
```diff
@@ -27,6 +27,57 @@ type DeltaUpdate struct {
...
+	du    DeltaUpdate
+	err   error
+}
+
+func (npr *nextPageResults[T]) NextPage() ([]T, bool, bool) {
+	if npr.pages == nil {
+		return nil, false, true
+	}
+
+	np := <-npr.pages
```
ashmrtn, line 62: this could also result in an NPE under some conditions (see other comment in `Results()`)

src/pkg/services/m365/api/item_pager.go
```diff
@@ -111,18 +165,40 @@ func enumerateItems[T any](
 		// get the next page of data, check for standard errors
 		page, err := pager.GetPage(ctx)
 		if err != nil {
-			return nil, graph.Stack(ctx, err)
+			npr.err = graph.Stack(ctx, err)
+			return
 		}
 
-		result = append(result, page.GetValue()...)
+		npr.pages <- nextPage[T]{items: page.GetValue()}
```
ashmrtn, line 172: nit: just for safety could you please expand this (and other attempts to send on the pager channel that are in this file) to check for context cancellation ?
```suggestion
		select {
			case <-ctx.Done():
				// Some error handling/logging
				
			case npr.pages <- nextPage[T]{items: page.GetValue()}:
		}
```

src/pkg/services/m365/api/item_pager.go
```diff
@@ -136,11 +212,19 @@ type DeltaPager[T any] interface {
 	ValidModTimer
 }
 
+// enumerates pages of items, streaming each page to the provided channel.
+// the DeltaUpdate, reset notifications, and any errors are also fed to the
+// same channel.
+// Returns false if conditions disallow making delta calls for the provided
+// pager.  Returns true otherwise, even in the event of an error.
```
ashmrtn, line 219: nit: this part of the comment is stale
```suggestion
```

src/pkg/services/m365/api/item_pager.go
```diff
@@ -27,6 +27,57 @@ type DeltaUpdate struct {
...
+type NextPager[T any] interface {
+	NextPage() (items []T, reset, done bool)
+}
+
+type nextPage[T any] struct {
+	items []T
+	// reset is only true on the iteration where the delta pager's Reset()
+	// is called.  Callers can use it to reset any data aggregation they
+	// currently use.  After that loop, it will be false again, though the
+	// DeltaUpdate will still contain the expected value.
```
ashmrtn, line 39: nit: could you please move this comment to be on `NextPage()` since that's what callers will actually use?

This comment also feels a bit underspecified. What is the behavior if `len(items) > 0 && reset`? Are those items from the page that should be discarded or from the page that should be kept? Or are we defining an API where it's considered impossible to get that result? (see golang docs for `io.Read` for a [similar situation](https://pkg.go.dev/io#Reader)). Whatever behavior we choose, we should at least document it in comments so callers know what to expect and how to handle edge cases

src/pkg/services/m365/api/item_pager.go
```diff
@@ -186,12 +276,37 @@ func deltaEnumerateItems[T any](
...
+		}
+		results = []T{}
+	)
+
+	go deltaEnumerateItems[T](ctx, pager, &npr, prevDeltaLink)
+
+	for is, reset, done := npr.NextPage(); !done; {
+		if reset {
+			results = []T{}
+		}
```
ashmrtn, line 302: just calling out that whatever behavior is decided for `NextPage` this should be updated to match (if needed)

ashmrtn (approved):
approving to unblock. Comments need addressed prior to merging though as there are several potential NPEs that are present in the current design

src/pkg/services/m365/api/drive_pager.go
ashmrtn: The updated design definitely help alleviate some of the issues the earlier design had. I do still worry about splitting the reset handling logic, but it seems reasonable to tackle that in a separate PR

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: Oh, huh, I forgot the close in close().  Yay for that.  Thanks for the catch.

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: sure, can do

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: Documented, thanks for the call out.

Fwiw i'm not _implementing_ an items>0 && reset == true possibility.  But the interface doesn't enforce it, so I've added expectations there.

src/pkg/services/m365/api/item_pager.go
ryanfkeepers: that's already compliant with expectations.

src/pkg/services/m365/api/item_pager.go
```diff
@@ -27,6 +27,84 @@ type DeltaUpdate struct {
...
+
+	np := <-npr.pages
+
+	return np.items, np.reset, false
+}
+
+func (npr *nextPageResults[T]) Results() (DeltaUpdate, error) {
+	if npr.pages != nil {
+		//nolint:revive
+		for range npr.pages {
```
ashmrtn, line 94: this can still generate NPEs for the same reason it could the first time around. I'd advise not setting the channel to `nil` until this function is called. Or just not setting it to nil at all and waiting for this struct to go out of scope prior to the channel being garbage collected

src/pkg/services/m365/api/item_pager.go
ashmrtn: >Fwiw i'm not implementing an items>0 && reset == true possibility

It's fine if it's not implemented. There is already other code that _assumes_ a specific implementation though (flagged in another comment) so that assumed behavior should be documented

PR #4416, opened 2023-09-30
apply channels and streaming to drive collections

ryanfkeepers:
Updates drive collection processing with the new pattern of streaming pages from the api using channels.

This is the last in a multipart update that has been separated for ease of review.  Everything should now pass.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
@@ -272,13 +273,6 @@ func (c *Collections) Get(
 			excludedItemIDs = map[string]struct{}{}
 			oldPrevPaths    = oldPrevPathsByDriveID[driveID]
 			prevDeltaLink   = prevDriveIDToDelta[driveID]
-
-			// itemCollection is used to identify which collection a
-			// file belongs to. This is useful to delete a file from the
-			// collection it was previously in, in case it was moved to a
-			// different collection within the same delta query
-			// item ID -> item ID
-			itemCollection = map[string]string{}
 		)
 
 		delete(driveTombstones, driveID)
@@ -295,13 +289,16 @@ func (c *Collections) Get(
 			"previous metadata for drive",
 			"num_paths_entries", len(oldPrevPaths))
 
-		items, du, err := c.handler.EnumerateDriveItemsDelta(
-			ictx,
+		du, newPrevPaths, err := c.PopulateDriveCollections(
+			ctx,
 			driveID,
+			driveName,
+			oldPrevPaths,
+			excludedItemIDs,
 			prevDeltaLink,
-			api.DefaultDriveItemProps())
+			errs)
 		if err != nil {
-			return nil, false, err
+			return nil, false, clues.Stack(err)
 		}
 
 		// It's alright to have an empty folders map (i.e. no folders found) but not
@@ -311,20 +308,6 @@ func (c *Collections) Get(
 		// for collections when not actually getting delta results.
 		if len(du.URL) > 0 {
 			driveIDToDeltaLink[driveID] = du.URL
-		}
-
-		newPrevPaths, err := c.UpdateCollections(
-			ctx,
-			driveID,
-			driveName,
-			items,
-			oldPrevPaths,
-			itemCollection,
-			excludedItemIDs,
-			du.Reset,
-			errs)
-		if err != nil {
-			return nil, false, clues.Stack(err)
 		}
 
 		// Avoid the edge case where there's no paths but we do have a valid delta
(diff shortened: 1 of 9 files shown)


src/internal/m365/service/onedrive/mock/handlers.go
```diff
@@ -316,7 +316,7 @@ func (edibd EnumerateItemsDeltaByDrive) EnumerateDriveItemsDelta(
 	_ api.CallConfig,
 ) api.NextPageResulter[models.DriveItemable] {
 	didp := edibd.DrivePagers[driveID]
-	return &didp
+	return didp
```
ashmrtn, line 319: this probably does need to return the pointer to the struct. Otherwise `NextPage` probably won't work as expected (which is why I'm guessing the diff for the index increment)

src/pkg/services/m365/api/drive_pager_test.go
```diff
@@ -199,7 +199,8 @@ func (suite *DrivePagerIntgSuite) TestEnumerateDriveItems() {
 				Select: api.DefaultDriveItemProps(),
 			})
 
-	for page, reset, done := pager.NextPage(); !done; {
+	page, reset, done := pager.NextPage()
```
ashmrtn, line 202: I'm curious as to why these variables need to exist outside the loop? I don't see any references to them in the code after the loop

src/internal/m365/collection/drive/collections.go
```diff
@@ -688,232 +670,297 @@ func (c *Collections) getCollectionPath(
...
 
-// UpdateCollections initializes and adds the provided drive items to Collections
-// A new collection is created for every drive folder (or package).
+// PopulateDriveCollections initializes and adds the provided drive items to Collections
+// A new collection is created for every drive folder.
 // oldPrevPaths is the unchanged data that was loaded from the metadata file.
 // This map is not modified during the call.
 // currPrevPaths starts as a copy of oldPaths and is updated as changes are found in
 // the returned results.  Items are added to this collection throughout the call.
 // newPrevPaths, ie: the items added during this call, get returned as a map.
```
ashmrtn, line 679: please update this comment to reflect the new set of parameters

src/internal/m365/collection/drive/collections.go
```diff
@@ -688,232 +670,297 @@ func (c *Collections) getCollectionPath(
...
+
+		if isFolder {
+			skip = fault.ContainerSkip(fault.SkipMalware, driveID, itemID, itemName, addtl)
 		}
 
-		switch {
-		case isFolder:
-			// Deletions are handled above so this is just moves/renames.
-			var prevPath path.Path
+		skipper.AddSkip(ctx, skip)
```
ashmrtn, line 784: honestly this feels like a bit of an anti-pattern. We're taking in (a part of) the fault bus, but only using it to add skips. Errors are returned like normal error handling with the assumption that the caller will add them to the bus

It feels like either this function should handle all the necessary steps to add stuff to the bus, both errors and skips, or the caller should

src/internal/m365/collection/drive/collections.go
```diff
@@ -688,232 +670,297 @@ func (c *Collections) getCollectionPath(
...
+		el               = errs.Local()
+		newPrevPaths     = map[string]string{}
+		invalidPrevDelta = len(prevDeltaLink) == 0
+
+		// currPrevPaths is used to identify which collection a
+		// file belongs to. This is useful to delete a file from the
+		// collection it was previously in, in case it was moved to a
+		// different collection within the same delta query
+		// item ID -> item ID
+		currPrevPaths = map[string]string{}
```
ashmrtn, line 698: seeing as this is per delta token it looks like we're assuming we only see each drive only once. We should probably patch the code that goes through drives to actually ensure that at some point

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -1446,25 +1469,146 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
...
+
+			enumerator: mock.EnumerateItemsDeltaByDrive{
+				DrivePagers: map[string]*mock.DriveItemsDeltaPager{
+					driveID1: {
+						Pages: []mock.NextPage{
+							{
+								Items: []models.DriveItemable{
+									driveRootItem("root"),
+									driveItem("folder", "folder", driveBasePath1, "root", false, true, false),
+									driveItem("file", "file", driveBasePath1+"/folder", "folder", true, false, false),
```
ashmrtn, line 1531: similar to the previous PR, could you please add another item in this result set that isn't in the result set after the reset. Just for context (if other folks see this without seeing the previous PR), this helps ensure resets are handled like we think they should be while removing the possibility that map key collisions mask improper handling

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -1735,16 +1885,34 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
...
+					driveID1: {
+						Pages: []mock.NextPage{
+							{
+								// on the first page, if this is the total data, we'd expect both folder and folder2
+								// since new previousPaths merge with the old previousPaths.
+								Items: []models.DriveItemable{
+									driveRootItem("root"),
+									driveItem("folder2", "folder", driveBasePath1, "root", false, true, false),
+									driveItem("file", "file", driveBasePath1+"/folder", "folder2", true, false, false),
+								},
```
ashmrtn, line 1899: this test should work even if this first page of results isn't sent. The important part is that we reset the delta token and return a page of results with a folder with the same name but a different ID
```suggestion
```

src/internal/m365/service/onedrive/mock/handlers.go
ryanfkeepers: That is a pointer.  See line 301: `DrivePagers map[string]*DriveItemsDeltaPager`

src/pkg/services/m365/api/drive_pager_test.go
ryanfkeepers: It's equivalent to:
```go
	for page, reset, done := pager.NextPage();
	!done;
	page, reset, done = pager.NextPage() {
```
or 
```go
page, reset, done := pager.NextPage()
for !done {
	//...
	page, reset, done = pager.NextPage()
}
```

The format I originally used (`for a, b, c := foo(); !c; {`) was a bug; it instantiates the values, but doesn't update them.  This just felt like the quickest fix.  chonky, but workable.

src/internal/m365/collection/drive/collections.go
ryanfkeepers: comments, my nemesis!

src/pkg/services/m365/api/drive_pager_test.go
ashmrtn: ah, I see. You can do the update in the loop as well. It just turns into
```go
for page, reset, done := pager.NextPage(); !done; page, reset, done = pager.NextPage() {
```

src/internal/m365/collection/drive/collections.go
ryanfkeepers: Grokked and aggreed.  I think I'm going to hold off on changes further here within this PR, since I'd like to revisit this code with some additional refactoring and this PR is already doing enough.  It'll be easier for me to work on design improvements there.  But I will get to it.

src/internal/m365/collection/drive/collections.go
ryanfkeepers: sounds reasonable

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Oh yeah!  Got lost in the refactor.

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Yes, it should.  However, this caught a bug that I'd missed in development.  For my own sanity, I'd like to keep the first part in place.

src/pkg/services/m365/api/drive_pager_test.go
ryanfkeepers: Dear santa, all I want for christmas is `for foo, bar, keepRanging := range baz() {`.

ryanfkeepers: also, wow, the linter does _not_ like the multiline style in my first example.

src/internal/m365/collection/drive/collections.go
ashmrtn: mostly concerned that we don't guarantee otherwise, especially given all the oddities we've seen with graph API over time

ashmrtn (approved):

src/internal/m365/collection/drive/collections_test.go
ashmrtn: could you make another test that doesn't have the additional page in it then? It'd be good to make sure both versions work

ryanfkeepers: Happily!

PR #4419, opened 2023-10-02
Skip graph call if the download url has expired

pandeyabs:
Builds on top of earlier PR #4417 to skip graph API call if the token has already expired. This is a performance optimization.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [x] Performance Opt

#### Issue(s)

* internal

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/item.go
```diff
@@ -121,6 +125,19 @@ func downloadFile(
 		return nil, clues.New("empty file url").WithClues(ctx)
 	}
 
+	// Precheck for url expiry before we make a call to graph to download the
+	// file. If the url is expired, we can return early and save a call to graph.
+	//
+	// Ignore all errors encountered during the check. We can rely on graph to
+	// return errors on malformed urls. Ignoring errors also future proofs against
+	// any sudden graph changes, for e.g. if graph decides to emb the token under a
```
ashmrtn, line 133: nit: not sure if `emb` is a typo or a term I'm unfamiliar with. Could you clarify it in the comment please?

src/internal/m365/collection/drive/item.go
```diff
@@ -193,3 +210,27 @@ func setName(orig models.ItemReferenceable, driveName string) models.ItemReferen
...
+func isURLExpired(
+	ctx context.Context,
+	url string,
+) (bool, error) {
+	// Extract the raw JWT string from the download url.
+	rawJWT, err := common.GetQueryParamFromURL(url, onedrive.JWTQueryParam)
+	if err != nil {
+		logger.CtxErr(ctx, err).Info("query param not found")
+
+		return false, clues.Stack(err)
```
ashmrtn, line 225: nit: could you please add the context's clues to the error in this return and the one below?
```suggestion
		return false, clues.Stack(err).WithClues(ctx)
```

src/internal/m365/service/onedrive/consts/consts.go
```diff
@@ -12,6 +12,10 @@ const (
 	RootPathDir = "root:"
 	// root id for drive items
 	RootID = "root"
+
+	// JWTQueryParam represents the query param embed in graph download URLs.
+	// It holds the JWT token.
+	JWTQueryParam = "tempauth"
```
ashmrtn, line 18: since this is used for all services that are backed by sharepoint/onedrive it feels like it would fit a bit better in the file it's used in (collection/drive/item.go)

ashmrtn (approved):

src/internal/m365/collection/drive/item.go
pandeyabs: thanks! it's a typo

pandeyabs:
Currently running scale test with this PR chain. Will merge once the test has concluded.

PR #4420, opened 2023-10-02
provide expected repo cfg in e2e tests

ryanfkeepers:
nightly tests were missing expected repo config due to, breaking api expectations, causing unexpected failures due to test env setup.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/helpers_test.go
+++ b/src/cli/backup/helpers_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/alcionai/corso/src/internal/tester/tconfig"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/repository"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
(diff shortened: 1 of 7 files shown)


src/cli/backup/helpers_test.go
```diff
@@ -132,6 +133,7 @@ type dependencies struct {
 func prepM365Test(
 	t *testing.T,
 	ctx context.Context, //revive:disable-line:context-as-argument
+	pst path.ServiceType,
```
ashmrtn, line 136: nit: can we name this `service` please? The current variable name is still referring to the type, not the actual usage of the thing

ashmrtn (approved):

src/cli/backup/helpers_test.go
ryanfkeepers: "service" is also a reference to the type and not its usage, is it not?  Just a more human way of writing it.

ashmrtn: service also describes what it is. `pst` alone gives no context as to what it's referring to. For all I know it's talking about files that you hook up to Exchange to read emails, contacts, and calendar events 😛

PR #4435, opened 2023-10-03
remove local bus struct

ryanfkeepers:
removes the local bus from fault in favor of a single Bus that operates both within a local and global instance, and can be passed downstream independent of context.  Also includes some code separation in the fault package for readability.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/fault/alert.go
+++ b/src/pkg/fault/alert.go
@@ -1,55 +1,8 @@
 package fault
 
 import (
-	"context"
-
-	"golang.org/x/exp/slices"
-
 	"github.com/alcionai/corso/src/cli/print"
-	"github.com/alcionai/corso/src/pkg/logger"
 )
-
-// Alerts returns the slice of alerts generated during runtime.
-// If the bus is a local alerts, this only returns the
-// local failure, and will not return parent data.
-func (e *Bus) Alerts() []Alert {
-	return slices.Clone(e.alerts)
-}
-
-// AddAlert appends a record of an Alert message to the fault bus.
-// Importantly, alerts are not errors, exceptions, or skipped items.
-// An alert should only be generated if no other fault functionality
-// is in use, but that we still want the end user to clearly and
-// plainly receive a notification about a runtime event.
-func (e *Bus) AddAlert(ctx context.Context, a *Alert) {
-	if a == nil {
-		return
-	}
-
-	e.mu.Lock()
-	defer e.mu.Unlock()
-
-	e.logAndAddAlert(ctx, a, 1)
-}
-
-// logs the error and adds an alert.
-func (e *Bus) logAndAddAlert(ctx context.Context, a *Alert, trace int) {
-	logger.CtxStack(ctx, trace+1).
-		With("skipped", a).
-		Info("recoverable error")
-	e.addAlert(a)
-}
-
-func (e *Bus) addAlert(a *Alert) *Bus {
-	e.alerts = append(e.alerts, *a)
-
-	// local bus instances must promote alerts to the root bus.
-	if e.parent != nil {
-		e.parent.addAlert(a)
-	}
-
-	return e
-}
 
 var _ print.Printable = &Alert{}
 
--- a/src/pkg/fault/alert.go
+++ b/src/pkg/fault/alert.go
@@ -1,0 +1,117 @@
+package fault
+
+import (
+	"context"
+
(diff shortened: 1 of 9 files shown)


src/pkg/fault/fault.go
```diff
@@ -435,68 +379,21 @@ func (pec printableErrCore) Values() []string {
...
+// in a constrained scope.  This allows the caller to review recoverable errors and
+// failures within only the current codespace, as opposed to the global set of errors.
+// The function that spawned the local bus should always return `bus.Failure()` to
+// ensure that hard failures are propagated back upstream.
+func (e *Bus) Local() *Bus {
+	parent := e
+
+	// always use the root bus reference, if e is not the root.
+	if e.parent != nil {
+		e = e.parent
```
meain, line 392: I guess both are fine, but I feel like `parent = e.parent` conveys the information better.

src/pkg/fault/fault.go
```diff
@@ -168,70 +170,12 @@ func (e *Bus) addRecoverableErr(err error) bool {
...
-// both the conditions and identifications of skips.
-// 2. Skipping avoids a permanent and consistent failure.  If
-// the underlying reason is transient or otherwise recoverable,
-// the item should not be skipped.
-func (e *Bus) AddSkip(ctx context.Context, s *Skipped) {
-	if s == nil {
-		return
+	// local bus instances must promote errors to the root bus.
+	if e.parent != nil {
+		e.parent.addRecoverableErr(err)
```
meain, line 175: We are not acquiring the lock on the parent for this change as that step is done in `AddRecoverableErr`(the public function). I think we should just move the lock acquiring part to the private function as the other thing is printing a log message which does not require a lock anyways.

meain (approved):
Few small things, but approving to unblock.

src/pkg/fault/alert.go
```diff
@@ -0,0 +1,117 @@
+package fault
```
pandeyabs, line 1: Is this purely a code movement from earlier PR in the chain? Could this change be done in the earlier PR itself?

src/pkg/fault/fault.go
```diff
@@ -379,68 +379,21 @@ func (pec printableErrCore) Values() []string {
...
-// both the conditions and identifications of skips.
-// 2. Skipping avoids a permanent and consistent failure.  If
-// the underlying reason is transient or otherwise recoverable,
-// the item should not be skipped.
-func (e *localBus) AddSkip(ctx context.Context, s *Skipped) {
-	if s == nil {
-		return
+	return &Bus{
+		mu:     &sync.Mutex{},
+		parent: parent,
```
ashmrtn, line 397: this is still assigning the original value of `e` as the parent, not the updated one. It would be better to just have `parent: e.parent` here I think

src/pkg/fault/fault.go
```diff
@@ -67,24 +88,16 @@ func (e *Bus) Failure() error {
...
 // doesn't require the entire process to end.
+// If the bus is a local instance, this only returns the
+// local recovered errors, and will not return parent data.
 func (e *Bus) Recovered() []error {
 	return slices.Clone(e.recoverable)
 }
 
-// Skipped returns the slice of items that were permanently
-// skipped during processing.
-func (e *Bus) Skipped() []Skipped {
```
ashmrtn, line 76: Can we please keep all functions for a struct in the same file? It makes it easer to determine what the full API of the struct is. The `Skipped()` function is still related to `Bus`, it just deals with a different aspect of it. There's already quite a few files in the codebase and splitting everything into the smallest possible unit per file just adds overhead

ryanfkeepers: Worth a design discussion sometime.  I, personally, find that splitting packages into small files representing semantic chunks is smaller cognitive load (I assume that's what you mean by 'overhead') than maintaining fewer, larger files.

For now, I'm happy to move these methods back.

src/pkg/fault/fault.go
ryanfkeepers: Yup, good call.

src/pkg/fault/alert.go
ryanfkeepers: It is.  Aplogies, I'll try to separate the two in future PRs.

PR #4437, opened 2023-10-04
Temporary removal of OneNote file size check

vkamra:
Temporarily remove the large file size check that controls skipping a OneNote file download that failed with a
503 error.

We've observed this error condition to be persistent even with smaller OneNote files.

This allows us to unblock backup for these resources while we investigate the issue.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collection.go
+++ b/src/internal/m365/collection/drive/collection.go
@@ -265,7 +265,10 @@ func (oc *Collection) getDriveItemContent(
 
 		// Skip big OneNote files as they can't be downloaded
 		if clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) &&
-			oc.scope == CollectionScopePackage && *item.GetSize() >= MaxOneNoteFileSize {
+			// TODO: We've removed the file size check because it looks like we've seen persistent
+			// 503's with smaller OneNote files also.
+			// oc.scope == CollectionScopePackage && *item.GetSize() >= MaxOneNoteFileSize {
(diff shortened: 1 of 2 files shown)


src/internal/m365/collection/drive/collection.go
```diff
@@ -265,7 +265,10 @@ func (oc *Collection) getDriveItemContent(
 
 		// Skip big OneNote files as they can't be downloaded
 		if clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) &&
-			oc.scope == CollectionScopePackage && *item.GetSize() >= MaxOneNoteFileSize {
+			// TODO: We've removed the file size check because it looks like we've seen persistent
+			// 503's with smaller OneNote files also.
+			// oc.scope == CollectionScopePackage && *item.GetSize() >= MaxOneNoteFileSize {
+			oc.scope == CollectionScopePackage {
```
meain, line 271: The fault name still in use is `SkipBigOneNote`. Maybe we can change to reflect this change.

meain (approved):
LGTM

src/internal/m365/collection/drive/collection.go
vkamra: I thought about that - will address when we figure out the final change here.

Issue #4441, opened 2023-10-04
Expose Corso error event


neha-Gupta1:
Send events to rudderstack in case or any failure/error or abrupt end of command. The events will include the reason of failure

PR #4446, opened 2023-10-05
add corso error event

neha-Gupta1:
add Corso Error event if a CLI command fails

- [ ] :no_entry: No

#### Type of change

#### Issue(s)

* https://github.com/alcionai/corso/issues/4441

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test

### DIFF
--- a/src/internal/events/events.go
+++ b/src/internal/events/events.go
@@ -32,7 +32,6 @@ const (
 	RepoConnect    = "Repo Connect"
 	BackupEnd      = "Backup End"
 	RestoreEnd     = "Restore End"
-	CorsoError     = "Corso Error"
 	ExportEnd      = "Export End"
 	MaintenanceEnd = "Maintenance End"
 
@@ -53,6 +52,7 @@ const (
 	StartTime        = "start_time"
 	Status           = "status"
 	Command          = "command"
+	ErrorMessage     = "error"
 
 	// default values for keys
 	RepoIDNotFound = "not_found"
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -248,34 +248,25 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		"disable_assist_backup", op.disableAssistBackup)
 
 	defer func() {
+		data := map[string]any{
+			events.BackupID:   op.Results.BackupID,
+			events.DataStored: op.Results.BytesUploaded,
+			events.Duration:   op.Results.CompletedAt.Sub(op.Results.StartedAt),
+			events.EndTime:    dttm.Format(op.Results.CompletedAt),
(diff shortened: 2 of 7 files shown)


neha-Gupta1:
For handling the `control+c` or similar signal we need to capture the signals using channel and then handling it forward. Something like- 
```
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	go func() {
		for sig := range c {
			op.Errors.Fail(clues.Wrap(err, "running backup"))
		}
	}()

```
@vkamra do we want to capture this signal?

neha-Gupta1:
For handling the `control+c` or similar signal we need to capture the signals using channel and then handling it forward. Something like- 
```
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	go func() {
		for sig := range c {
			op.Errors.Fail(clues.Wrap(err, "running backup"))
		}
	}()

```
@vkamra do we want to capture this signal?

ryanfkeepers:
@gmatev Is this the behavior you were hoping for?  Or did you want the error details attached to the `<Operation> End` event as metadata, rather than a separate event?

ryanfkeepers:
> do we want to capture this signal?

@neha-Gupta1 I don't think we want treat this signal as an error case.  In fact, I'm dubious that we need to handle it for eventing at all (can't remember if it was asked for).  Maybe something to skip for now?

gmatev:
> @gmatev Is this the behavior you were hoping for? Or did you want the error details attached to the `<Operation> End` event as metadata, rather than a separate event?

I'm not sure I quite follow what is proposed here. 

The desired behavior is roughly:
* For errors in otherwise tracked operations (e.g. Backup End) the error can be tracked as part of these events (I believe this is already the case). 
* For such operations, doing ctrl+c or equivalent should be captures as some type of error. for example as `status: cancelled` or `status: interrupted`
* `Corso Errored` should ideally be reserved for operations for commands that otherwise fail but would not be otherwise captured. For example, incorrect combination of commands, etc. For such we want to capture the command (without any sensitive information) and the error context. The idea would be to use such events to identify common points of confusion or friction.

neha-Gupta1:
Whats currently is handled here-
- send information about failure in the event if the action fails.

What's not handled in the PR-
-  Doing ctrl+c or equivalent should be captures as some type of error. for example as status: cancelled or status: interrupted
- Corso Errored should ideally be reserved for operations for commands that otherwise fail but would not be otherwise captured. For example, incorrect combination of commands, etc. For such we want to capture the command (without any sensitive information) and the error context. The idea would be to use such events to identify common points of confusion or friction.
Issue for the same - https://github.com/alcionai/corso/issues/4529

Issue #4311, opened 2023-09-20
Expose stats for exports


meain:
Expose some basic stats (eg: number of files used to produce the export) to the user. Users should be able to access a `Stats` object on `ExportOperation` which would have these stats populated after all the files in the export has been read.

This could enhance the usefulness of using it as "dry-run" restores to test the validity of backups.

PR #4461, opened 2023-10-10
Add stats to export operations

meain:
Have a way to gather stats about the exported data.

Users can now call `ExportOperation.GetStats()` at the end of the run to get the stats for the operations. The data will be in the format `map[path.CategoryType]data.KindStats` whre `KindStats` is:

```go
type KindStats struct {
	BytesRead     int64
	ResourceCount int64
}
```
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/4311

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/data/metrics.go
+++ b/src/internal/data/metrics.go
@@ -82,6 +82,10 @@ func ReaderWithStats(
 	kind details.ItemType,
 	stats *ExportStats,
 ) io.ReadCloser {
+	if reader == nil {
+		return nil
+	}
+
 	return &statsReader{
 		reader: reader,
 		kind:   kind,
--- a/src/internal/data/metrics.go
+++ b/src/internal/data/metrics.go
@@ -4,7 +4,7 @@ import (
 	"io"
 	"sync/atomic"
 
-	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/path"
 )
 
 type CollectionStats struct {
@@ -30,13 +30,12 @@ type KindStats struct {
 
 type ExportStats struct {
 	// data is kept private so that we can enforce atomic int updates
-	// TODO(meain): Should we be exposing details.ItemType?
-	data map[details.ItemType]KindStats
+	data map[path.CategoryType]KindStats
 }
 
-func (es *ExportStats) UpdateBytes(kind details.ItemType, bytesRead int64) {
+func (es *ExportStats) UpdateBytes(kind path.CategoryType, bytesRead int64) {
 	if es.data == nil {
-		es.data = map[details.ItemType]KindStats{}
+		es.data = map[path.CategoryType]KindStats{}
 	}
 
 	ks := es.data[kind]
@@ -44,9 +43,9 @@ func (es *ExportStats) UpdateBytes(kind details.ItemType, bytesRead int64) {
 	es.data[kind] = ks
 }
 
-func (es *ExportStats) UpdateResourceCount(kind details.ItemType) {
+func (es *ExportStats) UpdateResourceCount(kind path.CategoryType) {
 	if es.data == nil {
-		es.data = map[details.ItemType]KindStats{}
+		es.data = map[path.CategoryType]KindStats{}
 	}
 
 	ks := es.data[kind]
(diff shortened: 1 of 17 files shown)


src/internal/data/metrics.go
```diff
@@ -15,3 +22,73 @@ func (cs CollectionStats) IsZero() bool {
...
+type KindStats struct {
+	BytesRead     int64
+	ResourceCount int64
+}
+
+type ExportStats struct {
+	// data is kept private so that we can enforce atomic int updates
+	// TODO(meain): Should we be exposing details.ItemType?
+	data map[details.ItemType]KindStats
+}
```
ryanfkeepers, line 35: This all feels like a recreation of work that's already handled in https://github.com/alcionai/corso/blob/main/src/internal/stats/stats.go.  I've been meaning to revisit our stats gathering and replace them all with this.  Could you plug it in here instead of the bespoke solution?

I'll approve what's here if you want to stick with it, but that does mean I'll unwind it when I get around to the tech debt.

ryanfkeepers (approved):
Approved with a request.  See comment.

src/internal/data/metrics.go
```diff
@@ -15,3 +22,73 @@ func (cs CollectionStats) IsZero() bool {
...
+	return es.data
+}
+
+type statsReader struct {
+	reader io.ReadCloser
+	kind   details.ItemType
+	stats  *ExportStats
+}
+
+func (sr *statsReader) Read(p []byte) (n int, err error) {
```
ashmrtn, line 67: nit: this one is purely preference on my part, but can we keep explicit returns of data instead of named return variables. It takes my mind a few extra seconds to go through the logic of empty return statement -> something that still returns values -> using named return values 😅

src/internal/data/metrics.go
```diff
@@ -15,3 +22,73 @@ func (cs CollectionStats) IsZero() bool {
...
+	atomic.AddInt64(&ks.ResourceCount, 1)
+	es.data[kind] = ks
+}
+
+func (es *ExportStats) GetStats() map[details.ItemType]KindStats {
+	return es.data
+}
+
+type statsReader struct {
+	reader io.ReadCloser
```
ashmrtn, line 62: nit: this is stylistic again and not required, but you could just embed the ReadCloser. If you do that you can remove the implementation for `Close()` unless you feel the need for error wrapping there

```suggestion
	io.ReadCloser
```

src/internal/data/metrics.go
```diff
@@ -15,3 +22,73 @@ func (cs CollectionStats) IsZero() bool {
...
+	kind   details.ItemType
```
ashmrtn, line 63: can we actually switch this to use path.CategoryType and/or path.ServiceType? That'll get us the benefit of being able to use the newly added `HumanString()` (at least for category, may need to add something similar for service) which helps keep output across various parts of the program consistent. We've also already got quite a bit of infrastructure that ensures those values can be compared/switched on as needed

If you need to make them into a map key (i.e. want both service and category) I suggest making a function to take a service/category pair and output a string or something. These are all in-memory values, so as long as the output is consistent for a given run of corso we can update the generated combined value as needed

src/internal/data/metrics.go
meain: I thought of reusing that, but the data that we need is a bit different here. The main difference being that we need separate stats for individual `ItemType`. Also, I am kinda on the fence about exposing exposing `details.ItemType`. Any thoughts on that?

src/internal/data/metrics.go
meain: Ahh, forgot about `path.CategoryType`. Let me switch to that.

src/internal/data/metrics.go
ryanfkeepers: I'm a doof 🤦 .  I meant to link this library: https://github.com/alcionai/corso/blob/main/src/pkg/count/count.go

You can get per-type stats by creating local instances for each one.  A nice byproduct is that the stats will naturally centralize into a total at the same time.

Again, I'm happy to handle this later.

meain: I'm gonna defer it to a later PR as of now. If you don't get it before me, I'll create the PR to do the switch.

Issue #4457, opened 2023-10-09
Update to directory iterator based version of kopia


ashmrtn:
Kopia recently merged a change that changing the directory API to use [iterators instead of callbacks](https://github.com/kopia/kopia/pull/3365). This is a breaking change that will require changes to how Corso passes the virtualfs tree to kopia

PR #4466, opened 2023-10-10
Rewrite Corso's kopia snapshot tie-ins to use directory iterators

ashmrtn:
kopia/kopia#3365 switched kopia to use iterators
for traversing directory entries instead of
callbacks

This PR adds a new iterator struct that uses a
state machine to traverse the different data
sources we use to back a single directory passed
to kopia. While the state machine execution
logic is new, much of the logic for returning
individual entries is adapted from existing code

Recommend viewing by commit and possibly
ignoring whitespace changes

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4457

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
 
-func collectionEntries(
-	ctx context.Context,
-	ctr func(context.Context, fs.Entry) error,
-	streamedEnts data.BackupCollection,
-	progress *corsoProgress,
-) (map[string]struct{}, error) {
-	if streamedEnts == nil {
-		return nil, nil
+// These define a small state machine as to which source to return an entry from
```
ryanfkeepers, line 238:
```suggestion
// These define a small state machine describing which source to return an entry from
```

src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
+// next. Since these are in-memory only values we can use iota. Phases are
```
ryanfkeepers, line 239: Understanding that we _can_ use iota here with relative safety, I still think it's better to avoid it by standard (ie: add a linter that says no iotas).  We've hit bugs too many times for it to feel comfortable.

src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
+// traversed in the order defined unless the underlying data source isn't
+// present. If an underlying data source is missing, the non-pre/post phase
+// associated with that data source is skipped.
+//
+// Since some phases require initialization of the underlying data source we
+// insert additional phases to allow that. Once initialization is completed the
+// phase should be updated to the next phase.
+//
+// A similar tactic can be used to handle tearing down resources for underlying
+// data sources if needed.
```
ryanfkeepers, line 249: The explanation is much appreciated.  Thanks for the comments!

src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
+	// merging).
+	staticEnts       []fs.Entry
+	globalExcludeSet prefixmatcher.StringSetReader
+	progress         *corsoProgress
+
+	// endSpan is the callback to stop diagnostic span collection for iteration.
+	endSpan func()
+
+	// seenEnts contains the encoded names of entries that we've already streamed
+	// so we can skip returning them again when looking at base entries.
```
ryanfkeepers, line 278: what's the format of the key here?  Just the immediate name?  Not a path?

src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
+	seenEnts map[string]struct{}
+	// locationPath contains the human-readable location of the underlying
+	// collection.
+	locationPath *path.Builder
+
+	// excludeSet is the individual exclude set to use for the longest prefix for
+	// this iterator.
```
ryanfkeepers, line 285: This comment doesn't entirely make sense.  It makes it sound like we use it to exclude the longest prefix.  But that doesn't seem right.

src/internal/kopia/upload.go
```diff
@@ -235,38 +235,229 @@ func (cp *corsoProgress) get(k string) *itemDetails {
...
+	excludeSet map[string]struct{}
+
+	// traversalPhase is the current state in the state machine.
+	traversalPhase int
```
ryanfkeepers, line 289:
```suggestion
	currentPhase int
```

ryanfkeepers (approved):

PR #4470, opened 2023-10-10
Builder pattern for BackupBase struct

ashmrtn:
Will be useful for later testing that needs BackupBase

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #3943

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/mock_backup_base.go
+++ b/src/internal/kopia/mock_backup_base.go
@@ -197,3 +197,63 @@ func (bb *MockBackupBases) MockDisableMergeBases() *MockBackupBases {
 	bb.DisableMergeBases()
 	return bb
 }
+
+// -----------------------------------------------------------------------------
+// Functions for BackupBase creation
+// -----------------------------------------------------------------------------
+
+func NewBackupBaseBuilder(idPrefix string, id int) *BackupBaseBuilder {
+	bIDKey, _ := makeTagKV(TagBackupID)
+	baseID := fmt.Sprintf("%sID%d", idPrefix, id)
+
+	return &BackupBaseBuilder{
+		b: &BackupBase{
+			Backup: &backup.Backup{
+				BaseModel: model.BaseModel{
+					ID: model.StableID(baseID + "-backup"),
+				},
+				SnapshotID:    baseID + "-item-data",
+				StreamStoreID: baseID + "-stream-store",
+			},
+			ItemDataSnapshot: &snapshot.Manifest{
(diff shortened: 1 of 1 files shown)


src/internal/kopia/mock_backup_base.go
```diff
@@ -118,3 +126,63 @@ func (bb *MockBackupBases) MockDisableMergeBases() *MockBackupBases {
...
+	if bbb.b.Backup.Tags == nil {
+		bbb.b.Backup.Tags = map[string]string{}
+	}
+
+	bbb.b.Backup.Tags[model.BackupTypeTag] = model.AssistBackup
+
+	return bbb
+}
+
+func (bbb *BackupBaseBuilder) WithReasons(reasons ...identity.Reasoner) *BackupBaseBuilder {
```
pandeyabs, line 180: nit: Do we need `WithReasons`? Wondering if callers can just use `AppendReasons`.

src/internal/kopia/mock_backup_base.go
```diff
@@ -118,3 +126,63 @@ func (bb *MockBackupBases) MockDisableMergeBases() *MockBackupBases {
...
+func (bbb *BackupBaseBuilder) Build() BackupBase {
+	return *bbb.b
+}
+
+func (bbb *BackupBaseBuilder) MarkAssistBase() *BackupBaseBuilder {
+	if bbb.b.Backup.Tags == nil {
+		bbb.b.Backup.Tags = map[string]string{}
+	}
+
+	bbb.b.Backup.Tags[model.BackupTypeTag] = model.AssistBackup
```
pandeyabs, line 175: nit: `bbb.b` is a little hard to read :) Could we rename `bbb` to `builder` or similar?

pandeyabs (approved):

src/internal/kopia/mock_backup_base.go
ryanfkeepers: I'd rather we use the terminology `SetFoo` here.  `With` does sound like "in addition to"- ie: append.  And is also how WithFoo is used in clues, so there could be some confusion around that.  `SetReasons` makes for a clearer indicator that the provided reasons clobber any existing ones.

ryanfkeepers (approved):

PR #4480, opened 2023-10-11
refer to env instead of secrets in ci

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

### DIFF
--- a/.github/workflows/longevity_test.yml
+++ b/.github/workflows/longevity_test.yml
@@ -36,7 +36,7 @@ jobs:
       CORSO_LOG_DIR: ${{ github.workspace }}/src/testlog
       CORSO_LOG_FILE: ${{ github.workspace }}/src/testlog/run-longevity.log
       RESTORE_DEST_PFX: Corso_Test_Longevity_
-      TEST_USER: ${{ github.event.inputs.user != '' && github.event.inputs.user || secrets.CORSO_M365_TEST_USER_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/longevity_test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       CORSO_LOG_DIR: ${{ github.workspace }}/src/testlog
       CORSO_LOG_FILE: ${{ github.workspace }}/src/testlog/run-longevity.log
       RESTORE_DEST_PFX: Corso_Test_Longevity_
-      TEST_USER: ${{ github.event.inputs.user != '' && github.event.inputs.user || secrets.CORSO_M365_TEST_USER_ID }}
+      TEST_USER: ${{ github.event.inputs.user != '' && github.event.inputs.user || env.CORSO_M365_TEST_USER_ID }}
```
ashmrtn, line 39: I think this should actually be `vars` not `env` (I've had similar issues remembering the different contexts). This [section](https://docs.github.com/en/actions/learn-github-actions/variables#using-the-vars-context-to-access-configuration-variable-values) of the page has how to access vars while this [section](https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows) talks about setting vars

ashmrtn (approved):
approved to unblock, but pretty sure it needs fixups

PR #4482, opened 2023-10-12
use count bus in kopia backups

ryanfkeepers:
uses the count bus in the kopia backup package.
This currently duplicates counts that we're getting
from the kopia stats.  A later pr will remove the old
stats entirely in favor of the counter.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/operations/backup.go
```diff
@@ -202,6 +204,7 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 	ctx, flushMetrics := events.NewMetrics(ctx, logger.Writer{Ctx: ctx})
 	defer flushMetrics()
 
+	// for cases where we can't pass the counter down as part of a func call.
 	ctx = count.Embed(ctx, op.Counter)
```
ashmrtn, line 208: I'm not comfortable with this pattern. This is starting to promote context as the sort of "catch all" for data passing that we actually need. It makes it difficult to determine what the actual dependencies of a function are because it's not obvious what's required and where the required things come from

ryanfkeepers: Agreed.  It's not my preference either.  The bus _should_ be getting passed through as part of func params or struct properties.  I can make the comment sound more forceful if wanted.  As for the context embedding that's already in place here, fixing that is a lower priority than getting the bus passed around, but is something worth coming back to after the rest of the implementation is in place.

ashmrtn (approved):

PR #4485, opened 2023-10-12
Fail backups if we see an item has a zero mod time

ashmrtn:
This is a big hammer solution to keep bad state from getting into the repo. It will keep us getting into a situation in which we've made a backup that was marked as a successful backup (and thus a merge base) but has items which we won't be able to do details merging on if we do use it as a merge base.

Long term, we should probably tag these backups with some sort of special tag that says they shouldn't be used as bases at all since we'll fail to merge details. We'll fail details merging even if the base is used as an assist base.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/upload.go
+++ b/src/internal/kopia/upload.go
@@ -148,6 +148,10 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 			Label(fault.LabelForceNoBackupCreation))
 
 		return
+	} else if info.Modified().IsZero() {
+		cp.errs.AddRecoverable(ctx, clues.New("mod time of zero").
+			WithClues(ctx).
+			Label(fault.LabelForceNoBackupCreation))
 	}
 
 	err = cp.deets.Add(d.repoPath, d.locationPath, info)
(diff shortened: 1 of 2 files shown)


src/internal/kopia/upload.go
```diff
@@ -148,6 +148,10 @@ func (cp *corsoProgress) FinishedFile(relativePath string, err error) {
 			Label(fault.LabelForceNoBackupCreation))
 
 		return
+	} else if info.Modified().IsZero() {
+		cp.errs.AddRecoverable(ctx, clues.New("mod time of zero").
```
ryanfkeepers, line 152: Nit: "of zero" is kinda confusing.
```suggestion
		cp.errs.AddRecoverable(ctx, clues.New("zero-valued mod time").
```

ryanfkeepers (approved):

PR #4481, opened 2023-10-11
filter out system messages

ryanfkeepers:
adds a filter that removes system messages from
channel message backups.

Also extends the preview size from 16 to 128 characters.

---

#### Does this PR need a docs update or release note?

- [x] :no_entry: No

#### Type of change

- [x] :broom: Tech Debt/Cleanup

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

PR #4486, opened 2023-10-12
ensure api layer provides non-empty modtime

ryanfkeepers:
enforce the following:
* no mod times returned from the api layer are zero-valued
* dttm.OrNow() provides a zero value time check
* ptr.OrNow() calls dttmOrNow on non-nil time values.
* itemPagers for getAddedAndRemoved call ptr.OrNow()
* callers to getAddedAndRemoved provide a transformer func which produces the mod time from the paged item, instead of assuming all items use the same interface for mod time.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Catch and report cases where a protected resource is locked out of access.  SDK consumers have a new errs sentinel that allows them to check for this case.
+- Fix a case where missing item LastModifiedTimes could cause incremental backups to fail.
 
 ## [v0.14.0] (beta) - 2023-10-09
 
--- a/src/internal/common/dttm/dttm.go
+++ b/src/internal/common/dttm/dttm.go
@@ -98,10 +98,18 @@ func Now() string {
 	return FormatNow(Standard)
 }
 
+func OrNow(t time.Time) time.Time {
+	if t.IsZero() {
+		return time.Now().UTC()
+	}
+
+	return t
+}
+
 // FormatNow produces the current time in UTC using the provided
 // time format.
 func FormatNow(fmt TimeFormat) string {
-	return FormatTo(time.Now(), fmt)
+	return FormatTo(time.Now().UTC(), fmt)
 }
 
 // FormatTo produces the a datetime with the given format.
--- a/src/internal/common/dttm/dttm_test.go
+++ b/src/internal/common/dttm/dttm_test.go
@@ -58,6 +58,47 @@ func (suite *DTTMUnitSuite) TestParseTime() {
 	require.Error(t, err, clues.ToCore(err))
 }
 
+func (suite *DTTMUnitSuite) TestOrNow() {
+	var (
+		oneMinuteAgo = time.Now().Add(-1 * time.Minute)
+		inOneHour    = time.Now().Add(1 * time.Hour)
+		empty        time.Time
+	)
+
+	table := []struct {
+		name   string
+		input  time.Time
+		expect func(t *testing.T, result time.Time)
+	}{
+		{
+			name:  "a minute ago",
+			input: oneMinuteAgo,
+			expect: func(t *testing.T, result time.Time) {
+				assert.True(t, oneMinuteAgo.Equal(result), "equal")
+			},
+		},
+		{
+			name:  "in an hour",
+			input: inOneHour,
+			expect: func(t *testing.T, result time.Time) {
+				assert.True(t, inOneHour.Equal(result), "equal")
+			},
+		},
+		{
+			name:  "empty",
+			input: empty,
+			expect: func(t *testing.T, result time.Time) {
+				assert.WithinDuration(t, time.Now(), result, time.Minute)
+			},
+		},
+	}
+	for _, test := range table {
+		suite.Run(test.name, func() {
+			test.expect(suite.T(), dttm.OrNow(test.input))
+		})
+	}
+}
+
 func (suite *DTTMUnitSuite) TestExtractTime() {
 	comparable := func(t *testing.T, tt time.Time, shortFormat dttm.TimeFormat) time.Time {
 		ts := dttm.FormatToLegacy(tt.UTC())
--- a/src/internal/common/ptr/pointer.go
+++ b/src/internal/common/ptr/pointer.go
@@ -1,6 +1,10 @@
 package ptr
 
-import "time"
+import (
+	"time"
+
+	"github.com/alcionai/corso/src/internal/common/dttm"
+)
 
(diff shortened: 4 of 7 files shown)


src/internal/common/dttm/dttm_test.go
```diff
@@ -58,6 +58,49 @@ func (suite *DTTMUnitSuite) TestParseTime() {
 	require.Error(t, err, clues.ToCore(err))
 }
 
+func (suite *DTTMUnitSuite) TestOrNow() {
+	var (
+		oneMinuteAgo = time.Now().Add(-1 * time.Minute)
+		inOneMinute  = time.Now().Add(1 * time.Minute)
+		inOneHour    = time.Now().Add(1 * time.Hour)
```
ashmrtn, line 65: nit: there's no constant multiple required if you're just adding 1 of each

src/internal/common/dttm/dttm_test.go
```diff
@@ -58,6 +58,49 @@ func (suite *DTTMUnitSuite) TestParseTime() {
...
+		empty        = time.Time{}
```
ashmrtn, line 66: nit: we've generally not been explicitly setting to the zero value but just giving the type

src/internal/common/dttm/dttm_test.go
```diff
@@ -58,6 +58,49 @@ func (suite *DTTMUnitSuite) TestParseTime() {
...
+			expect: func(t *testing.T, result time.Time) {
+				assert.True(t, inOneHour.Equal(result), "equal")
+			},
+		},
+		{
+			name:  "empty",
+			input: empty,
+			expect: func(t *testing.T, result time.Time) {
+				assert.True(t, oneMinuteAgo.Before(result), "lower bound")
+				assert.True(t, inOneMinute.After(result), "upper bound")
```
ashmrtn, line 93: would you be interested in the `WithinDuration` [assertion](https://pkg.go.dev/github.com/stretchr/testify/assert#WithinDuration)?

src/internal/common/ptr/pointer.go
```diff
@@ -41,7 +45,7 @@ func OrNow(t *time.Time) time.Time {
 		return time.Now().UTC()
 	}
```
ashmrtn, line 46: at this point it sort of feels like we should be setting t and feeding it into `dttm.Now` as well. I'd like to be able to treat the call to the dttm package as a black box and that means I don't want to assume it's not doing additional formatting or such
```suggestion
		t = &time.Time{}
	}
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Catch and report cases where a protected resource is locked out of access.  SDK consumers have a new errs sentinel that allows them to check for this case.
+- Fix a case where missing item LastModifiedTimes could put incremental backups in a permanently failing state.
```
ashmrtn, line 16: nit: wording, it's not permanently breaking the repo, we have a way out of it. Could you please refine this to make it sound not as dire

src/internal/common/dttm/dttm.go
```diff
@@ -98,10 +98,18 @@ func Now() string {
 	return FormatNow(Standard)
 }
 
+func OrNow(t time.Time) time.Time {
+	if (time.Time{}).Equal(t) {
```
ashmrtn, line 102: there's an `IsZero` function for time 😄

src/internal/common/ptr/pointer_test.go
```diff
@@ -99,3 +99,43 @@ func (suite *PointerSuite) TestVal() {
...
+	}
+	for _, test := range table {
+		suite.Run(test.name, func() {
+			t := suite.T()
+			result := ptr.OrNow(test.p)
+
+			if test.expectEqual {
+				assert.Equal(t, *test.p, result)
+			} else {
+				assert.True(t, oneMinuteAgo.Before(result), "lower bound")
```
ashmrtn, line 136: could you also use the `WithinDuration` assertion here?

src/pkg/services/m365/api/item_pager.go
```diff
@@ -206,7 +207,10 @@ func deltaEnumerateItems[T any](
 // shared enumeration runner funcs
 // ---------------------------------------------------------------------------
 
-type addedAndRemovedHandler[T any] func(items []T) (map[string]time.Time, []string, error)
+type addedAndRemovedHandler[T any] func(
+	items []T,
+	getModTime getModTimer[T],
```
ashmrtn, line 212: I'm not really a fan of this design. I think a cleaner solution would be to wrap the items we're iterating through with our own struct that provides the function we want. We already do something similar with calendar events in another place in the code because we wanted `DisplayName` or some such function

src/internal/common/dttm/dttm.go
```diff
@@ -98,10 +98,18 @@ func Now() string {
 	return FormatNow(Standard)
 }
 
+func OrNow(t time.Time) time.Time {
+	if t == (time.Time{}) {
```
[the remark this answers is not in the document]
ryanfkeepers: huh, TIL

src/internal/common/dttm/dttm.go
ashmrtn: oh, this is a duplicate due to the PR getting an update and me hitting refresh and my old comment not being shown

src/internal/common/dttm/dttm_test.go
ryanfkeepers: I know, it's just a standard pattern for me.

src/internal/common/dttm/dttm_test.go
ryanfkeepers: hey, that's nifty!

ryanfkeepers: Not a bad idea.

src/internal/common/ptr/pointer.go
ryanfkeepers: Well, it's permanently breaking Incrementals.  Nonetheless, doomsaying reduced.

src/internal/common/dttm/dttm.go
ryanfkeepers: done

src/internal/common/ptr/pointer_test.go
ryanfkeepers: But then we need a transformer for that wrapper instead.  It's a transformer either way.  Just a matter of where we want the transformation to occur.  Between this change and https://github.com/alcionai/corso/pull/4481 I'll probably revisit this for cleanliness.  But I don't see the difference as a clear win at this stage.

ashmrtn (approved):

PR #4490, opened 2023-10-12
Log the categories participating in an operation

ashmrtn:
To assist in debugging via logs, add the set of
categories an operation is acting on to the
clues set in the context

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/restore.go
+++ b/src/internal/operations/restore.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/alcionai/clues"
 	"github.com/google/uuid"
+	"golang.org/x/exp/slices"
 
 	"github.com/alcionai/corso/src/internal/common/crash"
 	"github.com/alcionai/corso/src/internal/common/dttm"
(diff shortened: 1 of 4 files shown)


src/pkg/selectors/selectors.go
```diff
@@ -293,6 +293,27 @@ func (s Selector) PathCategories() (selectorPathCategories, error) {
...
 
+// AllPathCategories returns the sets of include and filter path categories
+// across all scope sets.
+func (s Selector) AllPathCategories() ([]path.CategoryType, error) {
+	all, err := s.PathCategories()
+	if err != nil {
+		return nil, clues.Stack(err)
+	}
+
+	res := map[string]path.CategoryType{}
```
ryanfkeepers, line 304: Why use the string value here instead of the enum for the key?

src/internal/operations/restore.go
```diff
@@ -138,11 +139,29 @@ func (op *RestoreOperation) Run(ctx context.Context) (restoreDetails *details.De
...
+	allCats, err := op.Selectors.AllPathCategories()
+	if err != nil {
+		// No need to exit over this, we'll just be missing a bit of info in the
+		// log.
+		logger.CtxErr(ctx, err).Info("getting categories for restore")
+	} else {
+		for _, cat := range allCats {
+			cats = append(cats, cat.HumanString())
+		}
+	}
```
ryanfkeepers, line 153: Since this is the same code in all three locations, and the AllPathCategories was created specifically for this, how about either 1) packing the humanString and sort into AllPathCategories() or 2) adding an AllHumanCategories() func that calls AllPathCategories() and runs the transform/sort in that func?

ryanfkeepers (approved):
some nits, lgtm

src/pkg/selectors/selectors.go
ashmrtn: that's an excellent point. My brain died so I didn't think of it 🙃

Issue #4254, opened 2023-09-14
too many service comparisons


ryanfkeepers:
# Issue

One of the results of the repo's organic growth up to this point is the proliferation of service-specific switches and conditionals.  While Corso only covered two services, this was largely a non issue.  After three services we ignored the problem.  Now in the process of adding Groups and Teams we're seeing just how many places in the code we need to acknowledge the existence of the service in order for the support to behave as expected.

This is a death-by-1000-papercuts maintenance issue.  It requires multiple tiny changes to implement high level concepts, increases cognitive load and code familiarity requirements, and will scale poorly with future feature additions.  Further, this problem creates a barrier for code contribution: the domain knowledge required for large-scale additions already races against contributor enthusiasm.  Adding a hundred points of minutia to update will antagonize the experience.

The various implementations of this issue appear as three basic design smells:

1. Domain-isolated data containers
2. Reactive handling to generic structures
3. One-off service conditionals

Across the solution suggestions for these three issues you'll see a couple common design principles:
* For behavior and categorization, interfaces should be prioritized to ensure service implementations can self-describe their state.
* For actual data, generic data structures should be prioritized over Getter interfaces.
* Packages with clear knowledge and ownership of the utilization of common structures should also own the transformations required by common endpoints.
* Proactive is better than reactive.  New implementation should comply with expected, identified conditions to minimize the risk of missing a needed implementation.

## Domain-isolated data containers

Example: [Details ItemInfo](https://github.com/alcionai/corso/blob/main/src/pkg/backup/details/iteminfo.go#L69)

Domain isolation occurs when we juggle multiple structs to identify expected metadata, as opposed to a generic data structure (such as a map) and/or a single interface value.

In the example above, data is split acros five different `DomainInfo` entries (four for services, one for folders).  In order to utilize an ItemInfo struct, all code must first check which of these structs is actually populated (if any), either by comparing each value or by utilizing some external identifier such as path service type.  In addition, many of these types express overlapping data (ex: four of the five contain DriveID and DriveName values).  The result of this is not only poor scaling and code duplication, but also potential gotchas around constraints like having multiple populated types or seeing different in property keys between structs for similar values.

An ideal solution would reduce the ItemInfo to a single DomainInfo interface.  The interface could describe common data categorization checks (WhatCategory, WhatService, IsFolder, IsDriveItem, etc), and should be able to produce a common container of data with well-known property keys (ie: either map or single, standardized struct).

## Reactive handling to generic structures

[...]

PR #4491, opened 2023-10-12
Create service specific handlers that know how to run an export

ashmrtn:
First step in reducing the number of places we have to check the service type manually. Create a way to get a handle to a service specific handler and implement exports for those handlers

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4254

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/export.go
```diff
@@ -1,89 +1,30 @@
...
-		err = clues.Wrap(clues.New(sels.Service.String()), "service not supported")
+	case path.GroupsService:
+		return groups.NewGroupsHandler(opts)
 	}
 
-	ctrl.incrementAwaitingMessages()
-	ctrl.UpdateStatus(status)
-
-	return expCollections, err
+	return &inject.BaseServiceHandler{}
```
ryanfkeepers, line 29: Not particularly a fan of deferring this to a base handler here.  The expectation, I assume, is for the base handler to error if used.  If we're in this code, we already know the caller intends to use the handler but has provided incorrect parameters.  An error up front feels more appropriate.

src/internal/m365/service/sharepoint/export.go
```diff
@@ -16,16 +17,41 @@ import (
...
+) *sharepointHandler {
+	return &sharepointHandler{
+		opts:               opts,
+		backupDriveIDNames: idname.NewCache(nil),
+	}
+}
+
+type sharepointHandler struct {
+	opts control.Options
+
```
ryanfkeepers, line 33:
```suggestion
```

src/internal/operations/export.go
```diff
@@ -266,10 +264,6 @@ func (op *ExportOperation) do(
 		return nil, clues.Stack(err)
 	}
 
-	opStats.ctrl = op.ec.Wait()
```
ryanfkeepers, line 269: I don't see this getting moved anywhere.  But also not seeing where we've dropped the wait counter.  Do we no longer need the wait?

src/internal/operations/inject/inject.go
```diff
@@ -116,4 +113,34 @@ type (
...
+
+func (h BaseServiceHandler) ProduceExportCollections(
+	ctx context.Context,
+	backupVersion int,
+	exportCfg control.ExportConfig,
+	dcs []data.RestoreCollection,
+	stats *data.ExportStats,
+	errs *fault.Bus,
+) ([]export.Collectioner, error) {
+	return nil, clues.New("not implemented")
```
ryanfkeepers, line 143: I'd prefer to avoid implementations and interface fulfillment within the inject package, if possible.  Better for us to let it only declare structures and interfaces used elsewhere.

Per my other comment, this doesn't seem like it has much purpose outside of delaying an error return.  Better off, imo, to remove it altogether if possible.

ryanfkeepers (commented):
I'm a bit concerned about a couple pieces of the design, but generally it looks good.

src/internal/operations/export.go
ashmrtn: nope, don't need it. We [were](https://github.com/alcionai/corso/blob/7b6484663b6f8f11e004105eba34dd0dcf60d299/src/internal/m365/export.go#L85-L86) technically telling the status handler that we added 1 collection and then immediately "completing" the one collection we told it about, with an empty status

Since it was returning all zeros anyway and we can't actually get any stats until the collections are consumed by whatever is doing the export (since we don't decide output here) it seemed less foot gun-ish to just remove the thing that could cause stuff to hang

src/internal/operations/inject/inject.go
ashmrtn: we're probably going to want something like this in the long run, at least to help with bootstrapping and/or interfaces changes going forward. There was a similar debate in kopia a bit back about how best to expand the blob storage interface. Turns out if all things embed a "default" handler then you can add no-op implementations and get compliance immediately. Things that need overrides can override it

We can remove it from this PR, but there's probably a good chance we'll end up with it again

ryanfkeepers: mmkay.  Lets keep it.  Though, per the inject package, maybe it could live in pkg/export?  Or collections.  Or whatevs, its' all good.  Also, how's about a TODO to remove it when migration is complete?

ryanfkeepers (approved):

PR #4494, opened 2023-10-13
add a resource getter setter to newServiceHandler

ryanfkeepers:
This sets us up to remove the service declaration
from the two-step repo creation + init/config, and allows sdk consumers to return to ignoring service until acting on the operation.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn (commented):
couple of high-level questions just so I understand the scope of the change and where we want to take it in the future since I'm not as familiar with this code (numbered for ease of reference):
1. this doesn't change the initialization for non-export operations correct? Right now `NewServiceHandler` is only for export so any changes to overall initialization could cause backup or restore issues
2. right now it seems like we're still planning on storing this resource getter in the controller. Is that the best place for it long-term? How much of the stuff in the controller can we pull out and place in the service handlers themselves as we update the architecture?

ryanfkeepers:
> couple of high-level questions 

1. Correct.  That coverage will only happen as serviceHandler spreads throughout operations usage.
2. Could we move the code out of controller?  Sure, that seems fine.  Not particularly front of mind for me until we get the handlers fully implemented.

ashmrtn:
just noting that this can be rebased for final review now (sorry for delay in getting the base PR merged)

ryanfkeepers:
> just noting that this can be rebased for final review now (sorry for delay in getting the base PR merged)

No worries, and thanks for the heads up!

ryanfkeepers:
@ashmrtn ready for final review

src/internal/m365/export.go
```diff
@@ -17,6 +17,8 @@ func (ctrl *Controller) NewServiceHandler(
 	opts control.Options,
 	service path.ServiceType,
 ) (inject.ServiceHandler, error) {
+	ctrl.setResourceHandler(service)
```
ashmrtn, line 20: nit: just for now can you add a comment that this is more or less a noop? Export isn't using the resource client at all. Restore and backup will, but we haven't pushed the changes through that far yet

ashmrtn (approved):

src/internal/m365/export.go
ryanfkeepers: certainly can

ryanfkeepers:
@ashmrtn something went wonky with the commit history in this branch.  I'm swapping it out with https://github.com/alcionai/corso/pull/4666

Issue #4462, opened 2023-10-10
Rework groups restore


meain:
In the current implementation, the user is able to a "restore" on an entire group. This, as of now restore all the SharePoint sites to back to where they were backed them up from. This model, while it sounds simple, can be unintuitive to the user. The idea is to change the behaviour and to **restrict the restore of only a single resource** which the user has to explicitly specify. For example, the user will have to explicitly state which kind of resource and in case of sites, the site to be restored. This should also take care of  https://github.com/alcionai/corso/issues/4433 .

**This will be a breaking change in the restore behaviour.**

Related: https://github.com/alcionai/corso/issues/3988

PR #4496, opened 2023-10-13
Enforce site filter for Groups restore

meain:
This PR reworks the groups restore to and makes the site to restore to mandatory. This also updates some missing filtering capabilities in groups export.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/4462

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/flags/groups.go
+++ b/src/cli/flags/groups.go
@@ -7,9 +7,10 @@ import (
 const DataMessages = "messages"
 
 const (
-	ChannelFN = "channel"
-	GroupFN   = "group"
-	MessageFN = "message"
+	ChannelFN   = "channel"
+	GroupFN     = "group"
+	MessageFN   = "message"
+	GroupSiteFN = "site"
 
 	MessageCreatedAfterFN    = "message-created-after"
 	MessageCreatedBeforeFN   = "message-created-before"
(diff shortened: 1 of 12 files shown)


src/cli/utils/groups.go
```diff
@@ -101,11 +101,18 @@ func MakeGroupsOpts(cmd *cobra.Command) GroupsOpts {
...
+func ValidateGroupsRestoreFlags(backupID string, opts GroupsOpts, isRestore bool) error {
 	if len(backupID) == 0 {
 		return clues.New("a backup ID is required")
 	}
 
+	// The user has to explicitly specify which resource to restore. In
+	// this case, since we can only restore sites, the user is supposed
+	// to specify which site to restore.
+	if isRestore && len(opts.Site) == 0 {
+		return clues.New("a site ID or web URL is required")
```
ryanfkeepers, line 113: Our standard in sharepoint is to separate siteIDs from weburls with the --site-id and --site flags, respectively. Reducing it to --site works, but i'm a bit worried about confusing end users.  Mostly speaking out loud.  We should probably find a way to remove the --site-id flag for sharepoint to keep things uniform across service commands.

ryanfkeepers (approved):
nits about flag duplication, but approving to unblock

src/cli/flags/groups.go
```diff
@@ -22,12 +23,32 @@ var (
 	GroupFV   []string
 	MessageFV []string
 
+	// Have to create a separate one for restore to make sure we
+	// create on which accepts only one value
+	GroupSiteFV string
```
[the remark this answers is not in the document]
meain: I guess that should work. The only limitation seems to be that we cannot have a "required" flag at the cobra level as we not have two possible options for the user to use.

src/cli/utils/groups.go
meain: Yeah, I think we could have just had a single flag. I don't wanna change that behaviour in this PR. I think that should be a separate larger change.

PR #4497, opened 2023-10-13
remove ctx embedding for counter

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/utils/users.go
```diff
@@ -17,23 +18,28 @@ func UsersMap(
...
 	}
 
 	return au.GetAllIDsAndNames(ctx, errs)
 }
 
-func makeUserAPI(acct account.Account, co control.Options) (api.Users, error) {
+func makeUserAPI(
+	acct account.Account,
+	counter *count.Bus,
+	co control.Options,
```
ashmrtn, line 35: nit: this feels a little weird since it's got the opposite ordering for `counter` and `co` than `m365.NewController`

src/pkg/services/m365/groups.go
```diff
@@ -33,8 +34,9 @@ func GroupByID(
 	ctx context.Context,
 	acct account.Account,
 	id string,
+	counter *count.Bus,
```
ashmrtn, line 37: this is an API that SDK users may interact with. Is this going to cause issues if they update the version of corso that they're using?

ashmrtn (commented):
mainly just a question to discuss prior to merging

src/pkg/services/m365/groups.go
ryanfkeepers: It'll change the api signature, yes.

src/cli/utils/users.go
ryanfkeepers: aye, will fix

ryanfkeepers: All ordering should now follow:
1. regular params
2. busses
  a. counter
  b. fault 
3. variadics

src/pkg/services/m365/groups.go
ashmrtn: is there a way around that? API updates are kind of a hassle and can land on people unexpectedly when they try to update corso

ryanfkeepers: we could revoke the capacity for callers to pass in their own counter.  I suppose I'll do that.  If anyone asks for it we can add it back later.

ryanfkeepers: alrighty, that's done

ashmrtn (approved):
LGTM, thanks for the updates!

PR #4498, opened 2023-10-13
replace col scope with bool

ryanfkeepers:
We aren't using this property as an enum, nor does it look like we have reason to do so in the future.  Currently, the enumeration values obscure the point of tracking the property.  This value change should make that more clear.

Next PR will handle identifying parent-child package relationships.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collection.go
+++ b/src/internal/m365/collection/drive/collection.go
@@ -286,7 +286,7 @@ func (oc *Collection) getDriveItemContent(
 			// oc.isPackageOrChildOfPackage && *item.GetSize() >= MaxOneNoteFileSize {
 			// TODO: We've removed the file size check because it looks like we've seen persistent
 			// 503's with smaller OneNote files also.
-			oc.isPackageOrChildOfPackage || strings.EqualFold(itemMimeType, oneNoteMimeType)) {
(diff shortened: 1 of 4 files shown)


src/internal/m365/collection/drive/collection.go
```diff
@@ -72,13 +72,14 @@ type Collection struct {
...
-	// files within them belong to a folder.
-	scope collectionScope
+	// true if this collection, or a parent directory of this collection,
+	// is marked as a package.
+	// packages are only marked on the top-level directory, but special-case
+	// handling need apply to all subfolders.  Therefore it is necessary to cascade
+	// that identification to all affected collections, not just those that identify
+	// as packages themselves.
+	// see: https://learn.microsoft.com/en-us/graph/api/resources/package?view=graph-rest-1.0
+	isPackageOrChildOfPackage bool
```
meain, line 82: How about just `isPackage`?

meain (approved):

src/internal/m365/collection/drive/collection.go
ryanfkeepers: going to keep for now, since it clearly idicates the conditions for the flag.  two-way door; so if we hate it in the future, we can rename as needed.

PR #4501, opened 2023-10-13
mark complete subtrees as packages

ryanfkeepers:
marks the full subtree of a drive package item
as being within the package.  This ensures
that all collections whose directory is a subdir of a package-type folder are themselves marked as package-typed.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/collections.go
```diff
@@ -794,6 +804,17 @@ func (c *Collections) UpdateCollections(
...
+			isPackage := item.GetPackageEscaped() != nil
+			if isPackage {
+				// mark this path as a package type for all other collections.
+				// any subfolder should get marked as a childOfPackage below.
+				topLevelPackages[collectionPath.String()] = struct{}{}
+			}
+
+			childOfPackage := filters.
+				PathPrefix(maps.Keys(topLevelPackages)).
+				Compare(collectionPath.String())
```
meain, line 816: I guess this could be moved into an else block of the previous if.

meain (approved):
Small nit, but approving to unblock.

PR #4509, opened 2023-10-16
Allow restore data from sites in groups backup to different site

meain:
This adds the option restore a site to a different one from the one that it was backed up from.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* fixes https://github.com/alcionai/corso/issues/4511

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/utils/restore_config.go
+++ b/src/cli/utils/restore_config.go
@@ -11,6 +11,7 @@ import (
 	. "github.com/alcionai/corso/src/cli/print"
 	"github.com/alcionai/corso/src/internal/common/dttm"
 	"github.com/alcionai/corso/src/pkg/control"
+	"github.com/alcionai/corso/src/pkg/path"
 )
 
 type RestoreCfgOpts struct {
@@ -23,10 +24,18 @@ type RestoreCfgOpts struct {
 	ProtectedResource string
 	SkipPermissions   bool
 
+	// SubService is useful when we are trying to restore Groups where
+	// the user can only restore a single service at any point instead of
+	// the entire backup.
+	SubServiceType path.ServiceType
+	SubService     string
+
 	Populated flags.PopulatedFlags
 }
 
-func makeRestoreCfgOpts(cmd *cobra.Command) RestoreCfgOpts {
+// makeBaseRestoreCfgOpts fills in all RestoreCfgOpts except for sub
+// service details. That is filled by the caller.
+func makeBaseRestoreCfgOpts(cmd *cobra.Command) RestoreCfgOpts {
 	return RestoreCfgOpts{
 		Collisions:        flags.CollisionsFV,
 		Destination:       flags.DestinationFV,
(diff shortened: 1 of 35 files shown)


src/internal/m365/controller.go
```diff
@@ -197,88 +198,6 @@ func (ctrl *Controller) CacheItemInfo(dii details.ItemInfo) {
 	}
 }
 
-// ---------------------------------------------------------------------------
```
meain, line 200: Move as is to `src/internal/m365/service/common/lookup.go`.

src/internal/operations/test/helper_test.go
```diff
@@ -398,6 +398,21 @@ func generateContainerOfItems(
...
 	restoreCfg.IncludePermissions = true
 
+	if sel.Service == selectors.ServiceGroups {
+		restoreCfg.SubService.Type = path.SharePointService
+		restoreCfg.SubService.ID = siteID
+	}
+
+	var protectedResource idname.Provider = sel
+
+	// In case of  groups, we use the root site as the restore target
```
meain, line 408: Do we have to make it configurable by caller? I didn't think it was required as this is test code.

src/cli/utils/restore_config.go
```diff
@@ -23,10 +24,18 @@ type RestoreCfgOpts struct {
 	ProtectedResource string
 	SkipPermissions   bool
 
+	// SubService is useful when we are trying to restore Groups where
+	// the user can only restore a single service at any point instead of
+	// the entire backup.
+	SubServiceType path.ServiceType
+	SubService     string
```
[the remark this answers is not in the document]
ryanfkeepers: This seems to duplicate the selector's ownership.  Moving these out of there and into the restore config opens up the potential for conflict between these props and the selector.  Similarly, it might lead to building this data into the selector downstream anyway.  Iirc, the selector already has the capacity to specify the secondary resource; seems like this identification could be owned there instead.

ryanfkeepers (changes requested):
Looks functional, but I don't fully agree with the design as it splits ownership across multiple systems (configuration, selection) instead of holding to the current pattern of using selectors to source restore data, which would centralize control within a singular interface.

Separately, this PR has too much going on to easily review.  There's a lot of refactoring, plus changes across the cli, operations, and service layers.  To review it properly these things need to get split out across several PRs.  I'd recommend:

* first: change selectors to best accomodate subservice declaration and identification according to an interface (we may want a design doc for this).
* second: change the operations and service layers to consume the new selectors interface.
* third: expose these changes via the CLI and SDK boundaries.

interpolate large refactors as separate PRs along the way, whenever appropriate.

src/internal/operations/restore.go
```diff
@@ -227,7 +227,11 @@ func (op *RestoreOperation) do(
 		return nil, clues.Wrap(err, "getting backup and details")
 	}
 
-	restoreToProtectedResource, err := chooseRestoreResource(ctx, op.rc, op.RestoreCfg, bup.Selector)
+	restoreService, restoreToProtectedResource, err := op.rc.GetRestoreResource(
```
ashmrtn, line 230: is there a way the selector and this service can differ? Could we remove the service return value and continue to use the value from the selector? Seems like it would be less error prone

meain:
Temporarily moving over to implementing exports for email as that has moved up in priority. Will get back to rework this after that.

meain:
Closing this PR. Will create a new PR when revisiting this issue.

PR #4518, opened 2023-10-17
add passphrase flag

neha-Gupta1:
add `--new-passphrase` flag for `update-passphrase` command  

- [ ] :no_entry: No

#### Type of change

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/repo/repo.go
+++ b/src/cli/repo/repo.go
@@ -60,6 +60,7 @@ func AddCommands(cmd *cobra.Command) {
 	flags.AddForceMaintenanceFlag(maintenanceCmd)
 	flags.AddMaintenanceUserFlag(maintenanceCmd)
 	flags.AddMaintenanceHostnameFlag(maintenanceCmd)
+	flags.AddCorsoUpdatePassphraseFlags(updatePassphraseCmd)
 
 	for _, addRepoTo := range repoCommands {
 		addRepoTo(initCmd)


src/cli/repo/repo.go
```diff
@@ -60,6 +60,7 @@ func AddCommands(cmd *cobra.Command) {
 	flags.AddForceMaintenanceFlag(maintenanceCmd)
 	flags.AddMaintenanceUserFlag(maintenanceCmd)
 	flags.AddMaintenanceHostnameFlag(maintenanceCmd)
+	flags.AddCorsoUpdatePassphraseFlags(updatePassphraseCmd)
```
meain, line 63: We should make the flag required as there is no point in the user calling without it.

PS: I don't think we need the word `Corso` in the function name, but we'll keep it since it is already there anyways.

meain (changes requested):

src/cli/flags/repo.go
```diff
@@ -76,12 +76,16 @@ func AddCorsoPassphaseFlags(cmd *cobra.Command) {
...
+func AddUpdatePassphraseFlags(cmd *cobra.Command, require bool) {
 	fs := cmd.Flags()
 	fs.StringVar(&NewPhasephraseFV,
 		NewPassphraseFN,
 		"",
 		"update Corso passphrase for repo")
+
+	if require {
+		cobra.CheckErr(cmd.MarkFlagRequired(NewPassphraseFN))
+	}
```
meain, line 88: I guess we can always mark it as required. I don't see why we need a flag.

meain (approved):
Small nit, but approving to unblock.

src/cli/repo/repo.go
neha-Gupta1: We were using the Corso word everywhere. But since its removed from other places it makes sense to remove it from here too.
Always added flag to required.

src/cli/flags/repo.go
neha-Gupta1: Just to make it more flexible if its required in future.

PR #4521, opened 2023-10-17
Don't update prev delta url

pandeyabs:
URL cache should use the same delta token base used by `ProduceCollections`. This is because we want to update URLs for the set of items which was discovered by `ProduceCollections`. Otherwise, we'll have url cache misses & we'll have to resort to doing GetItem graph calls for each item which means we lose all goodness of the url cacche.

This reverts a change introduced in a recent [PR](https://github.com/alcionai/corso/commit/5215e907b05c25415e738c21a1292d33f74e3419#diff-f79ab9c9b31db3600d4e14b6df4d4c679b4356e32c3a68eaeaeb934c1c5a13a7R175) where we are updating cache's delta token base to a newer one. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/url_cache.go
+++ b/src/internal/m365/collection/drive/url_cache.go
@@ -176,7 +176,7 @@ func (uc *urlCache) refreshCache(
 		}
 	}
 
-	du, err := pager.Results()
+	_, err := pager.Results()
 	if err != nil {
 		return clues.Stack(err)
 	}
@@ -185,7 +185,6 @@ func (uc *urlCache) refreshCache(
 
 	// Update last refresh time
 	uc.lastRefreshTime = time.Now()
-	uc.prevDelta = du.URL
 
 	return nil
 }


ashmrtn (approved):

src/internal/m365/collection/drive/url_cache.go
```diff
@@ -176,7 +176,7 @@ func (uc *urlCache) refreshCache(
 		}
 	}
 
-	du, err := pager.Results()
+	_, err := pager.Results()
```
vkamra, line 179: Can we update `TestURLCacheBasic` to also validate uc.prevDelta didn't change after the refresh?

ashmrtn: yea, was just thinking of adding a comment about that

pandeyabs: Thank you! I'll add this.

PR #4525, opened 2023-10-20
Make the cli colorful and informational

meain:
This commit include a few things that improves the progresbars and other info messages are printed out during the CLI run.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/observe/observe.go
ryanfkeepers, line 1: I like what's going on here, but am going to be a little critical of the design, as I think it'll be difficult to grok and maintain in the future.  In short, the terminology and func production is causing me a bit more confusion than clarity, and I think the following suggestion will help make behavior more transparent.

Terms like `Section` and `SubMessage` and `Tip` can make sense during the implementation process, but ultimately add a lot of cognitive load for other maintainers to understand what those terms mean in implementation.  Additionally, they tend to require combinatorics of func production.  EG: should a `Section` expect a completion marker?  Do we also need a `SectionWithCompletion`?

Instead of adding an array of new funcs, we could probably condense this all into one or two primary funcs that take in a set of configurations.  Quick hack:
```go
type ProgressCfg struct {
  NewSection bool  // instead of Section
  Indent int              // instead of Sub-*
  CompletionMessage func() string // instead of *-Tip
}

// could probably remove this func entirely with a cfg prop
// like `AutoComplete bool`
func Message(ctx, progCfg, msgs...) {...}
func MessageWithCompletion(ctx, progCfg, msg) {...}
```

This allows the code to minimize the amount of boilerplate integration with mpb and, more importantly, shifts identification of formatting away from arbitrary naming systems and into a concise set of configurations.  Even if we use constructors to produce common designs (ex: `Message(ctx, IndentWithOnComplete(tip), msg...)`), that'll allow maintainers to look at the configuration itself to grok behavior ("Oh, this message indents once, and writes a specific message on completion") instead of needing to understand the mpb integration.

src/cli/backup/backup.go
```diff
@@ -187,8 +190,15 @@ func genericCreateCommand(
...
-			Errf(ictx, "%v\n", err)
+			cerr := clues.Wrap(err, owner).WithClues(ictx)
+			errs = append(errs, cerr)
+
+			clj, jmerr := json.Marshal(cerr.Core().Values)
+			if jmerr != nil {
+				clj = []byte("Unable to marshal error metadata")
+			}
+
+			Errf(ictx, color.Red("ERROR: Unable to complete backup")+" \nMessage: %v\nMetadata: %s\n", err, clj)
```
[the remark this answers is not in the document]
meain: Good catch. I've kept it to `Error` as that is what cobra also does. As for the reason I moved the `Error` string from `Errf`, for one it was inconsistent with `Err`, also, I wanted to just color the header part of the error and not the entire error.

As a middle ground I have introduced two new functions `SimpleError` and `OperationError` instead of `Err` and `Errf`.

src/cli/print/print.go
```diff
@@ -83,16 +83,14 @@ func Only(ctx context.Context, e error) error {
...
 func Err(ctx context.Context, s ...any) {
 	out(ctx, getRootCmd(ctx).ErrOrStderr(), s...)
 }
 
 // Errf prints the params to cobra's error writer (stdErr by default)
 // if s is nil, prints nothing.
-// Prepends the message with "Error: "
 func Errf(ctx context.Context, tmpl string, s ...any) {
-	outf(ctx, getRootCmd(ctx).ErrOrStderr(), "\nError: \n\t"+tmpl+"\n", s...)
+	outf(ctx, getRootCmd(ctx).ErrOrStderr(), tmpl, s...)
```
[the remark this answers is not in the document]
meain: Refer to https://github.com/alcionai/corso/pull/4525#discussion_r1374173931

src/internal/color/common.go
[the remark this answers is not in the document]
meain: I've moved it to `/common/color`. Using `print` was causing import cycles.

src/internal/m365/collection/site/backup.go
```diff
@@ -43,6 +47,15 @@ func CollectLibraries(
 			bpc.Options)
 	)
 
+	name := stdpath.Base(bpc.ProtectedResource.Name())
+	msg := fmt.Sprintf("%s (%s)", path.LibrariesCategory.HumanString(), name)
```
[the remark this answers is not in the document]
meain: Both looks fine to me, I'll change to your suggestion.

src/internal/m365/service/groups/backup.go
```diff
@@ -134,15 +131,26 @@ func ProduceBackupCollections(
 
 				dbcs = append(dbcs, cs...)
 			}
-
 		case path.ChannelMessagesCategory:
+			var (
+				cs                   []data.BackupCollection
+				canUsePreviousBackup bool
+				err                  error
```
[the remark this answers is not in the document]
meain: This is done so that we can capture `cs` in the closure passed to `SubMessageWithCompletionAndTip` [here](https://github.com/alcionai/corso/blob/3dec2fbd56d6901bdcfe268efd7c2a510da34098/src/internal/m365/service/groups/backup.go#L145).

src/internal/observe/observe.go
```diff
@@ -209,18 +245,55 @@ func Message(ctx context.Context, msgs ...any) {
...
+	ctx context.Context,
+	msg string,
+	tip func() string,
+) chan<- struct{} {
+	var (
+		plain    = color.Cyan(msg)
+		loggable = fmt.Sprintf("%v", msg)
+	)
+
+	return messageWithCompletion(ctx, plain, loggable, 2, 40, tip)
```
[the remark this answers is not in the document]
meain: I've changed it to `progressBarWidth`.

src/internal/observe/observe.go
meain: The reason to have a separate function for `Section`(instead of merging with `Message`) is primarily to make sure we take two values instead of `msgs...`. We need this to make sure we are able to format and color it appropriately.

As of `MessageWithCompletion`, let me condense it to a single function, although I think using function arguments would be better than a separate config struct as the structs created will only be used once.

Let me know your thoughts.

ryanfkeepers: I mean, you know my [preference on](https://github.com/alcionai/corso/blob/main/src/pkg/selectors/scopes.go#L219) [extensible](https://github.com/alcionai/corso/blob/main/src/internal/m365/graph/service.go#L187) [func args](https://github.com/alcionai/corso/blob/main/src/cli/utils/utils.go#L154).  And, personally, I think this is an excellent case for that patter.  Per `Section`, for example, an optional func arg lets you add both the flag for section formatting, and also the section string, without disrupting the basic func signature.

src/cli/print/print.go
```diff
@@ -83,16 +84,29 @@ func Only(ctx context.Context, e error) error {
...
 // Errf prints the params to cobra's error writer (stdErr by default)
 // if s is nil, prints nothing.
-// Prepends the message with "Error: "
+// You should ideally be using SimpleError or OperationError.
 func Errf(ctx context.Context, tmpl string, s ...any) {
-	outf(ctx, getRootCmd(ctx).ErrOrStderr(), "\nError: \n\t"+tmpl+"\n", s...)
+	outf(ctx, getRootCmd(ctx).ErrOrStderr(), tmpl, s...)
+}
+
+func SimpleError(ctx context.Context, message string) {
```
ryanfkeepers, line 99: SimpleError feels like the behavior we should be getting out of using `Err`.  I think we can ditch this wrapper to prefer the standard `Err` and `Errf` behavior.

src/cli/print/print.go
```diff
@@ -83,16 +84,29 @@ func Only(ctx context.Context, e error) error {
...
+	Err(ctx, color.Red("Error: "+message))
+}
+
+func OperationError(ctx context.Context, header string, body error, metadata map[string]any) {
+	meta, err := json.Marshal(metadata)
+	if err != nil {
+		meta = []byte("Unable to marshal error metadata")
+	}
+
+	Err(ctx, color.Red("Error: "+header)+" \nMessage: %v\nMetadata: %s\n", body, meta)
```
ryanfkeepers, line 109: Not much a fan of this, for the same principles as discussed with observe.  What you're providing here is basically a template. Plus formatting the verbose error values.  On that note, I wholly disagree with displaying the clues values to the end user.  That's largely debugging metadata and will likely make for a poor and difficult to understand experience.  What's left here without clues values is the following call:

```go
// assumes Errf adds the "Error:" prefix and colors things red.
Errf(ctx, "\nMessage: %v\n", err)
```

Per coloring: I'm a fan of coloring the entire output, as that clearly delineates where the error message content ends and the next non-error content begins.

If it feels like I'm being nitpicky about formatting, its because the only difference between `Info[f]` and `Err[f]` _is_ the formatting.  Othewise the funcs are the same.  Case in point, due to the bug with Err() not adding the "Error" prefix, it is _exactly_ the same as Info().  Any formatting we move out of these two funcs serves to erase their point of existance.

Which, you know, could be fine.  Because it's all templates in the end.  We'd be just as well off with only using Infof() and making various error templates like:
```go
const ErrWithHeader color.Red("Error: %v\nMessage: %v\n")
Infof(ctx, ErrWithHeader, err, header)
```

meain:
@ryanfkeepers this should now be good to go.

ryanfkeepers:
@meain so sorry, lost track of this.  Looks like it has a conflict; if you could fix that up I'll review it first thing tomorrow.  Thanks!

meain:
@ryanfkeepers this should now be good to go

src/cli/backup/backup.go
```diff
@@ -187,8 +190,15 @@ func genericCreateCommand(
...
-			Errf(ictx, "%v\n", err)
+			cerr := clues.Wrap(err, owner).WithClues(ictx)
+			errs = append(errs, cerr)
+
+			meta, err := json.Marshal(cerr.Core().Values)
+			if err != nil {
+				meta = []byte("Unable to marshal error metadata")
+			}
+
+			Errf(ictx, "%s\nMessage: %v\nMetadata:%s", "Unable to complete backup", err, meta)
```
ryanfkeepers, line 201: Still not a fan of displaying clues values to end users.  I do think it'll make the ux worse.  Maybe run it by @gmatev for a final eval?

src/cli/print/print.go
```diff
@@ -6,10 +6,12 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/fatih/color"
```
ryanfkeepers, line 9: could we wrap the `fatih/color` interface in `internal/common/color`?  Would be nicer to run everything through a single package's interface.  Will also make swapping out the package easier to do, if we ever need to go that routet.

src/cli/print/print.go
```diff
@@ -83,16 +85,22 @@ func Only(ctx context.Context, e error) error {
 
 // Err prints the params to cobra's error writer (stdErr by default)
 // if s is nil, prints nothing.
-// Prepends the message with "Error: "
+// You should ideally be using SimpleError or OperationError.
```
ryanfkeepers, line 88: these comments are out of date now, yeah?

ryanfkeepers (approved):
some nits on cli UX, but otherwise gtg

src/cli/backup/backup.go
meain: I see your point. Let me drop that.

PR #4533, opened 2023-10-24
introduce debug commands

ryanfkeepers:
adds the primary `debug` command to the cli as
a hidden command option.  Also includes the scaffold for
a `metadata-files` command that could be used to print
out the metadata files in the backup.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/cli/debug/debug.go
+++ b/src/cli/debug/debug.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/alcionai/corso/src/cli/flags"
 	. "github.com/alcionai/corso/src/cli/print"
 	"github.com/alcionai/corso/src/cli/utils"
 	"github.com/alcionai/corso/src/pkg/selectors"
@@ -31,6 +32,8 @@ func AddCommands(cmd *cobra.Command) {
 
 		for _, addTo := range debugCommands {
 			addTo(subCommand)
+			flags.AddAllProviderFlags(subCommand)
+			flags.AddAllStorageFlags(subCommand)
 		}
 	}
 }
--- a/src/cli/debug/debug.go
+++ b/src/cli/debug/debug.go
@@ -1,0 +1,102 @@
+package debug
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	. "github.com/alcionai/corso/src/cli/print"
+	"github.com/alcionai/corso/src/cli/utils"
+	"github.com/alcionai/corso/src/pkg/selectors"
+)
+
+var subCommandFuncs = []func() *cobra.Command{
+	metadataFilesCmd,
+}
+
+var debugCommands = []func(cmd *cobra.Command) *cobra.Command{
+	addOneDriveCommands,
+	addSharePointCommands,
+	addGroupsCommands,
+	addExchangeCommands,
+}
+
+// AddCommands attaches all `corso debug * *` commands to the parent.
+func AddCommands(cmd *cobra.Command) {
+	debugC, _ := utils.AddCommand(cmd, debugCmd(), utils.MarkDebugCommand())
+
+	for _, sc := range subCommandFuncs {
+		subCommand := sc()
+		utils.AddCommand(debugC, subCommand, utils.MarkDebugCommand())
+
+		for _, addBackupTo := range debugCommands {
+			addBackupTo(subCommand)
+		}
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Commands
+// ---------------------------------------------------------------------------
+
+const debugCommand = "debug"
+
+// The debug category of commands.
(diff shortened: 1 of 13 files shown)


src/cli/debug/debug.go
```diff
@@ -0,0 +1,102 @@
...
+
+// AddCommands attaches all `corso debug * *` commands to the parent.
+func AddCommands(cmd *cobra.Command) {
+	debugC, _ := utils.AddCommand(cmd, debugCmd(), utils.MarkDebugCommand())
+
+	for _, sc := range subCommandFuncs {
+		subCommand := sc()
+		utils.AddCommand(debugC, subCommand, utils.MarkDebugCommand())
+
+		for _, addBackupTo := range debugCommands {
```
pandeyabs, line 32: nit: `dc` instead of `addBackupTo`?

src/cli/flags/debug.go
```diff
@@ -0,0 +1 @@
+package flags
```
pandeyabs, line 1: Can we remove these empty files?

src/cli/debug/debug.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Handler for flat calls to `corso debug`.
+// Produces the same output as `corso debug --help`.
+func handledebugCmd(cmd *cobra.Command, args []string) error {
+	return cmd.Help()
+}
+
+// The backup metadataFiles subcommand.
+// `corso backup metadata-files <service> [<flag>...]`
```
pandeyabs, line 63:
```suggestion
// `corso debug metadata-files <service> [<flag>...]`
```

src/cli/debug/debug.go
```diff
@@ -0,0 +1,102 @@
...
+// ---------------------------------------------------------------------------
+
+const debugCommand = "debug"
+
+// The debug category of commands.
+// `corso debug [<subcommand>] [<flag>...]`
+func debugCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   debugCommand,
+		Short: "debug your service data",
```
pandeyabs, line 49: In future, this command hierarchy may include commands unrelated to service data(e.g. performance, So can this description be made more generic?

I don't have a good alternative - maybe `Debug & troubleshooting utilities for Corso`

src/cli/debug/debug.go
```diff
@@ -0,0 +1,102 @@
...
+func metadataFilesCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   metadataFilesCommand,
+		Short: "Display all the metadata file contents stored by the service",
+		RunE:  handleMetadataFilesCmd,
+		Args:  cobra.NoArgs,
+	}
+}
+
+// Handler for calls to `corso backup metadata-files`.
```
pandeyabs, line 75:
```suggestion
// Handler for calls to `corso debug metadata-files`.
```

pandeyabs: Also right below on line 76

src/cli/debug/exchange.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil
+	}
+
+	// opts := utils.MakeExchangeOpts(cmd)
+
+	if flags.RunModeFV == flags.RunModeFlagTest {
+		return nil
+	}
+
+	sel := selectors.NewExchangeBackup([]string{"me"})
```
pandeyabs, line 70: nit: Can we remove the hardcoded identifiers - `"me"`, `"Exchange"`, `"OneDrive"` etc

src/cli/debug/debug.go
```diff
@@ -0,0 +1,102 @@
...
+
+const debugCommand = "debug"
+
+// The debug category of commands.
+// `corso debug [<subcommand>] [<flag>...]`
+func debugCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   debugCommand,
+		Short: "debug your service data",
+		Long:  `debug the data stored in one of your M365 services.`,
```
pandeyabs, line 50: Can we add a note somewhere which says that these debug commands won't modify customer data / repo state?

pandeyabs (approved):

src/cli/debug/debug.go
pandeyabs: Nvm, I think we may want to keep the capability to modify data. Especially if being used by corso developers.

src/cli/flags/debug.go
ryanfkeepers: sure can

src/cli/debug/exchange.go
ryanfkeepers: Per services: that'll need a separate refactor for human readable service strings.
Per "me": a value here is necessary; but I'll make it more obvious that it's a placeholder.

PR #4538, opened 2023-10-25
[Bugfix] Fixing getting owner email

juleslasarte:
Owner data might have email and ID, or only one of them, or none. Fixing the code to handle this and adding a unit test. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/sites.go
+++ b/src/pkg/services/m365/sites.go
@@ -21,7 +21,7 @@ import (
 type SiteOwnerType string
 
 const (
-	SiteOwnerUnknown SiteOwnerType = ""
+	SiteOwnerUnknown SiteOwnerType = "unknown"
 	SiteOwnerUser    SiteOwnerType = "user"
 	SiteOwnerGroup   SiteOwnerType = "group"
 )
--- a/src/pkg/services/m365/sites.go
+++ b/src/pkg/services/m365/sites.go
@@ -65,7 +65,11 @@ func SiteByID(
 		Expand: []string{"drive"},
 	}
 
-	s, err := ac.Sites().GetByID(ctx, id, cc)
+	return getSiteByID(ctx, ac.Sites(), id, cc)
+}
+
+func getSiteByID(ctx context.Context, ga api.GetByIDer[models.Siteable], id string, cc api.CallConfig) (*Site, error) {
+	s, err := ga.GetByID(ctx, id, cc)
 	if err != nil {
 		return nil, clues.Stack(err)
 	}
@@ -145,15 +149,10 @@ func ParseSite(item models.Siteable) *Site {
 				return s
 			}
 
-			s.OwnerID, err = str.AnyValueToString("id", group)
-			if err != nil {
-				return s
-			}
-
-			s.OwnerEmail, err = str.AnyValueToString("email", group)
-			if err != nil {
-				return s
-			}
+			// ignore the errors, these might or might not exist
+			// if they don't exist, we'll just have an empty string
+			s.OwnerID, _ = str.AnyValueToString("id", group)
+			s.OwnerEmail, _ = str.AnyValueToString("email", group)
 		}
 	}
 
--- a/src/pkg/services/m365/sites.go
+++ b/src/pkg/services/m365/sites.go
@@ -13,6 +13,7 @@ import (
 	"github.com/alcionai/corso/src/internal/m365/graph"
 	"github.com/alcionai/corso/src/pkg/account"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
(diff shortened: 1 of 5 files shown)


src/pkg/services/m365/sites_test.go
```diff
@@ -81,10 +82,7 @@ func (suite *siteIntegrationSuite) TestSites_GetByID() {
 			assert.NotEmpty(t, site.WebURL)
 			assert.NotEmpty(t, site.ID)
 			assert.NotEmpty(t, site.DisplayName)
-			if site.OwnerType != SiteOwnerUnknown {
-				assert.NotEmpty(t, site.OwnerID)
-				assert.NotEmpty(t, site.OwnerType)
-			}
```
juleslasarte, line 87: we can't actually guarantee this is true.  We have separate test with well known groups for ID and email.

src/pkg/services/m365/sites.go
```diff
@@ -65,7 +65,11 @@ func SiteByID(
 		Expand: []string{"drive"},
 	}
 
-	s, err := ac.Sites().GetByID(ctx, id, cc)
+	return getSiteByID(ctx, ac.Sites(), id, cc)
+}
+
+func getSiteByID(ctx context.Context, ga api.GetByIDer[models.Siteable], id string, cc api.CallConfig) (*Site, error) {
```
ryanfkeepers, line 71: corso code standards prefer multiline func props for readability.
```suggestion
func getSiteByID(
    ctx context.Context,
    ga api.GetByIDer[models.Siteable],
    id string,
    cc api.CallConfig,
) (*Site, error) {
```

src/pkg/services/m365/sites.go
```diff
@@ -145,15 +149,10 @@ func ParseSite(item models.Siteable) *Site {
...
-			}
-
-			s.OwnerEmail, err = str.AnyValueToString("email", group)
-			if err != nil {
-				return s
-			}
+			// ignore the errors, these might or might not exist
+			// if they don't exist, we'll just have an empty string
+			s.OwnerID, _ = str.AnyValueToString("id", group)
+			s.OwnerEmail, _ = str.AnyValueToString("email", group)
```
ryanfkeepers, line 155: the linter should complain here (rightly) about swallowing errors.  The errors should at least get logged.  Or returned, if the error is not a "not found" case.

src/pkg/services/m365/sites_test.go
```diff
@@ -216,3 +222,146 @@ func (suite *siteUnitSuite) TestGetAllSites() {
 		})
 	}
 }
+
+func userIdentitySet(userID string, userEmail string) models.IdentitySetable {
```
ryanfkeepers, line 226: the `-Set` suffix is kinda confusing.  I think the names could do without.
```suggestion
func userIdentity(userID string, userEmail string) models.IdentitySetable {
```

src/pkg/services/m365/sites_test.go
```diff
@@ -216,3 +222,146 @@ func (suite *siteUnitSuite) TestGetAllSites() {
...
+		groupData["id"] = groupID
+	}
+
+	group := models.NewIdentitySet()
+	group.SetAdditionalData(map[string]any{"group": groupData})
+
+	return group
+}
+
+func dummySite(owner models.IdentitySetable) models.Siteable {
```
ryanfkeepers, line 258: This should preferably live in /api/mocks/site.go, or /api/testdata/site.go (though that package doesn't exist yet) so that it's easily accesible throughout the codebase.

src/pkg/services/m365/sites_test.go
```diff
@@ -216,3 +222,146 @@ func (suite *siteUnitSuite) TestGetAllSites() {
...
+	site.SetDrive(drive)
+
+	return site
+}
+
+func (suite *siteUnitSuite) TestGetSites() {
+	table := []struct {
+		name       string
+		mock       func(context.Context) api.GetByIDer[models.Siteable]
+		expectErr  func(*testing.T, error)
```
ryanfkeepers, line 275: you can simplify this by using the assert comparison func type:
```suggestion
		expectErr  assert.ErrorAssertionFunc
```
Then on line 364 call:
```go
    test.expectErr(t, err, clues.ToCore(err))
```

ryanfkeepers (approved):
handful of nits, but generally looks good.  Approving to unblock.  Thanks for the fixes.

src/pkg/services/m365/sites.go
juleslasarte: linter def did not complain. Happy to log it -- but is it really an error? we know there's (seemingly valid) cases where either of this might not be returned at all. Why clutter the error output?

ryanfkeepers: It could be an error.  And sentinels tend to be easier to manage than pairwise combinatorics of values+flags+errors.  But it doesn't look like I properly built in sentinels here, so an Info level log is probably sufficient.

juleslasarte: roger. adding this.

PR #4539, opened 2023-10-25
expose additional channel metadata

ryanfkeepers:
builds out more details for channel messages and replies

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #3988

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/backup/details/groups.go
+++ b/src/pkg/backup/details/groups.go
@@ -37,25 +37,32 @@ func NewGroupsLocationIDer(
 
 // GroupsInfo describes a groups item
 type GroupsInfo struct {
-	Created    time.Time `json:"created,omitempty"`
-	ItemName   string    `json:"itemName,omitempty"`
-	ItemType   ItemType  `json:"itemType,omitempty"`
-	Modified   time.Time `json:"modified,omitempty"`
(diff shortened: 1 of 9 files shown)


src/pkg/backup/details/groups.go
```diff
@@ -37,25 +37,32 @@ func NewGroupsLocationIDer(
 
 // GroupsInfo describes a groups item
 type GroupsInfo struct {
+	ItemType ItemType  `json:"itemType,omitempty"`
+	Modified time.Time `json:"modified,omitempty"`
+
+	// Channels Specific
+	Message   ChannelMessageInfo `json:"message"`
```
meain, line 44: Should we be worried about not being able read older details file? I guess it should be fine as we still mention that "groups" is still experimental, but wanted to bring it up in case this was something you had overlooked.

ryanfkeepers: I'm leaning heavily on "not yet released" here.  We've communicated to the CLI community that things may break or change before full release.  And the SDK community will receive similar communication.

meain (approved):

PR #4544, opened 2023-10-26
fix api/mock circular ref

ryanfkeepers:
No logic changes.  Just code movement/renaming.

---

- [x] :no_entry: No

#### Type of change


src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -40,9 +40,9 @@ type NextPager[T any] interface {
 	NextPage() (items []T, reset, done bool)
 }
 
-type nextPage[T any] struct {
-	items []T
-	reset bool
+type NextPage[T any] struct {
+	Items []T
+	Reset bool
```
ashmrtn, line 45: I only did some quick greps on the branch, but I don't think these fields need to be exported

src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -51,59 +51,59 @@ type NextPageResulter[T any] interface {
...
+var _ NextPageResulter[any] = &NextPageResults[any]{}
 
-type nextPageResults[T any] struct {
-	pages chan nextPage[T]
-	du    DeltaUpdate
-	err   error
+type NextPageResults[T any] struct {
+	Pages chan NextPage[T]
+	DU    DeltaUpdate
+	Err   error
```
ashmrtn, line 59: I don't think these fields really need to be exported either

src/pkg/services/m365/api/drive_pager.go
```diff
@@ -204,20 +205,20 @@ func (c Drives) EnumerateDriveItemsDelta(
...
 	deltaPager := c.newDriveItemDeltaPager(
 		driveID,
 		prevDeltaLink,
 		cc)
 
-	npr := &nextPageResults[models.DriveItemable]{
-		pages: make(chan nextPage[models.DriveItemable]),
+	npr := &pagers.NextPageResults[models.DriveItemable]{
+		Pages: make(chan pagers.NextPage[models.DriveItemable]),
 	}
```
ashmrtn, line 216: this is the only place `Pages` is accessed so it seems reasonable to just make an initializer function for the struct so we can keep the field unexported

ashmrtn (approved):
LGTM though would really like to see fields switched back to being unexported

src/pkg/services/m365/api/pagers/pagers.go
ryanfkeepers: done; was autopiloting a bit.  thanks for the check.

ashmrtn: Please be more attentive going forward. No matter how the code came about, as the code author and author of the PR you're responsible for vouching for the general correctness and quality of the code. Expecting reviewers to check for simple quality things places a _huge_ burden in terms of review time and effort on them if you also want actionable and correct review comments. Just figuring out what could and couldn't be exported in these structs required me to either remember all the usages of these variables or clone the branch myself and search through the code using other tools

The bottom line is if the team at large keeps doing this either the review quality will go down (non-actionable or incorrect comments), the code quality will go down (more difficult to make changes going forward), or the reviewers are going to become overwhelmed

Issue #4540, opened 2023-10-25
[Bug]: SharePoint backup failure if a site has an empty display name


ashmrtn:
### What happened?

Attempting to backup any SharePoint site in a tenant when at least one site has an empty display name results in `Error: Failed to retrieve M365 sites: retrieving sites: validating site: missing site display name`

This is due to a validation [check](https://github.com/alcionai/corso/blob/cb6c56783ba3944b55160acb3d285aa245266a56/src/pkg/services/m365/api/sites.go#L282) on site display names. We call this code when [populating](https://github.com/alcionai/corso/blob/cb6c56783ba3944b55160acb3d285aa245266a56/src/cli/backup/sharepoint.go#L162) the map of site ID -> site URL for all sites in the tenant for name resolution on CLI input

### Corso Version?

Corso v0.14.2

### Where are you running Corso?

N/A

ashmrtn:
It may be possible to remove the display name check for now. However prior to releasing that we need to ensure upstream consumers of the Corso API can handle potentially empty display names

ashmrtn:
Long term we should probably consider refactoring this code so it won't fail backups not related to the erroring site(s)

ashmrtn:
When following the below, note that it may take a few minutes for Microsoft graph to start returning the updated site information. Graph may also get into a situation where it keeps returning the site info without the display name even after you re-add it for some amount of time. I got one test site into that state and just ended up deleting the entire site because I didn't want to wait for Microsoft Graph state to catch up

to create a site with no display name, first create a site with some name. I followed the directions [here](https://support.microsoft.com/en-us/office/create-a-site-with-sharepoint-online-9fdfbbc4-1675-4d40-8df4-93b1340d5f34) and made a "communication site"

Next, open the site settings using the gear icon in the top right and click "Site information"
![image](https://github.com/alcionai/corso/assets/3891298/38ac3304-1a29-405b-8153-0df772d49361)

In the updated side panel, click "View all site settings" to go to a different page for the site
![image](https://github.com/alcionai/corso/assets/3891298/7f719acd-fe80-49e4-b7f6-a0489f238e55)

On the new page hit the "Title, description, and logo" link at the top of the right hand column
![image](https://github.com/alcionai/corso/assets/3891298/608b8a67-43e4-42b9-9a07-a3933e3c86c9)

On the page it takes you to, clear the "Title" field and then hit save
[...]

PR #4548, opened 2023-10-26
Refactor site exclusion logic

ashmrtn:
Pull the logic for excluding personal and search
sites into a separate function instead of overloading
the verify function for a site

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4540

merge after:
* #4547

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/sites.go
+++ b/src/pkg/services/m365/api/sites.go
@@ -11,7 +11,6 @@ import (
 	msgraphgocore "github.com/microsoftgraph/msgraph-sdk-go-core"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/microsoftgraph/msgraph-sdk-go/sites"
-	"github.com/pkg/errors"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/graph"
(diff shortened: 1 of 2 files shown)


src/pkg/services/m365/api/sites.go
```diff
@@ -255,9 +253,26 @@ func (c Sites) GetDefaultDrive(
...
-var ErrKnownSkippableCase = clues.New("case is known and skippable")
+const personalSitePath = "sharepoint.com/personal/"
 
-const PersonalSitePath = "sharepoint.com/personal/"
+func skippableSite(item models.Siteable) bool {
+	wURL := ptr.Val(item.GetWebUrl())
+	name := ptr.Val(item.GetDisplayName())
+
+	// personal (ie: oneDrive) sites have to be filtered out server-side.
+	if strings.Contains(wURL, personalSitePath) {
```
meain, line 263: Using `strings.Contains` makes me feel a bit uneasy. Should we parse out the URL and check against specific components. _I guess this is technically not a part of this PR, but thought I would mention._

meain (approved):

src/pkg/services/m365/api/sites.go
ashmrtn: hmmm that's a good point. Let me make a ticket for that and then we can follow up on it. I'm not sure if there's other locations we're doing similar things

ashmrtn: filed #4555

ryanfkeepers: As long as we're poking at this, could one of y'all take another stab at getting the query filter to work?  Doing this server-side in any capacity is a smell.  I don't know what I was missing with the original query filter, but it *should* work (according to postman).  Just wasn't giving me anything in go.

ryanfkeepers (approved):

src/pkg/services/m365/api/sites.go
ashmrtn: this was sort of the extent of the changes I was planning on making since it's mostly cleanup and there's other things I'm slated to work on right now. Could you please make a ticket about the query filtering if it doesn't already exist? That way we won't lose track of it

ryanfkeepers: no worries

PR #4554, opened 2023-10-26
Fix export stalling when run with progressbars

meain:
Exports were stalling when run with progressbars. Most of our tests(except for longevity tests) did not catch this as they were running without progressbars. This should fix that.

First introduced in 040257f8be95f924fce3adcb1159422a1f22b45c

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/export/export.go
+++ b/src/cli/export/export.go
@@ -104,9 +104,15 @@ func runExport(
 	// It would be better to give a progressbar than a spinner, but we
 	// have any way of knowing how many files are available as of now.
 	diskWriteComplete := observe.MessageWithCompletion(ctx, "Writing data to disk")
-	defer close(diskWriteComplete)
 
 	err = export.ConsumeExportCollections(ctx, exportLocation, expColl, eo.Errors)
+
+	// The progressbar has to be closed before we move on as the Infof
+	// below flushes progressbar to prevent clobbering the output and
+	// that causes the entire export operation to stall indefinitely.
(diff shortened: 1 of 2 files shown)


src/cli/export/export.go
```diff
@@ -104,13 +104,14 @@ func runExport(
...
 	// have any way of knowing how many files are available as of now.
 	diskWriteComplete := observe.MessageWithCompletion(ctx, "Writing data to disk")
-	defer close(diskWriteComplete)
 
 	err = export.ConsumeExportCollections(ctx, exportLocation, expColl, eo.Errors)
 	if err != nil {
 		return Only(ctx, err)
 	}
 
+	close(diskWriteComplete)
```
ashmrtn, line 113: could you please do 2 things for this:
1. move this up to just after the `err = export.ConsumeExportCollections` so it's between that and the error check. That ensures the progress bar is always closed even if we failed the export
2. please add a comment saying why we need this. It looks like the real reason is that `Infof` and friends [attempt to flush the progress bars](https://github.com/alcionai/corso/blob/8102523dc62c001b301cd2ab4e799f86146ab1a0/src/cli/print/print.go#L151) prior to printing so we don't clobber output

Issue #4546, opened 2023-10-26
Groups exports have unweildy message content


ryanfkeepers:
## Issue

Groups exports display messsage content as plaintext, even when the content is html type or contains markup tags like attachment references.  This creates a poor user experience.

## Acceptance Criteria
- [ ] Swap attachment tags for the text `[attachment:{{attachment.Name}}]`
- [ ] Use a html cleaner to reduce html to plain text messages.

PR #4556, opened 2023-10-26
make channel message html human readable

ryanfkeepers:
Adds two-step processing to the html previews for channel messages and replies.  First, all inline attachments are replaced with the string `[attachment:name]`.
Second, remaining html is stripped out, leaving only plaintext.

This transformation is applied to both the exported content and the preview content in details.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4546

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/channels_test.go
+++ b/src/pkg/services/m365/api/channels_test.go
@@ -679,7 +679,7 @@ func (suite *ChannelsAPIUnitSuite) TestStripChatMessageContent() {
 			msg.SetAttachments(test.attachments)
 
 			// not testing len; it's effectively covered by the content assertion
-			result, _, err := StripChatMessageContent(msg)
+			result, _, err := stripChatMessageHTML(msg)
 			assert.Equal(t, test.expect, result)
 			test.expectErr(t, err, clues.ToCore(err))
 		})
--- a/src/pkg/services/m365/api/channels_test.go
+++ b/src/pkg/services/m365/api/channels_test.go
@@ -1,9 +1,11 @@
 package api
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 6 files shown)


ntolia:
I would expect Exports to match what was captured. Preview is a different case.

ryanfkeepers:
> I would expect Exports to match what was captured. Preview is a different case.

Hmm, that's a good topic for discussion.  We waffle back and forth between whether exports should be exact or human readable.  For example: we're already slicing out most of the stored information in a standard export to create a more digestable experience (there's a flag for providing raw output on top).  Are we better off here going for full human consumption, or partial?

ntolia:
> > I would expect Exports to match what was captured. Preview is a different case.
> 
> Hmm, that's a good topic for discussion. We waffle back and forth between whether exports should be exact or human readable. For example: we're already slicing out most of the stored information in a standard export to create a more digestable experience (there's a flag for providing raw output on top). Are we better off here going for full human consumption, or partial?

Folks might pull these messages into other third-party systems. I think export should try to be as true to the original as possible/feasible for broader use. In particular, stripping out HTML, etc. is definitely not what someone would want but this applies to other components of the conversation too.

gmatev:
> > > I would expect Exports to match what was captured. Preview is a different case.
> > 
> > 
> > Hmm, that's a good topic for discussion. We waffle back and forth between whether exports should be exact or human readable. For example: we're already slicing out most of the stored information in a standard export to create a more digestable experience (there's a flag for providing raw output on top). Are we better off here going for full human consumption, or partial?
> 
> Folks might pull these messages into other third-party systems. I think export should try to be as true to the original as possible/feasible for broader use. In particular, stripping out HTML, etc. is definitely not what someone would want but this applies to other components of the conversation too.

@ryanfkeepers I think you are reading this too literally. It is not about no changes to the payload of the graph call. It is more about the `content` and I guess `subject` fields. If there are other fields returned in the graph call that are not that relevant for the content of the message it should be ok to sanitize some of these to simplify the JSON export. 

If you prefer we can discuss specific examples.

ryanfkeepers:
@gmatev @ntolia Thanks for the reviews.  All changes are now in place.

src/pkg/services/m365/api/channels_test.go
```diff
@@ -565,3 +567,121 @@ func (suite *ChannelsAPIUnitSuite) TestChannelMessageInfo() {
...
+		{
+			name:        "with html",
+			content:     "<body>body<div>in the div</div>end</body>",
+			attachments: attachments,
+			expect:      "body\nin the div\nend",
+			expectErr:   assert.NoError,
+		},
+		{
+			name:        "with html and attachments",
+			content:     "<body>body<div>" + attachML("id1") + attachML("id2") + attachML("idX") + "</div>end</body>",
```
vkamra, line 665: @ryanfkeepers - not a blocker here but could we add a couple of test cases with "real content" obtained from Graph API?

vkamra (approved):
LGTM. Approving to unblock merge. Request we do add a couple of test cases with "real content" from the Graph API while this is merging to increase confidence.

PR #4561, opened 2023-10-27
Groups version bump

ashmrtn:
Bump the backup version and force a full backup if
there's a backup for teams/groups that has base(s)
from an older version of corso

This will avoid propagating older details formats
forward. Those formats don't have all the data
newer formats do

This is mostly a stop-gap, a more robust solution
can be added later

Manually tested that it forces a full backup

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #4569

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/backup_bases.go
+++ b/src/internal/kopia/backup_bases.go
@@ -46,6 +46,9 @@ type BackupBases interface {
 	// MinBackupVersion returns the lowest version of all merge backups in the
 	// BackupBases.
 	MinBackupVersion() int
+	// MinAssisttVersion returns the lowest version of all assist backups in the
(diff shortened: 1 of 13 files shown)


vkamra (approved):

ashmrtn:
ok, thought up one more case for this: we need to add a check on the version of the assist bases and drop those separately since it's possible to have assist bases without a merge base. The other half of the reason for this check is that assist bases also source backup details and we don't want to pull forward older backup details versions for teams

PR #4568, opened 2023-10-27
Add tests on assist bases for drive-backed services for groups bump

ashmrtn:
We need to make sure we don't use old assist bases
for the groups bump this adds automated tests for
that

The OneDrive test is currently disabled because
it's not making an assist base as expected. This
can be debugged later

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4569

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/test/group_test.go
+++ b/src/internal/operations/test/group_test.go
@@ -186,6 +186,15 @@ func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_groups9VersionBum
 		"items written")
 }
 
+func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_assistBackup_groups9VersionBump() {
+	sel := selectors.NewGroupsBackup([]string{suite.its.group.ID})
+	sel.Include(
+		selTD.GroupsBackupLibraryFolderScope(sel),
+		selTD.GroupsBackupChannelScope(sel))
+
+	runDriveAssistBaseGroupsUpdate(suite, sel.Selector, false)
+}
+
 func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic() {
 	t := suite.T()
 
--- a/src/internal/operations/test/group_test.go
+++ b/src/internal/operations/test/group_test.go
@@ -186,7 +186,7 @@ func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_groups9VersionBum
 		"items written")
 }
 
-func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_assistBackup_groups9VersionBump() {
+func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsVersion9AssistBases() {
 	sel := selectors.NewGroupsBackup([]string{suite.its.group.ID})
 	sel.Include(
 		selTD.GroupsBackupLibraryFolderScope(sel),
--- a/src/internal/operations/test/onedrive_test.go
+++ b/src/internal/operations/test/onedrive_test.go
@@ -3,6 +3,8 @@ package test_test
 import (
 	"context"
 	"fmt"
+	"io"
+	"sync/atomic"
 	"testing"
 
 	"github.com/alcionai/clues"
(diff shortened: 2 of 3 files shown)


src/internal/operations/test/onedrive_test.go
```diff
@@ -212,6 +215,13 @@ func (suite *OneDriveBackupIntgSuite) TestBackup_Run_oneDriveBasic_groups9Versio
 		"items written")
 }
 
+//func (suite *OneDriveBackupIntgSuite) TestBackup_Run_oneDriveBasic_assistBackup_groups9VersionBump() {
```
ryanfkeepers, line 218: can this get removed?

src/internal/operations/test/group_test.go
```diff
@@ -186,6 +186,15 @@ func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_groups9VersionBum
 		"items written")
 }
 
+func (suite *GroupsBackupIntgSuite) TestBackup_Run_groupsBasic_assistBackup_groups9VersionBump() {
```
ryanfkeepers, line 189: Nit: clarity.  Also, `basic` is meant to mean `just a plain backup and nothing else`, which isn't the case here.
```suggestion
func (suite *GroupsBackupIntgSuite) TestBackup_Run_ TestBackup_Run_groupsVersion9AssistBases() {
```

src/internal/operations/test/sharepoint_test.go
```diff
@@ -150,6 +150,13 @@ func (suite *SharePointBackupIntgSuite) TestBackup_Run_sharePointBasic_groups9Ve
 		"items written")
 }
 
+func (suite *SharePointBackupIntgSuite) TestBackup_Run_sharePointBasic_assistBackup_groups9VersionBump() {
```
ryanfkeepers, line 153:
```suggestion
func (suite *SharePointBackupIntgSuite) TestBackup_Run_sharePointVersion9AssistBases() {
```

ryanfkeepers, line 153: Question: why is sharepoint getting a special test case here?  It shouldn't be affected by the bump.

Edit: now that I read the full test func, it seems like this is the negative case.  Do we want a negative case test for all services?

src/internal/operations/test/onedrive_test.go
```diff
@@ -911,6 +921,179 @@ func runDriveIncrementalTest(
...
 
+var (
+	_ io.ReadCloser                    = &singleFileFailExtension{}
+	_ extensions.CreateItemExtensioner = &createSingleFileFailExtension{}
+)
+
+// singleFileFailExtension fails the first read on a file being uploaded during
+// a snapshot. Only one file is failed during the snapshot even if it the
+// snapshot contains multiple files.
+type singleFileFailExtension struct {
```
ryanfkeepers, line 932: Nit: naming simplicity
```suggestion
type failFirstRead struct {
```

src/internal/operations/test/onedrive_test.go
```diff
@@ -911,6 +921,179 @@ func runDriveIncrementalTest(
...
+	firstFile *atomic.Bool
```
ryanfkeepers, line 933: seems like a case for [sync.Once](https://pkg.go.dev/sync#Once)

ryanfkeepers (approved):
just nits

src/internal/operations/test/onedrive_test.go
ashmrtn: sure, though we're going to want to add it back at some point since we should be testing it. The only reason we're not right now is there's something with the OneDrive implementation that is making it not make an assist base

src/internal/operations/test/sharepoint_test.go
ashmrtn: This attempts to add a negative test case for as many services as possible. With the current implementation, it turns into "as many services as support extensions" because we need to inject a recoverable error in the backup. Since OneDrive is misbehaving for other reasons that negative test case is commented out leaving us with tests running for SharePoint and groups/teams

src/internal/operations/test/onedrive_test.go
ashmrtn: not sure if that'll make things any easier since the return from the `Read` function will be complicated. `Do` doesn't return anything so I have no way to tell if it ran or not and thus don't know if I should return an error or not in `Read`

PR #4570, opened 2023-10-27
boilerplate introduction of the conversations api

ryanfkeepers:
adds basic get requests for the conversations api set to the api package.  Plust some other minor change for formatting, correctness, or necessity.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4536 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/operations/test/helper_test.go
+++ b/src/internal/operations/test/helper_test.go
@@ -738,6 +738,7 @@ func verifyExtensionData(
 	case path.OneDriveService:
 		detailsSize = itemInfo.OneDrive.Size
 	case path.GroupsService:
+		// FIXME: needs update for message.
 		detailsSize = itemInfo.Groups.Size
 	default:
 		assert.Fail(t, "unrecognized data type")
--- a/src/pkg/backup/details/groups.go
+++ b/src/pkg/backup/details/groups.go
@@ -42,8 +42,11 @@ type GroupsInfo struct {
 	Modified time.Time `json:"modified,omitempty"`
 
 	// Channels Specific
-	Message   ChannelMessageInfo `json:"message"`
-	LastReply ChannelMessageInfo `json:"lastReply"`
+	Message   ChannelMessageInfo `json:"message,omitempty"`
+	LastReply ChannelMessageInfo `json:"lastReply,omitempty"`
+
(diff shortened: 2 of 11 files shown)


src/pkg/backup/details/groups.go
```diff
@@ -42,8 +42,11 @@ type GroupsInfo struct {
 	Modified time.Time `json:"modified,omitempty"`
 
 	// Channels Specific
-	Message   ChannelMessageInfo `json:"message"`
-	LastReply ChannelMessageInfo `json:"lastReply"`
+	Message   ChannelMessageInfo `json:"message,omitempty"`
+	LastReply ChannelMessageInfo `json:"lastReply,omitempty"`
+
+	// Conversations Specific
+	Post ConversationPostInfo `json:"post,omitempty"`
```
meain, line 49: 1) Can't we just reuse `ChannelMessageInfo` (maybe rename it)
1) Do we even need a separate field? Can we just fold into `Message` with `ItemType` deciding the type?

src/internal/operations/test/helper_test.go
```diff
@@ -728,6 +728,7 @@ func verifyExtensionData(
 	case path.OneDriveService:
 		detailsSize = itemInfo.OneDrive.Size
 	case path.GroupsService:
+		// FIXME: needs update for message.
```
meain, line 731: Should we create an issue to track this so that this does not get missed.

meain (approved):
Few thoughts, but approving to unblock.

src/pkg/backup/details/groups.go
ryanfkeepers: I don't see the value in it.  A little bit of duplication here won't hurt us, and it'll help minimize confusion for SDK consumers.  Per itemType, that's a largely internal reference.  End consumers don't have a good way of knowing which available value do and do not get populated based on type, nor are we providing strict guarantees of stability there.

src/internal/operations/test/helper_test.go
ryanfkeepers: It gets handled later.

PR #4585, opened 2023-10-31
adds conversations to the CLI

ryanfkeepers:
adds conversations support to groups backup commands

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4536

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/groups_e2e_test.go
+++ b/src/cli/backup/groups_e2e_test.go
@@ -28,11 +28,6 @@ import (
 	storeTD "github.com/alcionai/corso/src/pkg/storage/testdata"
 )
 
-var (
-	channelMessages = path.ChannelMessagesCategory
-	libraries       = path.LibrariesCategory
-)
-
 // ---------------------------------------------------------------------------
 // tests that require no existing backups
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


meain (approved):

src/cli/backup/groups_e2e_test.go
```diff
@@ -293,19 +296,27 @@ func (suite *PreparedBackupGroupsE2ESuite) SetupSuite() {
...
 		)
 
 		switch set {
-		case channelMessages:
+		case path.ChannelMessagesCategory:
 			scopes = selTD.GroupsBackupChannelScope(sel)
 
-		case libraries:
+		case path.ConversationPostsCategory:
+			scopes = selTD.GroupsBackupChannelScope(sel)
```
pandeyabs, line 317: Do we need a scope for conversations?

src/cli/utils/groups.go
```diff
@@ -172,11 +179,13 @@ func AddGroupsFilter(
...
-		llf, lli    = len(opts.ListFolder), len(opts.ListItem)
-		lpf, lpi    = len(opts.PageFolder), len(opts.Page)
-		lg, lch, lm = len(opts.Groups), len(opts.Channels), len(opts.Messages)
+		groups     = opts.Groups
+		lfp, lfn   = len(opts.FolderPath), len(opts.FileName)
+		llf, lli   = len(opts.ListFolder), len(opts.ListItem)
+		lpf, lpi   = len(opts.PageFolder), len(opts.Page)
+		lg         = len(opts.Groups)
+		lch, lm    = len(opts.Channels), len(opts.Messages)
+		lcon, lpst = len(opts.Conversations), len(opts.Posts)
```
pandeyabs, line 188: These acronyms are hard to follow through this func. This is one of the cases where I think we should just use `len(opts.Foo)` wherever they are being used. Otherwise it's easy to miss a letter and cause bugs. 

Also, the test (`TestIncludeGroupsRestoreDataSelectors`) only checks for include length. We should improve it to match against expected selectors.

pandeyabs (approved):
Looks good. Left a comment about var naming in one place and test improvements.

src/cli/backup/groups_e2e_test.go
ryanfkeepers: Ooh, good catch!

src/cli/utils/groups.go
ryanfkeepers: Yeah, they're getting out of hand, aren't they?  I'll clean that up.

PR #4593, opened 2023-10-31
Exclude preview backups when looking for bases

ashmrtn:
Add a comment and test ensuring preview backups are excluded during
base selection as expected. It's still not possible to make a
backup tagged with the preview tag right now

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/base_finder.go
+++ b/src/internal/kopia/base_finder.go
@@ -238,6 +238,8 @@ func (b *baseFinder) findBasesInSet(
 			}
 
 		default:
+			// Preview backups automatically fall into this group so there's no need
+			// to add a case to skip them.
 			logger.Ctx(ictx).Infow(
 				"skipping backup with empty or invalid type for incremental backups",
 				"backup_type", bup.Type())
--- a/src/internal/kopia/base_finder.go
+++ b/src/internal/kopia/base_finder.go
@@ -237,9 +237,11 @@ func (b *baseFinder) findBasesInSet(
 				Reasons:          []identity.Reasoner{reason},
 			}
 
+		case model.PreviewBackup:
+			// Preview backups are listed here for clarity though they use the same
+			// handling as the default case because they can't be used as bases.
+			fallthrough
 		default:
-			// Preview backups automatically fall into this group so there's no need
-			// to add a case to skip them.
 			logger.Ctx(ictx).Infow(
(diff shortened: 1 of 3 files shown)


src/internal/kopia/base_finder.go
```diff
@@ -238,6 +238,8 @@ func (b *baseFinder) findBasesInSet(
 			}
 
 		default:
+			// Preview backups automatically fall into this group so there's no need
+			// to add a case to skip them.
```
ryanfkeepers, line 242: I'd feel a little more comfortable with an explicit fallthrough as opposed to a comment.  The clarity may help maintenance/bug avoidance down the road.
```suggestion
                case model.PreviewBackup:
                    fallthrough
		default:
			// Preview backups automatically fall into this group so there's no need
			// to add a case to skip them.
```

ryanfkeepers (approved):

src/internal/model/model.go
```diff
@@ -53,6 +53,16 @@ const (
...
 	// the manifest search completes.
 	MergeBackup = "merge-backup"
+	// PreviewBackup denotes that this backup contains a subset of information for
+	// the protected resource. PreviewBackups are used to demonstrate value but
+	// are not safe to use as merge bases for incremental backups. It's possible
+	// they could be used as assist bases since the only difference from a regular
+	// backup is the amount of data they contain.
+	//
+	// This tag should only be used for finding backups during a
+	// manifest search. It shouldn't be used to differentiate between backups once
```
pandeyabs, line 63: > It shouldn't be used to differentiate between backups once the manifest search completes.
Could you explain this a bit?

pandeyabs (approved):

src/internal/model/model.go
ashmrtn: sure! I'll update the comment in the code as well to make it more clear

Mostly I'm trying to say that it's alright to use these tags during `FindBases()` but we shouldn't use them after that. Outside of `FindBases()` we should really use something like the `BackupBases` struct's functions or something to differentiate how a previous backup will be used in the current backup

PR #4607, opened 2023-11-03
Create a ContainerResolver that returns items in priority order

ashmrtn:
Wrapper for existing graph.ContainerResolver structs that returns items
in priority order in the Items() function. Also has logic to filter out
containers if desired

This can help with things like preview backups where we only want to
backup a subset of items because it allows us to ensure we'll see the
containers we care about most first

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/collection/exchange/container_resolver.go
```diff
@@ -415,3 +416,125 @@ func (cr *containerResolver) populatePaths(
...
+		ictx := clues.Add(ctx, "container_id", id)
+
+		c, err := getter.GetContainerByID(ctx, userID, id)
+		if err != nil {
+			return nil, clues.Wrap(err, "getting ranked container").WithClues(ictx)
+		}
+
+		gotID := ptr.Val(c.GetId())
+		if len(gotID) == 0 {
+			return nil, clues.New("received ranked include container with empty ID").
```
pandeyabs, line 475: Curious, does this need to be a failure? Am I correct in assuming that if we return a failure here, it would result in a preview backup failure?

src/internal/m365/collection/exchange/container_resolver.go
```diff
@@ -411,3 +416,125 @@ func (cr *containerResolver) populatePaths(
...
 	return lastErr
 }
+
+// ---------------------------------------------------------------------------
+// rankedContainerResolver
+// ---------------------------------------------------------------------------
+
+type rankedContainerResolver struct {
+	graph.ContainerResolver
+	// resolvedIncluded is the ordered list of actual container IDs to add to the
```
pandeyabs, line 426:
```suggestion
	// resolvedInclude is the ordered list of resolved container IDs to add to the
```

src/internal/m365/collection/exchange/container_resolver.go
```diff
@@ -411,3 +416,125 @@ func (cr *containerResolver) populatePaths(
...
+			continue
+		}
+
+		res = append(res, c)
+	}
+
+	return res
+}
+
+func (cr *rankedContainerResolver) ItemByID(id string) graph.CachedContainer {
```
pandeyabs, line 534: Nitpicking on a corner case. If there is a possible overlap between include & exclude lists, `Items()` code above would return the overlapping container as part of `res`. However, `ItemByID()` will return `nil`. 

Should we have the same behavior across both APIs? Another thing we can do is to catch overlaps in caller's unit or integration test.

src/internal/m365/collection/exchange/container_resolver_test.go
```diff
@@ -53,6 +55,316 @@ func strPtr(s string) *string {
...
+		{
+			name:         "ReturnsRankedItems SomeExcluded",
+			includes:     []string{id2},
+			excludes:     []string{id3, id1},
+			expectPrefix: []string{id2},
+			expectErr:    assert.NoError,
+		},
+		{
+			name:                 "ReturnsRankedItems SomeIncludesNotInBase",
+			includes:             []string{id2, id1, idNotInBase},
```
pandeyabs, line 293: Curious, is it realistically possible to have `idNotInBase` in `includes` but not in base?

pandeyabs (commented):
In general looks good! Have left a few comments & questions, can approve after that.
Let's wait for @ryanfkeepers's review as well.

ryanfkeepers:
@ashmrtn I can get to this review as soon as the conflicts are fixed up.  Thanks!

src/internal/m365/collection/exchange/container_resolver.go
ryanfkeepers: It's a good safety measure.  While graph api gives us good guarantees about the presence of an id, suppose we one day replace the inner item with an interface-compliant internal struct, or accidentally omit the ID from the item properties by selection.  This ensures we'll, at minimum, fail some integration tests.

ryanfkeepers: nit: the error here is a little verbose
```suggestion
			return nil, clues.New("ranked include container missing ID").
```

src/internal/m365/collection/exchange/container_resolver.go
```diff
@@ -411,3 +416,125 @@ func (cr *containerResolver) populatePaths(
...
+		gotID := ptr.Val(c.GetId())
+		if len(gotID) == 0 {
+			return nil, clues.New("received ranked include container with empty ID").
+				WithClues(ictx)
+		}
+
+		cr.resolvedInclude = append(cr.resolvedInclude, gotID)
+	}
+
+	for _, id := range excludeIDs {
```
ryanfkeepers, line 482: might want a comment explaining that we need to get the excluded containers in order to get their id for comparison.  Otherwise its not obvious why we're getting excluded items.

ryanfkeepers (approved):

src/internal/m365/collection/exchange/container_resolver.go
pandeyabs: > This ensures we'll, at minimum, fail some integration tests.

Understood. My concern is around user facing errors, especially for preview backups. Wondering if it's better to show user some less data(say one less folder) vs showing them a failed backup. We can always catch errors using integ tests but treat them as ignorable errors in prod.

I understand failures would be mostly unlikely here, but wanted to have a UX discussion. Doesn't need to be addressed in this PR.

PR #4612, opened 2023-11-06
Handle folder being deleted and recreated in between a backup

meain:
Previously if a folder with the same name was deleted and recreated, we would error out. This fixes that by ensuring that only the final collection makes it into the final set.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/collection/drive/collections.go
```diff
@@ -856,6 +863,22 @@ func (c *Collections) processItem(
...
+		// This check is to ensure that if a folder was deleted and
+		// recreated multiple times between a backup, we only use the
+		// final one.
+		pid, found := seenFolders.IDOf(collectionPath.String())
+
+		recreated := found && pid != itemID
+		if recreated {
+			prevPath = nil
+			c.NumContainers--
+			c.NumItems--
```
ryanfkeepers, line 875: This feels a bit hazardous, but I'm not sure the alternatives are less clunky.  Thoughts?

src/internal/m365/collection/drive/collections.go
```diff
@@ -856,6 +863,22 @@ func (c *Collections) processItem(
...
+			delete(c.CollectionMap[driveID], pid)
+			delete(newPrevPaths, pid)
+		} else {
+			seenFolders.Add(itemID, collectionPath.String())
```
ryanfkeepers, line 879: The comment says we're only keeping the final creation of the collecitonPath.  Two concerns here with that:
1. This entry only tracks the first instance of the collectionPath.  Feels like it should always update to the last one.
2. An IDName cacher isn't really the right implementation for this, since it's build with the assumption that all additions are discreet, and may have unexpectd side effects for partially overlapping additions.  A map[collPath]ItemID should work here, right?

ryanfkeepers (approved):
some nits; approving to unblock

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -220,6 +220,30 @@ func (suite *OneDriveCollectionsUnitSuite) TestPopulateDriveCollections() {
 			expectedExcludes:         map[string]struct{}{},
 			expectedTopLevelPackages: map[string]struct{}{},
 		},
+		{
+			name: "Single Folder created twice", // deleted a created with same name in between a backup
+			items: []models.DriveItemable{
+				driveRootItem("root"),
+				driveItem("id1", "folder", testBaseDrivePath, "root", false, true, false),
```
ashmrtn, line 227: do we have any indication that this is what's actually being returned from graph? It's nice to be able to handle it if we absolutely need to, but I'm also a bit worried that it could somehow hide other things if we're not careful

ashmrtn (commented):
I'm still a bit confused as to what is actually being returned by the delta token and why some of the [code](https://github.com/alcionai/corso/blob/187edc41453e1044f4dc804fa5ef273fedbb669d/src/internal/m365/collection/drive/collections.go#L582)  for handling deleted items isn't handling this. The comments in that code make it seem like it should be able to deal with it. I feel it would be good to look into that a bit more so we have a clearer idea of what we're actually dealing with

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: +1 here.  Seems reasonable that this is the response we get, but worth validating.

ashmrtn: discussed out of band, does indeed seem like what we're getting back

PR #4613, opened 2023-11-06
Enable user to perform multiple backups by hash-based repository config file name

### LINKED ISSUE #4443: Allow kopia config file path overrides
### What happened?

Currently we place the [kopia config file](https://github.com/kopia/kopia/blob/45c1f0950f5fd983b38b9282f08a5d52f4cdfefe/site/content/docs/Reference/Command-Line/_index.md?plain=1#L68) at `/tmp/repository.config`. This file maintains repo connection params. This file is shared between all corso instances which might be running in parallel on the system, unless user is running the instances in separate containers.

This shared file model can be a cause of bugs. For e.g. consider a case where a user is doing multiple concurrent backups to different repositories. The kopia config file will only reflect the last written repo details. 

We've also seen reports from users where corso instances error out simultaneously writing to the config file. 
`connecting to repo: unable to write config file: error writing file: cannot replace "/tmp/repository.config" with tempfile "/tmp/repository.config2296101656": rename /tmp/repository.config2296101656 /tmp/repository.config: operation not permitted`

**Proposal:**
1. Allow corso cli users to override path via a `--kopia-config-file` flag or similar, or an environment var. 
2. Change the default location of the config file. Make it per $USER. Just like the corso config file. 

### Corso Version?

Corso v0.13.0

NA

### Relevant log output

_No response_

HiteshRepo:
The goal is to enable user/client to be able to perform multiple backups in a concurrent manner irrespective of storage provider (s3/filesystem) for different tenants (anything that differentiates one account from another).
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* fixes #4443 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+		if !areEqual {
 			err := clues.New("value of " + k + " (" + v + ") does not match corso configuration value (" + vv + ")")
 			return clues.Stack(ErrVerifyingConfigStorage, err)
 		}
 	}
 
 	return nil
 }
+
+func ArePathsEquivalent(path1, path2 string) (bool, error) {
```
ryanfkeepers, line 193: this function has no error case.  The return value should only be `bool`

ryanfkeepers (commented):
Handful of nit.  At broad things look good.

Biggest issue is the size of the PR and its complexity.  We strive to keep change requests small and contained to logical sets whenever possible.  This helps minimize context switching and cognitive load for reviewers, and tends to improve the quality of reviews overall.  In this case the PR could ideally be split into four smaller changes:

1. Additions to the Path package (IsValidPath() and ArePathsEquivalent()).
2. Addition of hash generators to the Storage package.
3. Refactors to the config package to relocate the kopia config.
4. Updates to the kopia wrapper and other places that tie the whole thing together.

Personally, I'd like to see the change split up along those lines when done drafting things.  But since you've already got it all written out, I won't block things there, either.

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+
+	if isDir {
+		err := os.Mkdir(path, 0644)
+		if err == nil {
+			os.Remove(path)
+			return true
+		}
+
+		if os.IsPermission(err) {
+			logger.Ctx(context.Background()).Info("directory could have been created but got permission error")
```
ryanfkeepers, line 218: avoid using context.Background() whenver possible.  Can we pass a ctx into this func?
```suggestion
			logger.Ctx(ctx).Info("permission restrictions prevented directory creation")
```

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+
+func IsValidPath(path string) bool {
+	if _, err := os.Stat(path); err == nil {
+		return true
+	}
+
+	isDir := strings.TrimSpace(filepath.Ext(path)) == ""
+
+	if isDir {
+		err := os.Mkdir(path, 0644)
```
ryanfkeepers, line 211: I don't think `IsValid` should be creating directories as a side effect; that should be owned by a different function.  That, or, it should be stated as the goal by the func name, eg: `FindOrCreateDirectory()`

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+		if err == nil {
+			os.Remove(path)
+			return true
+		}
+
+		if os.IsPermission(err) {
+			logger.Ctx(context.Background()).Info("directory could have been created but got permission error")
+			return true
```
ryanfkeepers, line 219: Why are we returning `true` (ie: valid) on a permission issue?

src/pkg/storage/storage.go
ryanfkeepers, line 218: Preferred handling is `func > err != nil > the rest`.    IE:

```suggestion
		err := os.Mkdir(path, 0644)
		if os.IsPermission(err) {
			logger.Ctx(context.Background()).Info("directory could have been created but got permission error")
		}
		
		if err != nil {
		    return false, clues.Stack(err).WithClues(ctx)
		}
		
		os.Remove(path)
		
		return true
```

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+	if _, err := os.Stat(path); err == nil {
+		return true
+	}
+
+	isDir := strings.TrimSpace(filepath.Ext(path)) == ""
+
+	if isDir {
+		err := os.Mkdir(path, 0644)
+		if err == nil {
+			os.Remove(path)
```
ryanfkeepers, line 213: why are we removing the paths we just created?

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+		}
+
+		if os.IsPermission(err) {
+			logger.Ctx(context.Background()).Info("directory could have been created but got permission error")
+			return true
+		}
+	} else {
+		dir := filepath.Dir(path)
+		err := os.Mkdir(dir, 0644)
+		if err != nil {
```
ryanfkeepers, line 224:
```suggestion
		if err := os.Mkdir(dir, 0644); err != nil {
```

src/pkg/storage/storage.go
```diff
@@ -166,13 +171,71 @@ func mustMatchConfig(
...
+			return false
+		}
+		defer os.Remove(dir)
+
+		_, err = os.Create(path)
+		if err == nil {
```
ryanfkeepers, line 230: again, prefer handling errors first, and later handling non-error states without a conditional.

src/cli/config/config.go
```diff
@@ -280,11 +283,15 @@ func getStorageAndAccountWithViper(
 	// possibly read the prior config from a .corso file
 	if readFromFile {
 		if err := vpr.ReadInConfig(); err != nil {
-			if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
+			_, configNotSet := err.(viper.ConfigFileNotFoundError)
```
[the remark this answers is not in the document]
HiteshRepo: sure @ryanfkeepers

src/cli/config/config_test.go
```diff
@@ -157,6 +158,9 @@ func (suite *ConfigSuite) TestWriteReadConfig() {
 		host   = "some-host"
 	)
 
+	ctx := context.Background()
```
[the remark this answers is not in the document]
HiteshRepo: yes I encountered this as a linting error

src/internal/kopia/conn.go
```diff
@@ -230,6 +236,17 @@ func blobStoreByProvider(
 	}
 }
 
+func configHashByProvider(ctx context.Context, s storage.Storage) (string, error) {
```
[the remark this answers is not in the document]
HiteshRepo: This makes sense @ryanfkeepers .
I intended to make the hash out of account and strorage details both. Which is not accomplished here. 
I think the better place would be `pkg/repository/repository.go` as both those details are accessible here. 
Hash would be created and passed to kopia while 'Initialize', 'Connect' and 'UpdatePassword' methods called.
But this will lead to changes in lot of files (mostly test files). Do you see any concerns?

src/pkg/storage/storage.go
HiteshRepo: @ryanfkeepers This is an error.

src/pkg/storage/storage.go
HiteshRepo: @ryanfkeepers The directory and files are being created as an empirical way of testing if they are valid or not. So it is not required to create those paths but rather to check if they can be created and therefore qualifies to be a valid path.

src/cli/config/config_test.go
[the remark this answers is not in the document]
ryanfkeepers: Heh, yeah, figured you saw the linter.  Mostly wanted to make sure you knew the intended usage.

src/internal/kopia/conn.go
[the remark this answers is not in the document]
ryanfkeepers: > I intended to make the hash out of account and strorage details both. Which is not accomplished here.

An alternative suggestion:
* add a Hashing func to the `str` package for consistency.
* have Storage and Account independently own a `Hashed` function (as described in the prior comment with ConfigHash())
* Let repository make a hash out of the concatenated Storage and Account hashes (or just concat the hashes without re-hashing) for the final file name.

This ensures that: 1/ storage and account clearly own their hashing behavior, and can build expectations around the handling there as described in the prior comment, and 2/ the production of a combined hash file in the repository is a decoupled implementation of that behavior.

src/pkg/storage/storage.go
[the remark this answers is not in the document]
ryanfkeepers: Sure, the introduction of these functions is just fine.  But their _location_ should be in the path package, not storage.

src/pkg/storage/storage.go
ryanfkeepers: feels like a unit test should be able to catch this.

src/pkg/storage/storage.go
ryanfkeepers: Hmm, I'm a little dubious about that handling, since it has the potential to leave behind dirty state under certain conditions.  Is there a way to get the same assurances without creation?

Actually, are we already auto-generating config file locations if they don't exist?  If we aren't, can we reduce the assertion to only ensure the specified location is already present?

src/internal/kopia/conn.go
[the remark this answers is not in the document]
HiteshRepo: > This ensures that: 1/ storage and account clearly own their hashing behavior, and can build expectations around the handling there as described in the prior comment, and 2/ the production of a combined hash file in the repository is a decoupled implementation of that behavior.

Yeah this should keep things in their own domains.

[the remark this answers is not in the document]
HiteshRepo: > I'm not sure I grok the context. Can you please give me some specific examples?

if we implement hash generation in `pkg/repository/repository.go` and pass the hashed string as parameter to `Initialize`, `Connect` and `UpdatePassword` methods that would mean the function signature has to change [from `Connect(ctx context.Context, opts repository.Options) error` to `Connect(ctx context.Context, opts repository.Options, hashStr string) error`] and that would lead to change in many places primarily test files. 

For example https://github.com/alcionai/corso/blob/ec1afa8c8489fded4bf074bc719a110ff62287a7/src/internal/kopia/model_store_test.go#L881

src/cli/config/config.go
```diff
@@ -112,26 +117,24 @@ func initWithViper(vpr *viper.Viper, configFP string) error {
 		vpr.AddConfigPath(configDir)
 		vpr.SetConfigType("toml")
 		vpr.SetConfigName(".corso")
+	} else {
+		ext := filepath.Ext(configFP)
+		if len(ext) == 0 {
```
meain, line 122: Do we really wanna enforce this limitation?

src/internal/common/str/str.go
```diff
@@ -90,3 +92,25 @@ func SliceToMap(ss []string) map[string]struct{} {
...
 	return m
 }
+
+func GenerateHash(input []byte, hashLength int) string {
+	h := md5.New()
+	h.Write([]byte(input))
+	hexstr := hex.EncodeToString(h.Sum(nil))
+
+	if hashLength > 0 {
+		return hexstr[0:hashLength]
```
meain, line 102: I'm not the biggest fan of trimming hashes. This increases the probability of collisions(not that it might has real world probability of happening in our case). Plus since this is not something that user will be seeing, I don't think there is any reason to shorten it. If we really need smaller hashes, we can maybe use [crc32](https://pkg.go.dev/hash/crc32) or maybe [adler32](https://pkg.go.dev/hash/adler32).

src/internal/common/str/str.go
```diff
@@ -90,3 +92,25 @@ func SliceToMap(ss []string) map[string]struct{} {
...
+	}
+
+	return hexstr
+}
+
+func Contains(slice []string, item string) bool {
```
meain, line 108: We can use [`slices.Contains`](https://pkg.go.dev/golang.org/x/exp/slices#Contains) instead.

src/internal/kopia/conn.go
```diff
@@ -170,7 +173,8 @@ func (w *conn) Connect(ctx context.Context, opts repository.Options) error {
 		cfg.KopiaCfgDir,
 		bst,
 		cfg.CorsoPassphrase,
-		defaultCompressor)
+		defaultCompressor,
+		hashStr)
```
meain, line 177: nit: this might be best located near to `cfg.KopiaCfgDir`.

src/internal/kopia/conn.go
```diff
@@ -179,6 +183,7 @@ func (w *conn) commonConnect(
 	configDir string,
 	bst blob.Storage,
 	password, compressor string,
+	hashStr string,
```
meain, line 186: same nit: ideally near to `configDir`.

src/internal/kopia/conn_test.go
```diff
@@ -27,9 +28,10 @@ func openLocalKopiaRepo(
 	ctx context.Context, //revive:disable-line:context-as-argument
 ) (*conn, error) {
 	st := storeTD.NewFilesystemStorage(t)
+	hashStr := uuid.NewString()[:7]
 
 	k := NewConn(st)
-	if err := k.Initialize(ctx, repository.Options{}, repository.Retention{}); err != nil {
+	if err := k.Initialize(ctx, repository.Options{}, repository.Retention{}, hashStr); err != nil {
```
meain, line 34: I'm starting to feel like we could probably have `hashStr` inside the `conn` struct. Any thoughts @ryanfkeepers ?

src/pkg/account/m365.go
```diff
@@ -57,6 +62,31 @@ func (a Account) M365Config() (M365Config, error) {
 	return c, c.validate()
 }
 
+func (a Account) GetM365ConfigForHashing() (map[string]any, error) {
```
meain, line 65: Why not just introduce something like `Hash() (string, error)` on repository?

src/pkg/path/path.go
```diff
@@ -127,6 +129,25 @@ type RestorePaths struct {
 	RestorePath Path
 }
 
+type FileSystem interface {
```
meain, line 132: I guess you created it for use in tests. Just for completeness, we call add `os.Stat` as well here I guess. The test implementation can be just a `panic` if we are not using it.

src/pkg/path/path.go
```diff
@@ -317,6 +338,60 @@ func Split(segment string) []string {
...
+	normalizedPath2 := strings.TrimSpace(filepath.Clean(path2))
+
+	normalizedPath1 = strings.TrimSuffix(normalizedPath1, string(filepath.Separator))
+	normalizedPath2 = strings.TrimSuffix(normalizedPath2, string(filepath.Separator))
+
+	return normalizedPath1 == normalizedPath2
+}
+
+func IsValidPath(path string, fs FileSystem) bool {
+	isDir := strings.TrimSpace(filepath.Ext(path)) == ""
```
meain, line 352: That feels sketchy. I feel like we should allow them to have filenames without an extension.

src/pkg/repository/repository.go
```diff
@@ -336,6 +283,101 @@ func (r repository) NewRetentionConfig(
...
+		"storage_provider", r.Storage.Provider.String())
+}
+
+func (r *repository) handleRecovery(ctx context.Context, err *error, label string) {
+	if crErr := crash.Recovery(ctx, recover(), label); crErr != nil {
+		*err = crErr
+	}
+}
+
+func (r *repository) setupKopia(
```
meain, line 330: Good call on abstracting this out :D. Thank you!

src/pkg/account/m365.go
HiteshRepo: @meain This is to get the part of M365Config (for exclusion of fields that we do not want to be cosidered for hash) for hash generation.
Simlarly for S3: https://github.com/alcionai/corso/blob/43fc461138842b9d2f431aa9c1ce3ed665722b3e/src/pkg/storage/s3.go#L78

, and Filesystem: https://github.com/alcionai/corso/blob/43fc461138842b9d2f431aa9c1ce3ed665722b3e/src/pkg/storage/filesystem.go#L39

In repository, we have: https://github.com/alcionai/corso/blob/43fc461138842b9d2f431aa9c1ce3ed665722b3e/src/pkg/repository/repository.go#L286

HiteshRepo:
Closing this PR as it is too big for single review.
Migrated the changes into 4 different PRs:
- https://github.com/alcionai/corso/pull/4635
- https://github.com/alcionai/corso/pull/4637
- https://github.com/alcionai/corso/pull/4638
- https://github.com/alcionai/corso/pull/4639

PR #4615, opened 2023-11-06
also handle delete -> create -> delete

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -492,7 +492,7 @@ func updateCollectionPaths(
 	var initialCurPath path.Path
 
 	col, found := cmap[driveID][itemID]
-	if found {
+	if found && col.FullPath() != nil {
 		initialCurPath = col.FullPath()
 		if initialCurPath.String() == curPath.String() {
 			return found, nil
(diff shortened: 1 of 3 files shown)


src/internal/m365/collection/drive/collections.go
```diff
@@ -866,19 +866,25 @@ func (c *Collections) processItem(
...
-		recreated := found && pid != itemID
-		if recreated {
+		alreadyHandledFolderID, collPathAlreadyExists := seenFolders[collectionPath.String()]
+		collPathAlreadyExists = collPathAlreadyExists && alreadyHandledFolderID != itemID
+
+		if collPathAlreadyExists {
+			// we don't have a good way of juggling multiple previous paths
+			// at this time.  If a path was declared twice, it's a bit ambiguous
+			// which prior data the current folder now contains.  Safest thing to
+			// do is to call it a new folder and ingest items fresh.
```
ashmrtn, line 876: is this safe in all cases? Does moving a folder to the spot of the deleted folder also lead to a missing delete record in the delta result? If it does then we could end up duplicating (part of) the moved folder's items and never actually put the folder in the proper place in the kopia hierarchy

ashmrtn: also I think we need to explore what happens if we don't see a delta record for a deleted folder that had children. If we also don't see delta records showing the deletion of the child folders/items then the previous paths map will get out of sync

ashmrtn (approved):

PR #4618, opened 2023-11-06
Implement a sliding window rate limiter

pandeyabs:
Create a sliding window rate limiter which provides 2 guarantees:
1. Ensures we stay within service limits for a given window duration. 
2. Allows token tracking and window sliding at a finer resolution. This allows for more accuracy at the cost of more frequent sliding & more memory usage.

This implementation is partially inspired by the centralized token grant mechanism in https://github.com/reugn/equalizer#slider. 

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pandeyabs:
Removing use of goleak package that I used earlier, as it leads to flakiness. Some details: 
* We invoke `defer s.Shutdown()` followed by `defer goleak.VerifyNone(t)`. Expectation here was that `Shutdown` will signal the channel and cause `slide` goroutine to stop, before `goleak` tests for goroutine leaks. 
* However, this may not always be true since we are signaling the channel in a non blocking way in `Shutdown()`. 

The value add provided by `goleak` package in this case is minimal. So I'm removing its usage.

src/internal/common/limiters/sliding_window.go
```diff
@@ -0,0 +1,197 @@
...
+	return nil
+}
+
+// Shutdown cleans up the slide goroutine. If shutdown is not called, the slide
+// goroutine will continue to run until the program exits.
+func (s *slidingWindow) Shutdown() {
+	select {
+	case s.stopTimer <- struct{}{}:
+	default:
+	}
```
ashmrtn, line 107: nit: if we're not worried about restarting the rate limited this can be changed to just `close(s.stopTimer)`. Right now it seems a little odd because it's a non-blocking call but I'm not sure if it's guaranteed to send the shutdown signal since the default case is also there

src/internal/common/limiters/sliding_window.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return s, nil
+}
+
+// Wait blocks a request until a token is available or the context is cancelled.
+// TODO(pandeyabs): Implement WaitN.
+func (s *slidingWindow) Wait(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
```
ashmrtn, line 90: nit: could you stack this so we know where it came from?
```suggestion
		return clues.Stack(ctx.Err())
```

src/internal/common/limiters/sliding_window.go
```diff
@@ -0,0 +1,197 @@
...
+
+	s.nextInterval()
+
+	for i := int(0); i < s.prev.count[s.currentInterval]; i++ {
+		select {
+		case s.permits <- token{}:
+		default:
+			// Skip if permits are at capacity
+			return
+		}
```
ashmrtn, line 167: this feels a little odd to me. If we're never creating more tokens once the limiter's been initialized then how can the channel be at capacity if the channel itself can buffer as many tokens as we have total? It seems like either it's never going to pick the default case because the channel is buffered or if it does pick the default case without us knowing then we'll permanently lose some capacity in the rate limiter because we'll never add those tokens back

There's a quick example [here](https://play.golang.com/p/8RHMrQO36oi) that you could play around with if you're interested. Adjusting the buffer on the channel can cause the total number of sends on it to change which causes the final "count" printed at the end to change

src/internal/common/limiters/sliding_window.go
```diff
@@ -0,0 +1,197 @@
...
+				s.slide()
+			case <-s.stopTimer:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+
+	// Prefill permits to allow tokens to be granted immediately
+	for i := int(0); i < s.capacity; i++ {
```
ashmrtn: nit: counters declared in a for-loop should already be of type `int` so I don't believe a cast is necessary

src/internal/common/limiters/sliding_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 2*windowSize)
+	defer cancel()
+
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+
+		err := s.Wait(ctx)
+		require.Equal(t, context.DeadlineExceeded, err)
```
ashmrtn, line 158: nit: there's a `require.ErrorIs` function that is a bit more flexible and plays better with the idea of error wrapping/stacking
```suggestion
		require.ErrorIs(t, err, context.DeadlineExceeded)
```

src/internal/common/limiters/sliding_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+	}()
+
+	wg.Wait()
+}
+
+func (suite *SlidingWindowUnitTestSuite) TestNewSlidingWindowLimiter() {
+	tests := []struct {
```
ashmrtn, line 165: since it's in the validate logic, could you please also add a test that checks what happens if the window size isn't evenly divisible by the interval?

ashmrtn (commented):
mostly looks good, just a few discussion points and some nits

src/internal/common/limiters/sliding_window.go
pandeyabs: > If we're never creating more tokens once the limiter's been initialized then how can the channel be at capacity if the channel itself can buffer as many tokens as we have total?

Great catch! The default case will never execute here. Like you said, total tokens at any time = tokens buffered in the `permits` channel + tokens consumed in the sliding window. If the permits channel is at capacity, then the `s.prev.count[s.currentInterval]` can never have non-zero value.

src/internal/common/limiters/sliding_window.go
pandeyabs: Missed this comment. Yeah! This is an aftermath of replacing `int64` with `int` everywhere, which is basically the same thing for 64 bit systems. I fixed this earlier today

ashmrtn (approved):

src/internal/common/limiters/sliding_window_test.go
```diff
@@ -0,0 +1,260 @@
...
+			// intervals & windows, so that we can test the sliding window logic better.
+			// Without this, the requests will be bunched up in the very first intervals
+			// of the 2 windows. Rest of the intervals will be empty.
+			time.Sleep(time.Duration(rand.Intn(1500)) * time.Millisecond)
+
+			err := s.Wait(ctx)
+			require.NoError(t, err)
+		}()
+	}
+	wg.Wait()
```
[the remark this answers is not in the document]
pandeyabs: Thanks! Great eye!

src/internal/common/limiters/sliding_window_test.go
```diff
@@ -0,0 +1,260 @@
...
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+
+		err := s.Wait(ctx)
+		require.ErrorIs(t, err, context.DeadlineExceeded)
+	}()
+
+	wg.Wait()
```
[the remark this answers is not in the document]
pandeyabs: yeah, i was leaning in favor of letting CI timeout. This should be a one time affair, as it's easy to see any regressions with context cancellations. 

Thanks for the recommendation though! Looks neat, I'll see if it's easy to use.

src/internal/common/limiters/sliding_window.go
```diff
@@ -0,0 +1,189 @@
...
+		defer s.mu.Unlock()
+
+		s.curr.count[s.currentInterval]++
+	}
+
+	return nil
+}
+
+// Shutdown cleans up the slide goroutine. If shutdown is not called, the slide
+// goroutine will continue to run until the program exits.
```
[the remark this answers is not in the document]
pandeyabs: Thanks! Will just stick with comment as it serves our need.

src/internal/common/limiters/sliding_window_test.go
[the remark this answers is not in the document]
pandeyabs: Switched to `assert.Eventually`. It works great but it fails all the tests in the suite in case of timeout. Which is fine. Better than blocking CI for 20 mins.

PR #4625, opened 2023-11-08
Allow limiting or cancelling pager enumeration

ashmrtn:
Add the ability to cancel a pager partway through enumeration. This
will allow us to later request a specific number of items from the
pager. In those cases we should ideally fetch one page beyond what's
required to get the requested number of items (due to the pager
running in another goroutine)

Does not change the result set or add the ability to request a certain
number of items right now

May be easier to review by commit

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -379,37 +382,116 @@ type addedAndRemovedHandler[T any] func(
 	error,
 )
 
+func getLimitedItems[T any](
```
ryanfkeepers, line 385: the name here doesn't feel quite right, since it sounds like the items are what's limited.
```suggestion
func batchWithMaxItemCount[T any](
```

ryanfkeepers (approved):

src/pkg/services/m365/api/pagers/pagers.go
pandeyabs: Could you also add a function description?

src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -410,6 +410,75 @@ type addedAndRemovedHandler[T any] func(
...
+		reset bool
+		page  []T
+
+		removed []string
+		added   = map[string]time.Time{}
+	)
+
+	// Can't use a for-loop variable declaration because the line ends up too
+	// long.
+	for !done && (itemLimit <= 0 || len(added) < itemLimit) {
```
pandeyabs, line 403: Whats the significance of `itemLimit <= 0`? Is this a way to do a complete enumeration?

src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -485,31 +485,41 @@ func GetAddedAndRemovedItemIDs[T any](
 	deltaPager DeltaHandler[T],
 	prevDeltaLink string,
 	canMakeDeltaQueries bool,
+	itemLimit int,
```
pandeyabs, line 460: Could you please add a description for this function?

pandeyabs (approved):

src/pkg/services/m365/api/pagers/pagers.go
ashmrtn: yup! since pagers are already used in a few places I was trying to avoid having to setup default options on the call config (e.x. use `-1` for "enumerate all"). Since 0 is the zero-value for int's that's what I was hoping to leverage in order to keep other changes smaller

PR #4626, opened 2023-11-08
Minor cleanup

ashmrtn:

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/pagers/pagers.go
+++ b/src/pkg/services/m365/api/pagers/pagers.go
@@ -76,7 +76,10 @@ func (npr *nextPageResults[T]) writeNextPage(
 
 	select {
 	case <-ctx.Done():
-		return clues.Wrap(context.Canceled, "writing next page")
+		return clues.Wrap(
+			clues.Stack(ctx.Err(), context.Cause(ctx)),
+			"writing next page")
+
 	case npr.pages <- nextPage[T]{
 		items: items,
 		reset: reset,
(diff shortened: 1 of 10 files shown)


src/pkg/services/m365/api/pagers/pagers.go
```diff
@@ -232,8 +228,6 @@ func EnumerateItems[T any](
 
 		pager.SetNextLink(nextLink)
 	}
-
-	logger.Ctx(ctx).Infow("completed delta item enumeration", "result_count", len(result))
```
ryanfkeepers, line 236: This has been a helpful log in some recent debugging.  Can we keep it in?  Or, if its bothersome, move it to Debug?

ryanfkeepers (approved):

src/pkg/services/m365/api/pagers/pagers.go
ashmrtn: we can keep the log that we're done enumerating things. The item count is not useful at all unless we add additional logic since we're no longer accumulating the result in this function

pandeyabs: We should keep the log at info level, it's useful. result count can be omitted since its empty.
Also, we should change the log in `EnumerateItems` to `completed item enumeration` instead of `completed delta item enumeration`

ashmrtn:
closing as this cleanup got done in a different PR for other reasons

PR #4629, opened 2023-11-08
support debug reading out delta files

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :muscle: Manual


src/pkg/repository/data_providers.go
```diff
@@ -30,6 +36,12 @@ type DataProviderConnector interface {
 		ctx context.Context,
 		pst path.ServiceType,
 	) error
+	// DataProvider retrieves the data provider.
+	DataProvider() DataProvider
+}
+
+func (r *repository) DataProvider() DataProvider {
+	return r.Provider
```
ashmrtn, line 44: is this still used? If not, please remove it to help keep interfaces smaller

src/internal/kopia/base_finder.go
```diff
@@ -55,6 +55,20 @@ type BackupBase struct {
 	Reasons []identity.Reasoner
 }
 
+func (bb BackupBase) GetReasons() []identity.Reasoner {
+	return bb.Reasons
+}
```
ashmrtn, line 60: since we're now implementing a function for this, would you please go through and make `bb.Reasons` unexported and replace calls to it with this function?

src/internal/kopia/base_finder.go
```diff
@@ -55,6 +55,20 @@ type BackupBase struct {
...
+func (bb BackupBase) GetSnapshotID() manifest.ID {
+	return bb.ItemDataSnapshot.ID
+}
+
+func NewFromBackup(bup *backup.Backup) *BackupBase {
+	return &BackupBase{
+		Backup: bup,
+	}
+}
+
```
ashmrtn, line 71: I'm really uneasy about having a function like this when we also have a function like `GetSnapshotID` which explicitly tries to access unpopulated fields. It feels like too much of a gotcha. If this function is unused please remove it. Otherwise, we should come up with a way to make this safer

ashmrtn (approved):

src/pkg/repository/data_providers.go
ryanfkeepers: still used, yes.

src/internal/kopia/base_finder.go
ryanfkeepers: It's being set by external packages in some places.  We can tackle that as tech debt another time.

src/internal/kopia/base_finder.go
ryanfkeepers: Oh!  well, you're in luck.  This is dead code, anyway.  I just forgot to remove it.

PR #4630, opened 2023-11-08
verify previous path state before backup

ryanfkeepers:
Drive previous paths have a prior bug or condition that allows multiple previousPath metadata folder ID entries within a single drive point to the same previous path string.
In these cases we need to treat the backup as having no prior metadata, and re-run the full delta query to correct the bad state.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -78,7 +78,7 @@ func deserializeAndValidateMetadata(
 ) (map[string]string, map[string]map[string]string, bool, error) {
 	deltas, prevs, canUse, err := DeserializeMetadata(ctx, cols)
 	if err != nil || !canUse {
-		return deltas, prevs, canUse, clues.Stack(err).OrNil()
+		return deltas, prevs, false, clues.Stack(err).OrNil()
 	}
 
 	// Go through and remove delta tokens if we didn't have any paths for them
@@ -90,8 +90,11 @@ func deserializeAndValidateMetadata(
 	// without a delta but not to have a delta without paths. This way ensures
 	// we check at least all the path sets for the deltas we have.
 	for drive := range deltas {
+		ictx := clues.Add(ctx, "drive_id", drive)
+
 		paths := prevs[drive]
 		if len(paths) == 0 {
+			logger.Ctx(ictx).Info("dropping drive delta due to 0 prev paths")
 			delete(deltas, drive)
 		}
 
@@ -101,7 +104,9 @@ func deserializeAndValidateMetadata(
 		// for other possibly incorrect folder paths.
 		for _, prevPath := range paths {
 			if len(prevPath) == 0 {
+				logger.Ctx(ictx).Info("dropping drive delta due to 0 len path")
 				delete(deltas, drive)
+
 				break
 			}
 		}
@@ -117,10 +122,10 @@ func deserializeAndValidateMetadata(
 						"collision_folder_id_1", fid,
 						"collision_folder_id_2", otherID,
 						"collision_drive_id", driveID,
-						"collision_prev_path", prev).
+						"collision_prev_path", path.LoggableDir(prev)).
 					Info("duplicate previous paths across different folder IDs")
 
-				// If a previous path collision occurs, we need to force a full backup
+				// If a previous path collision occurs, we need to force a full enumeration
 				// in order to correct the backup state, else we'll potentially hit
 				// unresolvable failures.
 				return map[string]string{}, map[string]map[string]string{}, false, nil
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -72,7 +72,71 @@ func NewCollections(
 	}
 }
 
-func deserializeMetadata(
+func deserializeAndValidateMetadata(
+	ctx context.Context,
+	cols []data.RestoreCollection,
+) (map[string]string, map[string]map[string]string, bool, error) {
+	deltas, prevs, canUse, err := DeserializeMetadata(ctx, cols)
+	if err != nil || !canUse {
+		return deltas, prevs, canUse, clues.Stack(err).OrNil()
+	}
+
+	// Go through and remove delta tokens if we didn't have any paths for them
+	// or one or more paths are empty (incorrect somehow). This will ensure we
+	// don't accidentally try to pull in delta results when we should have
+	// enumerated everything instead.
+	//
+	// Loop over the set of previous deltas because it's alright to have paths
+	// without a delta but not to have a delta without paths. This way ensures
+	// we check at least all the path sets for the deltas we have.
+	for drive := range deltas {
+		paths := prevs[drive]
+		if len(paths) == 0 {
+			delete(deltas, drive)
+		}
+
+		// Drives have only a single delta token. If we find any folder that
+		// seems like the path is bad we need to drop the entire token and start
+		// fresh. Since we know the token will be gone we can also stop checking
+		// for other possibly incorrect folder paths.
+		for _, prevPath := range paths {
+			if len(prevPath) == 0 {
+				delete(deltas, drive)
+				break
+			}
+		}
+	}
+
+	prevPathCollisions := map[string]string{}
+
+	for driveID, folders := range prevs {
+		for fid, prev := range folders {
+			if otherID, collision := prevPathCollisions[prev]; collision {
+				logger.Ctx(ctx).
+					With(
+						"collision_folder_id_1", fid,
+						"collision_folder_id_2", otherID,
+						"collision_drive_id", driveID,
+						"collision_prev_path", prev).
+					Info("duplicate previous paths across different folder IDs")
+
+				// If a previous path collision occurs, we need to force a full backup
+				// in order to correct the backup state, else we'll potentially hit
+				// unresolvable failures.
+				return map[string]string{}, map[string]map[string]string{}, false, nil
+			}
+
+			prevPathCollisions[prev] = fid
+		}
+
+		// reset the collision check for each drive
+		prevPathCollisions = map[string]string{}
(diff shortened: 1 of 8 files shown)


src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
 	}
 }
 
+func deserializeAndValidateMetadata(
+	ctx context.Context,
+	cols []data.RestoreCollection,
+) (map[string]string, map[string]map[string]string, bool, error) {
+	deltas, prevs, canUse, err := DeserializeMetadata(ctx, cols)
+	if err != nil || !canUse {
+		return deltas, prevs, canUse, clues.Stack(err).OrNil()
```
ashmrtn, line 81: should we always return false for `canUse`? Or are we also guaranteeing that the caller will always exit if err != nil?

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+	// don't accidentally try to pull in delta results when we should have
+	// enumerated everything instead.
+	//
+	// Loop over the set of previous deltas because it's alright to have paths
+	// without a delta but not to have a delta without paths. This way ensures
+	// we check at least all the path sets for the deltas we have.
+	for drive := range deltas {
+		paths := prevs[drive]
+		if len(paths) == 0 {
+			delete(deltas, drive)
```
ashmrtn, line 95: nit: we should probably add log messages when we drop stuff so that we can track this stuff down later if needed

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+
+	for driveID, folders := range prevs {
+		for fid, prev := range folders {
+			if otherID, collision := prevPathCollisions[prev]; collision {
+				logger.Ctx(ctx).
+					With(
+						"collision_folder_id_1", fid,
+						"collision_folder_id_2", otherID,
+						"collision_drive_id", driveID,
+						"collision_prev_path", prev).
```
ashmrtn, line 120: nit: directory path will have PII
```suggestion
						"collision_prev_path", path.LoggableDir(prev)).
```

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+					Info("duplicate previous paths across different folder IDs")
+
+				// If a previous path collision occurs, we need to force a full backup
```
ashmrtn, line 123:
```suggestion
				// If a previous path collision occurs, we need to force full enumeration
```

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+			if len(prevPath) == 0 {
+				delete(deltas, drive)
+				break
+			}
+		}
+	}
+
+	prevPathCollisions := map[string]string{}
+
+	for driveID, folders := range prevs {
```
ashmrtn, line 112: Could you please rearrange this? I find it confusing to initialize/clear the map this way instead of just initializing it in the first for-loop
```suggestion
	for driveID, folders := range prevs {
		prevPathCollisions := map[string]string{}

```

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+				// in order to correct the backup state, else we'll potentially hit
+				// unresolvable failures.
+				return map[string]string{}, map[string]map[string]string{}, false, nil
+			}
+
+			prevPathCollisions[prev] = fid
+		}
+
+		// reset the collision check for each drive
+		prevPathCollisions = map[string]string{}
```
ashmrtn, line 133:
```suggestion
```

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -2839,6 +2910,62 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
...
+					},
+				},
+			},
+			canUsePreviousBackup: false,
+			errCheck:             assert.NoError,
+			prevFolderPaths: map[string]map[string]string{
+				driveID1: {
+					"root":    rootFolderPath1,
+					"folder":  rootFolderPath1 + "/folder",
+					"folder2": rootFolderPath1 + "/folder",
```
ashmrtn, line 2938: could you please add another folder that doesn't appear in the result set and make sure it's marked as deleted

src/internal/m365/collection/drive/collections.go
```diff
@@ -72,6 +72,70 @@ func NewCollections(
...
+						"collision_folder_id_1", fid,
+						"collision_folder_id_2", otherID,
+						"collision_drive_id", driveID,
+						"collision_prev_path", prev).
+					Info("duplicate previous paths across different folder IDs")
+
+				// If a previous path collision occurs, we need to force a full backup
+				// in order to correct the backup state, else we'll potentially hit
+				// unresolvable failures.
+				return map[string]string{}, map[string]map[string]string{}, false, nil
```
ashmrtn, line 126: This is actually a bit over-conservative. We can simply drop the previous paths and delta for the offending drive instead of for all drives

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -1375,6 +1375,77 @@ func (suite *OneDriveCollectionsUnitSuite) TestDeserializeMetadata() {
...
+					}
+				},
+			},
+			expectedDeltas:       map[string]string{},
+			expectedPaths:        map[string]map[string]string{},
+			canUsePreviousBackup: false,
+			errCheck:             assert.NoError,
+		},
+		{
+			name: "DuplicatePreviousPaths_separateDrives",
```
ashmrtn, line 1403: could you please also add a test where one drive has duplicate paths and the other doesn't? We should only be dropping metadata info for the one that has the duplicate in the ideal case

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -2839,6 +2910,62 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
 				rootFolderPath2: true,
 			},
 		},
+		{
+			name:   "duplicate previous paths in metadata",
```
ashmrtn, line 2914: I'm unsure if we should add another test (or expand this one) to have two drives, one of which has duplicate paths. I think it would be good coverage to have, though of course does come with more maintenance requirements

src/internal/m365/collection/drive/collections.go
ryanfkeepers: fair, might as well

src/internal/m365/collection/drive/collections.go
ryanfkeepers: sure, sounds good

src/internal/m365/collection/drive/collections.go
ryanfkeepers: good catch, thank you

src/internal/m365/collection/drive/collections.go
ryanfkeepers: Heh, yeah, I already did that in the other PR.   Had a "why did I even do this?" moment.

src/internal/m365/collection/drive/collections.go
ryanfkeepers: fair point

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Hmm, well, it's not marked as deleted.  It's not included in the result set either.  I'll follow up on this in the next PR where we don't force a full enumeration and make sure it's handled the right way there.

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Going to skip this.  Same reason as skipping the per-drive metadata removal; we won't be forcing full enumeration on the next PR, so it won't be handled this way.

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Sure, that's easy to do.

ryanfkeepers:
@ashmrtn ready for final review

src/internal/m365/collection/drive/collections_test.go
ashmrtn: I think it's still worth adding a test for this so we can, at the very least, build on it later on. Adding it will help ensure we don't lose track of that case and that alerts act as expected in the next PR

src/internal/m365/collection/drive/collections_test.go
ashmrtn: I'm not sure if that's correct behavior or not. It would depend on what other data is returned. If we don't return a flag saying the base shouldn't be merged (which we still need to remove at some point) or mark the collection as deleted then it'll be pulled forward

I know this PR train isn't trying to solve that problem anymore, but I'm concerned about losing track of this particular test. It's important to have because we should handle it properly eventually, I just don't know the best way to record it

ashmrtn (approved):
would really like to see a few more tests added so we don't lose track of them

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: fwiw, there's already a test for that in the next PR

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: It's tracked on a test in the next PR.

PR #4631, opened 2023-11-08
Ensure each drive gets a unique url cache instance

pandeyabs:
This fixes an url cache bug. We were incorrectly attaching the same url cache instance to all drives belonging to a sharepoint library. If a library has n drives, this bug had an effect of turning off cache for n - 1 drives. 

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections.go
+++ b/src/internal/m365/collection/drive/collections.go
@@ -311,25 +311,36 @@ func (c *Collections) Get(
 		driveIDToPrevPaths[driveID] = map[string]string{}
 		maps.Copy(driveIDToPrevPaths[driveID], newPrevPaths)
 
+		numDriveItems := c.NumItems - numPrevItems
+		numPrevItems = c.NumItems
+
 		logger.Ctx(ictx).Infow(
 			"persisted metadata for drive",
 			"num_new_paths_entries", len(newPrevPaths),
 			"delta_reset", du.Reset)
 
-		numDriveItems := c.NumItems - numPrevItems
-		numPrevItems = c.NumItems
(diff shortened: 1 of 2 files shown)


src/internal/m365/collection/drive/collections.go
```diff
@@ -319,17 +319,28 @@ func (c *Collections) Get(
...
 		numPrevItems = c.NumItems
 
-		// Attach an url cache
+		// Attach an url cache to the drive if the number of discovered items is
+		// below the threshold. Attaching cache to larger drives can cause
+		// performance issues since cache delta queries start taking up majority of
+		// the hour the refreshed URLs are valid for.
 		if numDriveItems < urlCacheDriveItemThreshold {
-			logger.Ctx(ictx).Info("adding url cache for drive")
+			logger.Ctx(ictx).Infow(
```
ashmrtn, line 327: nit: maybe we should log the drive ID too? Seems like it would be good to know which we're attaching to
```suggestion
			logger.Ctx(ictx).Infow(
				"cache_drive_id", driveID,
```

ashmrtn (approved):

src/internal/m365/collection/drive/collections.go
pandeyabs: drive ID is already included in `ictx` on line 256 above.

PR #4632, opened 2023-11-08
error if backup produces invalid metadata

ryanfkeepers:
fail the backup if it produces a previous path collision in drive metadata.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] ❓❓❓

### DIFF
--- a/src/internal/m365/collection/drive/collections_test.go
+++ b/src/internal/m365/collection/drive/collections_test.go
@@ -3054,6 +3054,124 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
 			}),
 			doNotMergeItems: map[string]bool{},
 		},
+		{
+			name:   "out of order item enumeration causes prev path collisions",
+			drives: []models.Driveable{drive1},
+			enumerator: mock.EnumerateItemsDeltaByDrive{
+				DrivePagers: map[string]*mock.DriveItemsDeltaPager{
+					driveID1: {
+						Pages: []mock.NextPage{{
+							Items: []models.DriveItemable{
+								driveRootItem("root"),
+								driveItem("fanny2", "fanny", driveBasePath1, "root", false, true, false),
+								driveItem("file2", "file2", driveBasePath1+"/fanny", "fanny2", true, false, false),
+								driveItem("nav", "nav", driveBasePath1, "root", false, true, false),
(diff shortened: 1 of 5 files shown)


src/internal/m365/collection/drive/collections_test.go
```diff
@@ -2962,46 +2963,96 @@ func (suite *OneDriveCollectionsUnitSuite) TestGet() {
...
+			errCheck:             assert.NoError,
+			previousPaths: map[string]map[string]string{
+				driveID1: {
+					"root": rootFolderPath1,
+					"nav":  rootFolderPath1 + "/fanny",
+				},
+			},
+			expectedCollections: map[string]map[data.CollectionState][]string{
+				rootFolderPath1: {
+					data.NewState: {"fanny2"},
```
ashmrtn, line 3033: I don't believe we should have folder info showing up in the parent of the folder. From my understanding, this is saying the metadata for this folder is going to be in root which is something we've been trying to avoid for a while now. It should only appear in the folder it refers to

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -1477,7 +1477,7 @@ func (suite *OneDriveCollectionsUnitSuite) TestDeserializeMetadata() {
 					data.NoFetchRestoreCollection{Collection: mc}))
 			}
 
-			deltas, paths, canUsePreviousBackup, err := deserializeAndValidateMetadata(ctx, cols)
+			deltas, paths, canUsePreviousBackup, err := deserializeAndValidateMetadata(ctx, cols, fault.New(true))
```
ashmrtn, line 1480: can you please expand this test to check that we actually add alerts when we think we should? Right now this test isn't really checking all that much since we no longer modify the input metadata

ashmrtn (commented):
need resolution on one of the test outputs for directory metadata in TestGet

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Hmm, I think this is a topic that's separate to the task at hand.  This behavior is consistent with all the other tests.  So either the tests are malformed in a consistent manner, or we have a bug.  Neither case really overlaps with this change.  If you think it deems further investigation, would you mind creating an issue ticket for it?

ashmrtn (approved):

PR #4635, opened 2023-11-09
check path equivalency

HiteshRepo:
This change is to address the path related issue that comes up while `corso repo connect filesystem`.
Below are the steps to reciprocate:
1. Initialize repo with filesystem provider (Notice that the path has a trailing slash): 
command: `./corso repo init filesystem --path ~/Documents/Backups/`
output:
Logging to file: ~/.cache/corso/logs/2023-11-09T05-21-57Z.log
Connecting to M365:         done 
Initializing repository   
Initialized a repository at path ~/Documents/Backups/

2. Connect repo with filesystem provider:
command: `./corso repo connect filesystem --path ~/Documents/Backups/`
output:
Logging to file: ~/.cache/corso/logs/2023-11-09T05-23-30Z.log
Error: retrieving storage provider details: verifying storage configs in corso config file: verifying configs in corso config file: value of path (~/Documents/Backups) does not match corso configuration value (~/Documents/Backups/)

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/path/path.go
```diff
@@ -317,6 +319,21 @@ func Split(segment string) []string {
 	return res
 }
 
+func ArePathsEquivalent(path1, path2 string) bool {
```
ryanfkeepers, line 322: could we get some unit tests for this func before merging, please?

src/pkg/path/path.go
```diff
@@ -317,6 +319,21 @@ func Split(segment string) []string {
...
+	normalizedPath1 := strings.TrimSpace(filepath.Clean(path1))
+	normalizedPath2 := strings.TrimSpace(filepath.Clean(path2))
+
+	normalizedPath1 = strings.TrimSuffix(normalizedPath1, string(filepath.Separator))
+	normalizedPath2 = strings.TrimSuffix(normalizedPath2, string(filepath.Separator))
+
+	return normalizedPath1 == normalizedPath2
+}
+
+func IsValidPath(path string) bool {
```
ryanfkeepers, line 332: The name is a bit misleading, since many inputs to this func could hold valid paths without passing the stat check.  This name makes it a bit more clear what we're checking.
```suggestion
func DoesPathExistInOS(path string) bool {
```

src/pkg/storage/storage.go
```diff
@@ -168,7 +169,16 @@ func mustMatchConfig(
...
 		vv := cast.ToString(g.Get(tomlK))
-		if v != vv {
+
+		areEqual := false
+
+		if path.IsValidPath(v) && path.IsValidPath(vv) {
+			areEqual = path.ArePathsEquivalent(v, vv)
+		} else {
+			areEqual = v == vv
+		}
```
ryanfkeepers, line 179: It should be possible to skip the valid check and call `path.ArePathsEquivalent` for any pair of values, right?  Whether or not they clear os.Stat doesn't actually change how ArePathsEquivalent compares them.

ryanfkeepers (approved):
Some nits on design choices to evaluate, but nothing blocking.  Approving to unblock while other details get finalized.

src/pkg/storage/storage.go
HiteshRepo: The value pairs can be any string other than paths. So the idea was to compare paths only if they qualify as paths.

src/pkg/storage/storage.go
```diff
@@ -168,7 +169,16 @@ func mustMatchConfig(
 		}
 
 		vv := cast.ToString(g.Get(tomlK))
-		if v != vv {
+
+		areEqual := false
+
+		if path.DoesPathExistInOS(v) && path.DoesPathExistInOS(vv) {
```
pandeyabs, line 175: Could you please add a comment to describe this code addition? Also once concern I have is that there are non path related keys in the map. This code seems to be doing `os.stat()` call for  all the keys. One could argue that these could be valid file paths on a user's system, could be interpreted incorrectly as a snooping behavior.

Is there a way we could only do this check for file system path key? Btw, thanks for fixing the trailing slash bug! Good find!

HiteshRepo: @pandeyabs please check the current change. Do you concur with the current implementation?

PR #4636, opened 2023-11-09
Allow callers to request multiple limiter tokens

pandeyabs:
Adds `WaitN` support in sliding window limiter. `WaitN` blocks the request until all `N` tokens are acquired.  It won't be used for exchange service since exchange doesn't assign weights to API requests. 

This is mostly for interface conformance right now, so that we can support various limiter types behind the same `limiter` interface. 

`WaitN` might be useful for onedrive/sharepoint if we decide to use sliding window limiter for those services. Although the value add is debatable since limits are per min. It doesn't help us much perf-wise to front load requests for a minute long window. 

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/common/limiters/sliding_window.go
+++ b/src/internal/common/limiters/sliding_window.go
@@ -89,8 +89,13 @@ func (s *slidingWindow) Wait(ctx context.Context) error {
 	return s.WaitN(ctx, 1)
 }
 
-// Wait blocks a request until N tokens are available or the context gets
-// cancelled.
+// WaitN blocks a request until n tokens are available or the context gets
+// cancelled. WaitN should be called with n <= capacity otherwise it will block
+// forever.
+//
+// TODO(pandeyabs): Enforce n <= capacity check. Not adding it right now because
+// we are relying on capacity = 0 for ctx cancellation test, which would need
+// some refactoring.
 func (s *slidingWindow) WaitN(ctx context.Context, n int) error {
 	for i := 0; i < n; i++ {
 		select {
--- a/src/internal/common/limiters/sliding_window.go
+++ b/src/internal/common/limiters/sliding_window.go
@@ -91,8 +91,8 @@ func (s *slidingWindow) Wait(ctx context.Context) error {
 
 // Wait blocks a request until N tokens are available or the context gets
 // cancelled.
-func (s *slidingWindow) WaitN(ctx context.Context, N int) error {
-	for i := 0; i < N; i++ {
+func (s *slidingWindow) WaitN(ctx context.Context, n int) error {
+	for i := 0; i < n; i++ {
 		select {
 		case <-ctx.Done():
 			return clues.Stack(ctx.Err())
--- a/src/internal/common/limiters/sliding_window.go
+++ b/src/internal/common/limiters/sliding_window.go
@@ -84,16 +84,23 @@ func NewSlidingWindowLimiter(
 }
 
 // Wait blocks a request until a token is available or the context is cancelled.
-// TODO(pandeyabs): Implement WaitN.
+// Equivalent to calling WaitN(ctx, 1).
 func (s *slidingWindow) Wait(ctx context.Context) error {
-	select {
-	case <-ctx.Done():
-		return clues.Stack(ctx.Err())
-	case <-s.permits:
-		s.mu.Lock()
-		defer s.mu.Unlock()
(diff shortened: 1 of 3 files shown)


src/internal/common/limiters/sliding_window.go
```diff
@@ -84,16 +84,28 @@ func NewSlidingWindowLimiter(
...
+
+// WaitN blocks a request until n tokens are available or the context gets
+// cancelled. WaitN should be called with n <= capacity otherwise it will block
+// forever.
+//
+// TODO(pandeyabs): Enforce n <= capacity check. Not adding it right now because
+// we are relying on capacity = 0 for ctx cancellation test, which would need
+// some refactoring.
+func (s *slidingWindow) WaitN(ctx context.Context, n int) error {
+	for i := 0; i < n; i++ {
```
ashmrtn, line 100: I think this is the point where things get more interesting when it comes to stuff like (request) scheduling policies.

While this implementation works in some cases, it could have the effect of spreading out token acquisition for `WaitN` calls across many callers. Say each caller wants just 2 tokens and there's X callers. In the best case scenario, golang will wake each caller up 2x consecutively so things won't get spread out. However, in the worst case it will round-robin the callers, meaning each one will first get one token and then, after all X have one token, it will start giving each the second token. Since the goroutine that handles the channel receive is random, there's no real way to bound the amount of time a caller may wait on a call like this

Given enough callers, it may even be possible to deadlock if the sum of requested tokens across all callers is greater than the limiter capacity and the tokens are distributed across callers such that no caller has enough tokens to become unblocked. For example, take just two callers, one which has `n = 2` and another where `n = capacity`. In this case, if caller 1 gets one token and then caller 2 gets the remaining tokens the system will deadlock because no matter how long we wait there'll never be enough free tokens to satisfy either caller ("free" meaning not in the previous window and not held by a caller)

I'd be happy to bounce ideas around if that would be helpful to finding a solution. I haven't thought through things completely so I'm not sure what the best option would be right now

pandeyabs: > Given enough callers, it may even be possible to deadlock if the sum of requested tokens across all callers is greater than the limiter capacity and the tokens are distributed across callers such that no caller has enough tokens to become unblocked.

Thanks! I've thought about this. The limiter should not deadlock in this extreme case, although its performance will severely degrade. The reason limiter won't deadlock is because `slide()` routine will keep replenishing tokens every interval. So at some point, the requests will have enough tokens to go through. Although this may result in a flurry of requests to graph at once, causing more 429s. 

Like you said if max(num_concurrent requests) * N ~= capacity or more, we'll see this issue. So this can be an issue for OD/SP since they don't enforce concurrency limits and their requests carry weight. But switching to sliding window limiter for OD/SP doesn't make sense to begin with, since limits are tracked on a minute basis. It won't bring as any noticeable speedups.

I can add a caveat for `WaitN` to summarize this discussion.

ashmrtn: > The limiter should not deadlock in this extreme case, although its performance will severely degrade. The reason limiter won't deadlock is because slide() routine will keep replenishing tokens every interval.

Ah, I see now, token usage is added to the interval in which the token is acquired, not the interval in which it's actually consumed by the request

> So this can be an issue for OD/SP since they don't enforce concurrency limits and their requests carry weight. But switching to sliding window limiter for OD/SP doesn't make sense to begin with, since limits are tracked on a minute basis

Given that information and the limitations being discussed around 429s and degraded performance, does this PR make sense? I don't mean to rugpull this, but do we want to risk the possibility of difficult to find problems unless the engineer using it carefully reads the documentation?

pandeyabs: Agreed. I'll close this PR for now since it's misleading. The purpose of introducing `WaitN` here was to have interface compliance with golang rate limiter package. I'll add another PR to do it in a more controlled way ( only support if N == 1). 

Also thinking about how to implement `WaitN` correctly.  One way we can do so is by ensuring that all `N` tokens get served to each request atomically. This would require introducing a request level lock or similar. I'll work on this later, not super high priority right now.

PR #4637, opened 2023-11-09
abstract kopia setup code

HiteshRepo:
Refactors the repository code to abstract code responsible for kopia related setup.

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/repository/repository.go
```diff
@@ -345,6 +266,68 @@ func (r repository) Counter() *count.Bus {
 	return r.counter
 }
 
+func (r *repository) addContextInfo(ctx context.Context) context.Context {
```
ryanfkeepers, line 269: Nit: for specificity.  "info" is a little ambiguous.
```suggestion
func (r *repository) addContextClues(ctx context.Context) context.Context {
```

src/pkg/repository/repository.go
```diff
@@ -345,6 +266,68 @@ func (r repository) Counter() *count.Bus {
...
+	return clues.Add(
+		ctx,
+		"acct_provider", r.Account.Provider.String(),
+		"acct_id", clues.Hide(r.Account.ID()),
+		"storage_provider", r.Storage.Provider.String())
```
ryanfkeepers, line 274: clues will stringify values for you.  I assume this is from old code, so this is mostly informational for you.
```suggestion
		"acct_provider", r.Account.Provider,
		"acct_id", clues.Hide(r.Account.ID()),
		"storage_provider", r.Storage.Provider)
```

src/pkg/repository/repository.go
```diff
@@ -345,6 +266,68 @@ func (r repository) Counter() *count.Bus {
...
+}
+
+func (r *repository) handleRecovery(ctx context.Context, err *error, label string) {
+	if crErr := crash.Recovery(ctx, recover(), label); crErr != nil {
+		*err = crErr
+	}
+}
```
ryanfkeepers, line 281: Can you double check that this correctly produces a trace back to the calling func, and not to this func?  It'll be a regression if all callers of this func report line 278 as the "place where the panic was handled"

In general, I don't think the abstraction earns us much in this case.  We would also be fine keeping the code how it was.

```suggestion
func (r *repository) handleRecovery(ctx context.Context, err *error, label string) {
	if crErr := crash.Recovery(ctx, recover(), label); crErr != nil {
		*err = crErr
	}
}
```

src/pkg/repository/repository.go
```diff
@@ -204,53 +169,18 @@ type ConnConfig struct {
...
-
-	if r.ID == events.RepoIDNotFound {
-		rm, err := getRepoModel(ctx, r.modelStore)
-		if err != nil {
-			return clues.Wrap(err, "retrieving repo model info")
-		}
-
-		r.ID = string(rm.ID)
+	if err := r.setupKopia(ctx, ctrlRepo.Retention{}, false); err != nil {
+		return err
```
ryanfkeepers, line 183: Always stack errors, please
```suggestion
		return clues.Stack(err)
```

ryanfkeepers (approved):

PR #4638, opened 2023-11-09
set repository config location w.r.t corso config location

HiteshRepo:
Create repository config(s) in the same parent directory as corso config instead of creating repository config in `/tmp` directory.
Scenarios:

1. CORSO_CONFIG_DIR set to /test/directory, --config-file not provided
    - `.corso.toml` should be created at /test/directory
    - `repository.config` should be created at /test/directory
2. CORSO_CONFIG_DIR not set, --config-file provided /test/directory-2
    - `.corso.toml` should be created at /test/directory-2
    - `repository.config` should be created at /test/directory-2
3. CORSO_CONFIG_DIR not set, --config-file not provided
    - `.corso.toml` should be created at /home/user
    - `repository.config` should be created at /home/user

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #4443 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/config/config.go
```diff
@@ -89,11 +91,15 @@ func InitFunc(cmd *cobra.Command, args []string) error {
 		fp = configFilePath
 	}
 
-	err := initWithViper(GetViper(cmd.Context()), fp)
+	vpr := GetViper(cmd.Context())
+
+	err := initWithViper(vpr, fp)
 	if err != nil {
```
ryanfkeepers, line 97:
```suggestion
	if err := initWithViper(vpr, fp); err != nil {
```

src/cli/config/config.go
```diff
@@ -89,11 +91,15 @@ func InitFunc(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
 
+	SetViper(cmd.Context(), vpr)
```
ryanfkeepers, line 101: bug here: note the return value of SetViper.  `context.Context` is not mutable.  if you don't overwrite the local contxt value with the one returned here, you'll lose the viper embedding.

Proper handling would be to set `ctx := cmd.Context()` at the top of the func, and then:
```suggestion
	ctx = SetViper(ctx, vpr)
```

src/cli/config/config_test.go
```diff
@@ -157,6 +157,11 @@ func (suite *ConfigSuite) TestWriteReadConfig() {
 		host   = "some-host"
 	)
 
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	SetViper(ctx, vpr)
```
ryanfkeepers, line 163: need to handle the return value in all of these as well

ryanfkeepers (approved):
there's a bug that needs fixing, but otherwise LGTM

PR #4639, opened 2023-11-09
generate repository config name based on provider specific hash

HiteshRepo:
Enable user/client to be able to perform multiple backups in a concurrent manner irrespective of storage provider (s3/filesystem) for different tenants (anything that differentiates one account from another).

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #4443 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/account/account.go
```diff
@@ -88,3 +91,12 @@ func (a Account) ID() string {
...
 	return a.Config[a.Provider.String()+"-tenant-id"]
 }
+
+func (a Account) GetAccountConfigForHash() (map[string]any, error) {
+	switch a.Provider {
+	case ProviderM365:
+		return a.GetM365ConfigForHashing()
+	}
+
+	return nil, clues.New(fmt.Sprintf(invalidProviderErrTemplate, a.Provider.String()))
```
ryanfkeepers, line 101: Per error and logging conventions, lets keep templates out of the error message, please.  This can be handled as:
```suggestion
	return nil, clues.Stack(errInvalidProvider).With("provider", a.Provider)
```

src/internal/operations/maintenance_test.go
```diff
@@ -39,8 +40,9 @@ func getKopiaHandles(
 	ctx context.Context, //revive:disable-line:context-as-argument
 ) (*kopia.Wrapper, *kopia.ModelStore) {
 	st := storeTD.NewPrefixedS3Storage(t)
+	repoNameHash := uuid.NewString()[:7]
```
ryanfkeepers, line 43: `uuid.NewString()[:7]` might be a good candidate to move into a package like `/str/testData` func.  That way all callers can call `strTD.Hash()`, and if we ever need to change the behavior, we only need to change it in the testData func instead of every place that generates a repo name hash.

src/pkg/repository/repository.go
```diff
@@ -161,8 +163,13 @@ func (r *repository) Initialize(
 
 	observe.Message(ctx, "Initializing repository")
 
+	repoNameHash, err := r.GenerateHashForRepositoryConfigFileName()
+	if err != nil {
+		return clues.Wrap(err, "generate repo config")
```
ryanfkeepers, line 168:
```suggestion
		return clues.Wrap(err, "generating repo config")
```

src/pkg/repository/repository.go
```diff
@@ -345,6 +362,36 @@ func (r repository) Counter() *count.Bus {
 	return r.counter
 }
 
+func (r repository) GenerateHashForRepositoryConfigFileName() (string, error) {
+	accountHashConfig, err := r.Account.GetAccountConfigForHash()
+	if err != nil {
+		return "", clues.Wrap(err, "fetch account hash config")
+	}
```
ryanfkeepers, line 369: instead of having the account and storage return a config property set, why not have them just generate their own hash?  callers would have the option to concatenate hashes, or hash their union, both of which would still be sufficiently unique operations.  That would relieve a lot of the extra processing happening here (which also remove it as boilerplate if any other caller wants to use the funcs).

src/pkg/account/account.go
```diff
@@ -88,3 +91,12 @@ func (a Account) ID() string {
 
 	return a.Config[a.Provider.String()+"-tenant-id"]
 }
+
+func (a Account) GetAccountConfigForHash() (map[string]any, error) {
```
ryanfkeepers, line 95: can you please add a unit test that checks every account type to ensure it provides a hasher?  We want to make sure that on the next account type addition we'll automatically fail the test if it doesn't include a hashing func.

ryanfkeepers (approved):
Various minor things that need handling before merger, but good enough to approve and unblock.  Thanks for the work!

src/pkg/repository/repository.go
HiteshRepo: Yeah. Originally I has used 'md5' and not 'crc32'. So concatenating hashes would have made it big and trimming hash might not be a good idea as it might lead to collisions. But since I am using 'crc32' now, this should be a better choice. I will have this in the next commit.

Issue #3893, opened 2023-07-24
`eml` export from Exchange emails backup


meain:

meain:
Closing this as we are currently only planning for `pst` exports for exchange.

vkamra:
Re-opened - this is in-progress

meain:
`eml` export support is now available in `main`.

PR #4640, opened 2023-11-09
Converter for msg to eml

meain:
Add logic to convert from msg(backup available in corso) to eml

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3893

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/eml/eml_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// read test file into body as []bytes
+	body, err := os.ReadFile("testdata/email-with-attachments.json")
+	require.NoError(t, err, "reading test file")
+
+	msg, err := api.BytesToMessageable(body)
+	require.NoError(t, err, "creating message")
+
+	_, err = toEml(msg)
+	// TODO(meain): add more tests on the generated content
```
meain, line 34: Improvements in https://github.com/alcionai/corso/pull/4644

src/internal/converters/eml/eml.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/alcionai/clues"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	mail "github.com/xhit/go-simple-mail/v2"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/pkg/logger"
+)
+
+const (
+	fromFormat = "%s <%s>"
```
ashmrtn, line 24: nit: maybe a more generic name would work better since it's used for senders and receivers? Something like `nameAddressFormat` maybe? (idk, naming is hard)

src/internal/converters/eml/eml.go
```diff
@@ -0,0 +1,142 @@
...
+// toEml converts a Messageable to .eml format
+func toEml(data models.Messageable) (string, error) {
+	email := mail.NewMSG()
+
+	if data.GetFrom() != nil {
+		email.SetFrom(
+			fmt.Sprintf(
+				fromFormat,
+				ptr.Val(data.GetFrom().GetEmailAddress().GetName()),
+				ptr.Val(data.GetFrom().GetEmailAddress().GetAddress())))
```
ashmrtn, line 37: this applies to the other times we use this pattern as well, but do we need to format things differently if the name is empty? If we feel like this is something we're going to be tweaking in the future maybe it's worth making a function that goes from the result of `GetEmailAddress()` to a string ready to be inserted into the eml library? That would help ensure it stays consistent across fields

src/internal/converters/eml/eml.go
```diff
@@ -0,0 +1,142 @@
...
+		email.SetSubject(ptr.Val(data.GetSubject()))
+	}
+
+	if data.GetSentDateTime() != nil {
+		email.SetDate(ptr.Val(data.GetSentDateTime()).Format(dateFormat))
+	}
+
+	if data.GetBody() != nil {
+		if data.GetBody().GetContentType() != nil {
+			var mct mail.ContentType
```
ashmrtn, line 96: nit: could you please use a more descriptive variable name? I always find initialisms of types to be rather cryptic and unhelpful even if the type name itself is well named. Something like `contentType` or such would work fine

src/internal/converters/eml/eml_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+type EMLUnitSuite struct {
+	tester.Suite
+}
+
+func TestEMLUnitSuite(t *testing.T) {
+	suite.Run(t, &EMLUnitSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+func (suite *EMLUnitSuite) TestConvert_messageble_to_eml() {
```
ashmrtn, line 23: I'd really love to see more tests around the variety of input data we could get. For example CC, BCC, senders that don't have a `Name` field filled out (just the email address), and the error paths in the code that's in `eml.go`. If you could please add tests for those it will help make sure things work like we think they do and that we don't accidentally break something going forward

src/internal/converters/eml/eml.go
ashmrtn: edit: still making my way through PRs, but seems like this may get renamed later?

ashmrtn (commented):
would really like to see some better test coverage. Doesn't need to be in this PR per se, but there's a lot of functionality that isn't tested right now that we should be checking. More than happy to have a discussion about the best way to do so given the tools we have to work with if that would be helpful

src/internal/converters/eml/eml.go
meain: Should be fixed in https://github.com/alcionai/corso/pull/4651

src/internal/converters/eml/eml.go
meain: Fixed in https://github.com/alcionai/corso/pull/4651

src/internal/converters/eml/eml_test.go
meain: Fixed in https://github.com/alcionai/corso/pull/4651

ashmrtn (approved):

PR #4644, opened 2023-11-09
Improvements and tests for eml conversion

meain:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/3893

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/eml/eml.go
```diff
@@ -13,26 +13,31 @@ import (
...
 const (
-	fromFormat = "%s <%s>"
-	dateFormat = "2006-01-02 15:04:05 MST" // from xhit/go-simple-mail
+	addressFormat = "%s <%s>"
+	dateFormat    = "2006-01-02 15:04:05 MST" // from xhit/go-simple-mail
 )
 
-// ToEml converts a Messageable to .eml format
-func ToEml(data models.Messageable) (string, error) {
+// FromJSON converts a Messageable(as json) to .eml format
```
ashmrtn, line 28:
```suggestion
// FromJSON converts a Messageable (as json) to .eml format
```

src/internal/converters/eml/eml_test.go
```diff
@@ -27,11 +30,52 @@ func (suite *EMLUnitSuite) TestConvert_messageble_to_eml() {
...
+		assert.Contains(
+			t,
+			out,
+			fmt.Sprintf(
+				`Bcc: "%s" <%s>`,
+				ptr.Val(addr.GetEmailAddress().GetName()),
+				ptr.Val(addr.GetEmailAddress().GetAddress())))
+	}
+
+	// Only fist 30 chars as it the data will be split
```
ashmrtn, line 79: nit: wording
```suggestion
	// Only fist 30 chars as the data will be split
```

ashmrtn: can you elaborate on this a bit more? I'm unsure what the nature of the "split" in this case is. Does it mean there's other output somewhere else we need to handle?

meain: It is possible for the tool to introduce newlines to limit the column width. I've update the comments as well.

src/internal/converters/eml/eml_test.go
```diff
@@ -30,11 +33,55 @@ func (suite *EMLUnitSuite) TestConvert_messageble_to_eml() {
...
+			fmt.Sprintf(
+				`Bcc: "%s" <%s>`,
+				ptr.Val(addr.GetEmailAddress().GetName()),
+				ptr.Val(addr.GetEmailAddress().GetAddress())))
+	}
+
+	// Only fist 30 chars as the .eml generator can introduce a
+	// newline in between the text to limit the column width of the
+	// output. It does not affect the data, but can break our tests and
+	// so using 30 as a safe limit to test.
```
ashmrtn, line 85: thanks for the comment update! Is there maybe a little bit more low-hanging fruit around this we can use to check for validity while avoiding flakes? For instance, would something like an assertion that the generated body content length is at least as long as the original input's body content be stable?

ashmrtn (approved):

src/internal/converters/eml/eml_test.go
meain: I think I'll parse the generated `eml` file and test against the parsed data. That should make it really robust.
I'll rework that along with https://github.com/alcionai/corso/pull/4651#discussion_r1394538083

PR #4648, opened 2023-11-10
clean up variable naming and usage in collections test

ryanfkeepers:
the collections_test.go contains a lot of hardcoded string references, and a jumble of ways to use them.  Furthermore, it was impelemented from the beginning with many overlapping strings that are used both as IDs and non-id properties, which causes significant confusion for new maintainers.

The goals of this PR are the following:
1. remove common hardcoded string values
2. make id and non-id value cases clearly visible
3. centralize path production so path structure is easier to grok
4. make references uniform throughout the file.

the result is a code set that's a little more difficult to read at first blush, but significantly easier to understand and maintain without needing to dive deep into the test structure.

No logical changes, only renames/code movement.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/.golangci.yml
+++ b/src/.golangci.yml
@@ -125,6 +125,9 @@ issues:
       linters:
         - forbidigo
       text: "context.(Background|TODO)"
+    - path: internal/m365/collection/drive/collections_test.go
+      linters:
+        - lll
     - path: internal/m365/graph/betasdk
       linters:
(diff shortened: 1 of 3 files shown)


HiteshRepo (approved):
LGTM

PR #4651, opened 2023-11-10
Imrpove message -> eml conversion

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/eml/eml_test.go
```diff
@@ -23,6 +24,40 @@ func TestEMLUnitSuite(t *testing.T) {
...
+		email    string
+		expected string
+	}{
+		{
+			tname:    "different name and email",
+			name:     "John Doe",
+			email:    "johndoe@provider.com",
+			expected: `John Doe <johndoe@provider.com>`,
+		},
+		{
```
ashmrtn, line 42: could you please also add a case where the name isn't set?

ashmrtn (approved):

src/internal/converters/eml/eml.go
```diff
@@ -25,75 +26,105 @@ const (
...
-				fmt.Sprintf(
-					addressFormat,
-					ptr.Val(rts[0].GetEmailAddress().GetName()),
-					ptr.Val(rts[0].GetEmailAddress().GetAddress())))
+				With("reply_to_count", len(rts)).
+				Warn("more than 1 Reply-To, adding only the first one")
+		}
+
+		if len(rts) != 0 {
+			email.SetReplyTo(formatAddress(rts[0].GetEmailAddress()))
```
ashmrtn, line 106: do we need to check for `email.Error` here as well? The other assignments seem to do that

ashmrtn: oh, actually it seems there's a `GetError()` [function](https://pkg.go.dev/github.com/xhit/go-simple-mail/v2#Email.GetError). With that it may be possible to just add all the data and then check for errors once at the end (assuming the error message returned by the go-simple-email package is descriptive as to what failed

ashmrtn (approved):

src/internal/converters/eml/eml.go
meain: Yeah, it looks like the errors are descriptive enough. Let me just switch to a single error.

PR #4653, opened 2023-11-10
expose graph options to sdk clients

ryanfkeepers:
1. moves the m365/graph package from internal to pkg/services/api so that options are accessible to sdk users.
2. exposes graph.Options in the api client.Service call.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature


src/pkg/services/m365/api/client.go
```diff
@@ -93,8 +93,11 @@ func InitConcurrencyLimit(ctx context.Context, pst path.ServiceType) {
 // so that in-flight state within the adapter doesn't get clobbered.
 // Most calls should use the Client.Stable property instead of calling this
 // func, unless it is explicitly necessary.
-func (c Client) Service(counter *count.Bus) (graph.Servicer, error) {
-	return NewService(c.Credentials, counter)
+func (c Client) Service(
+	counter *count.Bus,
+	opts ...graph.Option,
```
ryanfkeepers, line 98: This is the critical change that exposes client options.  The rest of the changes are package movement.

juricacuric (approved):

src/.golangci.yml
```diff
@@ -128,7 +128,7 @@ issues:
     - path: internal/m365/collection/drive/collections_test.go
       linters:
         - lll
-    - path: internal/m365/graph/betasdk
+    - path: kg/services/m365/api/graph/betasdk
```
vkamra, line 131: @ryanfkeepers - is there a typo here?
```suggestion
    - path: pkg/services/m365/api/graph/betasdk
```

ryanfkeepers: whoops, yes, good catch

PR #4657, opened 2023-11-10
Wire up config for pager result limit

ashmrtn:
Allow callers of GetAddedRemovedItems to set the number of added items
they'd like returned from the call. Will eventually be leveraged for
Exchange preview backups

Doesn't change existing limits for pagers since everything uses the
default of zero for "get everything"

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/handlers.go
+++ b/src/internal/m365/collection/exchange/handlers.go
@@ -29,7 +29,7 @@ type addedAndRemovedItemGetter interface {
 	GetAddedAndRemovedItemIDs(
 		ctx context.Context,
 		user, containerID, oldDeltaToken string,
-		cc api.CallConfig,
+		config api.CallConfig,
 	) (pagers.AddedAndRemoved, error)
 }
 
--- a/src/pkg/services/m365/api/events_pager.go
+++ b/src/pkg/services/m365/api/events_pager.go
@@ -244,7 +244,7 @@ func (p *eventDeltaPager) ValidModTimes() bool {
 func (c Events) GetAddedAndRemovedItemIDs(
 	ctx context.Context,
 	userID, containerID, prevDeltaLink string,
-	cc CallConfig,
+	config CallConfig,
 ) (pagers.AddedAndRemoved, error) {
 	ctx = clues.Add(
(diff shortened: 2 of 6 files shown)


src/internal/m365/collection/exchange/handlers.go
```diff
@@ -29,7 +29,7 @@ type addedAndRemovedItemGetter interface {
 	GetAddedAndRemovedItemIDs(
 		ctx context.Context,
 		user, containerID, oldDeltaToken string,
-		cc api.CallConfig,
+		config api.CallConfig,
```
ryanfkeepers, line 32: request: lets standardize on a name and refactor all api.CallConfig in one full sweep, rather than piecemeal the renaming.  `config` is fine.  I like `cfg`, personally.  Feelings are not strong either way.

ryanfkeepers (approved):

src/pkg/services/m365/api/events_pager.go
```diff
@@ -244,7 +244,7 @@ func (p *eventDeltaPager) ValidModTimes() bool {
 func (c Events) GetAddedAndRemovedItemIDs(
 	ctx context.Context,
 	userID, containerID, prevDeltaLink string,
-	cc CallConfig,
+	config CallConfig,
```
pandeyabs, line 247: Should we add `LimitResults` and other configs to clues?

pandeyabs (approved):

src/pkg/services/m365/api/events_pager.go
ashmrtn: hmmm let's actually punt that to another PR. I forgot the config isn't passed through to the pagers so either it's going to be some moving structs around or adding lots of individual logging statements. Both of those will generate a fair number of changes so it'd be best to do it separate from these changes

ashmrtn: created #4681 to track this

PR #4661, opened 2023-11-11
use counter in exchange backup enum

ryanfkeepers:
Increase proliferation of the count bus to record runtime stats.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/exchange/backup.go
```diff
@@ -278,7 +296,7 @@ func populateCollections(
 
 	collections["metadata"] = col
 
-	logger.Ctx(ctx).Infow("produced collections", "count_collections", len(collections))
+	logger.Ctx(ctx).Infow("produced collections", "stats", counter.Values())
```
pandeyabs, line 299: Sanity check, since this is not a local count bus, we'll end up logging stats from parent funcs too right?

src/pkg/count/keys.go
```diff
@@ -1,42 +1,62 @@
...
+	ItemsAdded           Key = "items-added"
+	ItemsRemoved         Key = "items-removed"
+	MissingDelta         Key = "missing-delta-token"
+	NewDeltas            Key = "new-delta-tokens"
+	NewPrevPaths         Key = "new-previous-paths"
+	NoDeltaQueries       Key = "cannot-make-delta-queries"
+	PrevDeltas           Key = "previous-deltas"
 	// amounts reported by data providers
-	ProviderItemsRead key = "provider-items-read"
+	ProviderItemsRead Key = "provider-items-read"
```
pandeyabs, line 40: Could you please add a description for some of these?

src/pkg/count/keys.go
```diff
@@ -1,42 +1,62 @@
 package count
 
-type key string
+type Key string
```
pandeyabs, line 3: Curious, what's the reason for exporting this?

pandeyabs (approved):

src/internal/m365/collection/exchange/backup.go
ryanfkeepers: Ah, yeh.  I've swapped the counter passed into m365 `ProduceBackupCollections()` from operations to be a local instance, so that it more cleanly encapsulates the producer stats relative to other ops level stats.

src/pkg/count/keys.go
ryanfkeepers: I've added a few.  Kinda difficult from my perspective as implementor to know what makes sense and what doesn't.  If you have additional pointers to which ones are confusing and uncommented, I'll do what I can to clarify them.  

Fwiw, the goal is to have the variable names be self explanatory.  There's too much mental model to keep in place if they don't work that way.  So I'm also very open to improvements we can make with the names itself.

src/pkg/count/keys.go
ryanfkeepers: Hah, I don't entirely remember.  Maybe I used it?  If there's no exported use by the end of the pr chain I'll unexport it.

src/pkg/count/keys.go
pandeyabs: I was confused mostly by this particular key. Maybe we can call it `kopia-items-read`? or `stream-items-read`

ryanfkeepers: Sure, I'll wordsmith on it.

PR #4662, opened 2023-11-11
add count bus addRead and incRead funcs

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/pkg/count/count.go
+++ b/src/pkg/count/count.go
@@ -41,7 +41,7 @@ func (b *Bus) Inc(k Key) {
 	b.Add(k, 1)
 }
 
-// Inc increases the count by 1.
+// IncRead increases the count by 1.
 // returns the current local value of k.
 func (b *Bus) IncRead(k Key) int64 {
 	if b == nil {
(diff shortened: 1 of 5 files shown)


src/pkg/count/count.go
```diff
@@ -54,6 +66,18 @@ func (b *Bus) Add(k Key, n int64) {
 	}
 }
 
+// Add increases the count by n.
+// returns the current local value of k.
+func (b *Bus) AddRead(k Key, n int64) int64 {
```
pandeyabs, line 71: nit: `IncReadN` might a be more natural counterpart to `IncRead`?

pandeyabs (approved):

src/pkg/count/count.go
ryanfkeepers: Eh, it's supposed to mimic the base func.  Ex:
```
Inc() -> IncRead()
Add() -> AddRead()
```

That said, I'm honestly not happy with the `FooRead()` name structure in the first place.  Suggestions?

pandeyabs: We could remove `FooRead` and have `Inc` and `Add` return the current value of counter. The caller can discard return val if not needed. Is this go kosher?

ryanfkeepers: it's certainly go kosher.  I actually initially went that route for implementation, then backed off because I wasn't sure I was happy with the two-op on every increment.  But, you know, we can probably eat that cost and be fine.

PR #4663, opened 2023-11-11
use count.Bus in exchange coll item streaming

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/exchange/collection.go
```diff
@@ -194,13 +194,17 @@ func (col *prefetchCollection) streamItems(
...
 
 		user = col.user
-		log  = logger.Ctx(ctx).With(
-			"service", path.ExchangeService.String(),
-			"category", col.Category().String())
 	)
 
+	ctx = clues.Add(
+		ctx,
+		"item_stream_category", col.Category().String())
```
pandeyabs, line 201: nit: `category` is a well known label, should we keep it? Also add back the `service`?

src/internal/m365/collection/exchange/collection.go
```diff
@@ -299,6 +310,12 @@ func (col *prefetchCollection) streamItems(
 
 			stream <- item
 
+			col.Counter.Add(count.StreamBytesAdded, info.Size)
+
+			if col.Counter.IncRead(count.StreamItemsAdded)%1000 == 0 {
+				logger.Ctx(ctx).Infow("item stream progress", "stats", col.Counter.Values())
```
pandeyabs, line 316: Could you modify this log a bit so that we can distinguish item removals from additions?

src/internal/m365/collection/exchange/collection.go
```diff
@@ -234,6 +238,10 @@ func (col *prefetchCollection) streamItems(
 
 			stream <- data.NewDeletedItem(id)
 
+			if col.Counter.IncRead(count.StreamItemsRemoved)%1000 == 0 {
+				logger.Ctx(ctx).Infow("item stream progress", "stats", col.Counter.Values())
```
pandeyabs, line 242: Do we need to log all `counter.Values()` everywhere? Can we just log the count key which is mutating?

I see that the `defer` block is logging all values already, that should be sufficient.

pandeyabs (approved):
LGTM, other than some nits.

src/internal/m365/collection/exchange/collection.go
ryanfkeepers: Hmm, sure, I suppose we can.  The worry was that we're clobbering the category value established at a higher scope, but that probably isn't much of a worry.

src/internal/m365/collection/exchange/collection.go
ryanfkeepers: can do!

src/internal/m365/collection/exchange/collection.go
ryanfkeepers: Is there any drawback to logging the full set?

pandeyabs: Seems unnecessary, especially in this case. In a pathological case you'd have 100s of stats, all the same except for one which is relevant & mutating. Harder to inspect while reviewing logs, it's like finding needle in a haystack. Unless you know the key name you are looking for.  Also, more chatty logs = more scroll time. My fingers hurt!

Don't know if there is a huge penalty for serializing the map to logger every time. But given this is a concurrent map, there might be some? Still, an extreme case and I'm not raising this concern.

pandeyabs: Hmm. Beyond this particular case, my concern is that we will end up using `.Values()` pattern everywhere. If we are doing say an `Infow` log with a key, val pair, we shouldn't dump all the stats by default unless they are meaningful in that context.

PR #4674, opened 2023-11-14
Add limiter reset functionality

pandeyabs:
This PR adds `Reset` support to limiter. In summary, it provides 2 things:
1. It resets limiter state & refills tokens to original capacity.
2. Reset is transparent to request senders, they can continue to call `Wait` even when reset is in progress. This is done using a new request level lock. It'll be useful towards `WaitN` implementation as well in a following PR. 

Reason for adding `Reset`:
1. Corso CLI supports backing up the entire tenant. Since exchange service limits are defined on a per app per mailbox basis, if we don't reset the sliding window limiter in between  n exchange resources, it'll artificially slow down n-1 resource backups. 
2. This is because sliding window limiter front allows front loading the requests. This is not a big problem with token bucket limiter, since it refills tokens at a constant rate of 16 per second. 
3. Ideally the rate limiter instance should be scoped to an exchange mailbox, and no `Reset` should be required. However, this would need some major refactoring of graph controller. Currently we use global rate limiter instances, and also the graph controller stays common for all resource backup operations. `Reset` allows us to keep using the global approach. 

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/common/limiters/sliding_window.go
```diff
@@ -107,8 +115,42 @@ func (s *slidingWindow) Shutdown() {
 	})
 }
 
+// Reset resets the limiter to its initial state. Any pending requests will be
+// cancelled. Reset should only be called when the limiter is not in use.
+// No need to shutdown slide goroutine. prev and curr
```
ashmrtn, line 120: nit: hanging comment. Also since this is more about the implementation it should be moved into the function body itself so it doesn't show up in godoc (unless it's removed completely if it's not needed anymore)

src/internal/common/limiters/sliding_window_test.go
```diff
@@ -283,3 +283,114 @@ func (suite *SlidingWindowUnitTestSuite) TestShutdown() {
...
+	for i := 0; i < sw.numIntervals; i++ {
+		require.Equal(t, 0, sw.prev.count[i])
+		require.Equal(t, 0, sw.curr.count[i])
+	}
+}
+
+// TestResetDuringActiveRequests tests if reset is transparent to any active
+// requests and they are not affected. It also checks that limiter stays
+// within capacity limits post reset.
+func (suite *SlidingWindowUnitTestSuite) TestResetDuringActiveRequests() {
```
ashmrtn, line 340: nice, I like this test! It'll be useful going forward to make sure we don't accidentally get NPEs or use old references if we ever need to tinker with internal state of the limiter

ashmrtn (approved):

src/internal/common/limiters/sliding_window.go
pandeyabs: Ah thanks! I don't know how this comment ended here :O

PR #4680, opened 2023-11-14
Integrate sliding window limiter into exchange backup operation

pandeyabs:
2nd last PR in the chain. Wires up the new sliding limiter into exchange backup create flows. I'll add some e2e regression tests to wrap it up.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup.go
+++ b/src/internal/operations/backup.go
@@ -338,6 +338,7 @@ func (op *BackupOperation) do(
 	backupID model.StableID,
 ) (*details.Builder, error) {
 	lastBackupVersion := version.NoBackup
+	service := op.Selectors.PathService()
 
 	reasons, err := op.Selectors.Reasons(op.account.ID(), false)
 	if err != nil {
@@ -385,7 +386,7 @@ func (op *BackupOperation) do(
 	//   * the base finder code to skip over older bases (breaks isolation a bit
 	//     by requiring knowledge of good/bad backup versions for different
 	//     services)
-	if op.Selectors.PathService() == path.GroupsService {
+	if service == path.GroupsService {
 		if mans.MinBackupVersion() != version.NoBackup &&
 			mans.MinBackupVersion() < version.Groups9Update {
 			logger.Ctx(ctx).Info("dropping merge bases due to groups version change")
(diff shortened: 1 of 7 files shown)


src/internal/operations/backup.go
```diff
@@ -424,11 +425,20 @@ func (op *BackupOperation) do(
...
 	}
 
+	// Use sliding window limiter for Exchange if the feature is not explicitly
+	// disabled. For other services we don't use sliding window limiter.
+	enableSlidingLim := false
+	if service == path.ExchangeService &&
+		!op.Options.ToggleFeatures.DisableSlidingWindowLimiter {
+		enableSlidingLim = true
+	}
+
```
ryanfkeepers, line 435: We're getting into bad habits here.  The operation layer should not investigate the service being handled.  These sorts of conditionals need to be handled via an interface that's common to all backup providers.  Something like `ctx = op.backupProducer.SetRateLimiter(ctx, op.Selector, op.Options)`.

ryanfkeepers (changes requested):

src/internal/operations/backup.go
pandeyabs: Good point! I'll make the change.

pandeyabs:
@ryanfkeepers this is ready for a final review. Thank you!

src/internal/operations/backup.go
```diff
@@ -326,6 +322,12 @@ func (op *BackupOperation) Run(ctx context.Context) (err error) {
 		logger.Ctx(ctx).Infow("completed backup", "results", op.Results)
 	}
 
+	// For exchange, rate limits are enforced on a mailbox level. Reset the
+	// rate limiter so that it doesn't accidentally throttle following mailboxes.
+	// This is a no-op if we are using token bucket limiter since it refreshes
+	// tokens on a fixed per second basis.
+	graph.ResetLimiter(ctx)
```
ryanfkeepers, line 329: ctx aren't mutable, you're slicing out the changes:
```suggestion
	ctx = graph.ResetLimiter(ctx)
```

src/pkg/services/m365/api/graph/concurrency_middleware.go
```diff
@@ -193,6 +223,11 @@ func QueueRequest(ctx context.Context) {
 	}
 }
 
+func ResetLimiter(ctx context.Context) {
```
ryanfkeepers, line 226: Can you add a comment explaining how this affects the limiter, and when it should reasonably be used?

ryanfkeepers (approved):
You have a bug, but otherwise things look good.  Thanks!

PR #4683, opened 2023-11-14
clean up unused graph error sentinels

ryanfkeepers:
A handful of the error sentinel values in graph/errors.go were only around to comply with an old expectation that IsErr funcs have a matching sentinel.  That assumption is out of date, and there's no point in keeping around unused sentinels.

This PR makes three changes to clean up that issue:
1. remove unused sentinels (unused is qualifed as "not produced").
2. fix external pacakges which referenced those errors in tests alone by replacing those sentinels with odata errors.
3. centralize odata error test code to help support upstream package odata error creation.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/item_collector_test.go
+++ b/src/internal/m365/collection/drive/item_collector_test.go
@@ -6,7 +6,6 @@ import (
 	"github.com/alcionai/clues"
 	"github.com/google/uuid"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
-	"github.com/microsoftgraph/msgraph-sdk-go/models/odataerrors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
@@ -22,6 +21,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 	"github.com/alcionai/corso/src/pkg/services/m365/api/graph"
+	graphTD "github.com/alcionai/corso/src/pkg/services/m365/api/graph/testdata"
 	"github.com/alcionai/corso/src/pkg/services/m365/api/mock"
 )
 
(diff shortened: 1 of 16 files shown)


ashmrtn:
can we have a design discussion about this at some point? I really don't think making corso rely on graph errors outside the API package is going to be maintainable in the future. It also subverts some newer stuff like error label counters (because we're neither adding nor using error labels), spreads knowledge of specific graph errors through the codebase (making tests fragile because of it), doesn't allow easy grouping of errors/tests for those grouped errors (i.e. by translating to a sentinel error), and goes against the golang pattern of checks like `errors.Is` and sentinel errors

ryanfkeepers:
Tech debt ticket as mentioned by Ash above: https://github.com/alcionai/corso/issues/4685

src/pkg/services/m365/api/graph/errors.go
```diff
@@ -241,14 +222,12 @@ func IsErrConnectionReset(err error) bool {
...
-		errors.Is(err, ErrTokenExpired) ||
-		errors.Is(err, ErrTokenInvalid)
+		errors.Is(err, ErrTokenExpired)
 }
 
 func IsErrBadJWTToken(err error) bool {
 	return hasErrorCode(err, invalidAuthenticationToken) ||
-		errors.Is(err, ErrTokenExpired) ||
-		errors.Is(err, ErrTokenInvalid)
+		errors.Is(err, ErrTokenExpired)
```
pandeyabs, line 230: `ErrTokenExpired` can be removed from this. We created this error to indicate download URL jwt expiry. Adapter `Send` where `IsErrBadJWTToken` gets used doesn't do an inspection of download URL expiries.

src/internal/m365/collection/drive/item_collector_test.go
```diff
@@ -38,15 +38,6 @@ const (
 	userMysiteNotFound    = "ResourceNotFound User's mysite not found"
 )
 
-func odErr(code string) *odataerrors.ODataError {
```
pandeyabs, line 41: nice! Quick heads up, we also duplicate this func in services/*/enabled_test.go . See [example](https://github.com/alcionai/corso/blob/1cbabdb73a14854fc97fad87f835ab1edac5e7b7/src/internal/m365/service/exchange/enabled_test.go#L52).

These can be cleaned up as well.

pandeyabs (approved):

src/internal/m365/collection/drive/item_collector_test.go
ryanfkeepers: thanks for the heads up!

PR #4684, opened 2023-11-15
use clues new|wrap|stackWC()

ryanfkeepers:
replace all cases of new|wrap|stack(...).WithClues(ctx) with the flattened new|wrap|stackWC(ctx, ...) functions introduced in the latest clues bump.

Other changes added:
* remove WithClues builders when the error producer already called it.
* corrected some usages of ictx within loops.

no logic changes, just cleanup.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -187,7 +187,7 @@ func genericCreateCommand(
 
 		bo, err := r.NewBackupWithLookup(ictx, discSel, ins)
 		if err != nil {
-			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
+			errs = append(errs, clues.WrapWC(ictx, err, owner))
 			Errf(ictx, "%v\n", err)
 
 			continue
@@ -208,7 +208,7 @@ func genericCreateCommand(
 				continue
 			}
 
-			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
+			errs = append(errs, clues.WrapWC(ictx, err, owner))
 			Errf(ictx, "%v\n", err)
 
 			continue
--- a/src/pkg/export/consume.go
+++ b/src/pkg/export/consume.go
@@ -28,15 +28,15 @@ func ConsumeExportCollections(
 		folder := filepath.Join(exportLocation, col.BasePath())
 		ictx := clues.Add(ctx, "dir_name", folder)
 
-		for item := range col.Items(ctx) {
+		for item := range col.Items(ictx) {
 			if item.Error != nil {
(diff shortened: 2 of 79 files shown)


src/cli/backup/backup.go
```diff
@@ -187,7 +187,7 @@ func genericCreateCommand(
 
 		bo, err := r.NewBackupWithLookup(ictx, discSel, ins)
 		if err != nil {
-			errs = append(errs, clues.Wrap(err, owner).WithClues(ictx))
+			errs = append(errs, clues.WrapWC(ictx, err, owner))
```
ashmrtn, line 190: do you happen to know if it's possible to make a lint rule for this? It seems like one of those things that will be difficult to get everyone on the same page for without any instances of the old form getting committed

It's been a while since I added something to forbidigo, but I think it would only match on either `Wrap`/`Stack` or `WithClues`, but not both. I'm not sure what [tree-sitter](https://tree-sitter.github.io/tree-sitter/) (used for trailing `)`) would be able to match on. Maybe both since iirc it did seem to support a fairly complex grammar

src/pkg/export/consume.go
```diff
@@ -28,15 +28,15 @@ func ConsumeExportCollections(
 		folder := filepath.Join(exportLocation, col.BasePath())
 		ictx := clues.Add(ctx, "dir_name", folder)
 
-		for item := range col.Items(ctx) {
+		for item := range col.Items(ictx) {
 			if item.Error != nil {
-				el.AddRecoverable(ictx, clues.Wrap(item.Error, "getting item").WithClues(ctx))
+				el.AddRecoverable(ictx, clues.Wrap(item.Error, "getting item"))
```
ashmrtn, line 33: just checking, but will we lose `dir_name` (added on line 29) if we don't do `WrapWC` here?

ashmrtn (approved):

src/cli/backup/backup.go
ryanfkeepers: I'm a little hesitant about adding a linter here, since the rule is not `"only ever use WrapWC"`.  The rule is, actually, `"don't use WithClues, use FooWC(ctx, ...)`.  Though, now that I think about it, that's an easier linter anyway...

src/pkg/export/consume.go
ryanfkeepers: Standard assumption is that any corso-owned func that takes in a context (`Items(ctx)` in this case) is responsible for calling WithClues(ctx) on any error it generates.  In this case, we'll only lose `dir_name` if `Items()` is misbehaving.

PR #4687, opened 2023-11-15
Add functions used for restore to existing service-level handlers

ashmrtn:
This continues the push towards having service-level handlers that know
how to perform different operations. It adds the helper functions that
will be used during restore operations to the existing handler code

This logic is not currently used nor does this PR change the restore
call path

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4254

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/groups/export.go
+++ b/src/internal/m365/service/groups/export.go
@@ -26,7 +26,7 @@ var _ inject.ServiceHandler = &groupsHandler{}
 func NewGroupsHandler(
 	opts control.Options,
 	apiClient api.Client,
-	resourceClient idname.GetResourceIDAndNamer,
+	resourceGetter idname.GetResourceIDAndNamer,
 ) *groupsHandler {
 	return &groupsHandler{
 		baseGroupsHandler: baseGroupsHandler{
@@ -35,7 +35,7 @@ func NewGroupsHandler(
 			backupSiteIDWebURL: idname.NewCache(nil),
 		},
 		apiClient:      apiClient,
-		resourceClient: resourceClient,
+		resourceGetter: resourceGetter,
 	}
 }
 
@@ -159,7 +159,7 @@ func (h *baseGroupsHandler) ProduceExportCollections(
 type groupsHandler struct {
 	baseGroupsHandler
 	apiClient      api.Client
-	resourceClient idname.GetResourceIDAndNamer
+	resourceGetter idname.GetResourceIDAndNamer
 }
 
 func (h *groupsHandler) IsServiceEnabled(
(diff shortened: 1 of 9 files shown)


src/internal/m365/service/groups/export.go
```diff
@@ -17,20 +17,34 @@ import (
...
 
 func NewGroupsHandler(
 	opts control.Options,
-) *baseGroupsHandler {
-	return &baseGroupsHandler{
-		opts:               opts,
-		backupDriveIDNames: idname.NewCache(nil),
-		backupSiteIDWebURL: idname.NewCache(nil),
+	apiClient api.Client,
+	resourceClient idname.GetResourceIDAndNamer,
```
ryanfkeepers, line 29: naming nit, I'd prefer Getter since this isn't a separate client in the way that api.Client is.  Makes it a little clearer what the expectations are for the complying struct.  IE: that it only needs to fulfill the methods; not that it maintains a connection to a resource.
```suggestion
	resourceGetter idname.GetResourceIDAndNamer,
```

src/internal/m365/service/groups/export.go
```diff
@@ -134,3 +148,39 @@ func (h *baseGroupsHandler) ProduceExportCollections(
...
+	baseGroupsHandler
+	apiClient      api.Client
+	resourceClient idname.GetResourceIDAndNamer
+}
+
+func (h *groupsHandler) IsServiceEnabled(
+	ctx context.Context,
+	resource string,
+) (bool, error) {
+	// TODO(ashmrtn): Move free function implementation to this function.
```
ryanfkeepers, line 169: Question: what is `free function` in this context?

ryanfkeepers (approved):
lgtm

src/internal/m365/service/groups/export.go
ashmrtn: ah, `IsServiceEnabled()` in this case. Right now it's still setup to be called from something more generic. Since we're now working to centralize things in the service level handler it seems like it would be good to move that here too

In general "free function" is a function that's not on a struct/class (linking [SO](https://stackoverflow.com/a/4863328) cause I don't want to type a full code example)

ryanfkeepers: Huh, interesting.  I've always referred to that as the difference between `methods` (not free) and `functions` (free).

ashmrtn: yea, I think that's also terminology that's used. I'm honestly kind of terrible at remembering that though and tend to use "methods" and "functions" interchangeably 😅

ryanfkeepers: hah, you're not alone

Issue #4689, opened 2023-11-15
Introduce drive backup delta tree


ryanfkeepers:
Subset of work for https://github.com/alcionai/corso/issues/4688

- [x] Introduce the tree/node structures.
- [x] Node (aka: folder) apis and tests.
- [x] File apis and tests.
- [x] Previous-Paths apis and tests.
- [x] Post-Processing apis and tests.

ryanfkeepers:
Tracking forget-me-nots:
- [ ] cleanup delta tree interface to minimize direct value usage
- [ ] ensure clues traces are used
- [ ] ensure NewTombstoneCollection is used where possible
- [ ] remove any commented code
- [ ] add delta tree test for empty id on add file
- [ ] expand addFile tests to cover custom drive item and properties
- [ ] replace `42` with item size const

PR #4692, opened 2023-11-15
add delta tree

ryanfkeepers:
introduces the delta tree in drive collections

---

- [x] :no_entry: No

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -1,0 +1,80 @@
+package drive
+
+import (
+	"time"
+
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+// folderyMcFolderFace owns our delta processing tree.
+type folderyMcFolderFace struct {
+	// tenant/service/resource/category/driveID
+	// (or whatever variant the service defines)
+	// allows the tree to focus only on folder structure,
+	// and minimizes the possibility of multi-prefix path bugs.
+	prefix path.Path
+
+	// the root of the tree;
+	// new, moved, and notMoved collections
+	collections *nodeyMcNodeFace
+
+	// the majority of operations we perform can be handled with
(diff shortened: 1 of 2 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -0,0 +1,80 @@
...
+	// allows the tree to focus only on folder structure,
+	// and minimizes the possibility of multi-prefix path bugs.
+	prefix path.Path
+
+	// the root of the tree;
+	// new, moved, and notMoved collections
+	collections *nodeyMcNodeFace
+
+	// the majority of operations we perform can be handled with
+	// a folder id lookup instead of re-walking the entire tree.
```
ashmrtn, line 22: nit: just capitalization
```suggestion
	// a folder ID lookup instead of re-walking the entire tree.
```

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -0,0 +1,80 @@
+package drive
+
+import (
+	"time"
+
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+// folderyMcFolderFace owns our delta processing tree.
+type folderyMcFolderFace struct {
```
ashmrtn, line 10: I'm not sure if I should be happy or sad about winning by default (seems like other's didn't vote) 😂

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: Well that's their fault for not reading the proposal.

pandeyabs: Hope this name is temporary!

Issue #4690, opened 2023-11-15
Drive Delta Backup Tree CLI and Feature Flags


ryanfkeepers:
Subset of work for https://github.com/alcionai/corso/issues/4688:

- [x] Add a configuration feature-flag to toggle backup-by-tree.
- [x] Reverse the feature flag (now runs the prior code).
- [ ] Cleanup: remove the feature flag.

PR #4694, opened 2023-11-15
add toggle for using the delta tree via cli

ryanfkeepers:

- [x] :no_entry: No

#### Issue(s)

* #4690

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/cli/backup/onedrive.go
+++ b/src/cli/backup/onedrive.go
@@ -67,6 +67,12 @@ func addOneDriveCommands(cmd *cobra.Command) *cobra.Command {
 
 		flags.AddUserFlag(c)
 		flags.AddGenericBackupFlags(c)
+		fs.BoolVar(
+			&flags.UseDeltaTreeFV,
+			flags.UseDeltaTreeFN,
+			false,
+			"process backups using the delta tree instead of standard enumeration")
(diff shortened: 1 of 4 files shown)


ashmrtn:
I see this and agree we'll need at least a flag in `control.Options` so we can test this. I'm a little unsure about the need for a CLI flag in main since we don't need to expose this to users at all (I do realize it would make testing easier)

I'd also like to hold off approving/merging this PR out of an abundance of caution until we feel we're more ready to flip the switch. It feels a bit premature right now since we don't have any of the wiring or a full implementation of the delta tree right now

ryanfkeepers:
> I'd also like to hold off approving/merging this PR out of an abundance of caution until we feel we're more ready to flip the switch. It feels a bit premature right now since we don't have any of the wiring or a full implementation of the delta tree right now

@ashmrtn I understand the concerns, but I think you're being overly conservative about the situation.  Precautions will be taken through deployment to ensure that this cannot put users' backups into a bad state.

While I personal consider alternatives like env vars to have the same level of risk as hidden flags, I'm open to using that implementation (or any similar) instead of a flag if it eases concerns.

ashmrtn (approved):

PR #4701, opened 2023-11-16
move service api to client methods

ryanfkeepers:
sdk consumers need a way to configure the graph api client options when using the services api.  This is the first in a two-part change to expose those options.  This step moves the api to a client-based set of methods instead of free functions.  The next step will add client configuration to the service client.

---

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/m365.go
+++ b/src/pkg/services/m365/m365.go
@@ -11,6 +11,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/fault"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
+	"github.com/alcionai/corso/src/pkg/services/m365/api/graph"
 )
 
 type client struct {
@@ -20,8 +21,9 @@ type client struct {
 func NewM365Client(
 	ctx context.Context,
 	acct account.Account,
+	opts ...graph.Option,
 ) (client, error) {
-	ac, err := makeAC(ctx, acct)
+	ac, err := makeAC(ctx, acct, opts...)
 	return client{ac}, clues.Stack(err).OrNil()
 }
 
@@ -40,6 +42,7 @@ type getAller[T any] interface {
 func makeAC(
 	ctx context.Context,
 	acct account.Account,
+	opts ...graph.Option,
 ) (api.Client, error) {
 	// exchange service inits a limit to concurrency.
 	api.InitConcurrencyLimit(ctx, path.ExchangeService)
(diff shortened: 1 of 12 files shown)


src/pkg/services/m365/m365.go
```diff
@@ -28,9 +40,9 @@ type getAller[T any] interface {
 func makeAC(
 	ctx context.Context,
 	acct account.Account,
-	pst path.ServiceType,
 ) (api.Client, error) {
-	api.InitConcurrencyLimit(ctx, pst)
+	// exchange service inits a limit to concurrency.
+	api.InitConcurrencyLimit(ctx, path.ExchangeService)
```
ashmrtn, line 45: just checking: how does this work if called multiple times? At least for the test suite setup in `m365/groups_test.go` in this PR the test suite will call `graph.InitializeConcurrencyLimiter` and then it'll call `NewM365Client` which will call this and this will call `api.InitConcurrencyLimit` with a service that isn't technically what we want for a groups test

I haven't traced how the concurrency limiter would be set of backup or restore operations so I don't know if this could affect corso prod code

src/pkg/services/m365/m365_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (suite *userIntegrationSuite) TestNewM365Client_invalidCredentials() {
+	table := []struct {
+		name string
+		acct func(t *testing.T) account.Account
+	}{
+		{
+			name: "Invalid Credentials",
+			acct: func(t *testing.T) account.Account {
```
ashmrtn, line 34: there's actually a `tester.NewFakeM365Account` function you can use now if you'd be interested

ashmrtn (commented):
LGTM, but want confirmation on calls to initialize the concurrency limiter first

src/pkg/services/m365/m365_test.go
ryanfkeepers: Heyo!  Sure am interested.

src/pkg/services/m365/m365.go
ryanfkeepers: It should no-op.  That said, thanks for sighting the double inits.  No point in keeping them in.

ashmrtn: is this going to misconfigure prod limits if we're not careful though?

ryanfkeepers: Ehh, technically could, if the caller used this in the same process as an operations.Backup or restore.  I don't think that's happening anywhere, though.  Service api processes are all isolated.

ashmrtn: mkay, thanks for clarifying!

ashmrtn (approved):

src/pkg/services/m365/m365.go
ryanfkeepers: sure thing!

PR #4704, opened 2023-11-16
bump xsync to v3

ryanfkeepers:
three changes
1. bumps the xsync package to v3
2. creates a common package for synced maps
3. replaces all xsync MapOf imports with the new common/syncd package.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/count/count.go
+++ b/src/pkg/count/count.go
@@ -1,7 +1,7 @@
 package count
 
 import (
-	"github.com/puzpuzpuz/xsync/v2"
+	"github.com/puzpuzpuz/xsync/v3"
 )
 
 // Bus handles threadsafe counting of arbitrarily keyed metrics.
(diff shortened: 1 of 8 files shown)


src/pkg/count/count.go
```diff
@@ -12,7 +12,7 @@ type Bus struct {
 
 func New() *Bus {
 	return &Bus{
-		stats: xsync.NewMapOf[*xsync.Counter](),
+		stats: xsync.NewMapOf[string, *xsync.Counter](),
```
meain, line 15: Why are we not using `syncd` here?

meain (approved):
Need a small change, but approving to unblock.

src/pkg/count/count.go
ryanfkeepers: primarily because I need to pull in xsync for the counter anyway.  Personally I think its an acceptable caveat to the centralized ownership.  Do you think that'll cause excess confusion?

PR #4713, opened 2023-11-17
Return stats directly from restore function

ashmrtn:
Rework restore return status so that later PRs will have a smaller diff

Instead of returning a status and then waiting on the message just return the restore stats directly. The status getter was only setup to wait for one item anyway and was setup to wait after the entire restore operation already completed (at end of
m365/restore.go:ConsumeRestoreCollections())

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4254

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers (commented):
While I'm in favor of the _pattern_ here, the intent is to replace this with the count.Bus anyway.  What's your preference: keep this change and undo it later?  Or drop it and we'll get around to it with the count.Bus update?

ashmrtn:
Let's keep this for now as it makes switching to service-level handlers much simpler. We can rearrange/remove this after that point

ryanfkeepers (approved):

PR #4714, opened 2023-11-17
Use service-handler for restore operations

ashmrtn:
Switch restore code path to use service-level handlers. Does a few
things:
* switches existing service-level functions to be methods on
  service-level handlers
* update interfaces as necessary
* moves some logic from old controller-level restore function to either
  the new handlers or the operation-level function
* removes old code

May be easiest to review by commit

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4254

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/service/sharepoint/restore.go
```diff
@@ -18,24 +17,35 @@ import (
...
 	ctx context.Context,
 	rcc inject.RestoreConsumerConfig,
-	ac api.Client,
-	backupDriveIDNames idname.Cacher,
 	dcs []data.RestoreCollection,
 	errs *fault.Bus,
 	ctr *count.Bus,
 ) (*details.Details, *data.CollectionStats, error) {
+	// TODO(ashmrtn): We should stop relying on the context for rate limiter stuff
+	// and instead configure this when we make the handler instance. We can't
```
ryanfkeepers, line 32: Two requests.   
1/ can you please make a tech debt ticket for this, instead of todo comments?  
2/ while i'm fine that we avoid hijacking the context for this, I worry that the lack of a suggested alternative leaves this open to going back to global singletons as a solution, which I'd like to avoid.  We've been pretty loose (ie: not centralized) when it comes to handling our http client configuration, and that's caused a handful of problems around statefulness and leaky configs.  It'd be nice for a refactor to look at solving all of that, and I worry a quick todo resolution won't have the necessary breathing room to solve many problems beyond the ctx usage itself.

ryanfkeepers (approved):

src/internal/m365/service/sharepoint/restore.go
ashmrtn: can we switch up how we do http client init a little? From my understanding, we need the client in the CLI layer to do some name resolution and in the operation itself. Would it be alright to make those two separate clients instead of reusing the same one? That would get us out of needing a global config that's initialized before we get all the info we need

The one question there is if it would scale nicely for something like `corso backup create exchange --user "*"`. If we tie the client to the operation (or functions like this) it's unclear how garbage collection/reuse of the old client's connection pool will go

src/internal/operations/restore.go
```diff
@@ -393,9 +390,16 @@ func consumeRestoreCollections(
...
 	ctr *count.Bus,
 ) (*details.Details, *data.CollectionStats, error) {
+	if len(dcs) == 0 {
+		return nil, nil, clues.New("no data collections to restore")
+	}
+
 	progressBar := observe.MessageWithCompletion(ctx, observe.ProgressCfg{}, "Restoring data")
 	defer close(progressBar)
 
+	ctx, end := diagnostics.Span(ctx, "m365:restore")
```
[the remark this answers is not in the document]
ashmrtn: whoops, copy/paste error. I'll switch to something like `operations` for this annotation

PR #4716, opened 2023-11-20
add boilerplate for drive tree processing

ryanfkeepers:
begins implementation of the drive delta tree support by adding boilerplate funcs for getting the backup data, and adding a framework of unit tests that will be used bring testing parity up to par with current tests.

---

- [x] :no_entry: No

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test


src/internal/m365/collection/drive/collections_tree.go
```diff
@@ -0,0 +1,325 @@
...
+// (list-based) design, and the in-progress (tree-based) redesign.
+// see: https://github.com/alcionai/corso/issues/4688
+
+func (c *Collections) getTree(
+	ctx context.Context,
+	prevMetadata []data.RestoreCollection,
+	ssmb *prefixmatcher.StringSetMatchBuilder,
+	errs *fault.Bus,
+) ([]data.BackupCollection, bool, error) {
+	ctx = clues.AddTraceName(ctx, "GetTree")
```
ashmrtn, line 31: should we prefix this with the service/data type?

src/internal/m365/collection/drive/collections_tree.go
```diff
@@ -0,0 +1,325 @@
...
+
+		prevDrivePath, err := c.handler.PathPrefix(c.tenantID, driveID)
+		if err != nil {
+			err = clues.WrapWC(ctx, err, "making drive tombstone for previous path").Label(count.BadPathPrefix)
+			el.AddRecoverable(ctx, err)
+
+			continue
+		}
+
+		coll, err := NewCollection(
```
ashmrtn, line 261: nit: how about using the `data.NewTombstoneCollection()` helper to keep things consistent? Unless maybe we still rely on being able to cast to a onedrive specific implementation of the collection? In which case could you add a TODO here about that

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -0,0 +1,521 @@
...
+	states        map[data.CollectionState]*stateAssertion
+	excludedItems map[string]struct{}
+}
+
+type statesToItemIDs map[data.CollectionState][]string
+
+func newCollAssertion(
+	doNotMerge bool,
+	itemsByState statesToItemIDs,
+	excludeItems ...string,
```
ashmrtn, line 147: I don't think this is the correct place to be taking excludes. Excludes are global to the entire test whereas this seems tied to the particular collection. I think this will end up clobbering previous excludes if this helper is used for multiple collections

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -0,0 +1,521 @@
...
+	default:
+		p = coll.FullPath()
+	}
+
+	for {
+		itm, ok := <-itemCh
+		if !ok {
+			break
+		}
+
```
ashmrtn, line 201: nit: range over the item channel so channel close doesn't have to be handled separately
```suggestion
	for itm := range <-itemCh {
```

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -0,0 +1,521 @@
...
+	defer flush()
+
+	var (
+		itemCh  = coll.Items(ctx, fault.New(true))
+		itemIDs = []string{}
+	)
+
+	var p path.Path
+
+	switch coll.State() {
```
ashmrtn, line 189: nit: seems like you could use `FullOrPrevPath` here

ashmrtn (approved):

src/internal/m365/collection/drive/collections_tree.go
ryanfkeepers: the trace naturally composes a branch of values within a process tree.  So, ideally, a parent container would have named (or more likely, auto-traced) the service at a higher point in the process.

src/internal/m365/collection/drive/collections_tree.go
ryanfkeepers: I can look at that later.  This is purely copy-paste from the original code, which is also using `NewCollection`.  I'll add a TODO for it.

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: hmm, fair.  I think I'll follow up on this later, since it's unused for the next handful of PRs.  When I get into rebuilding all of the mocked e2e tests (ie: the Get() tests) it'll be a natural time to revisit the design.

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: oh, good call!

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: I feel like I should have known that already 🤔

PR #4719, opened 2023-11-21
introduce flat deletions to delta tree

ryanfkeepers:
there are certain enumeration cases where tombstoning isn't the correct behavior, and we want to delete a folder from the tree entirely.  This primarily occurrs when we have a create->delete pair of markers either within or across enumerations.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -252,3 +282,38 @@ func (face *folderyMcFolderFace) SetTombstone(
...
+// enumeration, then deleted on the next delta.  The folder will have no previous
+// path, and no presence in our storage layer, and can be safely removed from the
+// tree.
+// 2. as a safety measure in mid-enumeration mutation.  If, during a delta
+// enumeration, the end user deletes a folder before the pager visits the item,
+// then the folder won't show up in the enumeration at all.  However, the next
+// delta will still contain a delete marker.  If this happens and we have no
+// previousPath metadata for the folder id, then there's nothing to tombstone
+// and this method can be safely called with a no-op.
+func (face *folderyMcFolderFace) DeleteFolder(id string) {
```
ashmrtn, line 298: Design question based off skimming the next PR for how this function will be used: could we instead move the decision of making a tombstone collection vs. "deleting" something to when we generate collections instead of during tree traversal? Based on the examples given, it seems like this choice centers around whether there's a previous path for the entry. Is there harm in collecting these entries as tombstones and then, when making tombstone collections, checking if a previous path exists for the entry and skipping making a collection if there's no path? That seems like it could save us passing around the previous paths map some. It also avoid special casing some tree additions/removals in favor of a potentially less nested check on previous paths when we're going to be accessing previous paths info anyway

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -8,11 +8,49 @@ import (
...
 	"github.com/alcionai/corso/src/pkg/path"
 )
 
+// ---------------------------------------------------------------------------
+// helpers
+// ---------------------------------------------------------------------------
+
+var loc = path.NewElements("root:/foo/bar/baz/qux/fnords/smarf/voi/zumba/bangles/howdyhowdyhowdy")
+
+func treeWithRoot() *folderyMcFolderFace {
```
ashmrtn, line 21: nit: could you please go through this function and the helpers below it and switch it to use API functions to manipulate the tree structure instead of manipulating it directly? That will make it so that these helpers require less maintenance going forward as APIs tend to be more stable than internal implementations

For example, this function could use `SetFolder` to add the root node instead of adding it directly to the tree

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: Ah, I'll put a pin in that and get around to it later.  Will save me some merge conflicts down the chain.

PR #4734, opened 2023-11-27
add new funcs and tidy up drive limiter

ryanfkeepers:
adds some new functions to the drive limiter that will be used specifically in the tree-based backup process. Also updates the limiter tests to have separate versions for the tree and non-tree variations.  In this PR, the tree variation will definitely fail.  The next PR will be focused on backup process corrections needed to ensure that the limit handling is correct according to the existing tests.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/limiter.go
```diff
@@ -81,25 +74,58 @@ func (l pagerLimiter) atContainerItemsLimit(numItems int) bool {
...
+
+// ---------------------------------------------------------------------------
+// Used by the tree version limit handling
+// ---------------------------------------------------------------------------
+
+// hitLimit returns true if the limiter is enabled and meets any of the
+// conditions for max items, containers, etc for this backup.
+func (l pagerLimiter) hitLimit(
+	pageCount, containerCount, itemCount int,
+	totalBytesCount int64,
```
ashmrtn, line 100: I'd much prefer keeping these values in a struct so that it's both easy to update the set of things that's being checked without having to go update a bunch of function signatures and so that it's easier for the dev calling the function because they don't need to memorize/lookup the order of parameters each time, especially since so many of them are the same type. This is somewhat similar to the reasons we've been trying to move to returning structs instead of repeated values of the same type

src/internal/m365/collection/drive/limiter.go
```diff
@@ -81,25 +74,58 @@ func (l pagerLimiter) atContainerItemsLimit(numItems int) bool {
...
+}
+
+// hitContainerLimit returns true if the limiter is enabled and the number of
+// unique containers added so far is beyond the limit for this backup.
+func (l pagerLimiter) hitContainerLimit(containerCount int) bool {
+	return l.enabled() && containerCount >= l.limits.MaxContainers
+}
+
+// hitItemLimit returns true if the limiter is enabled and has reached the limit
+// for unique items, or their accumulated size in bytes, added to collections for this backup.
```
ashmrtn, line 121: nit: stale comment
```suggestion
// for unique items for this backup.
```

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -98,14 +77,10 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits() {
...
 					idx(drive, 1): {
-						Pages: []mock.NextPage{{
-							Items: []models.DriveItemable{
-								driveRootItem(rootID), // will be present, not needed
-								driveItemWithSize(idx(file, 1), namex(file, 1), parent(1), rootID, 7, isFile),
-								driveItemWithSize(idx(file, 2), namex(file, 2), parent(1), rootID, 1, isFile),
-								driveItemWithSize(idx(file, 3), namex(file, 3), parent(1), rootID, 1, isFile),
-							},
-						}},
+						Pages: pagesOf(rootAnd(
```
ashmrtn, line 80: I'd really prefer test structure updates were done in separate PRs. When test structure updates are mixed with other updates it becomes much more difficult to see if the test assertions were changed in some unexpected way

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -54,36 +51,18 @@ func TestLimiterUnitSuite(t *testing.T) {
...
-	table := []struct {
-		name       string
-		limits     control.PreviewItemLimits
-		drives     []models.Driveable
-		enumerator mock.EnumerateItemsDeltaByDrive
-		// Collection name -> set of item IDs. We can't check item data because
-		// that's not mocked out. Metadata is checked separately.
-		expectedCollections map[string][]string
-	}{
+func backupLimitTable(drive1, drive2 models.Driveable) []backupLimitTest {
```
ashmrtn, line 64: nit: Passing these values in makes the tests fragile. These values are directly returned via the `drives` struct field, but the output of the enumerator builds the drive ID manually. This means that if any of the passed in drives doesn't follow the expected drive ID format things could break. It would probably be better to just construct the drives in this function instead of taking them as parameters

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -664,118 +525,45 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits() {
...
+// checks that don't examine metadata, collection states, etc. They really just
+// check the expected items appear.
+func (suite *LimiterUnitSuite) TestGet_PreviewLimits_noTree() {
+	metadataPath, err := path.BuildMetadata(
+		tenant,
+		user,
+		path.OneDriveService,
+		path.FilesCategory,
+		false)
+	require.NoError(suite.T(), err, "making metadata path", clues.ToCore(err))
```
ashmrtn, line 541: nit: could you please move this to `runGetPreviewLimits` since it's not used in this function beyond that? The same will likely apply to making the drives, depending on whether they're actually passed into the function that generates the test cases or not

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -880,151 +771,225 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_Defaults() {
...
+
+	drive1 := models.NewDrive()
+	drive1.SetId(ptr.To(idx(drive, 1)))
+	drive1.SetName(ptr.To(namex(drive, 1)))
+
+	// The number of pages the test generates can be controlled by setting the
+	// number of containers. The test will put one (non-root) container per page.
```
ashmrtn, line 812: nit: stray comment

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -880,151 +771,225 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_Defaults() {
...
+	require.NoError(suite.T(), err, "making metadata path", clues.ToCore(err))
+
+	drive1 := models.NewDrive()
+	drive1.SetId(ptr.To(idx(drive, 1)))
+	drive1.SetName(ptr.To(namex(drive, 1)))
 
-				for itemIdx := 0; itemIdx < test.numItemsPerContainer; itemIdx++ {
-					itemSuffix := fmt.Sprintf("%d-%d", containerIdx, itemIdx)
+	// The number of pages the test generates can be controlled by setting the
+	// number of containers. The test will put one (non-root) container per page.
```
ashmrtn, line 855: stray comment

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -880,151 +771,225 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_Defaults() {
...
+		tenant,
+		user,
+		path.OneDriveService,
+		path.FilesCategory,
+		false)
+	require.NoError(suite.T(), err, "making metadata path", clues.ToCore(err))
+
+	drive1 := models.NewDrive()
+	drive1.SetId(ptr.To(idx(drive, 1)))
+	drive1.SetName(ptr.To(namex(drive, 1)))
```
ashmrtn, line 809: nit: all of this could be done in `runGetPreviewLimitsDefaults`. Could you please move it there instead of having it in each separate top-level test function?

ashmrtn (approved):
needs some test fixups prior to merging since there's some cleanup that would make things simpler to manage

src/internal/m365/collection/drive/limiter.go
ryanfkeepers: This func gets deleted in the next PR, since I realized it wasn't getting used.  This version of it came from an assumption of need that didn't pan out.  As for keeping counts owned by a single struct, I think you'll be happy with the implementation in the next PR.

src/internal/m365/collection/drive/limiter_test.go
ryanfkeepers: fair, will do.

src/internal/m365/collection/drive/limiter_test.go
ryanfkeepers: You're right.  I should have them separated.  I believe I have refactors/changes basically discrete from here on out.  But feel free to ping me to split the PR if it's getting too rough to review.

src/internal/m365/collection/drive/limiter_test.go
ryanfkeepers: done

src/internal/m365/collection/drive/limiter_test.go
ryanfkeepers: done

PR #4735, opened 2023-11-28
swap stats for in-tree item counts for limits

ryanfkeepers:
In the tree-based implementation of drive collections.Get, swaps out the driveEnumerationStats and instead uses the current state of the tree to count the items within and compare those counts to the expected preview limits.

Tests are known to still be failing.  Next PR will bring things back to green.

---

#### Does this PR need a docs update or release note?

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

PR #4745, opened 2023-11-28
fix up tests

ryanfkeepers:
gets the original limiter tests back to green.  tree-based tests will need to wait on tree post-processing implementation. Largely refactoring for clarity with one minor bugfix.

PR bookend for:
* https://github.com/alcionai/corso/pull/4734
* https://github.com/alcionai/corso/pull/4735

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers:
@ashmrtn this bookended PR set is rebsed and ready for review.

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -551,7 +552,7 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_noTree() {
 	for _, test := range backupLimitTable(drive1, drive2) {
 		suite.Run(test.name, func() {
 			runGetPreviewLimits(
-				suite.T(),
+				t,
```
ashmrtn, line 555: this needs to stay as `suite.T()` else it will be using the top-level test instance to report errors/failures and not the subtest instance. Since this is a rather large PR I'm just going to put a comment on all the instances I see so they don't get missed

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -586,7 +591,7 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_tree() {
...
```
ashmrtn, line 594: same here as one of the other comments, could you please change this back to `suite.T()` so test failures are properly reported?

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -807,19 +814,19 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_Defaults_noTree() {
...
+	drive1.SetName(ptr.To(name(drive)))
 
 	// The number of pages the test generates can be controlled by setting the
 	// number of containers. The test will put one (non-root) container per page.
 
 	for _, test := range defaultLimitsTable() {
 		suite.Run(test.name, func() {
 			runGetPreviewLimitsDefaults(
-				suite.T(),
+				t,
```
ashmrtn, line 829: please also keep this as `suite.T()`

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -864,7 +875,7 @@ func (suite *LimiterUnitSuite) TestGet_PreviewLimits_Defaults_tree() {
...
```
ashmrtn, line 878: `suite.T()` here too please

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -247,8 +247,12 @@ type collectionAssertions map[string]collectionAssertion
 // 	}
 // }
 
-func rootAnd(items ...models.DriveItemable) []models.DriveItemable {
-	return append([]models.DriveItemable{driveItem(rootID, rootName, parent(0), "", isFolder)}, items...)
+func pageItems(items ...models.DriveItemable) []models.DriveItemable {
```
ashmrtn, line 250: doesn't need to be done in this PR, but at some point we should consolidate the helper functions into a single file. Right now we're using helpers defined in both this file and `collections_test.go` (maybe others too..?)

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -247,8 +247,12 @@ type collectionAssertions map[string]collectionAssertion
...
+	return append([]models.DriveItemable{driveRootItem()}, items...)
+}
+
+func drivePageItems(driveID any, items ...models.DriveItemable) []models.DriveItemable {
+	return append([]models.DriveItemable{driveRootItem()}, items...)
 }
```
ashmrtn, line 256: since the root item isn't actually populating the drive ID in the parent ref portion of the item (even though graph seems to) this function is exactly the same as `pageItems`. Could you please either switch `driveRootItem` to fill out the drive ID or just remove this function completely?

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -347,63 +365,63 @@ func (suite *CollectionsUnitSuite) TestPopulateDriveCollections() {
 		{
 			name: "Invalid item",
 			items: []models.DriveItemable{
-				driveRootItem(rootID),
-				driveItem(id(item), name(item), parent(drive), rootID, -1),
+				driveRootItem(),
+				driveItem(id(item), name(item), driveParentDir(drive), rootID, -1),
```
ashmrtn, line 369: these calls and calls to similar functions seem like incorrect usages of the helper functions by the standards that exist now (perhaps they were written when expectations were different). Could you please make another PR to update them to adhere to the new standards? It seems like most of them should be switched to just `parentDir` (or equivalent function)

ashmrtn (approved):
approving to unblock, but definitely needs simplification of test helper functions and fixes to which test instance is passed into functions that setup/run the test

src/internal/m365/collection/drive/limiter_test.go
ryanfkeepers: ah, good catch, thanks.  find-replace failures.

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: Test cleanup begins happening here and continues, interpolated with the other changes, across a series of PRs.  This suggestion is implemented in the next one, iirc.

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: Will remove.  I think I removed it in one of the later cleanups already, but might as well make the change here.

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: Going to keep this around in this PR.  It will get addressed in a future PR where I clean up the drive handling specifically.

src/internal/m365/collection/drive/collections_test.go
```diff
@@ -243,7 +243,19 @@ func toPath(elems ...string) string {
 	}
 }
 
-func fullPath(driveID any, elems ...string) string {
+func fullPath(elems ...string) string {
```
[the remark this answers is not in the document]
ryanfkeepers: Totally agree about the design woes here.  I took a "shortest path to victory" approach during the implementation churn (such as in this PR) and intend to rewrite the drive stub handling altogether.  The aim will be to avoid the possibility of gotchas and have a system of factories whose basic design enforces correctness.  Small example: `driveStub.FullPath(...)` instead of `fullPath(arbitraryDriveID, ...)`.  

> switch all...

Heh, jinx: I was already implementing that change while you wrote this review.

> rename...

yup. Will look for ways to not need to know the drive ID outside of handling the drive stub itself.

PR #4746, opened 2023-11-28
require rootID on tree construction

ryanfkeepers:
Turns out the root ID name isn't an appropriate match for establishing the root node.  Instead, the backup hander is now extended with a getRootFolder method and will pass the expected root folder ID into the tree's constructor func to ensure we establish the correct root node.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -45,9 +48,11 @@ type folderyMcFolderFace struct {
 
 func newFolderyMcFolderFace(
 	prefix path.Path,
+	rootID string,
 ) *folderyMcFolderFace {
 	return &folderyMcFolderFace{
 		prefix:           prefix,
+		rootID:           rootID,
```
ashmrtn, line 55: I'm a little uncomfortable with this initialization because it feels like it's sort of violating some of the invariants we want for the tree (or perhaps I'm unsure what those invariants should be given the addition of the root ID?)

One thing that stands out is that we take in the root ID but we don't actually add it as a node in the tree. We put that off until a later point. It doesn't seem like a bit issue, but kind of feels like an inconsistency in how we handle things. I think I'd be more comfortable if we made a call to `setFolder` with the root node's info in this function. That way the result is a tree with a completely initialized root node

ashmrtn (approved):
minor nit about initialization but otherwise LGTM

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: Aye, I hear you.  The primary reason I didn't go for that are twofold: 1/ I didn't want to interfere with a delta response that does _not_ contain the root folder (ie: no changes have been made since the last delta).  2/ I wanted a tree reset to allow for an easy clean slate, and preserving the root is just a little weirder in that case.

PR #4747, opened 2023-11-29
add multiple delta iteration to getTree

ryanfkeepers:
expands the drive item enumeration by running multiple delta queries per backup, with the intent to reach a no-change state in the given delta.  Most of the code here modifies the test structures to accomodate the possibility of a multi-delta enumeration within the tree version.  Plus renaming and other refactoring for quality of life.  Full test coverage of the change won't be included until the next PR.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/collections_test.go
```diff
@@ -1175,11 +1175,13 @@ func (suite *CollectionsUnitSuite) TestPopulateDriveCollections() {
 				errs     = fault.New(true)
 			)
 
-			mbh.DriveItemEnumeration = mock.EnumerateItemsDeltaByDrive{
-				DrivePagers: map[string]*mock.DriveItemsDeltaPager{
+			mbh.DriveItemEnumeration = mock.EnumerateDriveItemsDelta{
```
ryanfkeepers, line 1178: lots of chaff here.  all changes in this file should be the following refactor:
* `mock.EnumerateItemsDeltaByDrive` -> `mock.EnumerateDriveItemsDelta`
* `DrivePagers: map[string]*mock.DriveItemsDeltaPager` -> `DrivePagers: map[string]*mock.DeltaIterator`
* addition of `}},` at the block end

src/internal/m365/collection/drive/limiter_test.go
```diff
@@ -74,17 +74,15 @@ func backupLimitTable(drive1, drive2 models.Driveable) []backupLimitTest {
...
-					id(drive): {
-						Pages: pagesOf(pageItems(
-							driveItemWithSize(idx(file, 1), namex(file, 1), parentDir(), rootID, 7, isFile),
-							driveItemWithSize(idx(file, 2), namex(file, 2), parentDir(), rootID, 1, isFile),
-							driveItemWithSize(idx(file, 3), namex(file, 3), parentDir(), rootID, 1, isFile))),
-						DeltaUpdate: pagers.DeltaUpdate{URL: id(delta)},
-					},
-				},
-			},
+			enumerator: mock.SingleDriveDeltaEnumerator(
```
ryanfkeepers, line 77: lots of chaff changes in this file.  Everything should look as follows:
```go
// before
			enumerator: mock.EnumerateItemsDeltaByDrive{
 				DrivePagers: map[string]*mock.DriveItemsDeltaPager{
 					id(drive): {
 						Pages: pagesOf(pageItems(

// after
			enumerator: mock.SingleDriveDeltaEnumerator(
 				id(drive),
 				mock.DeltaEnum(
 					id(delta),
 					aDelta(aPage(
``

File: src/internal/m365/collection/drive/collections_tree.go [hunk 3]
```diff
@@ -277,65 +277,94 @@ func (c *Collections) populateTree(
...
-	// essentially a noop.
-	pager.Cancel()
+		currDeltaLink = du.URL
 
-	du, err := pager.Results()
-	if err != nil {
-		return du, clues.Stack(err)
+		// 0 pages is never expected.  We should at least have one (empty) page to
+		// consume.  But checking pageCount == 1 is brittle in a non-helpful way.
+		finished = pageCount < 2 && pageItemCount == 0 && !hadReset
```
ashmrtn, line 367: I'm unsure if we actually need to check if there was a reset or not. Perhaps we should go clarify the interface provided by the pager at some point to make this more clear

From my current understanding, when a reset is seen in the pager, it will send the reset status as a result (possibly with other items from after the reset), reset the pager, and then continue enumerating. Assuming that's the behavior we want to keep in the interface going forward, it seems like there's a few outcomes we could get here based on whether the pager can return additional items in the same result as the reset

**items in same result, no items returned**
Here we'll get `hadReset = true`, `pageCount = 1`, `pageItemCount = 0`. In theory, running another delta should return one empty page but without the reset

**items in same result, some items**
Here we'll get `hadReset = true`, `pageCount = 1`, `pageItemCount > 0`. We'll set `finished = false` regardless of if we check `hadReset` or not because of `pageItemCount`

**items in next result**
I'm consolidating not getting items and getting items for this because in both cases we'll get `hadReset = true` and `pageCount >= 2` (one result for reset, one for the next query). In both those cases `finished = false` because of `pageCount` so we'll do another query anyway

ashmrtn (approved):

src/internal/m365/collection/drive/collections_tree.go
```diff
@@ -277,65 +277,94 @@ func (c *Collections) populateTree(
...
+			// before seeing we've passed the limit.
+			if limiter.hitPageLimit(int(counter.Get(count.PagesEnumerated))) {
+				hitLimit = true
 				break
 			}
 
-			el.AddRecoverable(ctx, clues.Stack(err))
+			pageCount++
+
+			pageItemCount += len(page)
```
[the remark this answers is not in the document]
ryanfkeepers: > Alternatively, we could maybe pass a local count bus...

already done, see: https://github.com/alcionai/corso/pull/4795/files#diff-e491cdc5bf60f41e6277827adecca352d81745e7435c79564fc72710779056cdR252

src/internal/m365/collection/drive/collections_tree.go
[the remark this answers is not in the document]
ryanfkeepers: Yup, can do.

ryanfkeepers: Unless I coded it wrong, the `hadReset` check only applies to that page, not to the full enumeration.   Checking the flag is to prevent `reset, no items -> items in next result` from escaping early, on the reset.  After that (unless another reset happens), `hadReset` should revert to false.

For all identifications of "any reset ever occurred", the tree tracks that as part of calling `tree.reset()`, and callers can compare `tree.hadReset` for the flag.

ryanfkeepers: D'oh, I grok now.  You're right, it's not a required check.

PR #4436, opened 2023-10-03
hand resource down to drive controller

ryanfkeepers:
hands the backup resource into the drive collection for the handler to use to record as the siteID

---

#### Does this PR need a docs update or release note?

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #3988

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

PR #4748, opened 2023-11-29
Remove sharepointids from delta query $select params

pandeyabs:
We are not making use of `GetSharepointIds()` anymore. It was removed a month back, see [PR](https://github.com/alcionai/corso/pull/4436/files#diff-fc955b9b2e400cdda711ab2b6bf459845369691c5abc3973bbfaa1c83a664688L43). So we don't need to fetch this per item-property from graph during delta queries.

This will reduce overall mem usage since we don't need to store references to the sharepoint id map. It can also make delta queries go slightly faster since we are transferring & processing fewer bytes over network.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ryanfkeepers (approved):

ashmrtn:
This PR sounds like it's alright to merge because we're not using this field, but we need to be really careful about changing delta query parameters because delta tokens encode them. Just merging this doesn't mean that all new backups we do will use the new set of parameters. If we do an incremental backup with an older token it may still fetch this parameter

PR #4756, opened 2023-11-30
Fix sanity tests for exchange emails export

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cmd/sanity_test/restore/exchange.go
+++ b/src/cmd/sanity_test/restore/exchange.go
@@ -29,11 +29,12 @@ func CheckEmailRestoration(
 		"source_container_id", sourceTree.ID,
 		"source_container_name", sourceTree.Name)
 
-	common.AssertEqualTrees[models.MailFolderable, any](
+	// NOTE: We cannot compare leaves as the IDs of the restored items
+	// differ from the original ones.
+	common.CompareDiffTrees[models.MailFolderable, any](
 		ctx,
 		sourceTree,
 		restoredTree.Children[envs.SourceContainer],
-		nil,
 		nil)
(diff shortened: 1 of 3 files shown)


src/cmd/sanity_test/restore/exchange.go
```diff
@@ -122,7 +123,7 @@ func recursivelyBuildTree(
 	}
 
 	for _, child := range childFolders {
-		if int(ptr.Val(child.GetTotalItemCount())) == 0 {
+		if int(ptr.Val(child.GetTotalItemCount()))+len(childFolders) == 0 {
```
ryanfkeepers, line 126: I'm okay with these changes, but I'm starting to worry that all of the amendments to "get counts" implies that those funcs are counting the wrong things.  This is the, like, sixth one-off fix to a count like this.  It's beginning to look brittle from a maintenance perspective.  Do we need to centralize these changes in the counter func itself?  Or maybe update the func signature to return more data?

ryanfkeepers (approved):

src/cmd/sanity_test/restore/exchange.go
meain: Hmm, I understand your concern, but I feel like we are just matching the behavior of export and restore with the restores. I've tried to add comments wherever it might not be obvious. But yeah, I'll keep an eye.

PR #4760, opened 2023-11-30
Rework binary publish failure notification

### LINKED ISSUE #4758: binary build job uses old slack notifier
Our binary build worksflow still uses the slack notifier to send messages if there's a failure. We've since switched to a different notifier so this needs updated. Searching for `SLACK_WEBHOOK_URL` in the `.github` directory should be enough to find the offending code

ashmrtn:
Switch notification to be per workflow instead of per action. This means that CI jobs that want to check the result of the binary publish step will need to do so manually after the step runs.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* closes #4758

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/actions/publish-binary/action.yml
+++ b/.github/actions/publish-binary/action.yml
@@ -13,9 +13,6 @@ inputs:
   rudderstack_data_plane_url:
     description: Data plane URL for RudderStack
     required: true
-  slack_webhook_url:
-    description: Slack webhook url
-    required: true
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


ashmrtn:
I'm not sure if this is how we want to handle the publish binary notifications so let me know if the old strategy for notification was better

meain (approved):

PR #4761, opened 2023-11-30
persist deleted file markers for exclude

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -36,6 +36,8 @@ type folderyMcFolderFace struct {
 	// will also be used to construct the excluded file id map
 	// during the post-processing step
 	fileIDToParentID map[string]string
+	// required for populating the excluded file id map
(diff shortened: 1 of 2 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -289,19 +292,11 @@ func (face *folderyMcFolderFace) addFile(
 	face.fileIDToParentID[id] = parentID
 	parent.files[id] = lastModifed
 
+	delete(face.deletedFileIDs, id)
```
ashmrtn, line 295: sanity check: if we're removing from the deleted files map here, is generating the prefix matcher output at the end going to walk both the set of live items and the set of explicitly deleted items?

ashmrtn (approved):

PR #4762, opened 2023-11-30
add prev-path updates to tree

ryanfkeepers:
introduces a func to the delta tree that allows a caller to set the previous-path on a folder by ID.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -250,6 +250,102 @@ func (suite *DeltaTreeUnitSuite) TestFolderyMcFolderFace_AddTombstone() {
...
+			name:            "no changes become a no-op",
+			id:              id(folder),
+			loc:             loc,
+			tree:            newFolderyMcFolderFace(nil, rootID),
+			expectErr:       assert.NoError,
+			expectLive:      false,
+			expectTombstone: false,
+		},
+		{
+			name:            "create tombstone after reset",
```
ashmrtn, line 273: could you please add one more test for a tree that has the node as live but has also been reset? I think this is a case worth calling out because incorrect implementation could lead to hard failures in kopia wrapper (i.e. if we accidentally generated collections for the live node and a tombstone but both had the previous path set we'd probably fail)

Given the code in delta_tree.go it's clear this will work, but this is more of a black box type test to ensure we don't accidentally make a future change that will lead to incorrect behavior

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: sure can!

PR #4777, opened 2023-12-01
add test.t and pfx to test tree generators

ryanfkeepers:
this is needed to standardize the presence of a path prefix in all test-helper trees, so that we can use standard test factory helpers for producing complete post-process data.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/helper_test.go
+++ b/src/internal/m365/collection/drive/helper_test.go
@@ -444,6 +444,13 @@ func fullPath(elems ...string) string {
 		elems...)...)
 }
 
+func fullPathPath(t *testing.T, elems ...string) path.Path {
+	p, err := path.FromDataLayerPath(fullPath(elems...), false)
+	require.NoError(t, err, clues.ToCore(err))
+
+	return p
+}
+
 func driveFullPath(driveID any, elems ...string) string {
 	return toPath(append(
 		[]string{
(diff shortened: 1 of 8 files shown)


ashmrtn:
just noting that either this needs up dated or other PRs after this point in the chain need to have their base updated

src/internal/m365/collection/drive/helper_test.go
```diff
@@ -778,40 +785,46 @@ func aReset(items ...models.DriveItemable) mock.NextPage {
...
+
+func newTree(t *testing.T) *folderyMcFolderFace {
+	return newFolderyMcFolderFace(fullPathPath(t), rootID)
+}
 
-func treeWithRoot() *folderyMcFolderFace {
-	tree := newFolderyMcFolderFace(nil, rootID)
+func treeWithRoot(t *testing.T) *folderyMcFolderFace {
+	tree := newFolderyMcFolderFace(fullPathPath(t), rootID)
 	rootey := newNodeyMcNodeFace(nil, rootID, rootName, false)
```
ashmrtn, line 798: I'm not sure if you have another PR to update these calls as I haven't looked at them all yet, but now that we're passing `t` in could you please switch calls that add things to the tree to use the existing tree manipulation functions where possible?

ashmrtn (approved):

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: it's a ways out, but I have that change here: https://github.com/alcionai/corso/pull/4821/files#diff-48bc25f85f8c885368934862a2813c6fbd7dfb71f309f91169f605230a85895eR463

PR #4783, opened 2023-12-02
Add custom drive item

pandeyabs:
Add a `custom.DriveItem` struct which partially replicates `DriveItemable` and only stores the fields corso makes use of during a backup operation. 

Sorry about the PR size, I don't have a way to split this into multiple PRs. Also it should be trivial to review because it's simply copying over fields to another struct.

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/custom/drive_item.go
+++ b/src/pkg/services/m365/custom/drive_item.go
@@ -1,8 +1,3 @@
-// Disable revive linter since any structs in this file will expose the same
-// funcs as the original structs in the msgraph-sdk-go package, which do not
-// follow some of the golint rules.
-//
-//nolint:revive
 package custom
 
 import (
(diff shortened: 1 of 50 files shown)


src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/common/str"
+)
+
+// Replica of models.DriveItemable
+type LiteDriveItemable interface {
+	GetId() *string
+	GetName() *string
+	GetSize() *int64
+	GetFolder() interface{}
```
ryanfkeepers, line 23: Prefer `any` instead of `interface{}`.
```suggestion
	GetFolder() any
```

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	"time"
+
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/common/str"
+)
+
+// Replica of models.DriveItemable
+type LiteDriveItemable interface {
```
ryanfkeepers, line 19: Nit: to help clarify the relationship between this struct and `models.DriveItemable`.  Since the "lightness" largely comes from the transformation to a local struct, and not a reduced schema.
```suggestion
type ModelDriveItemable interface {
```

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	"strings"
+	"time"
+
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
+
+	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/common/str"
+)
+
+// Replica of models.DriveItemable
```
ryanfkeepers, line 18: Does this struct need Setter funcs to completely comply with that interface?

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	GetLastModifiedByUser() userable
+	GetCreatedDateTime() *time.Time
+	GetLastModifiedDateTime() *time.Time
+	GetAdditionalData() map[string]interface{}
+}
+
+var _ LiteDriveItemable = &driveItem{}
+
+type driveItem struct {
+	id                   string
```
ryanfkeepers, line 43: Can these values be exported?  I know we're abiding the getter interface, but it at least seems like a shortcut to struct creation.

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	name                 string
```
ryanfkeepers, line 44: Any particular reason to not use pointers for these values, to match the API?  I worry that msoft's expectations around the difference between nil and zero-valued primitives could cause unexpected results.

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func (c *driveItem) GetLastModifiedDateTime() *time.Time {
+	return &c.lastModifiedDateTime
+}
+
+func (c *driveItem) GetAdditionalData() map[string]interface{} {
+	return c.additionalData
+}
+
```
ryanfkeepers, line 134: Comment separators would add a lot of readability between structures here.
```suggestion

// ---------------------------------------------------------------------------
// malwareable
// ---------------------------------------------------------------------------
```

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	return iis.identity
+}
+
+var _ identityable = &identity{}
+
+type identity struct {
+	additionalData map[string]interface{}
+}
+
+func (iu *identity) GetAdditionalData() map[string]interface{} {
```
ryanfkeepers, line 222: iu?  What's the u?

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+type user struct {
+	id string
+}
+
+func (u *user) GetId() *string {
+	return &u.id
+}
+
+// TODO(pandeyabs): This is duplicated from collection/drive package.
+// Move to common pkg
```
ryanfkeepers, line 237: api/graph would be a better owner than common.

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	"@content.downloadUrl",
+}
+
+func ToLiteDriveItemable(item models.DriveItemable) LiteDriveItemable {
+	if item == nil {
+		return nil
+	}
+
+	di := &driveItem{
+		id:                   strings.Clone(ptr.Val(item.GetId())),
```
ryanfkeepers, line 249: Is it necessary to clone the string?  ptr.Val should already produce a copy of the value.

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+		name:                 strings.Clone(ptr.Val(item.GetName())),
+		size:                 ptr.Val(item.GetSize()),
+		createdDateTime:      ptr.Val(item.GetCreatedDateTime()),
+		lastModifiedDateTime: ptr.Val(item.GetLastModifiedDateTime()),
+	}
+
+	if item.GetFolder() != nil {
+		di.folder = &struct{}{}
+	} else if item.GetFile() != nil {
```
ryanfkeepers, line 258: I know we think of these proparties as conditional in corso, but that doesn't mean microsoft *can't* give us an item that is both a package and a folder.  I think you're safer off handling each value separately instead of excluding some.

src/pkg/services/m365/custom/drive_item.go
```diff
@@ -0,0 +1,339 @@
...
+	di := &driveItem{
+		id:                   strings.Clone(ptr.Val(item.GetId())),
+		name:                 strings.Clone(ptr.Val(item.GetName())),
+		size:                 ptr.Val(item.GetSize()),
+		createdDateTime:      ptr.Val(item.GetCreatedDateTime()),
+		lastModifiedDateTime: ptr.Val(item.GetLastModifiedDateTime()),
+	}
+
+	if item.GetFolder() != nil {
+		di.folder = &struct{}{}
```
ryanfkeepers, line 257: we really don't use any properties in folder, huh?

ryanfkeepers, line 257: Would it be useful to have a func in `common/ptr` for this?  Something like:
```suggestion
	di.folder = ptr.EmptyOrNil(item.GetFolder())
```

src/pkg/services/m365/custom/drive_item.go
pandeyabs: Yeah, that's a good point. Thanks!

src/pkg/services/m365/custom/drive_item.go
pandeyabs: yep yep

src/pkg/services/m365/custom/drive_item.go
pandeyabs: Clarified in another comment. `LiteDriveItemable` doesn't implement the entire `DriveItemable` interface. Also, even if it did ( function wise), we will have to construct return interfaces (e.g. `ItemReferenceable`, `Malwareable` etc.) to fully comply with `DriveItemable`, which would beat the whole `lite` aspect.

src/pkg/services/m365/custom/drive_item.go
pandeyabs: Just to elaborate on this, if we want to set properties on this `driveItem` struct, we can first create a `models.NewDriveItem()` and then call `ToLiteDriveItemable` on it. We could have a shortcut like you said, but there is no need for it, so I think we can skip it for now.

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: hah, fair

src/pkg/services/m365/custom/drive_item.go
pandeyabs: True, but I didn't want to use `custom.DriveItemable` name since it can be easily confused with `models.DriveItemable`. What do you think? Is it a fair concern?

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: `&struct{}{}`.  I don't really have a better word for it.  `StructOrNil`?  Doesn't really matter all that much, mostly wondering if you think this would be valuable to centralize outside of this file.

src/pkg/services/m365/custom/drive_item.go
pandeyabs: True, but it'll add unnecessary mem copies in most places. Since memcpy has a cpu cost, i think we should avoid it unless really needed. We could however expose another function `ptr.ValCopy` or similar with `Clone()` in it.

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: Personal preference: if we don't intend to have this item comply with the complete `models.DriveItemable` interface (as you mentioned [here](https://github.com/alcionai/corso/pull/4783#discussion_r1414296674)), then we don't need this interface at all.  Seems like we're fine to transform the graph object into a concrete struct (`custom.DriveItem`) and hand that around directly.

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: Okay, sounds good.  I don't think we need a `ValCopy`.  It's too generic a func, anyway, and would have lots of implications beyond just string value handling.  If we find ourselves going back to it often, `common/str` is probably the right owner.  `str.PtrVal(s)` or somesuch.

src/pkg/services/m365/custom/drive_item.go
pandeyabs: Yeah, that's a reasonable thing to do. Thanks! Let me make this change. I think Vaibhav also had a similar comment.

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: That seems like a lot of roundabout just to populate a struct.

pandeyabs: There is no current need to directly populate this struct anywhere. We can export the fields if a need arises in future.

src/pkg/services/m365/custom/drive_item.go
pandeyabs: I don't see a huge value add tbh. I'm going to skip this.

src/pkg/services/m365/custom/drive_item.go
ryanfkeepers: Okedoke!  Sounds good.

PR #4794, opened 2023-12-04
produce global excludes using tree

ryanfkeepers:
produce the global excludes list using the delta tree by aggregating all file IDs whose parents directories are not tombstoned.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -4,12 +4,12 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
-	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
 	"github.com/alcionai/corso/src/pkg/logger"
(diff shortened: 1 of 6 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -450,6 +451,27 @@ func walkTreeAndBuildCollections(
...
+			// those will get handled in the delete cascade.
+			continue
+		}
+
+		result[iID+metadata.DataFileSuffix] = struct{}{}
+		result[iID+metadata.MetaFileSuffix] = struct{}{}
+	}
+
+	for iID := range face.deletedFileIDs {
+		result[iID] = struct{}{}
```
ashmrtn, line 469: nit: these should be suffixed as well iirc. They need to match exactly with what kopia will be examining

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: Yup, that should be fixed in a later PR (when I round up all the fix changes).

PR #4796, opened 2023-12-04
swap tree file types for models.DriveItemable

ryanfkeepers:
drive collections curently pass around models.DriveItemables for their item data.  The reduction to a modtime and size in the delta tree was a design miss, and needs to be corrected.

This PR should put the tests into a green state again.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -4,11 +4,12 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
-	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
 	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/path"
+	"github.com/alcionai/corso/src/pkg/services/m365/custom"
 )
 
 // folderyMcFolderFace owns our delta processing tree.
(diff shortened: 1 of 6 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -101,16 +102,11 @@ func newNodeyMcNodeFace(
 		id:        id,
 		name:      name,
 		children:  map[string]*nodeyMcNodeFace{},
-		files:     map[string]fileyMcFileFace{},
+		files:     map[string]models.DriveItemable{},
 		isPackage: isPackage,
 	}
 }
 
-type fileyMcFileFace struct {
```
ashmrtn, line 109: 🪦 fileyMcFileFace

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -317,8 +313,7 @@ func (face *folderyMcFolderFace) setPreviousPath(
 // this func will update and/or clean up all the old references.
 func (face *folderyMcFolderFace) addFile(
 	parentID, id string,
```
ashmrtn, line 315: not sure if we want to do this, but it seems like we could just replace all the args with `file` and source the parentID and itemID from that too

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: we barely knew ye

Issue #4782, opened 2023-12-01
Allow configuring minimum kopia index epoch duration


ashmrtn:
Internally, kopia maintains an index of contents so it can find things. The index has an epoch-based system where each epoch has one or more blobs denoting changes for it. When no client is known to be using an epoch anymore, the individual changes are compacted into a larger file

Not switching epochs enough can lead to a high number of API calls as kopia must read many smaller files to generate a complete view of the epoch. Therefore, we should allow the minimum epoch time to be configurable so we have a bit more control over how many files corso may load during a backup

PR #4799, opened 2023-12-05
Wire up setting min epoch duration

ashmrtn:
Add the required wiring for SDK consumers to set the min epoch duration
(and eventually other persistent config info). This is just boiler-plate
with some tests at the various layers to make sure that layer is run as
expected. More nuanced tests about setting various config values are
left to the code in `internal/kopia/conn_test.go`

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #4782

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/operations/persistent_config.go
```diff
@@ -0,0 +1,78 @@
...
+func (op *PersistentConfigOperation) Run(ctx context.Context) (err error) {
+	defer func() {
+		if crErr := crash.Recovery(ctx, recover(), "persistent_config"); crErr != nil {
+			err = crErr
+		}
+	}()
+
+	op.Results.StartedAt = time.Now()
+
+	// TODO(ashmrtn): Send telemetry?
```
ryanfkeepers, line 59: what sorts of telemetry are you thinking of?

src/internal/kopia/conn.go
```diff
@@ -38,6 +39,24 @@ const (
...
 	ErrorRepoAlreadyExists  = clues.New("repo already exists")
+
+	// minEpochDurationLowerBound is the minimum corso will allow the kopia epoch
+	// duration to be set to. This number can still be tuned further, right now
+	// it's just to make sure it's not set to something totally wild.
+	//
+	// Note that there are still other parameters kopia checks when deciding if
+	// the epoch should be changed. This is just the min amount of time since the
+	// last epoch change that's required.
+	minEpochDurationLowerBound = 2 * time.Hour
```
ryanfkeepers, line 50: question: what's an epoch / what does this configure?

ryanfkeepers (approved):

src/internal/operations/persitent_config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (op *PersistentConfigOperation) Run(ctx context.Context) (err error) {
+	defer func() {
+		if crErr := crash.Recovery(ctx, recover(), "persistent_config"); crErr != nil {
+			err = crErr
+		}
+	}()
+
+	op.Results.StartedAt = time.Now()
```
pandeyabs: nit: Move start time capture to `do()` as well?

src/internal/operations/persitent_config.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (op *PersistentConfigOperation) do(ctx context.Context) error {
+	defer func() {
+		op.Results.CompletedAt = time.Now()
+	}()
+
+	err := op.operation.kopia.UpdatePersistentConfig(ctx, op.configOpts)
+	if err != nil {
+		op.Status = Failed
+		return clues.Wrap(err, "running retention config operation")
```
pandeyabs: nit: Change to `running persistent config operation`?

src/internal/kopia/conn.go
ashmrtn: this is for kopia's epoch manager which holds the index mapping content IDs -> (S3 blob, blob offset). The epoch manager is kind of like a log structured merge tree, it keeps the most recent epochs "uncompacted" with lots of little files denoting the changes in the epoch (additions/removals of content). All files corresponding to an epoch need to be loaded to create a complete view of the index and access kopia's data

Once the epoch is old enough, it gets compacted into a summary of sorts. The summary is a single S3 blob that contains all the index changes for one or more epochs (kopia compacts compacted epochs to further reduce the number of S3 GETs)

This parameter denotes the minimal amount of time that has to pass before switching to a new epoch. By changing this, one can tune how many S3 blobs may need to be fetched in order to access kopia's data. This works by reducing the number of possible blobs that can accrue in an epoch prior to switching to a new epoch (and thus compacting an older one) by letting kopia switch epochs more often

src/internal/operations/persistent_config.go
ashmrtn: oh mostly just stuff like what we have for other operation types. Not sure if we'll ever want to do something like that for this

pandeyabs (approved):
🚀 🚢

PR #4805, opened 2023-12-05
clean up drive unit test helpers

ryanfkeepers:
Adds some cleanups to the drive unit test helpers.  Goals were as follows:
1. remove idx() and namex() funcs, replace with id() and name()
2. minimize factory duplication in helper_test.go
3. improve readability in id() and name() calls by adding file and folder variants to handle the 99% of cases we use them in.

No logic changes in this PR.  Only test func updates.
---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test


src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -843,73 +844,94 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_PopulateTree_singleDelta(
 				err:            require.NoError,
 				numLiveFiles:   3,
 				numLiveFolders: 4,
-				sizeBytes:      3 * defaultItemSize,
+				sizeBytes:      3 * 42,
```
ashmrtn, line 847: not sure if there's another PR to handle the default (haven't looked ahead yet), but this is technically reverting the code we'd like to have

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -958,13 +980,79 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_PopulateTree_singleDelta(
...
-					id(folder),
+					folderID(),
+				},
+				treeContainsFileIDsWithParent: map[string]string{
+					fileID(): folderID(),
+				},
+			},
+		},
+		{
+			name: "delete->create folder with previous path",
```
ashmrtn, line 994: are these new tests from a later PR? It's not clear from the test right now that there's previous paths associated with the deleted folder so I'm a little confused

ashmrtn (approved):
looks good for the most part. There are some instances where `driveItem` is used still instead of `driveFile` or `driveFolder` (I think mostly `driveFile`) I think specifically when the item is nested. Both the name and ID of the item in some of the remaining calls to `driveItem` that can be swapped match which if I grok things is the requirement. There's multiple calls in `collections_test.go` and I think one in `collections_tree_test.go`. Not sure if it's worth digging through stuff to go find them at the moment though

src/internal/m365/collection/drive/helper_test.go
```diff
@@ -181,257 +181,249 @@ func driveItem(
...
+
+	c.SetMalware(mal)
+
+	return c
 }
 
-func fileAt(
-	parentX any,
+// delItem creates a DriveItemable that is marked as deleted. path must be set
+// to the base drive path.
```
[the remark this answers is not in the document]
ryanfkeepers: can do!

src/internal/m365/collection/drive/url_cache_test.go
```diff
@@ -389,47 +373,37 @@ func (suite *URLCacheUnitSuite) TestGetItemProperties() {
...
-					Reset: true,
-				},
-				{
-					Items: []models.DriveItemable{
-						fileItem("4", "file4", "root", "root", "https://dummy4.com", false),
-						fileItem("5", "file5", "root", "root", "https://dummy5.com", false),
-					},
-				},
+				aPage(
+					fileWURLAtRoot(aURL(1), false, 1),
```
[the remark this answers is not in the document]
ryanfkeepers: sure, done

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: ah, bargle.  rebases.  I'll clean this up again at the end of the bookend.

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: Ah, I renamed the test in a following PR.  It's kind of imaginging an nth delta enumeration where a prior delta introduced the file, then the next delta has a delete/recover pair of operations.

ryanfkeepers (commented):
Thanks for the review.  I think I clean up many of the `driveItem` calls in collections_test.go in a later PR, but fwiw: I've been avoiding updating every line in that folder to match new refactors because it's, ultimately, toil without value to upkeep the file.  Some things make it through, especially when necessary.  Other things slip through.  the `collections_tree_test.go` should avoid driveItem unless necessary, however.

PR #4816, opened 2023-12-07
minor mail attachment cleanup

ryanfkeepers:
started down the path of changing behavior under error conditions involving server-side nil reference errors.  Issue turned out to be transient.  I'm commiting the cleanup without the error case handling.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/mail.go
```diff
@@ -285,106 +285,179 @@ func (c Mail) GetItem(
...
+	userID, mailID string,
+	immutableIDs bool,
+	errs *fault.Bus,
+) ([]models.Attachmentable, int64, error) {
+	var (
+		result    = []models.Attachmentable{}
+		totalSize int64
+		cfg       = &users.ItemMessagesItemAttachmentsRequestBuilderGetRequestConfiguration{
+			QueryParameters: &users.ItemMessagesItemAttachmentsRequestBuilderGetQueryParameters{
+				Select: idAnd(),
```
pandeyabs, line 361: We need size parameter too right?

src/pkg/services/m365/api/mail.go
```diff
@@ -285,106 +285,179 @@ func (c Mail) GetItem(
...
+		logger.CtxErr(ctx, err).Info("falling back to fetching attachments by id")
+
+		attachments, totalSize, err = c.getAttachmentsIterated(ctx, userID, mailID, immutableIDs, errs)
+		if err != nil {
+			return nil, nil, clues.Stack(err)
+		}
 	}
 
-	attached, err := c.LargeItem.
+	size += totalSize
```
pandeyabs, line 302: nit: We can lose this and just use `totalSize`.

pandeyabs (approved):
LGTM! @vkamra FYI, this PR fixes the expand issue.

pandeyabs:
Could you also add a changelog for the `$expand` bug fix?

src/pkg/services/m365/api/mail.go
ryanfkeepers: not quite; see line 280.  we need to add in the size of the message body first.

src/pkg/services/m365/api/mail.go
ryanfkeepers: Nope, because this is just listing the IDs of each attachment.  We get the size when we look up the item itself on 392.

src/pkg/services/m365/api/mail.go
pandeyabs: Ah, thanks!

PR #4817, opened 2023-12-07
4689 cleanup 2

ryanfkeepers:
The next couple PRs are a distributed set of changes with the goal
of improving drive stubbing/management within testing, so that
the identification and use of a drive is more failsafe and maintainable,
and less reliant on obtuse drive id references.

-----

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/service/onedrive/mock/handlers.go
+++ b/src/internal/m365/service/onedrive/mock/handlers.go
@@ -1,5 +1,16 @@
 package mock
 
+// ---------------------------------------------------------------------------
+// 								>>> TODO <<<
+//              https://github.com/alcionai/corso/issues/4846
+// This file's functions are duplicated into /drive/helper_test.go, which
+// should act as the clear primary owner of that functionality.  However,
+// packages outside of /drive (such as sharepoint) depend on these helpers
+// for test functionality.  We'll want to unify the two at some point.
+// In the meantime, make sure you're referencing and updating the correct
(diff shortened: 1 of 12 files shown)


src/internal/m365/service/onedrive/mock/handlers.go
```diff
@@ -359,29 +356,73 @@ func (en EnumerateDriveItemsDelta) EnumerateDriveItemsDelta(
...
+
+	return &DeltaDrive{
+		ID:   driveID,
+		Able: able,
+	}
+}
+
+func (dd *DeltaDrive) NewEnumer() *DeltaDriveEnumerator {
+	copy := &DeltaDrive{}
+	*copy = *dd
```
ashmrtn, line 403: is a copy here strictly necessary? Also this looks like it may end up being a shallow copy since `dd.Able` is likely a pointer

ashmrtn (approved):

src/internal/m365/service/onedrive/mock/handlers.go
ryanfkeepers: no, not strictly.  I just wanted to have the little extra safety. Avoidig accidental dereferencing helps me sleep at night.  You're right that `able` is still exposed here, but since we tend to use that with only a getter interface, I feel a little more calm about that one.

PR #4818, opened 2023-12-07
move mocks into drive helper tests

ryanfkeepers:
will be used for easier readability

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/helper_test.go
+++ b/src/internal/m365/helper_test.go
@@ -1,5 +1,15 @@
 package m365
 
+// ---------------------------------------------------------------------------
+// 								>>> TODO <<<
+// This file's functions are duplicated into /drive/helper_test.go, which
+// should act as the clear primary owner of that functionality.  However,
+// packages outside of /drive (such as sharepoint) depend on these helpers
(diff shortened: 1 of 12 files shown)


src/internal/m365/helper_test.go
```diff
@@ -1,5 +1,15 @@
 package m365
 
+// ---------------------------------------------------------------------------
+// 								>>> TODO <<<
+// This file's functions are duplicated into /drive/helper_test.go, which
```
ashmrtn, line 5: can we get a tech debt ticket for this as well please? That way we have something outside the code that lets us know something needs fixed

ashmrtn (approved):

src/internal/m365/helper_test.go
ryanfkeepers: heh, doubly glad you mentioned this, because I put the comment in the wrong file, too...

https://github.com/alcionai/corso/issues/4846

PR #4819, opened 2023-12-08
use a drive to identify all file paths

ryanfkeepers:
this will ensure that the drive id is correct and consistent for each drive item, regardless of how the drive itself was composed. Minimizes the cognitive load on the test maintainer for which drive ID to use and where.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections_test.go
+++ b/src/internal/m365/collection/drive/collections_test.go
@@ -40,6 +40,7 @@ func TestCollectionsUnitSuite(t *testing.T) {
 
 func (suite *CollectionsUnitSuite) TestPopulateDriveCollections() {
 	t := suite.T()
+	d := drive()
 
 	tests := []struct {
 		name                     string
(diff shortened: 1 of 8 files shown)


src/internal/m365/collection/drive/helper_test.go
```diff
@@ -1543,46 +1547,31 @@ func toPath(elems ...string) string {
...
+func (dd *deltaDrive) fullPath(t *testing.T, elems ...string) path.Path {
+	p, err := path.FromDataLayerPath(dd.strPath(elems...), false)
 	require.NoError(t, err, clues.ToCore(err))
 
 	return p
 }
 
-func driveFullPath(driveID any, elems ...string) string {
+// produces a complete path prefix up to the drive root folder, and including any
+// elements passed in.
```
ashmrtn, line 1571: nit: minor rewording for clarity
```suggestion
// produces a complete path prefix up to the drive root folder with any
// elements passed in appended to the generated prefix.
```

src/internal/m365/collection/drive/collections_test.go
ashmrtn, line 1: I'm curious as to where this is heading in the long run. This file seems to still be at a sort of halfway point since it still has many calls where it's explicitly getting the path prefix from the drive struct instead of using functions on the drive struct to create items

ashmrtn (approved):

src/internal/m365/collection/drive/collections_test.go
ryanfkeepers: that should all get wrapped up in the next PR.

ryanfkeepers: oh, wait, spoke too soon.   So, per the collections_test.go file: i'm not concentrating as much energy on keeping that file updated, since the collections_tree_test will eventually take it over wholesale.  Effort to refactor there is largely sunk cost.

PR #4824, opened 2023-12-08
use cuustom.DriveItem, fix errors

ryanfkeepers:
cleanup following rebase for *-10 bookend PR set.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/alcionai/clues"
-	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
@@ -315,7 +314,7 @@ func (face *folderyMcFolderFace) setPreviousPath(
 // this func will update and/or clean up all the old references.
 func (face *folderyMcFolderFace) addFile(
 	parentID, id string,
-	file models.DriveItemable,
+	file *custom.DriveItem,
 ) error {
 	if len(parentID) == 0 {
 		return clues.New("item added without parent folder ID")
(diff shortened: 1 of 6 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -437,9 +438,15 @@ func walkTreeAndBuildCollections(
 			With("path_prefix", pathPfx, "path_suffix", parentPath.Elements())
 	}
 
+	customFiles := map[string]*custom.DriveItem{}
+
+	for fID, file := range node.files {
+		customFiles[fID] = custom.ToCustomDriveItem(file)
```
pandeyabs, line 444: Can be done in a future PR, but I wanted to add a note here.

Instead of converting it to custom drive item here, can we do the conversion a little earlier during `enumeratePageOfItems` i.e. before `addFileToTree` and `addFolderToTree` calls?

This way the `custom.DriveItem` can be the source of truth for all drive item properties. Otherwise, if we have a bug in `ToCustomDriveItem`, it'll be hard to discover if we use both `DriveItemable` (during enumeration), and `custom.DriveItem` during item download phase.

It'll also help to track memory usage more accurately, since the usage won't be split between `DriveItemable` and `custom.DriveItem`.

ryanfkeepers: sure, don't see why not.

ryanfkeepers: @pandeyabs done

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: can do

pandeyabs: Hmm. We can do that, but we'd still need `DriveItemable` for restore & export operations.

PR #4834, opened 2023-12-12
Skip emails that cannot be retrieved

vkamra:
Adds a skip condition for emails that can be enumerated but are not returned from the server (Exchange) because
it believes they are corrupt/invalid.

This handles the `ErrorInvalidRecipients` condition we believe is hit when exchange finds an email that pre-dates
M365 mailbox creation (either a server corruption or triggered by on-prem->M365 migration)

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/errors.go
+++ b/src/pkg/services/m365/api/graph/errors.go
@@ -43,6 +43,11 @@ const (
 	emailFolderNotFound      errorCode = "ErrorSyncFolderNotFound"
 	ErrorAccessDenied        errorCode = "ErrorAccessDenied"
 	errorItemNotFound        errorCode = "ErrorItemNotFound"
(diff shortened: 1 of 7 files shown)


src/pkg/services/m365/api/graph/errors.go
```diff
@@ -201,6 +206,10 @@ func IsErrUserNotFound(err error) bool {
 	return false
 }
 
+func IsErrInvalidRecipients(err error) bool {
```
pandeyabs, line 209: This could use a unit test

pandeyabs (approved):
Looks good!

PR #4839, opened 2023-12-13
fix previous paths handling

ryanfkeepers:
new previous paths generation doesn't update child paths.  This creates a specific prev-path post processing function to ensure that all unseen child dirs inherit changes from their parent.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/m365/collection/drive/collections_tree_test.go
+++ b/src/internal/m365/collection/drive/collections_tree_test.go
@@ -603,7 +603,11 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_TurnTreeIntoCollections()
 			},
 		},
 		{
-			name:           "no folders moved",
+			name: "nothing in the tree was moved " +
+				"but there were some folders in the previous paths that " +
+				"didn't appear in the delta so those have to appear in the " +
+				"new previous paths but those weren't moved either so " +
+				"everything should have the same path at the end",
 			tree:           fullTree,
 			enableURLCache: true,
 			prevPaths: map[string]string{
@@ -648,7 +652,12 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_TurnTreeIntoCollections()
 			},
 		},
 		{
-			name:           "all folders not moved - path separator string check",
+			name: "nothing in the tree was moved " +
+				"but there were some folders in the previous paths that " +
+				"didn't appear in the delta so those have to appear in the " +
+				"new previous paths but those weren't moved either so " +
+				"everything should have the same path at the end " +
+				"- the version with path separators chars in the directory names",
 			tree:           fullTreeWithNames("pa/rent", "to/mbstone"),
 			enableURLCache: true,
 			prevPaths: map[string]string{
--- a/src/internal/m365/collection/drive/collections_tree_test.go
+++ b/src/internal/m365/collection/drive/collections_tree_test.go
@@ -657,7 +657,7 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_TurnTreeIntoCollections()
 				folderID():             d.strPath(t, folderName("pa/rent"), folderName()),
 				folderID("pr/ev"):      d.strPath(t, folderName("pa/rent"), folderName("pr/ev")),
 				folderID("prev/chld"):  d.strPath(t, folderName("pa/rent"), folderName("pr/ev"), folderName("prev/chld")),
-				folderID("to/mbstone"): d.strPath(t, folderName("tombstone/prev")),
+				folderID("to/mbstone"): d.strPath(t, folderName("to/mbstone")),
 			},
 			expect: expected{
 				prevPaths: map[string]string{
@@ -683,7 +683,7 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_TurnTreeIntoCollections()
 							d.fullPath(t, folderName("pa/rent"), folderName()),
 							d.fullPath(t, folderName("pa/rent"), folderName()),
 							fileID()),
-						aColl(nil, d.fullPath(t, folderName("tombstone/prev"))))
+						aColl(nil, d.fullPath(t, folderName("to/mbstone"))))
 				},
 				globalExcludedFileIDs: makeExcludeMap(
 					fileID("r"),
--- a/src/internal/m365/collection/drive/collections_tree_test.go
+++ b/src/internal/m365/collection/drive/collections_tree_test.go
@@ -278,7 +278,9 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_MakeDriveCollections() {
 			enumerator: driveEnumerator(
 				d.newEnumer().with(
 					delta(id(deltaURL), nil).with(
-						aPage(d.folderAtRoot(), d.fileAt(folder))))),
+						aPage(
+							d.folderAt(root),
+							d.fileAt(folder))))),
 			prevPaths: map[string]string{},
 			expectCounts: countTD.Expected{
 				count.PrevPaths: 0,
@@ -290,7 +292,9 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_MakeDriveCollections() {
 			enumerator: driveEnumerator(
 				d.newEnumer().with(
 					delta(id(deltaURL), nil).with(
-						aPage(d.folderAtRoot(), d.fileAt(folder))))),
+						aPage(
+							d.folderAt(root),
+							d.fileAt(folder))))),
 			prevPaths: map[string]string{
 				folderID(): d.strPath(t, folderName()),
 			},
@@ -333,7 +337,9 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_MakeDriveCollections() {
 				d.newEnumer().with(
 					deltaWReset(id(deltaURL), nil).with(
 						aReset(),
-						aPage(d.folderAtRoot(), d.fileAt(folder))))),
+						aPage(
+							d.folderAt(root),
+							d.fileAt(folder))))),
 			prevPaths: map[string]string{},
 			expectCounts: countTD.Expected{
 				count.PrevPaths: 0,
@@ -346,7 +352,9 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_MakeDriveCollections() {
 				d.newEnumer().with(
 					deltaWReset(id(deltaURL), nil).with(
 						aReset(),
-						aPage(d.folderAtRoot(), d.fileAt(folder))))),
+						aPage(
+							d.folderAt(root),
+							d.fileAt(folder))))),
 			prevPaths: map[string]string{
 				folderID(): d.strPath(t, folderName()),
 			},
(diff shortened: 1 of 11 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -459,6 +461,111 @@ func walkTreeAndBuildCollections(
...
+	}
+
+	// next, create a slice of tuples representing any
+	// old prevPath entry whose ID isn't already bound to
+	// a collectable.
+
+	unseenPrevPaths := []idPrevPathTup{}
+
+	for id, p := range prevPaths {
+		if _, ok := results[id]; !ok {
```
ashmrtn, line 507: nit: I think it would be more clear here if paths that are known to be tombstoned are also removed from `unseenPrevPaths` here instead of below. Technically those paths are bound to a collectable like the comment above mentions and they shouldn't be updated or added to the set of output paths

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -459,6 +461,111 @@ func walkTreeAndBuildCollections(
...
+		parent := pb.String()
+
+		// if the parent was tombstoned, skip it
+		if _, ok := tombstoned[parent]; ok {
+			// add the current string to the tombstoned list, that'll allow it to cascade to all children.
+			tombstoned[un.prevPath] = struct{}{}
+			continue
+		}
+
+		// if the parent wasn't moved, add it to the result set
```
ashmrtn, line 546: nit: wording to make it clear it's adding the current path not the parent path
```suggestion
		// if the parent wasn't moved, add the current path to the result set
```

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -459,6 +461,111 @@ func walkTreeAndBuildCollections(
...
+		elems = elems[:len(elems)-1]
+
+		pb, err := path.Builder{}.UnescapeAndAppend(elems...)
+		if err != nil {
+			return nil, err
+		}
+
+		parent := pb.String()
+
+		// if the parent was tombstoned, skip it
```
ashmrtn, line 539: nit: wording for clarity
```suggestion
		// if the parent was tombstoned, skip the current path
```

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -459,6 +461,111 @@ func walkTreeAndBuildCollections(
...
+
+		elems := path.NewElements(un.prevPath)
+		elems = elems[:len(elems)-1]
+
+		pb, err := path.Builder{}.UnescapeAndAppend(elems...)
+		if err != nil {
+			return nil, err
+		}
+
+		parent := pb.String()
```
ashmrtn, line 537: nit: prefer functions on `path.Builder` rather than trimming the slice down manually
```suggestion
		pb, err := path.Builder{}.UnescapeAndAppend(elems...)
		if err != nil {
			return nil, err
		}

		parent := pb.Dir().String()
```

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -1205,3 +1205,187 @@ func (suite *DeltaTreeUnitSuite) TestFolderyMcFolderFace_GenerateCollectables()
...
+			expect: map[string]string{
+				rootID:         d.strPath(t),
+				folderID():     d.strPath(t, folderName("n/ew")),
+				folderID("c1"): d.strPath(t, folderName("n/ew"), folderName("c1")),
+				folderID("c2"): d.strPath(t, folderName("n/ew"), folderName("c2")),
+				folderID("c3"): d.strPath(t, folderName("n/ew"), folderName("c2"), folderName("c3")),
+			},
+		},
+		{
+			name: "tombstoned files get removed",
```
ashmrtn, line 1336:
```suggestion
			name: "tombstoned folders get removed",
```

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -1205,3 +1205,187 @@ func (suite *DeltaTreeUnitSuite) TestFolderyMcFolderFace_GenerateCollectables()
...
+				folderID("c1"): d.strPath(t, folderName("old"), folderName("c1")),
+				folderID("c2"): d.strPath(t, folderName("old"), folderName("c2")),
+				folderID("c3"): d.strPath(t, folderName("old"), folderName("c2"), folderName("c3")),
+			},
+			expect: map[string]string{
+				rootID: d.strPath(t),
+			},
+		},
+		{
+			name: "mix of moved and tombstoned",
```
ashmrtn, line 1353: could you please make a similar test but for multiple folder moves? For example:
```text
start with:
- root
  - old
    - c1
      - c2
        - c3
```
do the equivalent of:
```text
mv root/old root/folder
mkdir root/old
mkdir root/old/c1
mv root/folder/c1/c2 root/old/c1/c2
```

expect:
```text
- root
  - old
    - c1
      - c2
        - c3
  - folder
    - c1
```

src/internal/m365/collection/drive/collections_tree_test.go
```diff
@@ -635,6 +647,51 @@ func (suite *CollectionsTreeUnitSuite) TestCollections_TurnTreeIntoCollections()
 					fileID()),
 			},
 		},
+		{
+			name:           "all folders not moved - path separator string check",
```
ashmrtn, line 651: nit: for this and other tests that were updated could you please update the names so they more closely match what's happening? It's not quite clear now what things are not moved because they're not mentioned and what things are not moved because the "new" path in the tree matches the old path

ashmrtn (approved):
this solution works and has good test coverage so I think we should move forward with it to get this finished off. However, I am a little hesitant about the fact that it's continuing to use string replacement for path updates. I believe we were hoping to move away from that with the tree architecture we picked

ryanfkeepers:
> this solution works and has good test coverage so I think we should move forward with it to get this finished off. However, I am a little hesitant about the fact that it's continuing to use string replacement for path updates. I believe we were hoping to move away from that with the tree architecture we picked

@ashmrtn To speak about that a little: it does feel a bit like a step back, but there's not much of an easy way to handle it otherwise.  Specifically because the prevPaths metadata entries don't contain a reference to their parent by ID.  If they did, I could order the old previous paths from root to leaf, iterate through them, add unseen children by extending their parent.  

Without the parent reference, I have to jump through hoops.  There were two ways of approaching it.  First, the way I did it here.  Second, build a separate tree from the previousPaths metadata and use that tree to handle parent reference lookup to perform the above.  I'm happy to refactor towards that design, if we want to be completely free of string replacements.  But the effort/complexity is going to be roughly the same as what's here.

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: hierarchy makes that a little cumbersome.  I'd have to walk up the directory in every previous entry looking for a parent that was tombstoned.  The way I have it now, the tombstone map grows in sync with the root-to-leaf ordering that we get from this list.  I think the result would be about on par in terms of clarity.

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: arg, I knew there was something like this.  Couldn't remember what it was.  Thanks!

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: hmm, nothing should be moved here.  I don't know how to make that more obvious.  Suggestions?

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: yeah, happy to!

ashmrtn:
> Specifically because the prevPaths metadata entries don't contain a reference to their parent by ID.

Yes, the lack of parent references does make it a bit more challenging in this case

> Without the parent reference, I have to jump through hoops. There were two ways of approaching it. First, the way I did it here. Second, build a separate tree from the previousPaths metadata and use that tree to handle parent reference lookup to perform the above. I'm happy to refactor towards that design, if we want to be completely free of string replacements. But the effort/complexity is going to be roughly the same as what's here.

I was actually thinking of a slightly different implementation though it would be similar to the second option you mentioned:
1. populate nodes in the tree based on the previous paths map from the base backup. Use the path to create a node in the proper place by walking the tree and add new nodes with the given ID at the proper place. Mark all these nodes with some flag denoting they're from the previous paths map
2. do all the fun things with delta updates. For every node explicitly mentioned in the delta update clear the flag set in (1)
3. walk the tree and generate collectables and (new) previous paths. Every node in the tree with the flag from (1) not set will generate a collectable. All nodes in the tree will generate a previous paths entry

The solution above still relies on string processing to build the initial tree, but all path updates are handled by changes to the structure of the tree. That allows us to avoid string replace operations when building the new previous paths because we'll just be reading all paths off the tree itself

(you could also reverse the polarity of the flag mentioned in the above if that would make more sense. I just picked an arbitrary way to set it)

src/internal/m365/collection/drive/delta_tree.go
ashmrtn: I'm not sure I follow, and maybe I was a little unclear. I was talking about moving lines 524-527 below up to here. I don't believe that should require hierarchy traversal since it's a direct comparison on the path

src/internal/m365/collection/drive/collections_tree_test.go
ashmrtn: hmm maybe `all tree folders not moved all nontree folders not moved`? It's getting a bit wordy though

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -1205,3 +1205,231 @@ func (suite *DeltaTreeUnitSuite) TestFolderyMcFolderFace_GenerateCollectables()
...
+					currPath: d.fullPath(t, folderName("bar")),
+				},
+				folderID(2): {
+					currPath: d.fullPath(t, folderName("foo")),
+				},
+				folderID("2c1"): {
+					currPath: d.fullPath(t, folderName("foo"), folderName("c1")),
+				},
+				folderID("c2"): {
+					prevPath: d.fullPath(t, folderName("bar"), folderName("c1"), folderName("c2")),
```
ashmrtn, line 1400: I think actually this previous path should still be `root/foo/c1/c2` since this is all done by folder name not ID?

ashmrtn (approved):

src/internal/m365/collection/drive/collections_tree_test.go
ryanfkeepers: hmm, how about: `nothing in the tree was moved but there were some folders in the previous paths that didn't appear in the delta so those have to appear in the new previous paths but those weren't moved either so everything should have the same path at the end`

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: oh yeah, that works.  My brain is going off into the weeds.

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: right!  right.

PR #4841, opened 2023-12-13
ensure creation of .dirmeta files

ryanfkeepers:
the tree wasn't properly populating the dirmeta file for folders.  This change is primarily for the addition of that file, but also contains a minor refactor for taking in the folder drive item instead of only getting its properties.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/alcionai/clues"
+	"golang.org/x/exp/maps"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/m365/collection/drive/metadata"
@@ -66,7 +67,7 @@ func newFolderyMcFolderFace(
 
 // reset erases all data contained in the tree.  This is intended for
 // tracking a delta enumeration reset, not for tree re-use, and will
-// cause the tree to flag itself as dirty in order to appropriately
+// cause the tree to flag itfolder as dirty in order to appropriately
 // post-process the data.
 func (face *folderyMcFolderFace) reset() {
 	face.hadReset = true
@@ -80,33 +81,29 @@ type nodeyMcNodeFace struct {
 	// required for mid-enumeration folder moves, else we have to walk
 	// the tree completely to remove the node from its old parent.
 	parent *nodeyMcNodeFace
-	// the microsoft item ID.  Mostly because we might as well
-	// attach that to the node if we're also attaching the dir.
-	id string
-	// single directory name, not a path
-	name string
+	// folder is the actual drive item for this directory.
+	// we save this so that, during post-processing, it can
+	// get moved into the collection files, which will cause
+	// the collection processor to generate a permissions
+	// metadata file for the folder.
+	folder *custom.DriveItem
 	// contains the complete previous path
 	prev path.Path
 	// folderID -> node
 	children map[string]*nodeyMcNodeFace
 	// file item ID -> file metadata
 	files map[string]*custom.DriveItem
-	// for special handling protocols around packages
-	isPackage bool
 }
 
 func newNodeyMcNodeFace(
 	parent *nodeyMcNodeFace,
-	id, name string,
-	isPackage bool,
+	folder *custom.DriveItem,
 ) *nodeyMcNodeFace {
 	return &nodeyMcNodeFace{
-		parent:    parent,
-		id:        id,
-		name:      name,
-		children:  map[string]*nodeyMcNodeFace{},
-		files:     map[string]*custom.DriveItem{},
-		isPackage: isPackage,
+		parent:   parent,
+		folder:   folder,
+		children: map[string]*nodeyMcNodeFace{},
+		files:    map[string]*custom.DriveItem{},
 	}
 }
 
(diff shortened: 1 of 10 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -66,7 +67,7 @@ func newFolderyMcFolderFace(
 
 // reset erases all data contained in the tree.  This is intended for
 // tracking a delta enumeration reset, not for tree re-use, and will
-// cause the tree to flag itself as dirty in order to appropriately
+// cause the tree to flag itfolder as dirty in order to appropriately
```
ashmrtn, line 70: 🤔 not sure how this one popped out, but I think it's safe to revert this change 😉

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -186,9 +191,8 @@ func (face *folderyMcFolderFace) setFolder(
 	if zombey, tombstoned := face.tombstones[id]; tombstoned {
 		delete(face.tombstones, id)
 
-		zombey.parent = parent
-		zombey.name = name
-		parent.children[id] = zombey
+		zombey.parent = parentNode
```
ashmrtn, line 194: should this also be updating the contained folder for the node like is done with the root above?

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -65,98 +69,107 @@ func (suite *DeltaTreeUnitSuite) TestNewNodeyMcNodeFace() {
...
-			tname:     "add package",
-			tree:      treeWithRoot,
-			parentID:  rootID,
-			id:        folderID(),
-			name:      folderName(),
-			isPackage: true,
+			tname: "add package",
+			tree:  treeWithRoot,
+			folder: func() *custom.DriveItem {
+				return custom.ToCustomDriveItem(
```
ashmrtn, line 108: nit: since I think there's a helper can this be `custom.ToCustomDriveItem(d.pkgAtRoot())`?

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -65,98 +69,107 @@ func (suite *DeltaTreeUnitSuite) TestNewNodeyMcNodeFace() {
...
+			},
 			expectErr: assert.Error,
 		},
 		{
-			tname:     "missing parentID",
-			tree:      treeWithRoot,
-			id:        folderID(),
-			name:      folderName(),
-			isPackage: true,
+			tname: "missing parentID",
```
ashmrtn, line 141: nit
```suggestion
			tname: "missing parent",
```

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -168,30 +181,37 @@ func (suite *DeltaTreeUnitSuite) TestFolderyMcFolderFace_SetFolder() {
...
-			assert.Equal(t, test.isPackage, result.isPackage)
 
-			_, ded := tree.tombstones[test.id]
+			var (
+				expectID        = ptr.Val(folder.GetId())
+				expectName      = ptr.Val(folder.GetName())
+				expectIsPackage = folder.GetPackageEscaped() == nil
+				resultID        = ptr.Val(folder.GetId())
+				resultName      = ptr.Val(folder.GetName())
+				resultIsPackage = folder.GetPackageEscaped() == nil
```
ashmrtn, line 202: looks like copy/paste error here
```suggestion
				resultID        = ptr.Val(result.GetId())
				resultName      = ptr.Val(result.GetName())
				resultIsPackage = result.GetPackageEscaped() == nil
```

src/internal/m365/collection/drive/delta_tree_test.go
```diff
@@ -453,11 +477,15 @@ func (ts tombs) compare(
...
+		expectID := ptr.Val(entombed.self.folder.GetId())
+
+		zombey := tombstones[expectID]
 		require.NotNil(t, zombey, "tombstone must exist")
 		assert.Nil(t, zombey.parent, "tombstoned node should not have a parent reference")
 
-		t.Run("assert_tombstones/"+zombey.id, func(_t *testing.T) {
+		resultID := ptr.Val(zombey.folder.GetId())
+
+		t.Run("assert_tombstones/"+resultID, func(_t *testing.T) {
```
ashmrtn, line 488: nit: I think something besides `/` in the test name would be better. It'll probably cause odd escaping errors when trying to run individual tests

ashmrtn (approved):

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: no one expects the find&replace!

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: good catch, yes.

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: right you are.

src/internal/m365/collection/drive/delta_tree_test.go
ryanfkeepers: big oof

PR #4844, opened 2023-12-13
fix-ups based on last PR

ryanfkeepers:
makes two changes:
1. has `addFile` take in only a drive item, instead of ids and an item.
2. removes the file/folderAtRoot() generators in favor of *At(root, ...)

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/helper_test.go
+++ b/src/internal/m365/collection/drive/helper_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
@@ -315,10 +316,9 @@ func aColl(
 		ids = append(ids, fID+metadata.MetaFileSuffix)
 	}
 
-	// should always expect the folder to have a
-	// dir meta file for permissions.  Not expected
-	// to get added for tombstones.
-	if curr != nil {
+	// should expect all non-root, non-tombstone collections to contain
+	// a dir meta file for storing permissions.
+	if curr != nil && !strings.HasSuffix(curr.Folder(false), root) {
 		ids = append(ids, metadata.DirMetaFileSuffix)
 	}
 
--- a/src/internal/m365/collection/drive/helper_test.go
+++ b/src/internal/m365/collection/drive/helper_test.go
@@ -1471,7 +1471,11 @@ func (dd *deltaDrive) folderAt(
 	return driveItem(
 		folderID(folderSuffixes...),
 		folderName(folderSuffixes...),
-		dd.dir(),
+		// we should be putting in the full location here, not just the
+		// parent suffix.  But that full location would be unused because
+		// our unit tests don't utilize folder subselection (which is the
+		// only reason we need to provide the dir).
+		dd.dir(folderName(parentSuffix)),
 		folderID(parentSuffix),
 		isFolder)
 }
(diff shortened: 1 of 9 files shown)


src/internal/m365/collection/drive/helper_test.go
```diff
@@ -1369,22 +1359,20 @@ func (dd *deltaDrive) fileAt(
...
+			isFile)
+	}
 
-func (dd *deltaDrive) fileAtRoot(
-	fileSuffixes ...any,
-) models.DriveItemable {
 	return driveItem(
 		fileID(fileSuffixes...),
 		fileName(fileSuffixes...),
 		dd.dir(),
```
ashmrtn, line 1374: I'm not sure I grok the full usage of this function, but there's a couple things that feel a bit off with the current implementation. First, this seems like it's always going to produce a folder path at the root for the parent reference right now because it's not passing the parentSuffix to `dir`. Second, the function as a whole seems like it's only capable of producing files with one level of nesting. This is because it'll pass multiple suffixes to the file name/ID functions, but, assuming this should be `dd.dir(parentSuffix)` it'll only pass one suffix to the path initializer. This means that although calls like `fileAt("some", "folder", "path", "fileSuffix")` will compile and run, they'll generate unexpected results

If this function is going to continue to have the same set of input parameters, it may be a good idea to have some sort of check for the number of fileSuffixes passed in. From my understanding, it was kind of expected to have 0 or 1 suffixes passed in. Anything besides that seems like a bit of an error given the example I presented above

(sorry, I keep adding more thoughts to this)
I'd guess the reason why the path thing hasn't been an issue for nested paths is the tree relies on IDs to find parents, not paths and path generation at the end is done by walking the tree instead of relying on parent references in nodes. I think the only time we really examine the parent path is for seeing if the item matches a selector (not sure if we still need this support and/or if it's been added)

The same comment also applies to the `folderAt` function below since it has a similar implementation

ashmrtn (commented):
Updates to function calls in tests make sense, but more clarity is needed on the helper function implementation

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: @ashmrtn Great topics, thanks for seeking out the clarity.  I agree it's not obvious at this time.

> First, this seems like it's always going to produce a folder path at the root for the parent reference right now because it's not passing the parentSuffix to dir.

That's correct.  I'll add a comment, in short: we don't investigate the parent dir for files; we only use the ID.  This property is only used for folders.

On that note: we should be putting in the _full_ location for folders, and not just the parent suffix.  If you look at only the [first commit](https://github.com/alcionai/corso/pull/4844/commits/9cdf4eec6ed2d73c54925aada58d7432672ee385) in this PR, I actually did that.  I removed it because it was 1/ wordy, and 2/ unused because our unit tests don't utilize folder subselection (which is the only reason we need to provide the dir).  I've been thinking that I could actually use the tree for the path and drop the need for the parent path entirely.  May still do that later.

> Second, the function as a whole seems like it's only capable of producing files with one level of nesting.

It does seem that way, but due to the above, "levels of nesting" is entirely controlled by the parent folder ID reference.  We don't need files to describe their directory location.  Only the parent in which they appear.

(aside: I wrote this before reading your last paragraph.  Seems like you grok.)

> it may be a good idea to have some sort of check for the number of fileSuffixes passed in

Sure, I'm down with that.

ryanfkeepers: added a hacky-ish check for suffix len.

ashmrtn: Thanks for the detailed response and code updates. Could you please also add a comment to this function and the folder one giving a brief overview of the parent path situation for generated items? I just want to make sure it's documented somewhere so folks aren't surprised/confused by it if they go read the tests but haven't happened to see this comment thread

ashmrtn (approved):

PR #4845, opened 2023-12-14
ensure the tree limiter handling is correct

ryanfkeepers:
enables limit testing on the tree version of drive backups, along with any changes or fixes necessary to make the limiter follow (as closely as possible) the behavior from the non-tree backup process.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/collections_tree.go
+++ b/src/internal/m365/collection/drive/collections_tree.go
@@ -638,17 +638,25 @@ func (c *Collections) addFileToTree(
 	if parentNotNil && !alreadySeen {
 		countSize := tree.countLiveFilesAndSizes()
 
+		// Tell the enumerator to exit if we've already hit the total
+		// limit of bytes or items in this backup.
+		if limiter.alreadyHitTotalBytesLimit(countSize.totalBytes) ||
+			limiter.hitItemLimit(countSize.numFiles) {
+			return nil, errHitLimit
+		}
+
 		// Don't add new items if the new collection has already reached it's limit.
 		// item moves and updates are generally allowed through.
-		if limiter.atContainerItemsLimit(len(parentNode.files)) || limiter.hitItemLimit(countSize.numFiles) {
+		if limiter.atContainerItemsLimit(len(parentNode.files)) {
 			return nil, errHitCollectionLimit
 		}
 
 		// Don't include large files that don't fit within the size limit.
-		// Unlike the other checks, which see if we're already at the limit, this check
-		// needs to be forward-facing to ensure we don't go far over the limit.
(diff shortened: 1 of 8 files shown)


src/internal/m365/collection/drive/collections_tree.go
```diff
@@ -633,15 +641,17 @@ func (c *Collections) addFileToTree(
...
-		// Skip large files that don't fit within the size limit.
-		// unlike the other checks, which see if we're already at the limit, this check
+		// Don't include large files that don't fit within the size limit.
+		// Unlike the other checks, which see if we're already at the limit, this check
 		// needs to be forward-facing to ensure we don't go far over the limit.
 		// Example case: a 1gb limit and a 25gb file.
 		if limiter.hitTotalBytesLimit(fileSize + countSize.totalBytes) {
-			return nil, errHitLimit
+			// don't return errHitLimit here; we only want to skip the
+			// current file.  We may not want to skip files after it.
```
ashmrtn, line 653: hmm this may require an additional check either here or in the caller. It's likely there's a subtle bug in the original implementation as well. The main concern here is that if we do meet or get close to the limit then this will always skip the file but we'll continue to enumerate items. In the worst case where all other limits are set very high, we'll enumerate everything in the drive

We can tune this behavior in another PR, but maybe something better would be something like a check of `bytesSoFar + fileSize >= bytesLimit * <some small fudge factor>` and then a check for `bytesSoFar >= bytesLimit` in another place. That would make it more likely to actually reach the limit (or at least be close to it) without forcing enumeration of all items to try to get it exactly

src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -468,11 +468,15 @@ func walkTreeAndBuildCollections(
...
 	}
 
-	// add the folder itself to the list of files inside the folder.
-	// that will cause the collection processor to generate a metadata
-	// file to hold the folder's permissions.
-	files := maps.Clone(node.files)
-	files[id] = node.folder
+	files := node.files
+
+	if !isRoot {
```
ashmrtn, line 473: does adding the root cause failures with Graph or is this to be more inline with the existing implementation? Not having a special case for this would be a bit cleaner if it doesn't have adverse effects and isn't difficult to get test code setup for

ashmrtn (approved):

src/internal/m365/collection/drive/collections_tree.go
ryanfkeepers: patched here: https://github.com/alcionai/corso/pull/4863

Issue #4754, opened 2023-11-30
Sharepoint Lists Backup & Restore


HiteshRepo:
**Sharepoint support is not enabled yet**
**Backup Disabled**: 
- https://github.com/alcionai/corso/pull/4960
- https://github.com/alcionai/corso/blob/b49c12451209f312c86340865a2ee7ce8b570dba/src/cli/utils/sharepoint.go#L72
**Restore Disabled**: https://github.com/alcionai/corso/blob/13f549c6cbac5437417f4ec1e544055101c9b659/src/cli/restore/sharepoint.go#L99
**Export Disabled**: https://github.com/alcionai/corso/blob/13f549c6cbac5437417f4ec1e544055101c9b659/src/cli/export/sharepoint.go#L99

## Action Items
---

### Enumeration - https://github.com/alcionai/corso/pull/4917
- [x] All lists
- [x] All list items
- [x] All columns
- [x] All contentTypes
- [x] All columnLinks
- [x] All contentTypeColumns

### Interfaces - https://github.com/alcionai/corso/pull/4916

- [x] ListInfo - stores List related details like Name, Template, Item count, Modified time and Created time 
- [x] SharepointInfo - to accommodate ListInfo, give appropriate headers and values

### APIs - https://github.com/alcionai/corso/pull/4917
Package: src/pkg/services/m365/api/lists.go & src/pkg/services/m365/api/lists_pager.go

- [x] GetLists - to fetch all lists (call config to request list-id and last modified time)
- [x] GetListByID - to fetch a list by id, also make requests to fetch associated relations (columns, contentTypes) and list items
- [x] PostList - to create a new list from stored list bytes
- [x] DeleteList - to delete a list

### Packages
- [x] `internal/m365/collection/site` - to accommodate lists backup and restore handlers - https://github.com/alcionai/corso/pull/4919
- [x] `internal/m365/collection/site/backup.go` - creates new collections from fetching all lists - https://github.com/alcionai/corso/pull/4920
- [x] `internal/m365/collection/site/collection.go` - updates to use list backup handler to fetch list by id and stream items for all fetched collections - https://github.com/alcionai/corso/pull/4920
- [x] `internal/m365/collection/site/restore.go` - creates stored list for a particular backup - https://github.com/alcionai/corso/pull/4923
[...]

PR #4850, opened 2023-12-14
enables sharepoint lists category for backup

HiteshRepo:
- enables sharepoint lists category for backup
- tests at cli and operation level

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/sharepoint_e2e_test.go
+++ b/src/cli/backup/sharepoint_e2e_test.go
@@ -1,12 +1,14 @@
 package backup_test
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/alcionai/clues"
 	"github.com/google/uuid"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
@@ -22,12 +24,12 @@ import (
 	"github.com/alcionai/corso/src/internal/tester/tconfig"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/selectors"
-	"github.com/alcionai/corso/src/pkg/selectors/testdata"
+	selTD "github.com/alcionai/corso/src/pkg/selectors/testdata"
 	storeTD "github.com/alcionai/corso/src/pkg/storage/testdata"
 )
 
 // ---------------------------------------------------------------------------
-// tests with no prior backup
+// tests that require no existing backups
 // ---------------------------------------------------------------------------
 
 type NoBackupSharePointE2ESuite struct {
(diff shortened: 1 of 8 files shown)


src/cli/utils/sharepoint.go
```diff
@@ -66,6 +66,33 @@ func MakeSharePointOpts(cmd *cobra.Command) SharePointOpts {
 	}
 }
 
+func SharePointAllowedCategories() map[string]struct{} {
```
ryanfkeepers, line 69: Correct me if I'm wrong, because I can't remember if we made the decision (during a sync meeting) for `Lists` or `Conversations`, but did we want to actually expose lists here, or have it be hidden/not part of the backup unless specified?  I thought we picked the latter while we're finishing up development?

src/cli/utils/sharepoint.go
```diff
@@ -66,6 +66,33 @@ func MakeSharePointOpts(cmd *cobra.Command) SharePointOpts {
...
+	return map[string]struct{}{
+		flags.DataLibraries: {},
+		flags.DataLists:     {},
+		flags.DataPages:     {},
```
ryanfkeepers, line 73: I think you're carrying over old code here, but we do _not_ support backup of pages at this time, and should not display that category in our supported set.  Can you please ensure pages is not allowed (or displayed as supported) in the cli?

src/cli/backup/sharepoint_e2e_test.go
```diff
@@ -216,3 +490,23 @@ func (suite *BackupDeleteSharePointE2ESuite) TestSharePointBackupDeleteCmd_NoBac
...
+// ---------------------------------------------------------------------------
+
+func buildSharepointBackupCmd(
+	ctx context.Context,
+	configFile, site, category string,
+	recorder *strings.Builder,
+) (*cobra.Command, context.Context) {
+	cmd := cliTD.StubRootCmd(
+		"backup", "create", "sharepoint",
+		"--config-file", configFile,
```
ryanfkeepers, line 505: does `config-file` not have a flags entry?  if it doesn't, could you please add a tech debt ticket to fix that up?

ryanfkeepers (changes requested):
generally looks good, but we need to ensure we're exposing the correct set of support.

src/cli/utils/sharepoint.go
HiteshRepo: yeah correct. I will remove 'pages' from here, so that even if we explicitely try to backup pages, we get:
`Error: pages is an unrecognized data type; either  libraries or lists`

src/cli/utils/sharepoint.go
HiteshRepo: I believe the intention was to merge basic backup and restore functionality for lists to main and then work on other supports like: Delta, Metadata,

That is the reason we need to hide lists untill atleast `Incrementals` are in place.

Sure I can comment it here for time being, but If we were to test from cli, we can uncomment it that time. Does that work?

src/cli/backup/sharepoint_e2e_test.go
HiteshRepo: We [have](https://github.com/alcionai/corso/blob/e3363aaa46ca57c5081fd425da3c071d37d95173/src/cli/config/config.go#L77), but I think here we are tying up a configFile path as well for test.

ryanfkeepers: I'm not sure I follow that explanation

ryanfkeepers (approved):

PR #4851, opened 2023-12-14
updates sharepoint info to accommodate site-url and item-template 

HiteshRepo:
updates sharepoint info to accommodate site-url and item-template for lists backup details

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/site/lists.go
```diff
@@ -21,13 +21,19 @@ func ListToSPInfo(lst models.Listable) *details.SharePointInfo {
...
-		ItemType:  details.SharePointList,
-		ItemName:  name,
-		ItemCount: int64(count),
-		Created:   created,
-		Modified:  modified,
-		WebURL:    webURL,
+		ItemType:     details.SharePointList,
+		ItemName:     name,
+		ItemCount:    int64(count),
+		ItemTemplate: template,
```
ryanfkeepers, line 33: I'm a little confused by the value naming here.  The list has N (0 or more) items.  Those are counted in `ItemCount`.  But we have three other properties: `ItemType, ItemName, ItemTemplate` with only one value.  I think that the `Item`, in this case, is supposed to reference the list itself.  But that's a conflict of naming, because the list is not an item, and it's going to cause confusion.  Not only will it be confusing when trying to figure out whether an `Item` refers to a list or not, it will also be difficult to know if other types (ie: libraries) are supposed to populate ItemTemplate and ItemCount and etc.

Since we're only introducing support now, I think it's the right time to move this data to a substruct within sharePointInfo like:

```go
type details.SharePointInfo{
  List            *ListInfo
  ItemType   enum
}

type ListInfo struct {
  Name           string
  ItemCount   int64
  Template     string
  Created       time.TIme
  WebURL      string
  Modified      time.Time
  ...
}
```

This will minimize cross-contamination and ensure we have clear and identifiable properties specifically for list items themselves.  It'll have some structural overlap with other SharePointInfo properties, and that's fine.  Better to be duplicated and easy to understand than to be confusing.

src/pkg/backup/details/sharepoint.go
```diff
@@ -58,7 +61,7 @@ func (i SharePointInfo) Headers() []string {
 	case SharePointLibrary:
 		return []string{"ItemName", "Library", "ParentPath", "Size", "Owner", "Created", "Modified"}
 	case SharePointList:
-		return []string{"ListName", "ListItemsCount", "Owner", "Created", "Modified"}
+		return []string{"ListName", "ListItemsCount", "SiteURL", "Template", "Owner", "Created", "Modified"}
```
ryanfkeepers, line 64: There's a couple conflicts between this and our standard display:
* we don't normally show the SiteURL, since that's already identified by the backup resource.
* `Owner` is shown here, but not populated as part of the item info.
* These names should be very human readable.  Prefer words like `List` over `ListName`, and `Items` over `ListItemsCount`.

ryanfkeepers, line 64: Can you post an example of what the `Template` is going to look like when written to the CLI in this tabular output?  Templates are usually not human-friendly.  I'm dubious about whether we'll want to show it at all.

In general, when updating any CLI display, it's helpful if you add one or more example outputs in comments within the PR so that reviewers can see what the changes will look like on the user's end.

src/pkg/backup/details/sharepoint.go
```diff
@@ -120,3 +125,28 @@ func (i *SharePointInfo) updateFolder(f *FolderInfo) error {
 
 	return clues.New("unsupported non-SharePoint ItemType").With("item_type", i.ItemType)
 }
+
+func getSiteURL(itemWebURL string) string {
```
ryanfkeepers, line 129: This looks very similar to some other code we have in the `/m365/api` package.  Can you please find a central location for it that deduplicates handling?  You can make that change in a separate PR.  Just wanted it noted here.

ryanfkeepers (changes requested):
I have general concerns about the design of SharePointInfo data related to Lists.  I probably should have raised those in an earlier PR/design doc, but it's only now that these things occurred to me.  I think the changes proposed in comments will be best for maintainability in the long run.

Also, since updates are being made here to the CLI output, an example or two of how things look on the terminal would be really helpful.

src/internal/m365/collection/site/lists.go
HiteshRepo: Yes, here the Item represent the `list` itself.
Yes, this contract is much better. Initially I thought about the same, just like they way in `Conversation` and `Channels` in GroupsInfo.
But I got a bit influenced by `ExchangeInfo` having `Events`, `Contacts`, `Mail` fields in same struct. But their fields are quite distinctive to the categories.
I shall make the changes accordingly.

src/pkg/backup/details/sharepoint.go
HiteshRepo: I will drop the `Owner` and make the changes the column names.

HiteshRepo: @ryanfkeepers 
Here is how it would look like. We need `Template` to either not backup `` template lists or not restore it. Becaue it gives `Invalid template used` while creating with that template. I belive it is system generated and not for users to create and those lists are not even visible in the site.

```
 ID            List                                                    Items  SiteURL                                                    Template         Created               Modified            
  091a4e6cb4f8  Corso_Restore_14-Dec-2023_17-37-40_Events               0      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  events           2023-12-14T17:37:58Z  2023-12-14T17:38:01Z
  d623f7f6a8e9  Events                                                  0      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  events           2023-12-10T06:10:56Z  2023-12-13T17:19:57Z
  7ef15e85ea25  Corso_Restore_14-Dec-2023_17-37-40_Documents            0      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  documentLibrary  2023-12-14T17:37:56Z  2023-12-14T17:37:57Z
  9b8a6eed824a  Corso_Restore_14-Dec-2023_17-37-40_list-with-yes-no     2      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:15Z  2023-12-14T17:38:16Z
  e52ef055db79  list-with-datetime                                      1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T18:40:03Z  2023-12-14T07:30:29Z
  d09086fef4e7  Documents                                               0      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  documentLibrary  2023-12-10T06:10:54Z  2023-12-10T06:10:54Z
  ce68f43d5cf1  Corso_Restore_14-Dec-2023_17-37-40_list-with-multiline  1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:11Z  2023-12-14T17:38:11Z
  ed98d30966d0  new-books                                               2      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T09:23:06Z  2023-12-14T07:30:29Z
  719c04a93e14  list-with-yes-no                                        2      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T18:34:52Z  2023-12-14T07:30:35Z
  0a5afc266a59  Corso_Restore_14-Dec-2023_17-37-40_list-with-metadata   1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:09Z  2023-12-14T17:38:10Z
  6ff9af316541  link-with-choice                                        1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T11:56:52Z  2023-12-14T07:30:29Z
  bd9a5e592ca6  list-with-multiline                                     1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T18:41:51Z  2023-12-14T07:30:29Z
  4eaa2bbf521f  list-with-lookup                                        1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T18:44:56Z  2023-12-14T07:30:29Z
  5e4f541906db  list-with-metadata                                      1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T18:46:17Z  2023-12-14T07:30:30Z
  28d9be63ff0d  Corso_Restore_14-Dec-2023_17-37-40_list-with-person     1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:13Z  2023-12-14T17:38:14Z
  b96b04147533  Corso_Restore_14-Dec-2023_17-37-40_new-books            2      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:17Z  2023-12-14T17:38:19Z
  8738089bb528  Corso_Restore_14-Dec-2023_17-37-40_list-with-datetime   1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:05Z  2023-12-14T17:38:05Z
  49940bd33ed7  Corso_Restore_14-Dec-2023_17-37-40_list-with-lookup     1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:07Z  2023-12-14T17:38:08Z
  915f535a99dc  Corso_Restore_14-Dec-2023_17-37-40_link-with-choice     1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-14T17:38:03Z  2023-12-14T17:38:03Z
  7276fbf86223  list-with-person                                        1      https://10rqc2.sharepoint.com/sites/site-4754-small-lists  genericList      2023-12-13T12:19:52Z  2023-12-14T07:30:33Z

```

src/pkg/backup/details/sharepoint.go
HiteshRepo: found it [here](https://github.com/alcionai/corso/blob/e3363aaa46ca57c5081fd425da3c071d37d95173/src/pkg/services/m365/api/sites.go#L172-L191), but I guess it is not exactly what I am trying to do here.

src/pkg/backup/details/sharepoint.go
ryanfkeepers: > I thought we could use SiteURL as a selector and hence it would be useful for it to be displayed

Ah, good question.  That has two use cases.
1. when showing `sharepoint` app details, each backup is unique to the resource owner- IE: the site.  Therefore, displaying the site URL is redundant, because you cannot select any other site.
2. when showing `groups` app details, since each backup may contain multiple sites (because the group is the resource, and groups many contain many sites), we might want to show the site URL in that case.

Also, we probably want to name the column `Site`, instead of `SiteURL`.

ryanfkeepers: > We need Template to either not backup webTemplateExtensionsList template lists or not restore it.

I grok.  But I don't think we should display it.  That means the end user needs to know the difference between backup items that are, and are not, restorable.  It pushes complexity out onto the user, instead of handling it ourselves, which isn't up to our standards.

What you should do, instead, is to filter all non-restorable template types out of the backup details list.  That will ensure end users are able to restore any item we display.  If we want to view the unfiltered list, then passing in the `--skip-reduce` string should display all list items without filtering them.

That way we can keep the template in the details info, but not display it to the end user.

src/pkg/backup/details/sharepoint.go
ryanfkeepers: hmm, fair.  I think you could still unify the two.  But it's a low priority.

src/pkg/backup/details/sharepoint.go
HiteshRepo: I will implement `--skip-reduce` for backup details display in a different PR.
And filter `webTemplateExtensionsList` lists while restoring.

src/cli/utils/sharepoint.go
```diff
@@ -66,6 +66,30 @@ func MakeSharePointOpts(cmd *cobra.Command) SharePointOpts {
...
+func SharePointAllowedCategories() map[string]struct{} {
+	return map[string]struct{}{
+		flags.DataLibraries: {},
+		// flags.DataLists:     {},
+	}
+}
+
+func AddCategories(sel *selectors.SharePointBackup, cats []string) *selectors.SharePointBackup {
+	if len(cats) == 0 {
+		sel.Include(sel.LibraryFolders(selectors.Any()), sel.Lists(selectors.Any()))
```
ryanfkeepers, line 78: If lists isn't an allowed category (line 72), then we shouldn't include it in this selector.

In the future, when it's enabled, you should be able to use this func instead.
```suggestion
		sel.Include(sel.AllData())
```

ryanfkeepers (approved):
@HiteshRepo I feel like I"m seeing some code I've reviewed already.  Does this PR contain a merger from other PRs that I've approved?  If so, please hold off on merging branches together until all branches (both the base and the feature) have approval.  Otherwise it becomes significantly more difficult to isolate changes for review.

PR #4853, opened 2023-12-14
provision to post and delete lists or list items

HiteshRepo:
- adds methods to Lists client wrapper to create list and list items
- adds methods to Lists client wrapper to delete list
- handles graph error: General exception while processing by column validation nil check
- handles graph error: Unable to identify column definition
- handles location type of columns and its data

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/lists_test.go
```diff
@@ -68,6 +69,389 @@ func (suite *ListsUnitSuite) TestBytesToListable() {
...
+
+			require.NotEmpty(t, newCd)
+			assert.True(t, test.checkFn(newCd))
+		})
+	}
+}
+
+func (suite *ListsUnitSuite) TestLegacyColumnsAreNotSet() {
+	listName := "test-list"
+	contentTypeColumnName := "ContentType"
```
ryanfkeepers, line 240: suggestion: turn these values into constants.  Use those consts as both the keys in your list structure maps, as well as here in the tests.

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+	ctx context.Context,
+	siteID string,
+	listName string,
+	oldListByteArray []byte,
+) (models.Listable, error) {
+	newListName := listName
+
+	oldList, err := BytesToListable(oldListByteArray)
+	if err != nil {
+		return nil, clues.WrapWC(ctx, err, "creating old list")
```
ryanfkeepers, line 190:
```suggestion
		return nil, clues.WrapWC(ctx, err, "generating list from stored bytes")
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+	}
+
+	if name, ok := ptr.ValOK(oldList.GetDisplayName()); ok {
+		nameParts := strings.Split(listName, "_")
+		if len(nameParts) > 0 {
+			nameParts[len(nameParts)-1] = name
+			newListName = strings.Join(nameParts, "_")
+		}
+	}
```
ryanfkeepers, line 199: This behavior is opaque.  Can you please add a comment explaining what's being handled here, and why that handling is necessary?  Comments like that will help all future maintainers grok the implementation.

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+
+	// this ensure all columns, contentTypes are set to the newList
+	newList := ToListable(oldList, newListName)
+
+	// Restore to List base to M365 back store
```
ryanfkeepers, line 204:
```suggestion
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+	restoredList, err := c.Stable.Client().Sites().BySiteId(siteID).Lists().Post(ctx, newList, nil)
```
ryanfkeepers, line 205: nit: prefer multiline on func chains for code readability.
```suggestion
	restoredList, err := c.Stable.
	  Client().
	  Sites().
	  BySiteId(siteID).
	  Lists().
	  Post(ctx, newList, nil)
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "restoring list")
+	}
+
+	return restoredList, nil
```
ryanfkeepers, line 210:
```suggestion
	return restoredList, graph.Wrap(ctx, err, "creating list").OrNil()
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+		}
+	}
+
+	// this ensure all columns, contentTypes are set to the newList
+	newList := ToListable(oldList, newListName)
+
+	// Restore to List base to M365 back store
+	restoredList, err := c.Stable.Client().Sites().BySiteId(siteID).Lists().Post(ctx, newList, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "restoring list")
```
ryanfkeepers, line 207: Nit: "restoring" is the total process of taking data from backup and posting it to graph api.  It's a high level concept.  What failed here is "posting" (ie: creating) the list.  It's helpful to keep the two notions separate, othewise it can be difficult to grok the layer at which errors occur.

This suggestion applies to all other error messages similar to this one, too.

```suggestion
		return nil, graph.Wrap(ctx, err, "creating list")
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+}
+
+func (c Lists) PostListItem(
+	ctx context.Context,
+	siteID, listID string,
+	oldListByteArray []byte,
+) ([]models.ListItemable, error) {
+	oldList, err := BytesToListable(oldListByteArray)
+	if err != nil {
+		return nil, clues.WrapWC(ctx, err, "creating old list to get list items")
```
ryanfkeepers, line 220: Nit: the term "creating" should be saved for talking about the creation of stored data.  IE: POST requests.
```suggestion
		return nil, clues.WrapWC(ctx, err, "generating list item from stored bytes")
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+			Client().
+			Sites().
+			BySiteId(siteID).
+			Lists().
+			ByListId(listID).
+			Items().
+			Post(ctx, lItem, nil)
+		if err != nil {
+			return nil, graph.Wrap(ctx, err, "restoring list items").
+				With("restored_list_id", listID)
```
ryanfkeepers, line 241: you can assume the caller already embedded the list id in the ctx clues
```suggestion
			return nil, graph.Wrap(ctx, err, "creating item in list")
```

src/pkg/services/m365/api/lists.go
```diff
@@ -143,6 +177,91 @@ func (c Lists) getListContents(ctx context.Context, siteID, listID string) (
...
+		Client().
+		Sites().
+		BySiteId(siteID).
+		Lists().
+		ByListId(listID).
+		Delete(ctx, nil); err != nil {
+		return graph.Wrap(ctx, err, "deleting list")
+	}
+
+	return nil
```
ryanfkeepers, line 262: two nits for code standards:
1. don't embed multi-line func calls in `if` conditionals.
2. you can use `OrNil()` to simplify the error return.
```suggestion
	err := c.Stable.
		Client().
		Sites().
		BySiteId(siteID).
		Lists().
		ByListId(listID).
		Delete(ctx, nil)
		
		return graph.Wrap(ctx, err, "deleting list").OrNil()
```

src/pkg/services/m365/api/lists.go
```diff
@@ -218,44 +332,138 @@ func ToListable(orig models.Listable, displayName string) models.Listable {
...
+
+	validation := orig.GetValidation()
+	if validation != nil {
+		newColumn.SetValidation(validation)
+	}
 
 	return newColumn
 }
 
+func setColumnType(newColumn *models.ColumnDefinition, orig models.ColumnDefinitionable) {
```
ryanfkeepers, line 372: I think this func could be simplified a lot by using a blank switch statement.  ex:
```go
switch {
case orig.GetText() != nil:
  newColumn.SetText(orig.GetText())
case orig.GetBoolean() != nil:
  newColumn.SetText(orig.GetBoolean())
...
default:
  newColumn.SetText(models.NewTextColumn())
}
```

src/pkg/services/m365/api/lists.go
```diff
@@ -290,25 +498,67 @@ func CloneListItem(orig models.ListItemable) models.ListItemable {
...
+		if nestedFields, ok := value.(map[string]any); ok &&
+			hasRequiredFields(nestedFields, addressFieldNames) &&
+			hasRequiredFields(additionalData, readonlyAddressFieldNames) &&
+			key != "GeoLoc" {
+			return true
+		}
+	}
+
+	return false
+}
```
ryanfkeepers, line 554: There's a lot of code packed into this conditional, and it's not very easy to read.  Suggestion for layout:
```suggestion
  nestedFields, ok := value.(map[string]any)
  if !ok || key == "GeoLoc" {
    continue
  }
  if map.HasAllFields(
  
		if ; ok &&
			hasRequiredFields(nestedFields, addressFieldNames) &&
			hasRequiredFields(additionalData, readonlyAddressFieldNames) &&
			key != "GeoLoc" {
			return true
		}
```

```suggestion
func hasAddressFields(additionalData map[string]any) bool {
	if !hasRequiredFields(additionalData, readOnlyAddressFieldNames) {
		return false
	}
	
	for key, value := range additionalData {
		nestedFields, ok := value.(map[string]any)
		if !ok || key == "GeoLoc" {
			continue
		}
		if hasRequiredFields(nestedFields, addressFieldNames) {
			return true
		}
	}

	return false
}
```

src/pkg/services/m365/api/lists.go
```diff
@@ -290,25 +498,67 @@ func CloneListItem(orig models.ListItemable) models.ListItemable {
...
+
+func hasRequiredFields(data map[string]any, checkFieldNames map[string]struct{}) bool {
```
ryanfkeepers, line 556: Suggestion: this seems like a helpful func to have in a common package.  something like `/common/map` with the signature:
```go
func HasKeys(data map[string]any, keys ...string) bool {
```

That would both give it a central location, and make it something that's easy to test independent of the api code.

src/pkg/services/m365/api/lists.go
```diff
@@ -290,25 +498,67 @@ func CloneListItem(orig models.ListItemable) models.ListItemable {
...
 
+	retainPrimaryAddressField(additionalData)
+
 	fields.SetAdditionalData(additionalData)
 
 	return fields
 }
+
+func retainPrimaryAddressField(additionalData map[string]any) {
+	if !hasAddressFields(additionalData) {
```
ryanfkeepers, line 534: question: is this check necessary?  Either the additionalData has those fields, or it doesn't.  If it does have them, they need to get deleted (line 539).  If they don't, then we can run a no-op delete without issue.

src/pkg/services/m365/api/lists.go
```diff
@@ -290,25 +498,67 @@ func CloneListItem(orig models.ListItemable) models.ListItemable {
...
+		// .           -> @odata.etag : Embedded link to Prior M365 ID
+		// -- String Match: Read-Only Fields
+		// -> id : previous un
+		for key, value := range fieldData {
+			_, isReadOnlyField := readOnlyFieldNames[key]
+
+			if strings.HasPrefix(key, "_") ||
+				strings.HasPrefix(key, "@") ||
+				isReadOnlyField ||
+				strings.Contains(key, "LinkTitle") ||
```
ryanfkeepers, line 517: nit: hitting my ocd here 😁 
```suggestion
			if isReadOnlyField ||
				strings.HasPrefix(key, "_") ||
				strings.HasPrefix(key, "@") ||
				strings.Contains(key, "LinkTitle") ||
```

src/pkg/services/m365/api/lists.go
```diff
@@ -290,25 +498,67 @@ func CloneListItem(orig models.ListItemable) models.ListItemable {
...
+				strings.Contains(key, "ChildCount") {
```
ryanfkeepers, line 518: hardcoded strings should get const values, please.  This applies to all cases in this PR, not only this line.

src/pkg/services/m365/api/lists.go
```diff
@@ -11,6 +11,40 @@ import (
 	"github.com/alcionai/corso/src/pkg/services/m365/api/graph"
 )
 
+var legacyColumns = map[string]struct{}{
```
ryanfkeepers, line 14: I like your impulse to centralize these values.  It's good for code readability and cleanliness.  Looking at the usage of these maps, however, I think there's a couple extra protections that would be nice.  In general, this suggestion comes from how using global maps can produce a smell, since alterations to those maps at runtime can be difficult to detect and debug.

For the most part, it seems like you're using the maps purely as a check for whether a key exists in a certain set.  For those uses, I'd suggest getter funcs like `hasKey(string) bool`.

In one other case, you're actually looking for a slice of the keys, and are hacking the map to do the same.  In that case I'd suggest a  func like `keys() []string`.

Putting it together, you can easily get both cases with a type wrapper, like so:
```go
type keySet map[string]struct {}

func (ks keySet) hasKey(k string) bool {...}
func (ks keySet) keys() []string {...}

var legacyColumns = keySet{...}
```

That makes it safe and easy for implementations to do the same check you're currently doing by instead calling `if legacyColumns.hasKey(k) {...}`

src/pkg/services/m365/api/lists_test.go
```diff
@@ -68,6 +69,389 @@ func (suite *ListsUnitSuite) TestBytesToListable() {
 	}
 }
 
+func (suite *ListsUnitSuite) TestColumnDefinitionValidationSetWhenProvided() {
+	tests := []struct {
+		name     string
+		getOrig  func() models.ColumnDefinitionable
+		checkNil bool
```
ryanfkeepers, line 76: you can pack the assert call into the property itself here.
```suggestion
		expect assert.ValueAssertionFunc
```

This lets you change the test value to:
```go
expect: assert.Nil // or assert.NotNil
```

and verify the result like so
```go
newCd := cloneColumnDefinitionable(orig)
require.NotEmpty(t, newCd)
test.expect(t, newCd.GetValidation())
```

src/pkg/services/m365/api/lists_test.go
```diff
@@ -68,6 +69,389 @@ func (suite *ListsUnitSuite) TestBytesToListable() {
 	}
 }
 
+func (suite *ListsUnitSuite) TestColumnDefinitionValidationSetWhenProvided() {
```
ryanfkeepers, line 72: For test naming, code standard is to follow golang's testable example naming schemes.  see: https://go.dev/blog/examples#example-function-names

In short, it goes something like `TestStructName_FuncName_subtestExplanation`.

so in this case you're testing:
```suggestion
func (suite *ListsUnitSuite) TestColumnDefinitionable_GetValidation() {
```

src/pkg/services/m365/api/lists_test.go
```diff
@@ -267,3 +651,180 @@ func (suite *ListsAPIIntgSuite) TestLists_GetListByID() {
...
+	defer writer.Close()
+
+	oldListID := "old-list"
+	listItemID := "list-item1"
+	textColumnDefID := "list-col1"
+	textColumnDefName := "itemName"
+
+	textColumn := models.NewTextColumn()
+
+	txtColumnDef := models.NewColumnDefinition()
```
ryanfkeepers, line 732: This is a lot of chaff code.  Is there any way to centralize it, or parts of it, for ease of re-use across multipl tests?

src/internal/common/maps/keyset.go
```diff
@@ -0,0 +1,21 @@
+package maps
```
ryanfkeepers, line 1: as a warning, this may collide with the golang `maps` package, which we use elsewhere.  I don't know a better name off the top of my head.  Maybe `keys`?  Then you can change the value from `keys.KeySet` to just Set, for `keys.Set`?

src/internal/common/maps/keyset_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+type KeySetTestSuite struct {
+	tester.Suite
+}
+
+func TestKeySetTestSuite(t *testing.T) {
+	suite.Run(t, &KeySetTestSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+func (suite *KeySetTestSuite) Test_HasKey() {
```
ryanfkeepers, line 20: Nit: no underscore needed when naming the function under test directly.  Underscores are for methods, variables, and extensions.  eg:
```go
func TestStruct()
func TestStruct_Method()
func TestFunc()
func TestStruct_hasSpecificProperty()
func TestStruct_Method_doesSomethingSpecific()
func TestFunc_doesSomethingSpecific()
```
```suggestion
func (suite *KeySetTestSuite) TestHasKey() {
```

src/internal/common/maps/maps.go
```diff
@@ -0,0 +1,11 @@
+package maps
+
+func HasKeys(data map[string]any, keys ...string) bool {
```
ryanfkeepers, line 3: I think you could move this into the `keyset.go` file, since it's very small.

ryanfkeepers (approved):
looking good!  Thank you for the updates.

PR #4854, opened 2023-12-14
implement lists restore handler

HiteshRepo:
adds lists restore handler to:
- create list
- create list items
- delete list

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/site/handlers.go
+++ b/src/internal/m365/collection/site/handlers.go
@@ -23,7 +23,6 @@ type getItemser interface {
 
 type restoreHandler interface {
 	PostLister
-	PostListItemer
 	DeleteLister
 }
 
@@ -31,16 +30,8 @@ type PostLister interface {
 	PostList(
 		ctx context.Context,
 		listName string,
-		oldListByteArray []byte,
+		storedListData []byte,
 	) (models.Listable, error)
-}
-
-type PostListItemer interface {
-	PostListItem(
-		ctx context.Context,
-		listID string,
-		oldListByteArray []byte,
-	) ([]models.ListItemable, error)
 }
 
 type DeleteLister interface {
--- a/src/internal/m365/collection/site/handlers.go
+++ b/src/internal/m365/collection/site/handlers.go
@@ -20,3 +20,32 @@ type getItemByIDer interface {
 type getItemser interface {
 	GetItems(ctx context.Context, cc api.CallConfig) ([]models.Listable, error)
 }
+
+type restoreHandler interface {
+	PostLister
+	PostListItemer
+	DeleteLister
+}
+
+type PostLister interface {
+	PostList(
+		ctx context.Context,
+		listName string,
+		oldListByteArray []byte,
+	) (models.Listable, error)
+}
+
+type PostListItemer interface {
+	PostListItem(
+		ctx context.Context,
+		listID string,
+		oldListByteArray []byte,
+	) ([]models.ListItemable, error)
+}
+
+type DeleteLister interface {
+	DeleteList(
+		ctx context.Context,
+		listID string,
+	) error
+}
--- a/src/pkg/services/m365/api/lists.go
+++ b/src/pkg/services/m365/api/lists.go
@@ -255,12 +255,26 @@ func (c Lists) PostList(
 		return nil, graph.Wrap(ctx, err, "creating list")
 	}
 
-	listItems, err := c.PostListItems(ctx, siteID, restoredList)
+	listItems := make([]models.ListItemable, 0)
+
+	for _, itm := range oldList.GetItems() {
+		temp := CloneListItem(itm)
+		listItems = append(listItems, temp)
+	}
+
+	err = c.PostListItems(
+		ctx,
+		siteID,
+		ptr.Val(restoredList.GetId()),
+		listItems)
 	if err == nil {
 		restoredList.SetItems(listItems)
 		return restoredList, nil
 	}
 
+	// [TODO](hitesh) double check if we need to:
+	// 1. rollback the entire list
+	// 2. restore as much list items possible and add recoverables to fault bus
 	// rollback list creation
 	err = c.DeleteList(ctx, siteID, ptr.Val(restoredList.GetId()))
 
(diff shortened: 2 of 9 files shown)


src/internal/m365/collection/site/handlers.go
```diff
@@ -20,3 +20,32 @@ type getItemByIDer interface {
...
+		oldListByteArray []byte,
+	) (models.Listable, error)
+}
+
+type PostListItemer interface {
+	PostListItem(
+		ctx context.Context,
+		listID string,
+		oldListByteArray []byte,
+	) ([]models.ListItemable, error)
```
ryanfkeepers, line 43: hmm, I have some api conerns here.  The name suggests you're posting one item.  The return value has many items.  That doesn't match up.

It looks like (according to the name) that you're feeding this func the same byte slice as you give the PostList func.  That means that one of the following is true:
1. all items are contained in that slice, and their creation should be handled within `PostList` instead of running a two-step update function.
2. this is a different slice of bytes, and should be named differently to match.

I assume 1/ is true.  I would suggest packing the full restore process into that function instead of splitting it out.  That way this layer is only concerned with the complete creation of a list using the original byte slice, and the api can deal with the details.

src/internal/m365/collection/site/handlers.go
```diff
@@ -20,3 +20,32 @@ type getItemByIDer interface {
...
+	PostLister
+	PostListItemer
+	DeleteLister
+}
+
+type PostLister interface {
+	PostList(
+		ctx context.Context,
+		listName string,
+		oldListByteArray []byte,
```
ryanfkeepers, line 34: naming nits:
1. `old` implies that there's a `newListByteArray` somewhere.  I don't think that's the case.
2. this is a `slice`, not an `array`.

```suggestion
		listData []byte,
```

ryanfkeepers (commented):
Some review of the handler/restore process api needed.

Separate from the work in this PR: this looks like it'll give us proper support for restore using sharepoint.  What about accessing this data via groups and teams?  Do those services even have list data to backup?  (I assume they do).

src/internal/m365/collection/site/handlers.go
HiteshRepo: Yes the first point is correct. 
Sure let me put the PostItem into PostList. 
I will change the function schema as well:
- Name of the function to `PostListItems`
- Accepts `models.Listable` and not `listID` & `oldListByteArray`

src/pkg/services/m365/api/lists.go
```diff
@@ -251,42 +251,49 @@ func (c Lists) PostList(
...
+
+	listItems, err := c.PostListItems(ctx, siteID, restoredList)
+	if err == nil {
+		restoredList.SetItems(listItems)
+		return restoredList, nil
+	}
 
-	return restoredList, graph.Wrap(ctx, err, "creating list").OrNil()
+	// rollback list creation
+	err = c.DeleteList(ctx, siteID, ptr.Val(restoredList.GetId()))
```
ryanfkeepers, line 265: can you please add a `TODO` comment here to say that we need to investigate the behavior we want?

src/pkg/services/m365/api/lists.go
```diff
@@ -251,42 +251,49 @@ func (c Lists) PostList(
 		BySiteId(siteID).
 		Lists().
 		Post(ctx, newList, nil)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "creating list")
+	}
+
+	listItems, err := c.PostListItems(ctx, siteID, restoredList)
```
ryanfkeepers, line 258: Question: if the restored list already contains the items, were they already included in the restore when you POST the newList?

src/pkg/services/m365/api/lists.go
```diff
@@ -251,42 +251,49 @@ func (c Lists) PostList(
...
+	return nil, graph.Wrap(ctx, err, "deleting restored list after items creation failure").OrNil()
 }
 
-func (c Lists) PostListItem(
+func (c Lists) PostListItems(
 	ctx context.Context,
-	siteID, listID string,
-	oldListByteArray []byte,
+	siteID string,
+	restoredList models.Listable,
```
ryanfkeepers, line 273: Nit: although it seems like a natural flow to pass in a Listable, it doesn't make sense as an api design.  Since the func only restores list items, it should only take in a slice of list items.  That reduces caller overhead by not needing a Listable, since the Listable isn't actually used except to get its slice of items and ID.
```suggestion
	siteID, listID string,
	items []models.ListItemable,
```

src/pkg/services/m365/api/lists.go
```diff
@@ -251,42 +251,49 @@ func (c Lists) PostList(
...
-	if err != nil {
-		return nil, clues.WrapWC(ctx, err, "generating list item from stored bytes")
-	}
+	listItems := make([]models.ListItemable, 0)
 
-	contents := make([]models.ListItemable, 0)
-
-	for _, itm := range oldList.GetItems() {
+	for _, itm := range restoredList.GetItems() {
 		temp := CloneListItem(itm)
```
ryanfkeepers, line 278: hmm, can you please explain why cloning the item is necessary?

ryanfkeepers (approved):
Some design nits, and we still need to double check the delete-on-restore-failure behavior expectations.  But generally looks good enough to approve and unblock.

src/pkg/services/m365/api/lists.go
HiteshRepo: When we post the newList, we do not include the items within it.
We create the `storedList` from `storedListBytes`, but we create new empty list instance and populate only metadata items like columns, contentTypes and other list attributes to the newList.

src/pkg/services/m365/api/lists.go
HiteshRepo: Instead of directly passing all the attributes from the `storedListItemBytes`, we check and filter them to pass only the attributes and data enough to create the list item.

Else we might encounter errors while creation of list item:
- 'Invalid request' --> if CreatedByUser / LastModifiedByUser is set with nil
- 'Invalid request' --> if number of columns do not match number of field data of the list item
- 'ColumnDefinition unidentifiable' --> In cases of certain columns like location, there is no way to identify if is a location column or not. So we default to text instead of not sending any column. The address value is then populated as text while list item creation.
- 'General exception while processing' --> if DefaultValue and Validation attributes of models.ColumnDefinitionable is set with nil

PR #4855, opened 2023-12-14
enables restore paths creation and some cleanup

HiteshRepo:
- removes delete lists function from collections package
- enables restore paths creation
- relocates restore lists collection test

- [x] :no_entry: No

#### Type of change

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/operations/pathtransformer/restore_path_transformer_test.go
```diff
@@ -156,23 +156,24 @@ func (suite *RestorePathTransformerUnitSuite) TestGetPaths() {
...
-					ItemInfo: details.ItemInfo{
-						SharePoint: &details.SharePointInfo{
-							ItemType: details.SharePointList,
-						},
-					},
-				},
-			},
-			expectErr: assert.Error,
-		},
+		// TODO(hitesh): fix this test
```
ryanfkeepers, line 159: What's the issue here?  the comment isn't very helpful.

src/internal/m365/collection/site/collection_test.go
```diff
@@ -282,66 +279,3 @@ func (suite *SharePointCollectionSuite) TestCollection_streamItems() {
 		})
 	}
 }
-
-// TestRestoreListCollection verifies Graph Restore API for the List Collection
-func (suite *SharePointCollectionSuite) TestListCollection_Restore() {
```
ryanfkeepers, line 287: were any lines changed with this movement?

HiteshRepo: No @ryanfkeepers ,
This was moved to `/restore_test.go` under a new suite `SharePointRestoreSuite`.
The suite-setup and imports resulted in 70 additional lines.

meain:
@ryanfkeepers @HiteshRepo I'm a bit OOTL on the discussion around the path format. Was there any reasoning behind using the id as the filename? Should we just use a fixed value like `data`. I don't see using the id providing any advantage (unless we remove one more top level folder where we use the name).

I like the current format of having a folder for each list so that it is easier for us to add any additional metadata or split the current file in future. Using names for the folder also seems OK as SharePoint does not seem to let one create lists with duplicate names.

ryanfkeepers (approved):

HiteshRepo:
> @ryanfkeepers @HiteshRepo I'm a bit OOTL on the discussion around the path format. Was there any reasoning behind using the id as the filename? Should we just use a fixed value like `data`. I don't see using the id providing any advantage (unless we remove one more top level folder where we use the name).
> 
> I like the current format of having a folder for each list so that it is easier for us to add any additional metadata or split the current file in future. Using names for the folder also seems OK as SharePoint does not seem to let one create lists with duplicate names.

@meain changed kopia loc for list dir name from list-name to list-id and the file name is also going to be list-id itself instead of 'data'. https://github.com/alcionai/corso/pull/4893

PR #4857, opened 2023-12-14
Add config passing test that future PRs will leverage

ashmrtn:
We're currently passing some backup options during repo connect which
complicates how SDK users can connect to a corso repo. This PR starts
to lay the groundwork for passing a discrete set of config parameters
to backup operations

This PR adds a test around config values passed into the
NewBackupOperation function. It checks that the config values stored
in the backup operation match what's passed in. It also makes sure
the passed in config parameters have all non-zero values so that
future changes don't silently pass the test when they shouldn't

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/operations/backup_test.go
+++ b/src/internal/operations/backup_test.go
@@ -3,7 +3,9 @@ package operations
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	stdpath "path"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


src/internal/operations/backup_test.go
```diff
@@ -365,6 +369,109 @@ func TestBackupOpUnitSuite(t *testing.T) {
...
+	for i := 0; i < v.NumField(); i++ {
+		f := v.Field(i)
+
+		if err := checkPopulatedInner(f); err != nil {
+			errs = clues.Stack(errs, clues.Wrap(err, fmt.Sprintf("field at index %d", i)))
+		}
+	}
+
+	return errs.OrNil()
+}
```
ryanfkeepers, line 392: thoughts on moving this into /ptr?

src/internal/operations/backup_test.go
```diff
@@ -365,6 +369,109 @@ func TestBackupOpUnitSuite(t *testing.T) {
...
+	}
+
+	t := suite.T()
+
+	ctx, flush := tester.NewContext(t)
+	defer flush()
+
+	// This is a sanity check to make sure all fields on the input control.Options
+	// are populated. This helps ensure that this chunk of code stays updated as
+	// options are added to the struct.
```
ryanfkeepers, line 446: can you please move/duplicate this comment onto `checkPopulated` as well?

src/internal/operations/backup_test.go
```diff
@@ -365,6 +369,109 @@ func TestBackupOpUnitSuite(t *testing.T) {
...
+
+	return errs.OrNil()
+}
+
+func checkPopulated(t *testing.T, input control.Options) {
+	err := checkPopulatedInner(reflect.ValueOf(input))
+	require.NoError(t, err, clues.ToCore(err))
+}
+
+func (suite *BackupOpUnitSuite) TestBackupOperation_OptionPassing() {
```
ryanfkeepers, line 399: The purpose of this test is a little obtuse by code alone.  I think if you added a comment here that matches, more or less, what you wrote for the PR description, it would help convey purpose and clarity.

Separately: nit on test func naming: `NewBackupOperation` is what's being tested (thus uppercase).  `configsMatchOperationOptions` is the test focus, but is not a func or struct (thus lowercase).
```suggestion
func (suite *BackupOpUnitSuite) TestNewBackupOperation_configsMatchOperationOptions() {
```

ryanfkeepers (approved):
I'm

src/internal/operations/backup_test.go
ashmrtn: do we have a need for it in other places? If so then sure, otherwise I'd prefer to wait until we have at least another usage

PR #4860, opened 2023-12-14
Add struct definitions and CLI helpers for backup options

ashmrtn:
Define structs that will be used for backup options and add some CLI
helpers/tests to populate those structs with flag values

Rate limiter config is pulled out as a separate struct because it will
likely be used for backup and restore operations and it has values that
are [passed separately](https://github.com/alcionai/corso/blob/505c06441a223f84d6dd81b42a7a56c83ea30f7c/src/internal/m365/backup.go#L232) to the rate limiter config code

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/cli/utils/options.go
```diff
@@ -41,3 +41,27 @@ func ControlWithConfig(cfg config.RepoDetails) control.Options {
 
 	return opt
 }
+
+func BackupOptions() control.Backup {
```
ryanfkeepers, line 45: Nit: to make the purpose a little more clear.  `BackupOptions` sounds more like the name of a struct.
```suggestion
func DefaultBackupOptions() control.Backup {
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,96 @@
...
+			CollectionBuffer: 4,
+			ItemFetch:        4,
+		},
+	}
+}
+
+// Backup is the set of options used for backup operations. Each set of options
+// is only applied to the backup operation it's passed to. To use the same set
+// of options for multiple backup operations pass the struct to all operations.
+type Backup struct {
```
ryanfkeepers, line 22: Nit: `Backup` is a little generic.  Can we do this to match how we've named the restore counterpart?
```suggestion
type BackupConfig struct {
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,96 @@
...
+	// DeltaPageSize controls the quantity of items fetched in each page
+	// during multi-page queries, such as graph api delta endpoints.
+	DeltaPageSize        int32                              `json:"deltaPageSize"`
+	FailureHandling      FailurePolicy                      `json:"failureHandling"`
+	ItemExtensionFactory []extensions.CreateItemExtensioner `json:"-"`
+	Parallelism          Parallelism                        `json:"parallelism"`
+	ToggleFeatures       BackupToggles                      `json:"toggleFeatures"`
```
ryanfkeepers, line 29: I'm not fully on board with the naming/substruct here. These aren't really feature flags, and `Toggles` doesn't really mean anything.  It's too much a grab bag of "options that are arbitrarily nested instead of top level."  Do you mind fixing it up a bit, since you're here?

Looking at the struct, most of the properties are  related to incremental behavior.  A simple change would be to rename it `IncrementalsConfig`.  The struct could remain more or less where it is, but we'd want to move `ExchangeImmutableIDs` out- maybe to an ExchangeConfig struct?  Or maybe just top level- and move `DeltaPageSize` into it.

ryanfkeepers (approved):

src/pkg/control/backup.go
ashmrtn: I agree that the grouping is kind of arbitrary, but looking at what's in the current set of toggles I don't feel like I see a grouping that is less arbitrary. The main things in there related to incrementals is just `DisableIncrementals` and `ForceItemDataDownload`. `DisableDelta`, `RunMigrations`, and `UseDeltaTree` are all related to specific services/versions of backups, not really incrementals as a whole. The same logic kind of applies with `ExchangeImmutableIDs`

ryanfkeepers: > are all related to specific services/versions of backups

How much of that is due to incomplete support?  All services should probably support `DisableDelta`, `ForceItemDataDownload`, and `RunMigrations` (though that one probably needs a flag for each migration, really).  By and large we just haven't followed up on making sure that our one-off toggles get full support after they've achieved their v0 purpose.

After that, sure, `UseDeltaTree` stands out as being drive/sharepoint/groups specific (ie: 75% of our service coverage).  But it's still very much a "delta/incremental" behavior set.

No strong feelings here.  It's a minimal impact on our daily lives either way.

src/cli/utils/options.go
ashmrtn: hmmm I'll work on the naming, not sure I want to go with `Default` in this function though since it does parse values from the passed in flags

src/pkg/control/backup.go
ryanfkeepers: > A different external service may not have that sort of support or may not map neatly onto that concept

I've been wondering if we'll just end up calling the next provider's incremental changelog `Delta`s for ease of communication.

ashmrtn: after staring at stuff for a bit, I came up with the following based on what's already defined and the suggestions you presented earlier:
```go
type BackupM365Config struct {
  DeltaPageSize int32
  UseExchangeImmutableIDs bool
  DisableDelta bool
}

type BackupFeatureConfig struct {
  UseDeltaTree bool
  RunMigrations bool
}
```

This presents the delta tree update as a sort of "migration" (in this case a feature migration, not a data one) so groups it there. It also groups what to me feels like the M365-specific stuff into a single place. In addition to this we can make a struct specifically for the two incremental backup flags or just leave them in `BackupConfig`

I'm not really happy with the name of the second struct so if we go with this I'd appreciate name suggestions if you have any

Does this seem reasonable? I'm still a bit hesitant to move `DisableDelta` to be with the incremental backup stuff but don't really have strong feelings on the others

ashmrtn: ah, I see we commented at roughly the same time

> I've been wondering if we'll just end up calling the next provider's incremental changelog Deltas for ease of communication.

We may. I think the big unknown here is if other services will have a way to disable similar features and if it will have the same sort of behavior that M365 does. For us right now disabling deltas in M365 is more a workaround (due to mailbox being full) than an expectation or a code path that is a full "feature" for how backups should run

ashmrtn:
alright I've updated the PR to better match what was discussed offline. @ryanfkeepers could you please give this another look to ensure I properly captured things?

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+	MaxPages             int
+	Enabled              bool
+}
+
+// IncrementalsConfig contains options specific to incremental backups and
+// affects what data will be fetched from the external service being backed up.
+type IncrementalsConfig struct {
+	// ForceFullEnumeration prevents the use of a previous backup as the starting
+	// point for the current backup. All data in the external service will be
+	// discovered whether or not it's changed.
```
ryanfkeepers, line 96: Nit: for clarity.
```suggestion
	// point for the current backup. All data in the external service will be
	// enumerated whether or not it's changed.  Per-item storage will only
	//  get updated if changes have occurred.
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+	ForceFullEnumeration bool `json:"forceFullEnumeration,omitempty"`
+
+	// ForceItemDataRefresh causes data for all discovered items to be downloaded
+	// from the external service instead of using unchanged data from previous
+	// failed or successful backups where possible. Data dedupe will still occur
+	// if the redownloaded data matches data previously backed up by corso.
```
ryanfkeepers, line 102: Nit: for clarity.
```suggestion
	// ForceItemDataRefresh causes the data for all enumerated items to replace
	// stored data, even if no changes have been detected. Storage-side data
	// deduplication still applies, but that's after item download, and items are
	// always downloaded when this flag is toggled.
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+
+// BackupM365Config contains config options that are specific to backing up data
+// from M365 or Corso features that are only available during M365 backups.
+type BackupM365Config struct {
+	// DeltaPageSize controls the quantity of items fetched in each page during
+	// multi-page queries, such as graph api delta endpoints.
+	DeltaPageSize int32 `json:"deltaPageSize"`
+
+	// DisableDelta prevents backups from using delta based lookups,
+	// forcing a backup by enumerating all items. This is different
```
ryanfkeepers, line 52:
```suggestion
	// DisableDelta prevents backups from calling /delta endpoints,
	// and will force a full enumeration of all items. This is different
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+	// from IncrementalsConfig.ForceFullEnumeration in that this does not even
+	// make use of delta endpoints if a delta token is available. This is
+	// necessary when the user has filled up the mailbox storage available to the
+	// user as Microsoft prevents the API from being able to make calls
+	// to delta endpoints.
```
ryanfkeepers, line 57:
```suggestion
	// necessary when the user has filled up their mailbox storage as Microsoft
	// prevents the API from being able to make calls to /delta endpoints when a
	// mailbox is over storage limits.
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+// of options for multiple backup operations pass the struct to all operations.
+type BackupConfig struct {
+	FailureHandling      FailurePolicy                      `json:"failureHandling"`
+	ItemExtensionFactory []extensions.CreateItemExtensioner `json:"-"`
+	Parallelism          Parallelism                        `json:"parallelism"`
+	ServiceRateLimiter   RateLimiter                        `json:"serviceRateLimiter"`
+	Incrementals         IncrementalsConfig                 `json:"incrementalsConfig"`
+	M365                 BackupM365Config                   `json:"m365Config"`
+
+	// PreviewItemLimits defines the number of items and/or amount of data to
```
ryanfkeepers, line 32:
```suggestion
	// PreviewLimits defines the number of items and/or amount of data to
```

src/pkg/control/backup.go
```diff
@@ -0,0 +1,107 @@
...
+			CollectionBuffer: 4,
+			ItemFetch:        4,
+		},
+		M365: BackupM365Config{
+			DeltaPageSize: 500,
+		},
+	}
+}
+
+// Backup is the set of options used for backup operations. Each set of options
```
ryanfkeepers, line 21:
```suggestion
// BackupConfig is the set of options used for backup operations. Each set of options
```

ryanfkeepers (approved):

src/pkg/control/backup.go
ashmrtn: hah if only someone wrote (and used) a linter for something like this 😉

PR #4863, opened 2023-12-15
patch byte limit handling

ryanfkeepers:
in the tree version of backups, allow one item to hit or slightly overstep the total byte limits.  Also, stop all enumeration afterward if the backup has hit the total item limits.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


ashmrtn (approved):
Thanks for updating this!

PR #4865, opened 2023-12-15
deltaURL stub refactor

ryanfkeepers:
clean up the handling of deltaURL by using its own func instead of the id() func.  Also minimizes the delta stub constructor by using the suffix pattern used by id and folder constructors.

no logic changes, just a quick refactor.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/helper_test.go
```diff
@@ -1053,11 +1053,11 @@ type deltaQuery struct {
 }
 
 func delta(
```
ashmrtn, line 1055: thoughts on also updating `deltaWReset` to have the same set of parameters?

ashmrtn (approved):

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: oohh, good idea.

PR #4866, opened 2023-12-16
preparing for test expansion

ryanfkeepers:
some additions to test helpers to prepare for expanding unit tests

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/drive/helper_test.go
```diff
@@ -384,6 +395,14 @@ func (ecs expectedCollections) compareColl(t *testing.T, coll data.BackupCollect
 		p       = fullOrPrevPath(t, coll)
 	)
 
+	// check the metadata collection separately
+	if coll.FullPath() != nil && coll.FullPath().Equal(defaultMetadataPath(t)) {
```
ashmrtn, line 399: nit: could this be simplified to just `p` since that's going to be either the full path or the previous path?
```suggestion
	if p.Equal(defaultMetadataPath(t)) {
```

src/internal/m365/collection/drive/helper_test.go
```diff
@@ -996,6 +1022,78 @@ func (dd *deltaDrive) newEnumer() *DeltaDriveEnumerator {
...
+func (dd *deltaDrive) newPrevPaths(
+	idPathPairs ...string,
+) *drivePrevPaths {
+	dpp := drivePrevPaths{
+		id:                 dd.id,
+		folderIDToPrevPath: map[string]string{},
+	}
+
+	if len(idPathPairs)%2 == 1 {
+		dpp.folderIDToPrevPath["error"] = "idPathPairs had odd count of elements"
```
ashmrtn, line 1039: since this is a new function I'd prefer if it took `*testing.T` and used `require.Zero` or something like that. This seems like it could be error prone because not everyone may know they need to check for a specific key in the map

src/internal/m365/collection/drive/helper_test.go
```diff
@@ -996,6 +1022,78 @@ func (dd *deltaDrive) newEnumer() *DeltaDriveEnumerator {
...
+	for i := 0; i < len(idPathPairs); i += 2 {
+		dpp.folderIDToPrevPath[idPathPairs[i]] = idPathPairs[i+1]
+	}
+
+	return &dpp
+}
+
+// transforms 0 or more drivePrevPaths to a map[driveID]map[folderID]prevPathString
+func multiDrivePrevPaths(drivePrevs ...*drivePrevPaths) map[string]map[string]string {
+	msmss := map[string]map[string]string{}
```
ashmrtn, line 1052: nit: could you please use a more descriptive name? Or at least one where the reader isn't tempted to try to figure out what the random letters mean (I'm guess mapStringMapStringString?)? In my opinion even `res` would be more useful here because at least that has meaning. `msmss` doesn't have clear meaning just from looking at it

src/internal/m365/collection/drive/helper_test.go
```diff
@@ -996,6 +1022,78 @@ func (dd *deltaDrive) newEnumer() *DeltaDriveEnumerator {
...
+}
+
+// transforms 0 or more drivePrevPaths to a []data.RestoreCollection containing
+// a metadata collection.
+// DeltaURLs are currently always populated with {driveID: deltaURL()}.
+func multiDriveMetadata(
+	t *testing.T,
+	drivePrevs ...*drivePrevPaths,
+) []data.RestoreCollection {
+	drc := []data.RestoreCollection{}
```
ashmrtn, line 1068: nit: could you please use a more descriptive name for this and `dp` below too too? This function is starting to get complex enough that meaningful names would be useful to the reader. Something like `prevMetadataCols` and `prevPaths` could work

ashmrtn (approved):
Mostly nits around naming

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: seems reasonable

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: Heh, okay.

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: Can I name it `missMisses`?

This one gets updated in the next PR.

src/internal/m365/collection/drive/helper_test.go
ryanfkeepers: can do

PR #4870, opened 2023-12-16
ignore lists with webTemplateExtensionsList template while creation

HiteshRepo:
ignore lists with `webTemplateExtensionsList` template while creation in order to avoid `invalid template` graph error

#### Does this PR need a docs update or release note?
- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

PR #4871, opened 2023-12-17
filter out lists with template as 'webTemplateExtensionsList' 

HiteshRepo:
filter out lists with template as 'webTemplateExtensionsList' from details view

backup details without list type of `webTemplateExtensionsList` even though those lists are backed-up
```
./backup-corso backup details sharepoint --backup 7775976f-b0e1-4254-a507-1f4a5ba0bc57 
Logging to file: /home/hitesh/.cache/corso/logs/2023-12-19T04-54-54Z.log
Connecting to M365                               done 
Connecting to repository                     32s done 
  ID            List             Items  Created               Modified            
  6ec0b2fcd4e1  Events           0      2023-06-10T16:11:19Z  2023-06-10T16:11:19Z
  b5506781cf9d  Access Requests  1      2023-12-18T06:33:04Z  2023-12-18T06:33:07Z
  508314222e0a  test-list        1      2023-12-18T13:56:06Z  2023-12-18T13:57:06Z
  cf63e8c7d30f  DocuLIb          5      2023-08-25T10:46:09Z  2023-12-18T05:39:27Z
  f4c1f2b75a17  Documents        11     2023-06-10T16:11:17Z  2023-12-18T05:37:13Z
```

'Web Template Extensions' is the list that is intended to be hidden by default from CLI user and will not be restored.

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/flags/sharepoint.go
+++ b/src/cli/flags/sharepoint.go
@@ -14,25 +14,23 @@ const (
 )
 
 const (
-	LibraryFN      = "library"
-	ListFolderFN   = "list"
-	ListItemFN     = "list-item"
-	ListTemplateFN = "template"
-	PageFolderFN   = "page-folder"
-	PageFN         = "page"
-	SiteFN         = "site"    // site only accepts WebURL values
-	SiteIDFN       = "site-id" // site-id accepts actual site ids
+	LibraryFN    = "library"
+	ListFolderFN = "list"
+	ListItemFN   = "list-item"
+	PageFolderFN = "page-folder"
+	PageFN       = "page"
+	SiteFN       = "site"    // site only accepts WebURL values
+	SiteIDFN     = "site-id" // site-id accepts actual site ids
 )
 
 var (
-	LibraryFV      string
-	ListFolderFV   []string
-	ListItemFV     []string
-	ListTemplateFV []string
-	PageFolderFV   []string
-	PageFV         []string
-	SiteIDFV       []string
-	WebURLFV       []string
+	LibraryFV    string
+	ListFolderFV []string
+	ListItemFV   []string
+	PageFolderFV []string
+	PageFV       []string
+	SiteIDFV     []string
+	WebURLFV     []string
 )
 
 // AddSharePointDetailsAndRestoreFlags adds flags that are common to both the
@@ -83,11 +81,6 @@ func AddSharePointDetailsAndRestoreFlags(cmd *cobra.Command) {
 		ListItemFN, nil,
 		"Select lists by item name; accepts '"+Wildcard+"' to select all lists.")
 	cobra.CheckErr(fs.MarkHidden(ListItemFN))
-	fs.StringSliceVar(
-		&ListTemplateFV,
-		ListTemplateFN, []string{InvalidListTemplate},
-		"Select lists by template; accepts '"+Wildcard+"' to select all lists.")
-	cobra.CheckErr(fs.MarkHidden(ListTemplateFN))
 
 	// pages
 
--- a/src/cli/flags/sharepoint.go
+++ b/src/cli/flags/sharepoint.go
@@ -9,6 +9,7 @@ const (
 	DataPages     = "pages"
 	DataLists     = "lists"
 
+	// [TODO]: move to api package
 	InvalidListTemplate = "webTemplateExtensionsList"
 )
 
--- a/src/cli/flags/sharepoint.go
+++ b/src/cli/flags/sharepoint.go
@@ -8,26 +8,30 @@ const (
 	DataLibraries = "libraries"
 	DataPages     = "pages"
 	DataLists     = "lists"
+
+	InvalidListTemplate = "webTemplateExtensionsList"
 )
 
 const (
-	LibraryFN    = "library"
-	ListFolderFN = "list"
-	ListItemFN   = "list-item"
-	PageFolderFN = "page-folder"
-	PageFN       = "page"
-	SiteFN       = "site"    // site only accepts WebURL values
-	SiteIDFN     = "site-id" // site-id accepts actual site ids
+	LibraryFN      = "library"
+	ListFolderFN   = "list"
+	ListItemFN     = "list-item"
+	ListTemplateFN = "template"
+	PageFolderFN   = "page-folder"
+	PageFN         = "page"
+	SiteFN         = "site"    // site only accepts WebURL values
+	SiteIDFN       = "site-id" // site-id accepts actual site ids
 )
 
 var (
-	LibraryFV    string
-	ListFolderFV []string
-	ListItemFV   []string
-	PageFolderFV []string
-	PageFV       []string
-	SiteIDFV     []string
-	WebURLFV     []string
+	LibraryFV      string
+	ListFolderFV   []string
+	ListItemFV     []string
+	ListTemplateFV []string
+	PageFolderFV   []string
+	PageFV         []string
+	SiteIDFV       []string
+	WebURLFV       []string
 )
 
 // AddSharePointDetailsAndRestoreFlags adds flags that are common to both the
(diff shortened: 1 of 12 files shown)


src/cli/utils/sharepoint.go
```diff
@@ -105,6 +108,20 @@ func ValidateSharePointRestoreFlags(backupID string, opts SharePointOpts) error
...
 	}
 
+	if _, ok := opts.Populated[flags.ListTemplateFN]; ok {
+		found := false
+
+		for _, wu := range opts.ListTemplate {
+			if wu == flags.InvalidListTemplate {
+				found = true
+			}
+		}
```
ryanfkeepers, line 118: you can use the slices package here.  `slices.Contains(opts.ListTemplate, flags.InvalidListTemplate)`

src/cli/flags/sharepoint.go
```diff
@@ -8,26 +8,30 @@ const (
 	DataLibraries = "libraries"
 	DataPages     = "pages"
 	DataLists     = "lists"
+
+	InvalidListTemplate = "webTemplateExtensionsList"
```
ryanfkeepers, line 12: This is the wrong ownership for a variable like this.  It should be owned by the api package.

src/cli/utils/sharepoint.go
```diff
@@ -105,6 +108,20 @@ func ValidateSharePointRestoreFlags(backupID string, opts SharePointOpts) error
...
+
+		if !found {
+			return clues.New(fmt.Sprintf("%s list template not present", flags.InvalidListTemplate))
```
ryanfkeepers, line 121: per code standards, errors should not include format statements.

aside from that, I'm not sure what the purpose of this error is: to require the flag parameters to contain a certain value?  That's a poor UX for end users, since it's not normal to require values in a flag.  I would be confused if I told the CLI to use a certain (valid) flag value and was given an error in response due to my command missing some _other_ valid flag value.  It's too much opportunity for the user to give themselves a bad experience.

ryanfkeepers (changes requested):
I have concerns about the approach to filtering in this PR at large.  The approach isn't _wrong_.  That is to say, it isn't buggy or erroneous.  But I don't think it provides the correct UX, either.

To reiterate the problem that we're trying to solve here: the backup includes a type of list that cannot be restored.  Attempting to restore this type of list will result in errors, which may confuse the end user since we are showing them data from a backup, and the base assumption is that all backup data is either exportable or restorable, without any additional knoweldge required to make it work.  End user confusion is the problem we want to solve.

This problem can be solved in one of three ways:
1. don't include this data in the backup in the first place.
2. add this data to the backup, but don't show it in a `backup details` unless the user adds the `--skip-reduce` flag.  Also, never include it in a `restore`.
3. add this data to the backup, and show it to the user, but give them additional CLI flags that let them filter by List template.

1/ isn't something we've talked about, but it could be a viable option.  Is there anything critical about this data that we need to keep?  Is there any point to backing it up, ever?  Can we even export it?

3/ is what you're doing in this PR.  There's some minor differences (defaulting to hiding the template), but the choice is categorically the same.  This approach is not my preference for a couple reasons.  In general it increases the work the system performs and increases complexity upon users, instead of simplifying and reducing things (which should always be our goal).  I say this because the change is adding more tooling to solve the problem: more cli flags, more selectors, more rules, more data shown to the user, more data to backup, and more items getting restored (even though we know they can't be restored anyway).  Lots of "and more" here.

Compare this approach to 1/.  If we filter out this data from all backups, everything automatically works with no need for additional tooling.  That could be the most simple and least complex option.  If we're willing to slice data out of the backup (need to do our homework here to be sure).

In the middle is 2/.  We already [have a pattern](https://github.com/alcionai/corso/blob/main/src/cli/backup/backup.go#L397) around the `skip-reduce` flag and showing folders in the backup details display, because folder entries are _never_ shown in backup details, and thus are _never_ included in the restore details set, even though we store that data.  This acts as the next most simple option, with minimal need for additional tooling or knoweldge.  It ensures that, even though we back up this data, we don't let any systems interact with it.  The point of control automatically solves the need for additional cli flags, selectors, or etc.  

Hopefully, in comparison, you can see where this approach is creating more mental burden where we could, instead, be reducing it.  By going with option 1/ or 2/ we solve the problem for CLI users, SDK users, and corso developers and internal systems.  By choosing 3/ we have to hoist the knowledge and tooling to solve this issue up to the highest layers and end users.

I'm going to push for you to pick 1/ or 2/ instead of this approach.  It would be nice, regardless of your choice, to have a document describing what's critical/valuable/intersting about backing up webTemplateExentsions.  But I'll leave that up to you.

src/cli/flags/sharepoint.go
HiteshRepo: yes @ryanfkeepers 
In this [PR](https://github.com/alcionai/corso/pull/4870), I have put it in api package and intended to remove it from here post merges.

HiteshRepo:
> see where this approach is creating more mental burden where we could, instead, be redu

hi @ryanfkeepers I believe I have gone with the apporach '2' here. 
1. Adding it to the backup
2. Showing in details to user only when --skip-reduce is used: https://github.com/alcionai/corso/blob/e16dc4093b119da44a4dd5f16b02f6c8f1255afb/src/pkg/selectors/sharepoint.go#L244
3. [Ignoring the list when restore](https://github.com/alcionai/corso/pull/4870)

src/cli/utils/sharepoint.go
HiteshRepo: Instead of returning an error for not finding the 'InvalidListTemplate', maybe we can add it to the opts.ListTemplate.

ryanfkeepers:
>  I believe I have gone with the apporach '2' here

@HiteshRepo Ah, my apologies.  I wrote a lot, but didn't explain much, or very well.  Check out this code: https://github.com/alcionai/corso/blob/40d1d4fb2171a5a8fe83a7a6d3fb62c33e5d4a48/src/pkg/backup/details/model.go#L83

What's happening there is that we filter out any unusable details entries (currently just the folder entries) even before they get put through the selectors reducer.  This is, I think, the appropriate place to folder out entries such as the templates without needing to introduce additional flags or selectors.

Some care has to be taken with the implementation.  That linked code isn't aware of service-specific behavior, and we don't want to inject a conditional like "if entry is sharepoint and a list and has template of type...".  That's too much external domain knowledge for good maintenance.  We can, however, get a similar result by determining these parameters at backup creation time.

How about the following: add a new property to the `Entry` struct in `details/entry.go`.  A boolean var named something like: `NotRecoverable` or `DoNotUseForRecovery`.  When running a backup for Lists ensure that this boolean gets set properly for all generated item info.  IE: true for WebTemplateExtensions, false for all other templates.  Finally, in the linked code above, make sure we also filter out any entries where `DoNotUseForRecovery == true`.

This is more along the lines of what I was thinking for 2/.  It gives us the same filtering as the current PR without needing to add flags or selectors, because the filtering happens as part of details enumeration, _before_ flags or selectors get applied.  Even more important than being able to avoid the flags/selectors is that this ensures the entries read from details are always in a good state for display and recovery, regardless of who runs the query, and regardless of whether that caller follows up with a call to selectors.Reduce().  It makes the system more simple and clean overall by ensuring we only produce usable data, independent of the context.

Hopefully that explanation is more clear.  Apologies again; I didn't give you enough guidance with the first comment.  What's more, I myself had a hard time finding that linked code (had to get another person to help me find it again).  I'll try to give more complete suggestions in the future.

src/internal/m365/collection/site/collection.go
```diff
@@ -355,8 +356,17 @@ func (sc *Collection) handleListItems(
 	metrics.Bytes += size
 	metrics.Successes++
 
+	template := ""
+	if list != nil &&
+		list.GetList() != nil {
```
ryanfkeepers, line 361: nit: I think this is short enough to inline.
```suggestion
	if list != nil && list.GetList() != nil {
```

ryanfkeepers (approved):
Looking good, thank you!

Issue #3891, opened 2023-07-24
`vcf` export form Exchange contacts backup


meain:

meain:
Closing this as we are currently only planning for `pst` exports for exchange.

PR #4883, opened 2023-12-19
Contacts export for exchange

meain:
---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* closes https://github.com/alcionai/corso/issues/3891

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/vcf/vcf.go
ashmrtn, line 1: do we have an idea of what the range of valid inputs are? For example, is it possible that we'll get malformed emails, phone numbers, or addresses from the source data that could cause problems for the vcard code?

ashmrtn (commented):
generally looks good. Would like to see some info discussion around possible invalid inputs (if any) and some clarity/testing around the organization field

src/internal/converters/vcf/vcf.go
```diff
@@ -0,0 +1,206 @@
...
+		if len(orgFull) > 0 {
+			orgFull += ";"
+		}
+
+		orgFull += *profession
+	}
+
+	if len(orgFull) > 0 {
+		vc.Set(vcard.FieldOrganization, &vcard.Field{Value: orgFull})
+	}
```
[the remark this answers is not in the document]
meain: We will set the other fields even if the org name is not available. The `if` condition pretty much decides if we prefix a `;`. That said, now that I think about I think we should have an empty column to symbolize that the org name is empty.

The [spec](https://datatracker.ietf.org/doc/html/rfc6350) does not go into any detail around what we should do if the name is not available.

Also, let me add the necessary tests.

src/internal/converters/vcf/vcf.go
meain: Unlike in the case of emails, the vcard lib that we currently does not use any validation on the strings that pass to it and so whatever is the data that we get from graph goes into the output. Most values are free-form text. `TEL` is "free-form text value" in spec and `EMAIL` is "free-form UTF-8 text" and so I don't think it matter as far as spec is concerned.

ashmrtn (approved):

PR #4886, opened 2023-12-19
remove the rootID in the tree

ryanfkeepers:
Instead of comparing IDs, we can check the drive item's GetRoot() property, which is only non-nil for the root folder.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #4689

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/delta_tree.go
+++ b/src/internal/m365/collection/drive/delta_tree.go
@@ -87,17 +87,25 @@ type nodeyMcNodeFace struct {
 	children map[string]*nodeyMcNodeFace
 	// file item ID -> file metadata
 	files map[string]*custom.DriveItem
+	// when true, this flag means the folder appeared in enumeration,
+	// but was not included in the backup selection.  We include
+	// unselected folders in the tree so we don't have to hold stateful
+	// decisions (such as folder selection) as part of delta processing;
+	// we only need to evaluate them during post-processing.
+	isNotSelected bool
 }
 
 func newNodeyMcNodeFace(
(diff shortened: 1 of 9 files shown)


src/internal/m365/collection/drive/delta_tree.go
```diff
@@ -336,32 +335,33 @@ func (face *folderyMcFolderFace) addFile(
...
 
 	// in case of file movement, clean up any references
 	// to the file in the old parent
 	oldParentID, ok := face.fileIDToParentID[id]
 	if ok && oldParentID != parentID {
-		if nodey, ok := face.folderIDToNode[oldParentID]; ok {
+		if nodey := face.getNode(oldParentID); nodey != nil {
 			delete(nodey.files, id)
 		}
-
```
pandeyabs, line 356: Is this code removal intentional? Also this change doesn't look related to `rootID` removal, please correct me if wrong.

Going forward, can we please move unrelated changes to a different PR? :)

pandeyabs (approved):
LGTM, left a comment to sanity check on an unrelated code removal.

src/internal/m365/collection/drive/delta_tree.go
ryanfkeepers: Oh, it's all related at this point.  And, yes, this is intentional.

pandeyabs: Thanks for clarifying

Issue #4822, opened 2023-12-08
Add group mailbox exports


pandeyabs:
### What happened?

We have most of group mailbox backup functionality in main. Next step is to enable exports.

### Corso Version?

Corso v0.16.0

### Where are you running Corso?

NA

pandeyabs:
Basic support is ready(not merged yet). However, we need to make sure the export format is agreeable and includes all the details we'd like to have. I'll start an internal thread on this.

PR #4890, opened 2023-12-20
Exclude conversation selectors from export and restore

pandeyabs:
Group export sanity test is failing because we added conversations to `sel.AllData()`. But we don't have export support for conversations yet.

Adding conv selector to exclude so that it doesn't fail exports for groups libraries and channels. This is a short term change, and will be removed once we have export support https://github.com/alcionai/corso/issues/4822

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/restore/groups.go
+++ b/src/cli/restore/groups.go
@@ -7,6 +7,7 @@ import (
 	"github.com/alcionai/corso/src/cli/flags"
 	"github.com/alcionai/corso/src/cli/utils"
 	"github.com/alcionai/corso/src/pkg/dttm"
+	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
 // called by restore.go to map subcommands to provider-specific handling.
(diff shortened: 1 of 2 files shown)


src/cli/restore/groups.go
```diff
@@ -91,6 +92,10 @@ func restoreGroupsCmd(cmd *cobra.Command, args []string) error {
 	sel := utils.IncludeGroupsRestoreDataSelectors(ctx, opts)
```
pandeyabs, line 92: Quick note: I didn't add the exclude change inside `IncludeGroupsRestoreDataSelectors` because details cmd also uses it. I don't see a big impact here, since this is a short term change.

HiteshRepo (approved):

HiteshRepo (approved):

Issue #4685, opened 2023-11-15
Corso needs standardized errors


ryanfkeepers:
# Problem

The [graph package's errors](https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/api/graph/errors.go) constants, sentinels, and identification processes have (as a result of organic growth) leaked out of the api layer and appeared in many of the higher layers.  This was caused by a history of equating "graph" with "all http communication controls", and treating graph errors as http api errors at broad.

There are two core design issues with this state of the repo:
1. Graph error production can be ambiguous and non-idiomatic.  Graph package doesn't always produce a sentinel as part of error handling, and this causes all upstream consumers to need to abide that package's unique flavor of error comparison instead of using a more standardized set of sentinels to compare with `errors.Is`.
2. We don't have a clear separation between the api and upper layers.  There exists no set of standardized errors for graph package to wrap/stack/transform its errors into.

# Acceptance Criteria
1. Introduce a new package to own the generic error sentinels that lower layer packages can use for wrapping/stacking.  Maybe `/pkg/common/sentinels`?  /pkg is preferred over /internal since these sentinels may be useful for SDK consumers.
2. Ensure graph package error handling appropriately wraps/transforms its api-layer errors into common sentinels.

PR #4897, opened 2023-12-20
introduce errs/core

ryanfkeepers:
first step towards having a centralized set of error sentinels that can be passed around corso instead of re-using low level error sentinels like those found in the graph/errors.go file.

This PR works as a standalone, and only handles the lowest hanging fruit. SDK consumers will need to change their error enum references, but all api behavior should remain the same.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4685

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/errs/errs.go
+++ b/src/pkg/errs/errs.go
@@ -3,35 +3,18 @@ package errs
 import (
 	"errors"
 
+	"github.com/alcionai/corso/src/pkg/errs/core"
 	"github.com/alcionai/corso/src/pkg/repository"
 	"github.com/alcionai/corso/src/pkg/services/m365/api/graph"
-)
-
-// expose enums, rather than errors, for Is checks.  The enum should
-// map to a specific internal error that can be used for the actual
-// errors.Is comparison.
-type errEnum string
-
-const (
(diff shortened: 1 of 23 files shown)


src/pkg/errs/errs.go
```diff
@@ -3,35 +3,18 @@ package errs
...
-	BackupNotFound            errEnum = "backup-not-found"
-	InsufficientAuthorization errEnum = "insufficient-authorization"
-	RepoAlreadyExists         errEnum = "repository-already-exists"
-	ResourceNotAccessible     errEnum = "resource-not-accesible"
-	ResourceOwnerNotFound     errEnum = "resource-owner-not-found"
-	ServiceNotEnabled         errEnum = "service-not-enabled"
-)
-
 // map of enums to errors.  We might want to re-use an enum for multiple
 // internal errors (ex: "ServiceNotEnabled" may exist in both graph and
```
meain, line 12: Now that we have moved around, we might wanna update the example mentioning `ServiceNotEnabled`.

meain (approved):

PR #4898, opened 2023-12-20
Beef up Exchange export error handling

ashmrtn:
Add log statements when errors are encountered so we get full clues
output and make sure context clues are added to returned errors. The
additional logging is necessary because not all corso SDK consumers
will use clues which means they could miss out on valuable information
if they just log the returned errors normally

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/export.go
+++ b/src/internal/m365/collection/exchange/export.go
@@ -12,6 +12,7 @@ import (
 	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/export"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/metrics"
 	"github.com/alcionai/corso/src/pkg/path"
 )
@@ -59,9 +60,13 @@ func streamItems(
 			reader.Close()
 
 			if err != nil {
+				err = clues.Wrap(err, "reading export item")
+
+				logger.CtxErr(ctx, err).Info("processing collection item")
+
 				ch <- export.Item{
 					ID:    id,
-					Error: clues.Wrap(err, "reading data"),
+					Error: err,
 				}
 
 				continue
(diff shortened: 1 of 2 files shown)


ryanfkeepers (approved):

src/internal/m365/collection/exchange/export.go
```diff
@@ -59,19 +60,27 @@ func streamItems(
...
 
 			if err != nil {
+				err = clues.Wrap(err, "reading export item")
+
+				logger.CtxErr(ctx, err).Info("processing collection item")
+
 				ch <- export.Item{
 					ID:    id,
-					Error: clues.Wrap(err, "reading data"),
+					Error: err,
```
vkamra, line 69: Will clues include the itemID?

ashmrtn: let me tweak the log statement a bit. Right now it will include the shortref because it's added to the context up on line 48/49

vkamra: Thanks. Line 48/49 is already outside the inner loop so I'm assuming that's the collection/directory shortref and not the item?

ashmrtn: arg yes, hang on, try the new version with updated clues in the context and some extra `WrapWC` calls

src/internal/m365/collection/exchange/export.go
```diff
@@ -51,6 +52,8 @@ func streamItems(
 			id := item.ID()
 			name := id + ".eml"
 
+			itemCtx := clues.Add(ictx, "item_file_name", id)
```
ryanfkeepers, line 55: nit: "file name" is an interpretation of how the value will be used, not actually what it is.  that may drift over time/future implementation.
```suggestion
			itemCtx := clues.Add(ictx, "stream_item_id", id)
```

PR #4899, opened 2023-12-20
swap pkg/metrics to threadsafe map

ryanfkeepers:
---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/metrics/metrics.go
+++ b/src/pkg/metrics/metrics.go
@@ -23,6 +23,12 @@ type ExportStats struct {
 	data syncd.MapOf[path.CategoryType, *count.Bus]
 }
 
+func NewExportStats() *ExportStats {
+	return &ExportStats{
+		data: syncd.NewMapOf[path.CategoryType, *count.Bus](),
+	}
+}
+
 func (es *ExportStats) UpdateBytes(kind path.CategoryType, numBytes int64) {
 	es.getCB(kind).Add(bytesRead, numBytes)
 }
--- a/src/pkg/metrics/metrics.go
+++ b/src/pkg/metrics/metrics.go
@@ -2,8 +2,9 @@ package metrics
 
 import (
 	"io"
-	"sync/atomic"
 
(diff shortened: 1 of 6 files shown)


src/pkg/metrics/metrics.go
```diff
@@ -12,33 +13,47 @@ type KindStats struct {
...
+	toKindStats := map[path.CategoryType]KindStats{}
+
+	for k, cb := range es.data.Values() {
+		toKindStats[k] = KindStats{
+			BytesRead:     cb.Get(bytesRead),
+			ResourceCount: cb.Get(resources),
+		}
+	}
+
+	return toKindStats
```
meain, line 56: The current downstream implementations calls `GetStats` and pass around the map which would get updated only to pull the values out at this end. We should coordinate to ensure that we pass around `ExportStats` instead of the value got from `GetStats`.

ryanfkeepers: I don't see the map getting re-used.  `exportStats.GetStats` is called here: https://github.com/alcionai/corso/blob/main/src/internal/operations/export.go#L327.  That func is only called here in the CLI: https://github.com/alcionai/corso/blob/main/src/cli/export/export.go#L134, which prints, but doesn't update, the map.  Am I missing its usage somewhere?

meain: I meant the SDK consumers, not code within corso.

ryanfkeepers: I think that'll have to be fixed on the SDK side.  Those users can store the ExportStats as a `GetStatser` interface, and call `GetStats()` at the end of the operation.  That'll provide an equivalent usage.

meain (approved):

src/pkg/metrics/metrics.go
meain: @zackrossman tagging you just so that you are aware of an upcoming change.

PR #4901, opened 2023-12-21
Log and ignore sharing disabled errors when doing permissions restores

ashmrtn:
We've encountered an error where the user previously had sharing for
one or more external users on an item and then did a backup. They
subsequently disabled sharing for the SharePoint site in either the
tenant-wide config or for the site in particular. Once sharing was
disabled, they attempted a restore which subsequently failed while
trying to restore sharing for the external user

This PR logs and ignores errors about sharing being disabled when
doing a restore

This has been tested manually

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -475,6 +475,11 @@ func RestorePermissions(
 
 	previous, err := computePreviousMetadata(ctx, itemPath, caches.ParentDirToMeta)
 	if err != nil {
+		if graph.IsErrSharingDisabled(err) {
+			logger.Ctx(ctx).Info("sharing disabled, not restoring permissions")
+			return nil
+		}
+
 		return clues.Wrap(err, "previous metadata")
 	}
 
@@ -500,6 +505,11 @@ func RestorePermissions(
 		caches.OldPermIDToNewID,
 		errs)
 	if err != nil {
+		if graph.IsErrSharingDisabled(err) {
+			logger.Ctx(ctx).Info("sharing disabled, not restoring permissions")
+			return nil
+		}
+
 		return clues.Wrap(err, "updating permissions")
 	}
 
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -389,13 +389,13 @@ func UpdateLinkShares(
 			continue
 		}
 
-		if err != nil {
-			if graph.IsErrSharingDisabled(err) {
-				logger.Ctx(ctx).Info("unable to restore link share; sharing disabled")
-			} else {
-				el.AddRecoverable(ctx, clues.Stack(err))
-			}
+		if graph.IsErrSharingDisabled(err) {
+			logger.CtxErr(ictx, err).Info("unable to restore link share; sharing disabled")
+			continue
+		}
 
+		if err != nil {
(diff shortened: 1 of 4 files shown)


src/internal/m365/collection/drive/permission.go
```diff
@@ -390,7 +390,12 @@ func UpdateLinkShares(
 		}
 
 		if err != nil {
-			el.AddRecoverable(ctx, clues.Stack(err))
+			if graph.IsErrSharingDisabled(err) {
+				logger.Ctx(ctx).Info("unable to restore link share; sharing disabled")
```
pandeyabs, line 394: Should we add an alert as well, like [here](https://github.com/alcionai/corso/blob/e46119a81878304b2167bd36da549186e7186f5e/src/pkg/services/m365/api/mail.go#L441)? Also at line 510

src/internal/m365/collection/drive/permission.go
```diff
@@ -390,7 +390,12 @@ func UpdateLinkShares(
 		}
 
 		if err != nil {
-			el.AddRecoverable(ctx, clues.Stack(err))
+			if graph.IsErrSharingDisabled(err) {
```
pandeyabs, line 393: nit: These could be added as an `else ifs` at line 390.

pandeyabs (approved):
LGTM. Left a comment about adding an informational alert

pandeyabs:
Let's also wait for @meain and @ryanfkeepers since they have more context.

src/internal/m365/collection/drive/permission.go
meain: IMO, I don't think alert makes sense in this case as the sharing was disabled by the user. Even if we do alert, I don't think we should do it for every single item.

src/internal/m365/collection/drive/permission.go
ryanfkeepers: imo, else-if chains are difficult to read.  However, they could be unnested, since `graph.IsErrFoo` will return falsy for any nil error.  Ex:

```go
if graph.IsErrSharingDisabled(err) {
  logit...
  continue
}

if err != nil {
  addRecoverable
  continue
}
```

Imo, that's the easiest format to read.

src/internal/m365/collection/drive/permission.go
ryanfkeepers: I agree with Abin.  Alerts should be reserved for cases where we think the system has unintended consequences/conditions.  Abiding user configurations shouldn't cascade to user visible attention.

ryanfkeepers, line 394: Nit: no harm in including the error, even at an info level.
```suggestion
				logger.CtxErr(ctx, err).Info("unable to restore link share; sharing disabled")
```

ryanfkeepers: Also, should this use `ictx`?  I see that in use on line 388.

ryanfkeepers:
@ashmrtn @pandeyabs looks good to me.  Some formatting nits, and a worry about `ctx` vs `ictx`.  But the behavior is sound overall.

ryanfkeepers (approved):

Issue #4862, opened 2023-12-15
Support kopia assists for group mailbox backup


pandeyabs:
### What happened?

Group mailboxes don't support delta tokens. So as things currently stand, we'd be doing full backups every time. Having kopia assist support in v0 will make sure these full backups don't redownload all the data. Creating a ticket to track lazy item reader support for group mailboxes. 

### Corso Version?

Corso v0.17.0

### Where are you running Corso?

NA

PR #4907, opened 2023-12-21
Add feature flag to disable lazy item reader

pandeyabs:
AddDisableLazyItemReader disables lazy item reader, such that we fall
back to prefetch reader. This flag is currently only meant for groups
conversations backup. Although it can be utilized for other services
in future.

This flag should only be used if lazy item reader is the default choice
and we want to fallback to prefetch reader.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/4862

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/cli/backup/groups.go
+++ b/src/cli/backup/groups.go
@@ -75,6 +75,7 @@ func addGroupsCommands(cmd *cobra.Command) *cobra.Command {
 		flags.AddFetchParallelismFlag(c)
 		flags.AddDisableDeltaFlag(c)
 		flags.AddGenericBackupFlags(c)
+		flags.AddDisableLazyItemReader(c)
 
 	case listCommand:
 		c, fs = utils.AddCommand(cmd, groupsListCmd(), utils.MarkPreviewCommand())
--- a/src/cli/backup/groups_test.go
+++ b/src/cli/backup/groups_test.go
@@ -153,6 +153,7 @@ func (suite *GroupsUnitSuite) TestBackupCreateFlags() {
 				"--" + flags.CategoryDataFN, flagsTD.FlgInputs(flagsTD.GroupsCategoryDataInput),
 				"--" + flags.FetchParallelismFN, flagsTD.FetchParallelism,
 				"--" + flags.DisableDeltaFN,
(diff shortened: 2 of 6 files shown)


ryanfkeepers (approved):
Approving to unblock, if we want to go this route.  But I think we should be re-using the `force-item-data-download` instead, which is intended for the same purpose (though the lack of comments explaining the purpose of the flag make it easy to miss).  There's some nuanced mismatch: forceItemDataDownload doesn't remove the use of the lazy reader directly.  But it does remove kopia assist references, meaning the lazy reader will never lazily skip item downloads.  IMO, this is a preferable pattern, because it simplifies the low level architecture by keeping the reader wrapper implementation static, as opposed to configurable.

pandeyabs:
> There's some nuanced mismatch: forceItemDataDownload doesn't remove the use of the lazy reader directly. But it does remove kopia assist references, meaning the lazy reader will never lazily skip item downloads.

This flag only drops the assist bases. Any items from merge bases will not be downloaded. See description [here](https://github.com/alcionai/corso/blob/e7e29962f4aaea098f97fe7ac07018cae66112fe/src/pkg/control/options.go#L79)

So the lazy reader will not be invoked by kopia for downloading items which are present in merge base.

ryanfkeepers:
@pandeyabs
Hmm, that description conflicts with [this one](https://github.com/alcionai/corso/blob/e7e29962f4aaea098f97fe7ac07018cae66112fe/src/pkg/control/backup.go#L105), which is the parallel toggle in the config options that we're moving towards as of Ash's recent/ongoing refactoring.  

@ashmrtn can you please help us clarify behavior expectations here?

pandeyabs:
I think the `--disable-lazy-item-reader` flag serves a different purpose. Just like we noticed for exchange, this flag can come in handy if we run into COR-172("deleted" in flight) kind of issues with conversations. We can use this flag to disable lazy reader and switch to prefetch approach in that case. 

I am going to proceed with this killswitch. We can rework it later if needed.

PR #4925, opened 2023-12-22
JSON sanitizer and tests

ashmrtn:
We've found that it's possible to get malformed JSON back from
the graph server and graph SDK. This results in errors while trying
to deserialize objects that are malformed JSON

This adds a simple JSON sanitizer and tests to make sure that even if
graph and the graph SDK happen to generate malformed JSON we can get
back to a state where we can deserialize it again. The tests also
print info when the received content differs from the original

This PR does not change any of the logic that corso uses during
backups or restores, it just adds sanitization code and tests

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


vkamra (approved):

src/internal/common/sanitize/json.go
```diff
@@ -0,0 +1,58 @@
...
+			buf.WriteString(fmt.Sprintf(`\u%04X`, c))
+
+		default:
+			buf.WriteByte(c)
+		}
+	}
+
+	// Return a copy just so we don't hold a reference to internal bytes.Buffer
+	// data.
+	return slices.Clone(buf.Bytes())
```
pandeyabs, line 57: Not sure why we need this clone. Could you please explain a bit?

src/internal/common/sanitize/json_test.go
```diff
@@ -0,0 +1,104 @@
...
+	expectValid assert.BoolAssertionFunc
+}
+
+func generateCharacterTests() []jsonTest {
+	var (
+		res []jsonTest
+
+		baseTestName = "Escape0x%02X"
+		baseTestData = `{"foo":"ba%sr"}`
+		expect       = `{"foo":"ba%s\u00%02Xr"}`
```
pandeyabs, line 36: Could you please add some comments around these? e.g. it would help a new reader to know we are expecting escape chars in `\u0000` format here.

src/internal/common/sanitize/json_test.go
```diff
@@ -0,0 +1,104 @@
...
+		)
+	}
+
+	return res
+}
+
+func (suite *SanitizeJSONUnitSuite) TestJSONString() {
+	table := []jsonTest{
+		{
+			name:        "AlreadyValid NoSpecialCharacters",
```
pandeyabs, line 63: Should we add a test with unicode chars in input?

pandeyabs: Cool, sounds good

PR #4936, opened 2023-12-24
fix tests referencing top level of sharepointInfo for listInfo

HiteshRepo:
fix tests referencing top level of SharepointInfo for list details instead of nested ListInfo.
also disables a test that requires list backup to be enabled in CLI level.

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)
https://github.com/alcionai/corso/actions/runs/7312375792/job/19922878417#step:6:666

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/site/collection_test.go
```diff
@@ -233,9 +237,15 @@ func (suite *SharePointCollectionSuite) TestCollection_Items() {
...
 
-			assert.NotNil(t, info)
-			assert.NotNil(t, info.SharePoint)
-			assert.Equal(t, test.itemName, info.SharePoint.ItemName)
+			require.NotNil(t, info)
+			require.NotNil(t, info.SharePoint)
+
+			if test.scope.Category().String() == selectors.SharePointList.String() {
+				require.NotNil(t, info.SharePoint.List)
+				assert.Equal(t, test.itemName, info.SharePoint.List.Name)
```
meain, line 245: Didn't know we were not using the original `ItemName`.

src/cli/backup/sharepoint_e2e_test.go
```diff
@@ -341,7 +342,9 @@ func runSharepointDetailsCmdTest(suite *PreparedBackupSharepointE2ESuite, catego
 	foundList := 0
 
 	for _, ent := range deets.Entries {
-		if ent.SharePoint != nil && ent.SharePoint.ItemName != "" {
+		if ent.SharePoint != nil &&
```
meain, line 345: Looks like we ended up shipping it for items that are not lists.

meain (changes requested):

src/internal/m365/collection/site/collection_test.go
HiteshRepo: @meain  for lists only populating 'name' of [ListInfo](https://github.com/alcionai/corso/blob/43ca43910fdb241a8ae5d917a6500bcea274cab4/src/pkg/services/m365/api/lists.go#L623) and not the itemName in SharepointInfo

HiteshRepo: yeah. Let me make the change to populate `itemName` name instead.

meain: No need to change it if this has already been accepted to be how we do it. I was just thinking out loud.

src/cli/backup/sharepoint_e2e_test.go
HiteshRepo: @meain  attempted to make it generic instead of renaming the function

meain: Hmm, not sure if we are on the same page here. The function was previously checking for details of SharePoint drive items and hence `ent.SharePoint.ItemName` check, but now that case is never checked as `ent.SharePoint.List` will be `nil` for drive items.

pandeyabs (approved):

src/cli/backup/sharepoint_e2e_test.go
HiteshRepo: @meain please take a look at the current change. Now it caters to both sharepoint list and drive items.

PR #4921, opened 2023-12-22
Integrate lazy item reader into conversations backup

pandeyabs:
---

#### Does this PR need a docs update or release note?

- [x] :white_check_mark: Yes, it's included

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* Closes https://github.com/alcionai/corso/issues/4862

#### Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

PR #4939, opened 2023-12-27
Skip group mailbox items that were deleted in flight

pandeyabs:
Discovered an issue with how we were handling items deleted in flight with the lazy item reader. This needs to go in before integration PR. https://github.com/alcionai/corso/pull/4921 

Here are the details.

**Orig:**
* For such items, we were using the exchange approach to persist an empty data file (with serialized version) and not persist it in details file.
* Exchange discovers the deleted items during next backup during delta query and excludes them.

**Problem:**
* Group mailbox doesn't support delta queries. Also, there is no graph API to discover/fetch deleted items.
* So if we use the exchange lazy reader approach like above, the ongoing backup will succeed. However, the next incremental backup fails during details merge with ` running backup: merging details: incomplete migration of backup details`. 
* Failure is [here](https://github.com/alcionai/corso/blob/c58cd9302fd0e9bd011e7e7ecfac3211d2a7a74b/src/internal/operations/backup.go#L837). It's because the deleted item reporefs are still present in the merge base. Since there is no delta query, there is no `GetDeleted()` to exclude the item.

**Mod:**
* Using the approach drive lazy reader code uses to mark an item as skipped, so that kopia doesn't error out during upload.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/4862

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/groups/collection.go
+++ b/src/internal/m365/collection/groups/collection.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"net/http"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


meain (approved):
The changes looks good to me. I am just wondering why we did it the old way earlier :thinking:. I'm guessing that we did not have the idea of skippable items when we initially wrote this.

pandeyabs:
> The changes looks good to me. I am just wondering why we did it the old way earlier 🤔. I'm guessing that we did not have the idea of skippable items when we initially wrote this.

Thanks Abin. I think the skippable item code came earlier, followed by the exchange lazy item reader. I'm planning to bring this up in our next meeting. I think skipping approach is probably better. Exchange approach pushes the responsibility on to next backup and delta query. There might be some context behind why we decided to do it this way. Will discuss.

PR #4943, opened 2023-12-28
lists prefetech

HiteshRepo:
refactors sharepoint lists collection by:
- renaming `Collection` interface as `preFetchCollection` interface
- making `finishPopulation` method as independent function 
- seralizing list using `serializeContent` function

- [x] :no_entry: No

#### Type of change

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/site/collection.go
+++ b/src/internal/m365/collection/site/collection.go
@@ -50,7 +50,7 @@ var _ data.BackupCollection = &prefetchCollection{}
 // as the calls are identical for populating the Collection
 type prefetchCollection struct {
 	// stream is the container for each individual SharePoint item of (page/list)
-	stream chan data.Item
+	stream map[path.CategoryType]chan data.Item
 	// fullPath indicates the hierarchy within the collection
 	fullPath path.Path
 	// jobs contain the SharePoint.List.IDs or SharePoint.Page.IDs
@@ -77,7 +77,7 @@ func NewPrefetchCollection(
 		fullPath:      folderPath,
 		items:         make([]string, 0),
 		getter:        getter,
-		stream:        make(chan data.Item, collectionChannelBufferSize),
+		stream:        make(map[path.CategoryType]chan data.Item),
 		client:        ac.Sites(),
 		statusUpdater: statusUpdater,
 		category:      scope.Category().PathType(),
@@ -122,8 +122,13 @@ func (pc *prefetchCollection) Items(
 	ctx context.Context,
 	errs *fault.Bus,
 ) <-chan data.Item {
+	if _, ok := pc.stream[pc.category]; !ok {
+		pc.stream[pc.category] = make(chan data.Item, collectionChannelBufferSize)
+	}
+
 	go pc.streamItems(ctx, errs)
-	return pc.stream
+
+	return pc.stream[pc.category]
 }
 
 // streamItems utility function to retrieve data from back store for a given collection
@@ -157,7 +162,7 @@ func (pc *prefetchCollection) streamLists(
 
 	defer updateStatus(
 		ctx,
-		pc.stream,
+		pc.stream[path.ListsCategory],
 		pc.statusUpdater,
 		pc.fullPath,
 		&metrics,
@@ -211,7 +216,7 @@ func (pc *prefetchCollection) streamPages(
 
 	defer updateStatus(
 		ctx,
-		pc.stream,
+		pc.stream[path.PagesCategory],
 		pc.statusUpdater,
 		pc.fullPath,
 		&metrics,
@@ -276,7 +281,7 @@ func (pc *prefetchCollection) streamPages(
 			continue
 		}
 
-		pc.stream <- item
+		pc.stream[path.PagesCategory] <- item
 		progress <- struct{}{}
 	}
 }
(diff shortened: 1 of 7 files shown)


src/internal/m365/collection/site/collection.go
```diff
@@ -86,67 +86,47 @@ func NewCollection(
 	return c
 }
 
-func (sc *Collection) SetBetaService(betaService *betaAPI.BetaService) {
+func (sc *prefetchCollection) SetBetaService(betaService *betaAPI.BetaService) {
```
pandeyabs, line 89: nit: Should we rename this from `sc` to `col` or `pc`?

src/internal/m365/collection/site/collection.go
```diff
@@ -171,7 +151,13 @@ func (sc *Collection) streamLists(
...
 	)
 
-	defer sc.finishPopulation(ctx, &metrics)
+	defer finishPopulation(
+		ctx,
+		sc.stream,
+		sc.statusUpdater,
+		sc.fullPath,
+		metrics,
+	)
```
pandeyabs, line 160: I can't comment on unedited lines. But one comment about the `semaphoreCh`  on line 166. Although it has a size of 5, we are serializing all requests i.e. we call `sc.handleListItems` for one `listID` at a time. 

Maybe it's missing a `go` keyword in front? It could be handled in a separate PR, just wanted to bring it up.

src/internal/m365/collection/site/collection.go
```diff
@@ -198,7 +184,7 @@ func (sc *Collection) streamLists(
 	wg.Wait()
 }
 
-func (sc *Collection) retrievePages(
+func (sc *prefetchCollection) retrievePages(
```
pandeyabs, line 187: nit: Should we call this `streamPages`, to be consistent with `streamLists`?

src/internal/m365/collection/site/collection.go
```diff
@@ -366,3 +358,26 @@ func (sc *Collection) handleListItems(
...
 }
+
+func finishPopulation(
+	ctx context.Context,
+	stream chan data.Item,
+	su support.StatusUpdater,
+	fullPath path.Path,
+	metrics support.CollectionMetrics,
+) {
+	close(stream)
```
pandeyabs, line 354: Quick question unrelated to this PR:

If I do a backup with both `lists` and `pages` as selected categories, and say  `Items` is called for `lists` first. After finishing its going to close the stream. Given pages and lists both share the same object instance `sc`, when `Items` will be called for `pages` next, it's going to error out on closed channel in `retrievePages`. Is my understanding correct?

src/internal/m365/collection/site/collection.go
```diff
@@ -366,3 +358,26 @@ func (sc *Collection) handleListItems(
 	sc.stream <- item
 	progress <- struct{}{}
 }
+
+func finishPopulation(
```
pandeyabs, line 347: nit: Could you rename this to `updateStatus`, just to stay consistent with other services? I know the arg set has diverged, but still it helps while gleaning through the code.

pandeyabs (approved):
LGTM. Added some questions unrelated to this PR. It shouldn't block this PR from going through though.

src/internal/m365/collection/site/collection.go
HiteshRepo: yeah. that will be a problem.

Should we change `Items` to below:
```
stream := make(chan data.Item)
go sc.streamItems(ctx, errs, stream)
return stream
```

src/internal/m365/collection/site/collection.go
HiteshRepo: `go` keyword infornt of `sc.handleListItems...`?
you suspect if there are, say 1000 lists, it will block the `semaphoreCh` because we have a smaller channel siz of `5`?

src/internal/m365/collection/site/collection.go
pandeyabs: Approach sounds good. You'd probably need a buffered channel like we have today, i.e. `make(chan data.Item, collectionChannelBufferSize)`.

src/internal/m365/collection/site/collection.go
pandeyabs: > go keyword infornt of sc.handleListItems...?

yes.

> you suspect if there are, say 1000 lists, it will block the semaphoreCh because we have a smaller channel siz of 5?

No. Currently we are downloading 1 list item at a time with prefetch collection. Using a goroutine will help us do 5 at a time since that's the semaphore capacity.

Do note that once we switch to lazy fetch collection for lists, `semaphoreCh` will only be useful for parallelizing list item enumeration. It will no longer be relevant for item data download since that's deferred to lazy reader i.e. parallelism for  list **data** download will be dictated by kopia.

src/internal/m365/collection/site/collection.go
```diff
@@ -41,15 +42,15 @@ const (
...
 
 // Collection is the SharePoint.List or SharePoint.Page implementation of data.Collection.
 
 // SharePoint.Libraries collections are supported by the oneDrive.Collection
 // as the calls are identical for populating the Collection
-type Collection struct {
+type prefetchCollection struct {
 	// stream is the container for each individual SharePoint item of (page/list)
-	stream chan data.Item
+	stream map[path.CategoryType]chan data.Item
```
pandeyabs, line 53: nit: We should update description above.

pandeyabs (approved):
Looks good!

PR #4944, opened 2023-12-28
lists collection stores last modified time

HiteshRepo:
updates lists prefetch collection to store last modified times along with list ids.

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/site/backup.go
+++ b/src/internal/m365/collection/site/backup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	stdpath "path"
+	"time"
 
 	"github.com/alcionai/clues"
 
@@ -131,7 +132,7 @@ func CollectPages(
 			su,
 			bpc.Options)
 		collection.SetBetaService(betaService)
-		collection.AddItem(tuple.ID)
+		collection.AddItem(tuple.ID, time.Now())
 
 		spcs = append(spcs, collection)
 	}
@@ -154,7 +155,7 @@ func CollectLists(
 	var (
 		el   = errs.Local()
 		spcs = make([]data.BackupCollection, 0)
-		acc  = api.CallConfig{Select: idAnd("list")}
+		acc  = api.CallConfig{Select: idAnd("list", "lastModifiedDateTime")}
 	)
 
 	lists, err := bh.GetItems(ctx, acc)
@@ -170,6 +171,8 @@ func CollectLists(
 		if api.SkipListTemplates.HasKey(ptr.Val(list.GetList().GetTemplate())) {
 			continue
 		}
+
+		lmt := list.GetLastModifiedDateTime()
 
 		dir, err := path.Build(
 			tenantID,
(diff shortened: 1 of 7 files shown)


src/internal/m365/collection/site/backup.go
```diff
@@ -154,7 +155,7 @@ func CollectLists(
 	var (
 		el   = errs.Local()
 		spcs = make([]data.BackupCollection, 0)
-		acc  = api.CallConfig{Select: idAnd("list")}
+		acc  = api.CallConfig{Select: idAnd("list", "lastModifiedDateTime")}
```
pandeyabs, line 158: nit: Can we call this `cfg` instead of `acc`? Easier to understand and used in other places too.

src/internal/m365/collection/site/backup.go
```diff
@@ -171,6 +172,8 @@ func CollectLists(
 			continue
 		}
 
+		lmt := list.GetLastModifiedDateTime()
```
pandeyabs, line 175: nit: `modTime` instead of `lmt` might be more descriptive.

src/internal/m365/collection/site/collection.go
```diff
@@ -53,8 +55,9 @@ type prefetchCollection struct {
 	stream map[path.CategoryType]chan data.Item
 	// fullPath indicates the hierarchy within the collection
 	fullPath path.Path
-	// jobs contain the SharePoint.List.IDs or SharePoint.Page.IDs
-	items []string
+	// items contains the SharePoint.List.IDs or SharePoint.Page.IDs
+	// and their corresponding last modified time
+	items map[string]time.Time
```
pandeyabs, line 60: QQ: If both lists and pages are selected to be backed up together, `pc.items` will end up storing IDs for both lists and pages at some point. So if lists follows pages, then we might end up streaming page IDs as list IDs. Could you please confirm?

If true, a quick fix would be to clean up `items` map on `Items()` exit. Other options could be a) 2D map - `path.CategoryType -> map[string]time.Time`. Something similar to the change you made for `stream`. b) use generics like groups does to support channels & group conversations with common code for pages and lists. c) split out pages and lists into separate services.

pandeyabs (approved):
Looks good. Left a concern unrelated to this PR.

src/internal/m365/collection/site/collection.go
pandeyabs: Sorry, my bad. I just looked at the code, this is not possible since `prefetchCollection` instances are different for pages and lists. I was wrong. This is not needed.

src/internal/m365/collection/site/backup.go
HiteshRepo: made the change in #4945

src/internal/m365/collection/site/backup.go
HiteshRepo: made the change in #4945

PR #4945, opened 2023-12-28
implements lists lazy fetch

HiteshRepo:
implements lists lazy fetch

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/site/backup.go
+++ b/src/internal/m365/collection/site/backup.go
@@ -149,13 +149,16 @@ func CollectLists(
 	scope selectors.SharePointScope,
 	su support.StatusUpdater,
 	errs *fault.Bus,
+	counter *count.Bus,
 ) ([]data.BackupCollection, error) {
 	logger.Ctx(ctx).Debug("Creating SharePoint List Collections")
 
 	var (
-		el   = errs.Local()
-		spcs = make([]data.BackupCollection, 0)
-		acc  = api.CallConfig{Select: idAnd("list", "lastModifiedDateTime")}
+		collection data.BackupCollection
+		el         = errs.Local()
+		cl         = counter.Local()
+		spcs       = make([]data.BackupCollection, 0)
+		acc        = api.CallConfig{Select: idAnd("list", "lastModifiedDateTime")}
 	)
 
 	lists, err := bh.GetItems(ctx, acc)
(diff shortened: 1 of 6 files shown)


src/internal/m365/collection/site/backup.go
```diff
@@ -185,17 +188,33 @@ func CollectLists(
...
-			dir,
-			ac,
-			scope,
-			su,
-			bpc.Options)
-
-		collection.AddItem(
-			ptr.Val(list.GetId()),
-			ptr.Val(lmt))
+		if lmt.IsZero() {
```
pandeyabs, line 191: This may cause NPE if `lmt` is nil. You can use `ptr.Val(lmt).IsZero()`

pandeyabs: Also can you please explain the logic behind why we are using zero mod time to select between prefetch and lazy collections?

src/internal/m365/collection/site/backup.go
```diff
@@ -185,17 +188,33 @@ func CollectLists(
...
+				scope,
+				su,
+				bpc.Options)
+
+			prefetchCol.AddItem(
+				ptr.Val(list.GetId()),
+				ptr.Val(lmt))
+
+			collection = prefetchCol
+		} else {
```
pandeyabs, line 205: nit: A pattern we follow throughout the codebase is that instead of if else, we initialize with a certain type e.g. `collection := NewLazyFetchCollection{...`. And then we do `if condition { collection := NewPrefetchCollection{...`

src/internal/m365/collection/site/backup.go
HiteshRepo: @pandeyabs I believe kopia uses `modtime` to determine if it needs to read an item or not. So in case if we get an invalid modtime from graph, the idea is to use a prefetch coll for that. Probably prefetch will never be invoked its just for precautinary purpose.

src/internal/m365/collection/site/backup.go
HiteshRepo: Definitely. In this case `AddItem` method is tied to both collection separately and cannot be called on the `collection` var. 

Let me make a change to avoid the if-else.

src/internal/m365/collection/site/backup.go
pandeyabs: Thanks, makes sense.

src/internal/m365/collection/site/backup.go
```diff
@@ -185,17 +188,35 @@ func CollectLists(
...
+		if modTime.IsZero() {
+			prefetchCol := NewPrefetchCollection(
+				bh,
+				dir,
+				ac,
+				scope,
+				su,
+				bpc.Options)
+
+			prefetchCol.AddItem(
```
pandeyabs, line 214: nit: `AddItem` call can be moved out as common code for both lazy and prefetch col.

pandeyabs (approved):
🚀

PR #4947, opened 2023-12-28
parses metadata collections using previous paths

HiteshRepo:
parses metadata collections using previous paths

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/site/backup.go
```diff
@@ -160,9 +160,19 @@ func CollectLists(
 		currPaths = map[string]string{}
 	)
 
+	dps, canUsePreviousBackup, err := parseMetadataCollections(ctx, path.ListsCategory, bpc.MetadataCollections)
+	if err != nil {
+		return nil, false, err
+	}
+
+	// [TODO] utilise deltapaths to determine list's state
```
pandeyabs, line 168: nit: We use `// TODO(hiteshrepo): comment` as a convention for marking todos in code.

src/internal/m365/collection/site/backup.go
```diff
@@ -160,9 +160,19 @@ func CollectLists(
...
+	_ = dps
```
pandeyabs, line 169: nit: You could do this in line 163 itself i.e. `_, canUsePreviousBackup, err`

src/internal/m365/collection/site/metadata.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/data"
+	"github.com/alcionai/corso/src/pkg/backup/metadata"
+	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
+	"github.com/alcionai/corso/src/pkg/path"
+)
+
+func parseMetadataCollections(
```
pandeyabs, line 16: 1. This is missing a unit test. You can use `TestParseMetadataCollections` as a reference. 

2. Also, should we make this function `list` specific? Maybe call the file `list_metadata.go` as well. Reasons: 
- It can be confusing since it comes off as a generic func used for all `sites` categories. But it's only being used for lists.
- It doesn't handle delta file paths, which is a list behavior. Might not be true for other sites categories. Please correct me if wrong.

3. Could you please also file a tech debt issue to make `parseMetadataCollections` a generic function which works for all services? Right now this code is duplicated everywhere, but most of it is common for all services. We don't need to fix the tech debt right away, but it'd be good to track it as an issue. Let me know if you disagree :)

pandeyabs (approved):
Approving to unblock as the remaining PRs in the chain look good in general. 

Please feel free to add new PRs for missing test coverages if that makes it more convenient for you.

src/internal/m365/collection/site/metadata.go
HiteshRepo: @pandeyabs 
1. Yeah sure I will write the unit test for it.
2. I thought, when sharepoint page is being implemented, that would to take care of dividing/unifying the function. But lets do what you suggested. In that case while page is being implemented, we have to consider unifying the logic based on delta path availability for page.
3. Yeah. I will assimilate tech-debts arising from PRs in the notion page momentarily and then move them to appropriate location (GH/linear).

PR #4948, opened 2023-12-28
determine collection state by current and previous paths

HiteshRepo:
determines sharepoint collection state using current and previous paths

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/site/metadata.go
```diff
@@ -97,3 +97,12 @@ func parseMetadataCollections(
 
 	return cdp[cat], true, nil
 }
+
+func pathFromPrevString(ps string) (path.Path, error) {
```
pandeyabs, line 101: We should also add this to the tech debt item. This func is also duplicated in 3 places. It can be moved to util or similar pkg.

src/internal/m365/collection/site/backup.go
```diff
@@ -188,6 +189,20 @@ func CollectLists(
 
 		storageDir := path.Elements{listID}
 
+		var (
+			dp          = dps[storageDir.String()]
+			prevPathStr = dp.Path
+		)
+
+		if len(prevPathStr) > 0 {
+			if prevPath, err = pathFromPrevString(prevPathStr); err != nil {
```
pandeyabs, line 198: We should add a test for previous paths, or metadata collections in general. This can be at `ProduceBackupCollection` level. You can use `TestCreateGroupsCollection_SharePoint_InvalidMetadata` as a reference. 

Please ignore if your upcoming PRs already do this.

pandeyabs (approved):

src/internal/m365/collection/site/backup.go
HiteshRepo: yes I believe testing of this part is taken care in the last PR in the chain

PR #4952, opened 2023-12-29
Add a product behavior section in docs

meain:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


website/docs/support/known-issues.md
```diff
@@ -35,3 +35,5 @@ Below is a list of known Corso issues and limitations:
 * Groups and Teams support is available in an early-access status, and may be subject to breaking changes.
 
 * Restoring the data into a different Group from the one it was backed up from isn't currently supported
+
+* Multipart emails with both html and text version will only contain the html version in export.
```
ryanfkeepers, line 39:
```suggestion
* Exports of multipart emails containing both html and text versions will only produce the html version of the email.
```

website/docs/support/product-behaviour.md
```diff
@@ -0,0 +1,9 @@
+# Product behavior
+
+Below is a list of possibly unexpected behaviors/limitations of the project.
```
ryanfkeepers, line 3: Is the difference between these notes and the known-issues page significant enough to create another page?  I worry we're splitting ownership between things where the owner is unclear.  Not a big deal when adding details, but possibly confusing for end users who go looking for this sort of info.

Could we get deliver the same info in a better way, and in a single page, if we update the known issues page with more structure?

website/docs/support/product-behaviour.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Drive items (OneDrive/SharePoint)
+
+- Link shares when restored always generate different links from the original.
```
ryanfkeepers, line 7:
```suggestion
- Restored link shares always generate different links from the original.
```

website/docs/support/product-behaviour.md
```diff
@@ -0,0 +1,9 @@
...
+
+- Anonymous link shares(link shares which aren't associated with any user) aren't restored.
```
ryanfkeepers, line 9: Any user?  Or any user in the org?  Not sure if there's a technical difference, or just a clarification.
```suggestion
- Anonymous link shares (link shares which aren't associated with any user) aren't restored.
```

website/docs/support/product-behaviour.md
meain: > Could we get deliver the same info in a better way, and in a single page, if we update the known issues page with more structure?

The suggestion for a separate page came through @vkamra , but having it in a single page could also work out I guess. The primary difference here is that these are things which might be "unexpected" behaviors but the `known_issues` page is actual issues we indent to fix(if possible).

vkamra: We could try doing this on the same page as `Known Issues` - and see if that is cleaner. 

How about we rename `Known Issues` to `Known Issues and Limitations` -> and add the structure @meain  is proposing below.

meain: @ryanfkeepers @vkamra I've combined both the pages. There were a few items in "known issues" which would better fit under "limitations".

PR #4954, opened 2023-12-29
Update handling of permissions restore failures

meain:
- All permissions failures are now recoverable to the granularity of a single permission or link share
- Failure to delete child permissions because the permission was not restored on parent will now not produce any warnings

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -214,7 +214,13 @@ func UpdatePermissions(
 
 		pid, ok := oldPermIDToNewID.Load(p.ID)
 		if !ok {
-			return clues.NewWC(ictx, "no new permission id")
+			logger.Ctx(ictx).Info("Unable to delete permission")
+			continue
+		}
+
+		if len(pid) == 0 {
+			// permission was not restored on parent and cannot be deleted
+			continue
 		}
 
 		err := udip.DeleteItemPermission(
@@ -277,7 +283,8 @@ func UpdatePermissions(
 
 		newPerm, err := udip.PostItemPermissionUpdate(ictx, driveID, itemID, pbody)
 		if graph.IsErrUsersCannotBeResolved(err) {
-			logger.CtxErr(ictx, err).Info("Unable to restore link share")
+			oldPermIDToNewID.Store(p.ID, "") // empty to signify that we could not restore
+			logger.CtxErr(ictx, err).Info("Unable to restore permission")
 			continue
 		}
 
@@ -385,6 +392,7 @@ func UpdateLinkShares(
 
 		newLS, err := upils.PostItemLinkShareUpdate(ictx, driveID, itemID, lsbody)
 		if graph.IsErrUsersCannotBeResolved(err) {
+			oldLinkShareIDToNewID.Store(ls.ID, "") // empty to signify that we could not restore
 			logger.CtxErr(ictx, err).Info("Unable to restore link share")
 			continue
 		}
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -462,39 +462,44 @@ func RestorePermissions(
 	current metadata.Metadata,
 	caches *restoreCaches,
 	errs *fault.Bus,
-) error {
+) {
 	if current.SharingMode == metadata.SharingModeInherited {
-		return nil
+		return
 	}
+
+	var didReset bool
 
 	ctx = clues.Add(ctx, "permission_item_id", itemID)
 
 	previousLinkShares, err := computePreviousLinkShares(ctx, itemPath, caches.ParentDirToMeta)
 	if err != nil {
-		return clues.Wrap(err, "previous link shares")
+		errs.AddRecoverable(ctx, clues.WrapWC(ctx, err, "previous link shares"))
 	}
 
-	lsAdded, lsRemoved := metadata.DiffLinkShares(previousLinkShares, current.LinkShares)
+	if previousLinkShares != nil {
+		lsAdded, lsRemoved := metadata.DiffLinkShares(previousLinkShares, current.LinkShares)
 
-	// Link shares have to be updated before permissions as we have to
-	// use the information about if we had to reset the inheritance to
-	// decide if we have to restore all the permissions.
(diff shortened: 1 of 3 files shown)


src/internal/m365/collection/drive/permission.go
```diff
@@ -214,7 +214,13 @@ func UpdatePermissions(
 
 		pid, ok := oldPermIDToNewID.Load(p.ID)
 		if !ok {
-			return clues.NewWC(ictx, "no new permission id")
+			el.AddRecoverable(ictx, clues.NewWC(ictx, "finding updated permission id"))
```
ryanfkeepers, line 217:
```suggestion
			el.AddRecoverable(ictx, clues.NewWC(ictx, "no permission matches id"))
```

src/internal/m365/collection/drive/permission.go
```diff
@@ -277,7 +283,9 @@ func UpdatePermissions(
 
 		newPerm, err := udip.PostItemPermissionUpdate(ictx, driveID, itemID, pbody)
 		if graph.IsErrUsersCannotBeResolved(err) {
-			logger.CtxErr(ictx, err).Info("Unable to restore link share")
+			oldPermIDToNewID.Store(p.ID, "") // empty to signify that we could not restore
+			logger.CtxErr(ictx, err).Info("Unable to restore permission")
```
ryanfkeepers, line 287:
```suggestion
			logger.CtxErr(ictx, err).Info("unable to restore permission")
```

src/internal/m365/collection/drive/permission.go
```diff
@@ -385,7 +393,9 @@ func UpdateLinkShares(
 
 		newLS, err := upils.PostItemLinkShareUpdate(ictx, driveID, itemID, lsbody)
 		if graph.IsErrUsersCannotBeResolved(err) {
+			oldLinkShareIDToNewID.Store(ls.ID, "") // empty to signify that we could not restore
```
ryanfkeepers, line 396: Nit: a const might be easier to grok than a comment.
```suggestion
const nonRestorablePermission = ""
// ...
			oldLinkShareIDToNewID.Store(ls.ID, nonRestorablePermission)
```

src/internal/m365/collection/drive/permission.go
```diff
@@ -509,5 +519,9 @@ func RestorePermissions(
 		return nil
 	}
 
-	return clues.Wrap(err, "updating permissions").OrNil()
+	if err != nil {
+		errs.AddRecoverable(ctx, clues.WrapWC(ctx, err, "updating permissions"))
+	}
+
+	return nil
```
ryanfkeepers, line 526: Some concerns about the pattern of producing both recoverable errors and also returning an error here.  This is a usage we've tried to stay away from in the past because it makes error ownership ambiguous.

The base idea is recoverable errors allow the owner of a loop to complete the loop without exiting early from an error.  Non-looping code should simply return errors (as is idiomatic in golang) because that scope doesn't have full context about whether the caller considers that call recoverable.  Now that recoverable errors are expected to produce top-level operation failures (depending on configs), it can be risky for arbitrary funcs to assume identification and ownership of recoverability.

In this case, I think we should retain the original error handling instead of adding recoverables.  If the case is something we can move past without exiting the func over (ie, line 488), then an info log provides sufficient tracking.

src/internal/m365/collection/drive/restore.go
```diff
@@ -591,7 +591,13 @@ func CreateRestoreFolders(
...
 		errs)
 
-	return id, err
+	if err != nil {
+		// We should not ideally get any errors as we mostly skip
+		// over them with a recoverable error. This is just in case.
+		errs.AddRecoverable(ctx, clues.Wrap(err, "restoring folder permissions"))
+	}
+
+	return id, nil
```
ryanfkeepers, line 600: same concerns here.

src/internal/m365/collection/drive/permission.go
meain: Hmm, I understand your concern but this one is a little tricky. If we fail to create the folders, we can't proceed(and so we return an error), but we need both link share and permissions restore to go through even when there are errors. Also, we should not consider the restore a success with a link share or permissions restore error and so just using a log would not work here.

Here is what we want the code to do. Let me know if you have a suggestion on how to better do this:
- If folder creation fails, hard error
- If link share or permissions restore fails, continue but with recoverable error

meain: Ahh, you are right. There is no reason why `RestorePermissions` should be returning an error.

src/internal/m365/collection/drive/restore.go
meain: https://github.com/alcionai/corso/pull/4954#discussion_r1441566703

ryanfkeepers (approved):
see the last comment in unresolved conversation about error handling.  Approving to unblock.

src/internal/m365/collection/drive/permission.go
meain: Hmm, the problem is that within `RestorePermissins`, I want to ensure we restore permissions even if restore of link shares fail. If we return an error, this is not possible. One option would be to capture and return the link share error at the end if permissions succeed and and link shares fail, but this might look non-idiomatic as well.

ryanfkeepers: > Hmm, the problem is that within RestorePermissins, I want to ensure we restore permissions even if restore of link shares fail. If we return an error, this is not possible

In that case it shouldn't produce an error.  Only a log.  Maybe an alert?  Recoverable errors should be thought of as eqivalent to non-recoverable errors.  The only difference is that we might defer the handling, depending on runtime configuration.  IE: if we wouldn't exit the process with an error, we don't need a recoverable error.

meain: > ... if we wouldn't exit the process with an error, we don't need a recoverable error.

We have to exit the process with an error, but only after we have completed restoring everything we can.

PR #4955, opened 2023-12-29
Filter out non existent entities in permissions and link shares

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -397,7 +397,7 @@ func UpdateLinkShares(
 
 		newLS, err := upils.PostItemLinkShareUpdate(ictx, driveID, itemID, lsbody)
 		if graph.IsErrUsersCannotBeResolved(err) {
-			oldLinkShareIDToNewID.Store(ls.ID, "") // empty to signify that we could not restore
+			oldLinkShareIDToNewID.Store(ls.ID, nonRestorablePermission)
 			logger.CtxErr(ictx, err).Info("unable to restore link share")
 
 			continue
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -452,7 +452,7 @@ func UpdateLinkShares(
 func filterUnavailableEntitiesInLinkShare(
 	ctx context.Context,
 	linkShares []metadata.LinkShare,
-	availableEntities AvailableEntities,
+	availableEntities ResourceIDNames,
 	oldLinkShareIDToNewID syncd.MapTo[string],
 ) []metadata.LinkShare {
 	filtered := []metadata.LinkShare{}
@@ -505,7 +505,7 @@ func filterUnavailableEntitiesInLinkShare(
 func filterUnavailableEntitiesInPermissions(
 	ctx context.Context,
 	perms []metadata.Permission,
-	availableEntities AvailableEntities,
+	availableEntities ResourceIDNames,
 	oldPermIDToNewID syncd.MapTo[string],
 ) []metadata.Permission {
 	if availableEntities.Users == nil || availableEntities.Groups == nil {
--- a/src/internal/m365/collection/drive/permission.go
+++ b/src/internal/m365/collection/drive/permission.go
@@ -450,11 +450,18 @@ func UpdateLinkShares(
 }
 
 func filterUnavailableEntitiesInLinkShare(
+	ctx context.Context,
 	linkShares []metadata.LinkShare,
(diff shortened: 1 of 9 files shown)


src/pkg/services/m365/api/groups.go
```diff
@@ -50,6 +51,25 @@ func (c Groups) GetAll(
 	return getGroups(ctx, errs, service)
 }
 
+// GetAllIDsAndNames retrieves all groups in the tenant and returns them in an idname.Cacher
+func (c Groups) GetAllIDsAndNames(ctx context.Context, errs *fault.Bus) (idname.Cacher, error) {
```
ryanfkeepers, line 55: Request: can you please move this addition to a separate PR?  And, in that change, can you please remove this old func in services favor of this new one? https://github.com/alcionai/corso/blob/main/src/pkg/services/m365/groups.go#L160

src/internal/m365/collection/drive/restore_caches.go
```diff
@@ -21,6 +22,11 @@ type driveInfo struct {
 	rootFolderID string
 }
 
+type AvailableEntities struct {
```
ryanfkeepers, line 25: Prefer the term `Resource` over `Entity`, please.  Entity is too generic, and can reference non-resource entries.

```suggestion
type ResourceIDNames struct {
```

src/internal/m365/collection/drive/permission.go
```diff
@@ -445,6 +445,102 @@ func UpdateLinkShares(
 	return alreadyDeleted, nil
 }
 
+func filterUnavailableEntitiesInLinkShare(
```
ryanfkeepers, line 448: these two funcs could be merged pretty easily with generics.

```go
nc filterToExistingResourceEntities[T any](
 	ctx context.Context,
 	perms []T,
 	availableEntities AvailableEntities,
 	oldIDToNewID syncd.MapTo[string],
 ) []T {
```

ryanfkeepers (approved):
code implementation could be refined, and there's a request for breaking apart the PR.  But nothing blocking.

src/internal/m365/collection/drive/permission.go
meain: Hmm, maybe I am missing something, but I could not get it folded into one. They are slightly different enough to cause trouble.

src/pkg/services/m365/api/groups.go
meain: Moving this out to a separate PR might get a bit messy as we need it for this PR, but let me create a follow up to remove the old one.

meain: Created https://github.com/alcionai/corso/pull/4967

src/internal/m365/collection/drive/permission.go
ryanfkeepers: Ah, looks like you'd need to extend an interface.  Then it'd work.  Take it or leave it, no big deal.

```go
type idAndTyper interface {
  ID() string
  Type() metadata.GV2Type
}

func filterToExistingResourceEntities[T idAndTyper](
 	ctx context.Context,
 	orig []T,
 	availableEntities ResourceIDNames,
 	oldIDToNewID syncd.MapTo[string],
 ) []T {
 	if availableEntities.Users == nil || availableEntities.Groups == nil {
 		logger.Ctx(ctx).Info("no resources provided for filtering")
 		return orig
 	}

 	filtered := []T{}

 	for _, t := range orig {
 		available := false

 		switch t.Type() {
 		case metadata.GV2User:
 			_, ok := availableEntities.Users.NameOf(t.ID())
 			available = available || ok
 		case metadata.GV2Group:
 			_, ok := availableEntities.Groups.NameOf(t.ID())
 			available = available || ok
 		default:
 			// We only know about users and groups
 			// TODO: extend the check to other entity types
 			available = true
 		}

 		if available {
 			filtered = append(filtered, t)
 			continue
 		}

 		// If we have no entities, we can't restore the permission
 		// and so we have to mark it as not restored.
 		oldPermIDToNewID.Store(t.ID(), "")
 	}

 	return filtered
 }
```

meain: Yeah that works, but I think I'll skip it the primary reason being that the implementation within is also slightly different. Permissions has a single entry and link shares can have multiple entities. Don't think it will give a big advantage.

PR #4957, opened 2024-01-01
ignores accessRequest list templates from backup and restore

HiteshRepo:
ignores accessRequest list templates from backup and restore

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/lists.go
```diff
@@ -53,6 +53,7 @@ const (
 	// tracks to backup/restore supportability of `documentLibrary` templated lists
 	DocumentLibraryListTemplateName = "documentLibrary"
 	SharingLinksListTemplateName    = "sharingLinks"
+	AccessRequestsListTemplateName  = "accessRequest"
```
pandeyabs, line 56: minor nit: `Name` suffix is redundant for the templates and can be removed. I do see we are using the same convention for all var namings in this file. Usually we avoid such suffixes as it can make code verbose. 

Also, I don't think we need to export any of these vars, since their usage is restricted to api pkg.

pandeyabs (approved):

src/pkg/services/m365/api/lists.go
HiteshRepo: Using `WebTemplateExtensionsListTemplateName` and `DocumentLibraryListTemplateName` in `collection/site/restore_test.go`, hence exported. Others are exported for parity. I will remove the `Name` suffix.

pandeyabs: Ah thanks, I didn't notice that we were using a template in restore tests.

Issue #3890, opened 2023-07-24
`ics` export form Exchange events backup


meain:

meain:
Closing this as we are currently only planning for `pst` exports for exchange.

PR #4958, opened 2024-01-02
Events export as ICS

meain:
There are a few more pending items. I've added TODO entries for them. The PR was getting pretty big and so I thought I would address them in a followup PR.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/ics/ics_test.go
```diff
@@ -0,0 +1,1078 @@
...
+
+			e, err := FromJSON(ctx, bts)
+			require.NoError(t, err, "converting to ics")
+
+			tt.check(e)
+		})
+	}
+}
+
+// checkAttendee check the ATTENDEE field
```
ashmrtn, line 752: nit:
```suggestion
// checkAttendee checks the ATTENDEE field
```

src/internal/converters/ics/ics_test.go
```diff
@@ -0,0 +1,1078 @@
...
+// This is required instead of a string check as the fields might
+// always be in the same order
```
ashmrtn, line 754:
```suggestion
// not be in the same order
```

src/internal/converters/ics/ics.go
```diff
@@ -0,0 +1,444 @@
...
+	// this everywhere. This is the easiest way to ensure we
+	// have the correct time everywhere and is the same(need
+	// to same according to spec).
+	it, err := dttm.ParseTime(ts)
+	if err != nil {
+		return time.Now(), clues.Wrap(err, "parsing time")
+	}
+
+	timezone, ok := GraphTimeZoneToTZ[tz]
+	if !ok {
```
ashmrtn, line 90: you may need to check if Outlook allows creating events without a specified timezone (i.e. it just occurs at the same time of day in the local timezone no matter where the user is). The iCal spec seems to allow for this (called a "floating time" or something) but the code here seems like it may end up returning an error and skipping the item if Outlook allows creating an event like that

ashmrtn (approved):
generally looks good, most comments are nits or questions since I haven't gone through the RFC a whole lot

src/internal/m365/collection/exchange/export.go
```diff
@@ -111,6 +112,20 @@ func streamItems(
...
+				}
+			case path.EventsCategory:
+				outData, err = ics.FromJSON(ctx, content)
+				if err != nil {
+					err = clues.Wrap(err, "converting to ics")
+
+					logger.CtxErr(ctx, err).Info("processing collection item")
+
+					ch <- export.Item{
+						ID:    id,
```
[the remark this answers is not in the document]
meain: Ahh, you are right. I missed updating that.

src/internal/converters/ics/ics_test.go
meain: oops!

src/internal/converters/ics/ics.go
```diff
@@ -0,0 +1,444 @@
...
+		index := pat.GetIndex()
+		prefix := ""
+
+		if index != nil &&
+			(ptr.Val(freq) == models.RELATIVEMONTHLY_RECURRENCEPATTERNTYPE ||
+				ptr.Val(freq) == models.RELATIVEYEARLY_RECURRENCEPATTERNTYPE) {
+			prefix = fmt.Sprint(GraphToICalIndex[index.String()])
+		}
+
+		recurComponents = append(recurComponents, "BYDAY="+prefix+strings.Join(dowComponents, ","))
```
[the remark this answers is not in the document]
meain: I think this should be fine as we have `dow != nil` check above. It is supposed to at least have one entry.

src/internal/converters/ics/ics.go
meain: Does not look like from docs. When trying to create one without a timezone, I get the following error.

```json
{
  "error": {
    "code": "TimeZoneNotSupportedException",
    "message": "A valid TimeZone value must be specified. The following TimeZone value is not supported: ''."
  }
}
```

src/internal/converters/ics/consts.go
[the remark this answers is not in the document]
meain: I understand the concern, but these ones should be stable. They are converting from what graph send(which I don't expect to change) to what the RFC specifies. None of these are specific to the library in use.

src/internal/converters/ics/ics.go
```diff
@@ -0,0 +1,444 @@
...
+	if draft != nil {
+		event.SetStatus(ics.ObjectStatusDraft)
+	}
+
+	summary := data.GetSubject()
+	if summary != nil {
+		event.SetSummary(ptr.Val(summary))
+	}
+
+	// Emojies seem to mess up the ics file
```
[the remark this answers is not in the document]
meain: I'm merging this as is for now. I've changed it to a TODO, let me spent a bit more time on this.

meain:
I've [reverted](https://github.com/alcionai/corso/pull/4958/commits/7fdada3be1766cd9a7947051cde34b06276d916d) events export as of now. Will reenable with the next PR.

meain:
The first CI run ran into an issue which should be fixed via https://github.com/alcionai/corso/pull/4899

src/internal/converters/ics/ics.go
ashmrtn: great, thanks for checking that!

PR #4970, opened 2024-01-04
handles location columns and fields

HiteshRepo:
Location column of a list item is not identifiable from GRAPH's response.
Hence this is a maneouver to handle such fields until a column definition is introduced for `Location`.

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/lists.go
```diff
@@ -29,11 +30,18 @@ const (
...
+	NestedCountryFieldName    = "countryOrRegion"
+	NestedPostalCodeFieldName = "postalCode"
+	NestedStateFieldName      = "state"
+	NestedStreetFieldName     = "street"
+	CoordinatesFieldName      = "coordinates"
+	NestedLatitudeFieldName   = "latitude"
+	NestedLongitudeFieldName  = "longitude"
+	DisplayNameFieldName      = "displayName"
+	LocationURIFieldName      = "locationUri"
+	UniqueIDFieldName         = "uniqueId"
```
ryanfkeepers, line 44: this is getting a bit difficult to read, both here in the declaration and in usage.  I think you'd be okay to use some abbreviations to simplify "FieldName" to FN.  However, like stated in other comments, I think you're using the term `fieldName` in place of `key`, as in the key for a map entry.  I think we should avoid that vocabulary, since it's more difficult to grok while `key` is very simple.
```suggestion
	AddressKey          = "address"
	CoordinatesKey      = "coordinates"
	DisplayNameKey      = "displayName"
	LocationURIKey      = "locationUri"
	UniqueIDKey         = "uniqueId"
	
	// entries that are nested within a second layer
	CityKey       = "city"
	CountryKey    = "countryOrRegion"
	PostalCodeKey = "postalCode"
	StateKey      = "state"
	StreetKey     = "street"
	LatitudeKey   = "latitude"
	LongitudeKey  = "longitude"
```

src/pkg/services/m365/api/lists.go
```diff
@@ -529,6 +551,62 @@ func setAdditionalDataByColumnNames(
...
+		return concatenatedAddress
+	}
+
+	if ptrValue, ok := field[fieldName].(*string); ok {
+		concatenatedAddress = fmt.Sprintf("%s,%s", concatenatedAddress, ptr.Val(ptrValue))
+	} else if ptrValue, ok := field[fieldName].(*float64); ok {
+		concatenatedAddress = fmt.Sprintf("%s,%v", concatenatedAddress, ptr.Val(ptrValue))
+	}
+
+	return concatenatedAddress
```
ryanfkeepers, line 607: three suggestions:
1/ this should use the str package for the first check.
2/ the not-exists check is redundant.
3/ the more idiomatic way to concatenate strings is to build a slice and then Join it.

```suggestion
func concatenateField(
parts []string,
fields map[string]any,
key string,
 ) []string {
	if v, err := str.AnyValueToString(key, fields); err != nil {
		parts = append(parts, v)
	} else if fl, ok := field[fieldName].(*float64); ok {
		parts = append(parts, fmt.Sprintf("%v", fl))
	}

	return parts
	// the caller can later use strings.Join(parts, ",")
```

src/pkg/services/m365/api/lists.go
```diff
@@ -529,6 +551,62 @@ func setAdditionalDataByColumnNames(
...
+}
+
+func concatenateAddressFields(addressFields map[string]any) string {
+	concatenatedAddress := ""
+
+	if dispName, ok := addressFields[DisplayNameFieldName].(*string); ok {
+		concatenatedAddress = fmt.Sprintf("%s,%s", concatenatedAddress, ptr.Val(dispName))
+	}
+
+	if address, ok := addressFields[AddressFieldName].(map[string]any); ok {
```
ryanfkeepers, line 576: naming a map a singular name (ex: address, field, etc) makes it seem like the value isn't a map, because maps contain multiple values.  
```suggestion
	if fields, ok := addressFields[AddressFieldName].(map[string]any); ok {
```

src/pkg/services/m365/api/lists.go
```diff
@@ -529,6 +551,62 @@ func setAdditionalDataByColumnNames(
 	return filteredData
 }
 
+func hasAddressFields(additionalData map[string]any) (map[string]any, string, bool) {
+	for fieldName, value := range additionalData {
```
ryanfkeepers, line 555: nit: prefer `k` or `key` when you need a generic name for a map key.
```suggestion
	for k, value := range additionalData {
```

ryanfkeepers (commented):
I would like to better understand the design goals here before approving.

HiteshRepo:
> I would like to better understand the design goals here before approving.

@ryanfkeepers Agreed. Also I should have created this as a draft PR instead. I was unsure of the solution and wanted to have a disucssion, hence I had not added reviewers. Here is the notion link where I have jot down the issue details https://www.notion.so/alcion/Issue-Location-column-type-e6feff73f9284ac3a2b960c03dfabe26

Similar issue for hyperlink column as well: https://www.notion.so/alcion/Issue-Hyperlink-Column-type-f91b5c79cd4f49aeb868e0ae0d3b3d0e

src/pkg/services/m365/api/lists.go
```diff
@@ -504,6 +520,12 @@ func retrieveFieldData(orig models.FieldValueSetable, columnNames map[string]str
 	fields := models.NewFieldValueSet()
 
 	additionalData := setAdditionalDataByColumnNames(orig, columnNames)
+
+	if addressField, fieldName, ok := hasAddressFields(additionalData); ok {
+		concatenatedAddress := concatenateAddressFields(addressField)
+		additionalData[fieldName] = concatenatedAddress
```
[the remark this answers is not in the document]
HiteshRepo: This is how a list with `location` columns look like:
Screenshot from 2024-01-10 23-43-02

with the above workaround, the restored list will look like:
Screenshot from 2024-01-10 23-41-19

lists with `location` columns cannot be restored as the way it originally was created because of the issue described [here](https://github.com/microsoftgraph/msgraph-sdk-go/issues/638).

pandeyabs:
@HiteshRepo are the location column and hyperlink column PRs ready for review? Asking because they appear to be in draft stage.

HiteshRepo:
> @HiteshRepo are the location column and hyperlink column PRs ready for review? Asking because they appear to be in draft stage.

yes @pandeyabs

pandeyabs:
Cool! I'll review them tomorrow.

pandeyabs (approved):
🚢

PR #4972, opened 2024-01-04
centralize core err interpretation

ryanfkeepers:
now that we have a core error library, we can start swapping out the percolation of graph api errors with core error sentinels.  This PR begins that process by adding a centralized transformer to the graph service handlers that ensures all calls produce the same set of core errors according to some standard categorizations. This is just an initial step, and does not transfer all graph errors into core sentinels.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4685 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/http_wrapper.go
+++ b/src/pkg/services/m365/api/graph/http_wrapper.go
@@ -120,8 +120,12 @@ func (hw httpWrapper) Request(
 			break
 		}
 
-		if IsErrApplicationThrottled(err) {
-			return nil, Stack(ctx, clues.Stack(core.ErrApplicationThrottled, err))
+		err = stackErrsCore(ctx, err, 1)
+
+		// force an early exit on throttling issues.
+		// those retries are already handled in middleware.
+		if errors.Is(err, core.ErrApplicationThrottled) {
+			return nil, err
 		}
 
 		var http2StreamErr http2.StreamError
--- a/src/pkg/services/m365/api/graph/http_wrapper.go
+++ b/src/pkg/services/m365/api/graph/http_wrapper.go
@@ -120,7 +120,7 @@ func (hw httpWrapper) Request(
 			break
 		}
 
-		err = stackErrsCore(ctx, err, 1)
+		err = stackWithCoreErr(ctx, err, 1)
 
 		// force an early exit on throttling issues.
 		// those retries are already handled in middleware.
@@ -130,7 +130,7 @@ func (hw httpWrapper) Request(
 
 		var http2StreamErr http2.StreamError
 		if !errors.As(err, &http2StreamErr) {
-			return nil, Stack(ctx, err)
+			return nil, err
 		}
 
 		logger.Ctx(ctx).Debug("http2 stream error")
@@ -140,7 +140,7 @@ func (hw httpWrapper) Request(
 	}
 
 	if err != nil {
-		return nil, Stack(ctx, err)
+		return nil, err
 	}
 
 	logResp(ctx, resp)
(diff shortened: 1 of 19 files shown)


src/pkg/services/m365/api/graph/http_wrapper.go
```diff
@@ -120,13 +120,17 @@ func (hw httpWrapper) Request(
 			break
 		}
 
-		if IsErrApplicationThrottled(err) {
-			return nil, Stack(ctx, clues.Stack(core.ErrApplicationThrottled, err))
+		err = stackWithCoreErr(ctx, err, 1)
```
ashmrtn, line 123: hmmm I'm trying to figure out if a slight rearrangement of the code would be cleaner here. I'm not sure, but maybe it would help since we could have just a single if-block to check for errors? I was thinking of something like the following for the loop body:
1. make request
2. unconditionally call `stackWithCoreErr`
3. if the error is an http2StreamError: sleep and use `continue` to try again
4. otherwise: break out of the loop

src/pkg/services/m365/api/graph/service.go
```diff
@@ -392,11 +393,12 @@ func (aw *adapterWrap) Send(
 			break
 		}
 
+		err = stackWithCoreErr(ictx, err, 1)
```
ashmrtn, line 396: is this going to cause additional core errors to be added to the stack since from my understanding it's above the the stuff defined in in http_wrapper.go?

src/pkg/services/m365/api/graph/service.go
```diff
@@ -392,11 +393,12 @@ func (aw *adapterWrap) Send(
...
+
 		// force an early exit on throttling issues.
-		// those retries are well handled in middleware already. We want to ensure
-		// that the error gets wrapped with the appropriate sentinel here.
-		if IsErrApplicationThrottled(err) {
-			return nil, clues.StackWC(ictx, core.ErrApplicationThrottled, err).WithTrace(1)
+		// those retries are already handled in middleware.
+		if errors.Is(err, core.ErrApplicationThrottled) {
+			return nil, err
 		}
```
ashmrtn, line 402: is this explicit check still required? It seems like this would fit into the default case in the switch below

src/pkg/services/m365/api/groups.go
```diff
@@ -113,10 +114,6 @@ func (c Groups) GetTeamByID(
 
 	t, err := c.Stable.Client().Teams().ByTeamId(identifier).Get(ctx, nil)
 	if err != nil {
-		if graph.IsErrResourceLocked(err) {
-			return nil, graph.Stack(ctx, clues.Stack(graph.ErrResourceLocked, err))
-		}
-
 		return nil, graph.Wrap(ctx, err, "finding team by ID")
```
ashmrtn, line 117: nit: the errors in most/all of these calls with changes can just be clues calls now right? The lower layers should now add the error details when it calls `stackWithCoreErr` no?

```suggestion
		return nil, clues.Wrap(err, "finding team by ID")
```

src/pkg/services/m365/api/users_test.go
```diff
@@ -81,16 +81,16 @@ func (suite *UsersUnitSuite) TestEvaluateMailboxError() {
 		},
 		{
 			name: "mail inbox err - user not found",
-			err:  graphTD.ODataErr(string(graph.RequestResourceNotFound)),
+			err:  core.ErrResourceOwnerNotFound,
```
ashmrtn, line 84: nice!

ashmrtn (commented):
generally looks good, mostly just minor things to try to keep things clean and make sure the things we think are happening are actually happening

I would like to see some info about what changes, if any, SDK users will need to make since this looks like it starts to change some (more?) of the information returned in calls available to them

src/pkg/services/m365/api/graph/http_wrapper.go
ryanfkeepers: I think that's largely what we're doing, just a twist on how we continue/break.  This should also get cleaned up a little in the next PR.  I needed to deduplicate the error stacking already.

ryanfkeepers:
> I would like to see some info about what changes, if any, SDK users will need to make since this looks like it starts to change some (more?) of the information returned in calls available to them

@ashmrtn good question.  The general change is that sdk users will switch from `errs.Is(err, sentinel)` to `errors.Is(err, core.Sentinel)`.  No other changes should be necessary.

ashmrtn (approved):

PR #4973, opened 2024-01-04
deduplicate graph api errors

ryanfkeepers:
the graph api service wrapper is currently stacking every retried error, leading to errors that read like, "foo; bar; baz; baz; baz". This output is confusing and unhelpful for debugging.  The changes in this PR reduce that duplication while still adding the prior error cases into the returned context so that the returned errors are easier to grok.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (approved):
Nice!

PR #4975, opened 2024-01-05
make a `not found` core error

ryanfkeepers:
make a canonical 'not found' error for our core sentinels. This replaces the graph.DeletedInFlight error, which acted as an interpretation of "not found".

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4685

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/errs/core/core.go
+++ b/src/pkg/errs/core/core.go
@@ -51,16 +51,28 @@ func (e Err) Error() string {
 }
 
 var (
+	// occurs when creation of an entity (usually by restful POST or PUT) errors
+	// because some other entity already already exists with a conflicting identifier.
+	// The identifier is not always the id.  For example: duplicate filenames
+	// in the same directory will cause conflicts, even with different IDs.
+	ErrAlreadyExists = &Err{msg: "conflict: already exists"}
 	// currently we have no internal throttling controls.  We only try to match
 	// external throttling requirements.  This sentinel assumes that an external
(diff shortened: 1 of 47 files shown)


src/pkg/errs/core/core.go
```diff
@@ -61,6 +61,10 @@ var (
 	ErrBackupNotFound = &Err{msg: "backup not found"}
 	// a catch-all for downstream api auth issues.  doesn't matter which api.
 	ErrInsufficientAuthorization = &Err{msg: "insufficient authorization"}
+	// basically what it sounds like: we went looking for something by ID and
+	// it wasn't found.  This might be because it was deleted in flight, or
+	// because we had the id wrong, or some other reason.
+	ErrNotFound = &Err{msg: "not found"}
```
meain, line 67: Nit:
The changes look good, but I'm not sure sure about the part of the comment "... because we had the id wrong ...".
How would we end up with a wrong id? And even if we do, should it be considered same as "deleted in flight"?

meain (approved):
Have a small nit, approving to unblock.

src/pkg/errs/core/core.go
ryanfkeepers: > How would we end up with a wrong id?

Well, hopefully we never would, right?  Heh, I'm probably adding more confusion than I'm alleviating here.  I'll remove that part.  Thanks for pointing it out.

PR #4977, opened 2024-01-05
move conflictAlreadyExists to core errs

ryanfkeepers:
move the graph error for already-exists conflicts on item creation to the errs/core package as a more generic sentinel.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4685

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/errs/core/core.go
+++ b/src/pkg/errs/core/core.go
@@ -61,6 +61,11 @@ var (
 	// about what it sounds like: we tried to look for a backup by ID, but the
 	// storage layer couldn't find anything for that ID.
 	ErrBackupNotFound = &Err{msg: "backup not found"}
+	// occurs when creation of an entity (usually by restful POST or PUT) errors
+	// because some other entity already already exists with a conflicting identifier.
+	// The identifier is not always the id.  For example: duplicate filenames
+	// in the same directory will cause conflicts, even with different IDs.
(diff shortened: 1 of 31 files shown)


src/pkg/errs/core/core.go
```diff
@@ -61,6 +61,11 @@ var (
 	// about what it sounds like: we tried to look for a backup by ID, but the
 	// storage layer couldn't find anything for that ID.
 	ErrBackupNotFound = &Err{msg: "backup not found"}
+	// occurs when creation of an entity (usually by restful POST or PUT) errors
+	// because some other entity already already exists with a conflicting identifier.
+	// The identifier is not always the id.  For example: duplicate filenames
+	// in the same directory will cause conflicts, even with different IDs.
+	ErrConflictAlreadyExists = &Err{msg: "conflict: already exists"}
```
ashmrtn, line 68: I'm kind of on the fence with the error naming. At least for me reading it the first time it sounded like the conflict already existed, not the item that created the conflict. Could we maybe rename it to `ErrItemAlreadyExists` or `ErrItemConflict` (or something else if there's a better idea)?

ashmrtn (approved):

src/pkg/errs/core/core.go
ryanfkeepers: sounds good!

PR #4988, opened 2024-01-08
Extra logging and clues context for folder caches

ashmrtn:
Add extra logging and clues to errors when populating container caches for Exchange. This could help track down some things related to time spent getting data.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/contacts_container_cache.go
+++ b/src/internal/m365/collection/exchange/contacts_container_cache.go
@@ -2,6 +2,7 @@ package exchange
 
 import (
 	"context"
+	"time"
 
 	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
@@ -94,10 +95,12 @@ func (cfc *contactContainerCache) Populate(
 	baseID string,
 	baseContainerPath ...string,
 ) error {
+	start := time.Now()
+
 	logger.Ctx(ctx).Info("populating container cache")
 
 	if err := cfc.init(ctx, baseID, baseContainerPath); err != nil {
-		return clues.WrapWC(ctx, err, "initializing")
+		return clues.Wrap(err, "initializing")
 	}
 
 	el := errs.Local()
@@ -129,10 +132,12 @@ func (cfc *contactContainerCache) Populate(
 	}
 
 	if err := cfc.populatePaths(ctx, errs); err != nil {
-		return clues.WrapWC(ctx, err, "populating paths")
+		return clues.Wrap(err, "populating paths")
 	}
 
-	logger.Ctx(ctx).Info("done populating container cache")
+	logger.Ctx(ctx).Infow(
+		"done populating container cache",
+		"duration", time.Since(start))
 
 	return el.Failure()
 }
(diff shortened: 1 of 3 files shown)


src/internal/m365/collection/exchange/contacts_container_cache.go
```diff
@@ -104,8 +107,10 @@ func (cfc *contactContainerCache) Populate(
 		cfc.userID,
 		baseID,
 		false)
+	ctx = clues.Add(ctx, "num_enumerated_containers", len(containers))
```
ryanfkeepers, line 110: minor: I'd love to see this go into a `count.Bus`, but I don't know how much investment that would take.  nbd.

src/internal/m365/collection/exchange/contacts_container_cache.go
```diff
@@ -93,8 +94,10 @@ func (cfc *contactContainerCache) Populate(
 	baseID string,
 	baseContainerPath ...string,
 ) error {
+	logger.Ctx(ctx).Info("populating container cache")
+
 	if err := cfc.init(ctx, baseID, baseContainerPath); err != nil {
-		return clues.Wrap(err, "initializing")
+		return clues.WrapWC(ctx, err, "initializing")
```
ryanfkeepers, line 100: `init` (and, below, `enumerateContainers` and `populatePaths`) all take in Ctx.  Do they not attach the context to the error already?

src/internal/m365/collection/exchange/contacts_container_cache.go
```diff
@@ -124,8 +129,10 @@ func (cfc *contactContainerCache) Populate(
 	}
 
 	if err := cfc.populatePaths(ctx, errs); err != nil {
-		return clues.Wrap(err, "populating paths")
+		return clues.WrapWC(ctx, err, "populating paths")
 	}
 
+	logger.Ctx(ctx).Info("done populating container cache")
```
ryanfkeepers, line 135: Suggestion, incase we're looking for runtime heuristics:
```suggestion
        // top of func
        start := time.Now()
        // here
	logger.Ctx(ctx).With("duration", time.Now().Sub(start)).Info("done populating container cache")
```

ryanfkeepers (approved):
approving to unblock.  I think the fooWC() calls might be unnecessary.

src/internal/m365/collection/exchange/contacts_container_cache.go
ashmrtn: I think that is a good goal to have. I think it would be best done in a separate PR though since that's not directly related to this logging and the bus isn't present in the resolver already. Mostly I don't want this PR to snowball into something larger since it should be safe to merge by itself

src/internal/m365/collection/exchange/contacts_container_cache.go
ashmrtn: hmmm seems like they do

Issue #4991, opened 2024-01-08
Handle group mailbox items with very large number of attachments


pandeyabs:
### What happened?

Group mailbox(aka conversations) posts support attachments, just like exchange. There are 2 ways to get attachments for a post. 

1.  via attachment endpoint -  `GET conversations/conv-id/threads/thread-id/posts/post-id/attachments?$expand=microsoft.graph.itemattachment/item` - This requires delegated access. Similarly, getting attachment list requires delegated token too i.e. `GET conversations/conv-id/threads/thread-id/posts/post-id/attachments?$select=id`
2. via post endpoint -  `GET conversations/conv-id/threads/thread-id/posts/post-id?$expand=attachments` - This works with application token.

We can only use 2) in corso right now. There is a need for 1) in a special case. We've noticed in the past with exchange that if an email has too many attachments, it can fail with 503s. See COR-56 and [code](https://github.com/alcionai/corso/blob/584fc21d36f63f48c6bbf4530c972253ca219129/src/pkg/services/m365/api/mail.go#L293). In that case, we fallback to a) getting list of attachments with IDs, b) Downloading attachments 1 at a time. 

Fallback is not possible for group mailbox since we don't have delegated access. So if we run into the "too many attachments" failure with group mailbox, we currently don't have a way to handle it, other than failing item backup or skipping the attachments.

Creating an issue to track this.

### Corso Version?

Corso v0.18.0

### Where are you running Corso?

NA

PR #4992, opened 2024-01-09
Download attachments for group mailbox posts

pandeyabs:
* Attachment download needed an expand operation. Added that. 
* Added test coverage with gock. Tested with manual backup of posts which contain attachments(embedded/non-embedded). We can't add e2e tests with attachments, since API to create new conversations requires delegated access.
* Note that https://github.com/alcionai/corso/issues/4991 is still an open item. We don't have a resolution for it right now, since attachment endpoint requires delegated token. Defaulting to let the backup fail for "too many attachments" error case. We don't know yet if we'd see that with group mailboxes, and whether it'd be the same error that we saw with exchange (recurring 503s).
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/conversations.go
+++ b/src/pkg/services/m365/api/conversations.go
@@ -11,6 +11,7 @@ import (
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/common/str"
 	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/services/m365/api/graph"
 )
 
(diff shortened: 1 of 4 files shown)


vkamra:
@pandeyabs - let's add a note clarifying the known issue here with large number of attachments,

src/pkg/services/m365/api/conversations.go
```diff
@@ -63,7 +64,38 @@ func (c Conversations) GetConversationPost(
 		return nil, nil, graph.Stack(ctx, err)
 	}
 
-	return post, conversationPostInfo(post), graph.Stack(ctx, err).OrNil()
+	preview, contentLen, err := getConversationPostContentPreview(post)
+	if err != nil {
+		preview = "malformed or unparseable html" + preview
```
ryanfkeepers, line 69:
```suggestion
		preview = "malformed or unparseable content body: " + preview
```

ryanfkeepers (approved):
just nits; looks good!

pandeyabs:
Thanks @vkamra, good call! I'll add a known issue. Thanks Keepers for the review!

PR #4993, opened 2024-01-09
Add retries for transient graph 400 "invalidRequest" errors

pandeyabs:
* Adding a retry for 400 "invalidRequest" errors. Restricting it to GET requests to limit the scope of the fix.
* Retrying at adapter wrapper `Send` level instead of doing it in retry middleware. If we do it in middleware, we'll have to resort to a string search of "invalidRequest" in the http response, which is not the best way. kiota-http nicely parses the http response into an odata error after the request has finished. So doing retry after that in adapter wrapper `Send` is better as we have better guarantees on it being an odata error type. 
* Also added a small test optimization to reduce test run time.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/services/m365/api/graph/errors.go
```diff
@@ -557,6 +563,16 @@ func parseODataErr(err error) oDataErr {
...
 }
 
+// hasResponseCode checks if the error is an ODataError and carries the provided
+// http response code.
+func (ode oDataErr) hasResponseCode(err error, httpResponseCode int) bool {
+	if !ode.isODataErr {
+		return false
+	}
+
+	return ode.Resp.StatusCode == httpResponseCode
```
ryanfkeepers, line 573: nit: could be simplified
```suggestion
	return ode.isODataErr && ode.Resp.StatusCode == httpResponseCode
```

src/pkg/services/m365/api/graph/service.go
ryanfkeepers, line 1: question: have we ever seen this response when using the `http_wrapper`?  If so, this case will need to be duplicated in that file's retry handler as well.

src/pkg/services/m365/api/graph/service_test.go
```diff
@@ -55,6 +59,15 @@ func (suite *GraphIntgSuite) SetupSuite() {
 	suite.credentials = m365
 
 	InitializeConcurrencyLimiter(ctx, false, 0)
+
+	// We are going to validate retry attempts for a few errors as part of this
+	// test suite. Set retry delay to a low value to reduce suite runtime.
+	adapterRetryDelay = 10 * time.Millisecond
```
ryanfkeepers, line 65: good thinking

src/pkg/services/m365/api/graph/testdata/errors.go
```diff
@@ -57,6 +57,18 @@ func ODataInner(innerCode string) *odataerrors.ODataError {
 	return odErr
 }
 
+func ODataErrWithAPIResponse(
```
ryanfkeepers, line 60: 🙏

src/pkg/services/m365/api/graph/service_test.go
```diff
@@ -307,3 +320,102 @@ func (suite *GraphIntgSuite) TestAdapterWrap_retriesBadJWTToken() {
...
+				_, err = NewService(adpt).Client().Users().Post(ctx, u, &cfg)
+				return err
+			},
+			expectedRetries: 1,
+		},
+	}
+
+	for _, test := range table {
+		suite.Run(test.name, func() {
+			tt := suite.T()
```
ryanfkeepers, line 410: You don't have to worry about clobbering here, since you're creating a new var.
```suggestion
			t := suite.T()
```

ryanfkeepers (approved):

src/pkg/services/m365/api/graph/service.go
```diff
@@ -407,11 +411,18 @@ func (aw *adapterWrap) Send(
...
+			// for now as it's a transient issue. Restrict retries to GET requests only
+			// to limit the scope of this fix.
+			logger.Ctx(ictx).Debug("invalid request")
+			events.Inc(events.APICall, "invalidrequest")
 		default:
 			return nil, clues.StackWC(ictx, err).WithTrace(1)
 		}
 
-		time.Sleep(3 * time.Second)
+		time.Sleep(adapterRetryDelay)
```
ashmrtn, line 425: could we actually make this a field on the adapterWrapper struct? That would allow for changing it during tests (tests can modify the instance of the struct prior to running tests), allow having a const for it, and ensure isolation between tests by ensuring changes from one test don't affect anything else

src/pkg/services/m365/api/graph/service.go
pandeyabs: I haven't seen this issue yet with `http_wrapper`. We can add it if we start seeing them.

src/pkg/services/m365/api/graph/service.go
pandeyabs: Great point! Let me switch to this pattern.

src/pkg/services/m365/api/graph/service_test.go
pandeyabs: We have `t` defined above at L329. Didn't want to shadow that var name.

PR #4996, opened 2024-01-09
Cancelled and modified events in ics export

meain:
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/converters/ics/ics.go
+++ b/src/internal/converters/ics/ics.go
@@ -505,6 +505,7 @@ func updateEventProperties(ctx context.Context, event models.Eventable, iCalEven
 
 			startStr := splits[len(splits)-1]
 
+			// the data just contains date and no time which seems to work
 			start, err := getUTCTime(startStr, ptr.Val(startTimezone))
 			if err != nil {
 				return clues.Wrap(err, "parsing cancelled event date")
--- a/src/internal/converters/ics/ics.go
+++ b/src/internal/converters/ics/ics.go
@@ -3,6 +3,7 @@ package ics
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
@@ -33,7 +34,10 @@ import (
 // originalStartTimeZone, reminderMinutesBeforeStart, responseRequested,
 // responseStatus, sensitivity
 
-const iCalDateTimeFormat = "20060102T150405Z"
+const (
+	iCalDateTimeFormat = "20060102T150405Z"
+	iCalDateFormat     = "20060102"
+)
 
 func keyValues(key, value string) *ics.KeyValues {
 	return &ics.KeyValues{
@@ -194,7 +198,7 @@ func getRecurrencePattern(
 }
 
 func FromJSON(ctx context.Context, body []byte) (string, error) {
-	data, err := api.BytesToEventable(body)
+	event, err := api.BytesToEventable(body)
 	if err != nil {
 		return "", clues.Wrap(err, "converting to eventable")
 	}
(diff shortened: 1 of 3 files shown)


src/internal/converters/ics/ics.go
```diff
@@ -100,6 +106,7 @@ func getUTCTime(ts, tz string) (time.Time, error) {
 // https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10
 // https://learn.microsoft.com/en-us/graph/api/resources/patternedrecurrence?view=graph-rest-1.0
 // Ref: https://github.com/closeio/sync-engine/pull/381/files
+// FIXME: When we have daily repeating task the last one is not getting added (due to timezone differences)
```
meain, line 109: Let me take this up in a followup PR

src/internal/converters/ics/ics.go
```diff
@@ -429,19 +476,66 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
...
+		if err != nil {
+			return nil, clues.Wrap(err, "getting cancelled occurrence id")
+		}
+
+		splits := strings.Split(instance, ".")
+		if len(splits) < 2 {
+			return nil, clues.NewWC(ctx, "invalid cancelled occurrence id").With("id", instance)
+		}
+
+		startStr := splits[len(splits)-1]
```
ashmrtn, line 529: is there similar code for this in the restore path for events? If there is, could we pull that out into a helper function so we don't have to keep two instances of it the same?

src/internal/converters/ics/ics_test.go
```diff
@@ -1079,3 +1079,152 @@ func (suite *ICSUnitSuite) TestAttachments() {
...
+		suite.Run(tt.name, func() {
+			ctx, flush := tester.NewContext(suite.T())
+			defer flush()
+
+			// convert event to bytes
+			writer := kjson.NewJsonSerializationWriter()
+			defer writer.Close()
+
+			err := writer.WriteObjectValue("", tt.event())
+			require.NoError(suite.T(), err, "serializing contact")
```
ashmrtn, line 1219: nit
```suggestion
			require.NoError(suite.T(), err, "serializing event")
```

src/internal/converters/ics/ics_test.go
```diff
@@ -1079,3 +1079,152 @@ func (suite *ICSUnitSuite) TestAttachments() {
...
+}
+
+func (suite *ICSUnitSuite) TestEventExceptions() {
+	table := []struct {
+		name  string
+		event func() *models.Event
+		check func(string)
+	}{
+		{
+			name: "single exception",
```
ashmrtn, line 1145: could you please also add a test with > 1 exception? That will act as a regression test to make sure we don't accidentally lose the ability to handle multiple exceptions in the future

src/internal/converters/ics/ics_test.go
```diff
@@ -1079,3 +1079,152 @@ func (suite *ICSUnitSuite) TestAttachments() {
...
+
+			e := baseEvent()
+
+			e.SetIsCancelled(ptr.To(true))
+			e.SetAdditionalData(map[string]any{
+				"cancelledOccurrences": tt.cancelledIds,
+			})
+
+			err := writer.WriteObjectValue("", e)
+			require.NoError(t, err, "serializing contact")
```
ashmrtn, line 1125: nit

```suggestion
			require.NoError(t, err, "serializing event")
```

ashmrtn (approved):
Generally looks good, would like a bit of clarification on the expected behavior when adding event exceptions (and how the original acts) as well as some code consolidation with the restore path if possible (could be done in a separate PR)

src/internal/converters/ics/ics.go
```diff
@@ -33,6 +34,11 @@ import (
 // originalStartTimeZone, reminderMinutesBeforeStart, responseRequested,
 // responseStatus, sensitivity
 
+const (
+	iCalDateTimeFormat = "20060102T150405Z"
+	iCalDateFormat     = "20060102"
+)
+
```
ryanfkeepers, line 41: can you move this to the `dttm` package, please?

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
 
-	created := data.GetCreatedDateTime()
+	err = updateEventProperties(ctx, event, iCalEvent)
+	if err != nil {
+		return "", clues.Wrap(err, "updating event properties")
+	}
+
+	exceptionOcurrances := event.GetAdditionalData()["exceptionOccurrences"]
+	if exceptionOcurrances != nil {
+		for _, occ := range exceptionOcurrances.([]any) {
```
ryanfkeepers, line 220: less nesting this way:
```suggestion
	if exceptionOcurrances == nil {
	  return cal.Serialize(), nil
	}
	
	for _, occ := range exceptionOcurrances.([]any) {
```

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
+			instance, ok := occ.(map[string]any)
+			if !ok {
+				return "", clues.New("converting exception instance to map[string]any").
+					With("type", fmt.Sprintf("%T", instance))
```
ryanfkeepers, line 224:
```suggestion
					With("interface_type", fmt.Sprintf("%T", instance))
```

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
+			}
+
+			exBody, err := json.Marshal(instance)
+			if err != nil {
+				return "", clues.Wrap(err, "marshalling exception instance")
```
ryanfkeepers, line 229: WrapWC is needed since json.Marshal doesn't handle Ctx addition.  This goes for some errors below as well.
```suggestion
				return "", clues.WrapWC(ctx, err, "marshalling exception instance")
```

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
+			}
+
+			exception, err := api.BytesToEventable(exBody)
+			if err != nil {
+				return "", clues.Wrap(err, "converting to eventable")
+			}
+
+			exICalEvent := cal.AddEvent(id)
+			start := exception.GetOriginalStart() // will always be in UTC
```
ryanfkeepers, line 238: > will always be in UTC

Which system guarantees that?  Is it safer to convert to utc anyway?

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
+
+			exICalEvent.AddProperty(ics.ComponentProperty(ics.PropertyRecurrenceId), start.Format(iCalDateTimeFormat))
+
+			err = updateEventProperties(ctx, exception, exICalEvent)
+			if err != nil {
+				return "", clues.WrapWC(ctx, err, "updating exception event properties")
```
ryanfkeepers, line 244: WrapWC shouldn't be neededhere, since the func takes in a ctx.
```suggestion
				return "", clues.Wrap(err, "updating exception event properties")
```

src/internal/converters/ics/ics.go
```diff
@@ -429,19 +476,66 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
...
+
+		splits := strings.Split(instance, ".")
+		if len(splits) < 2 {
+			return nil, clues.NewWC(ctx, "invalid cancelled occurrence id").With("id", instance)
+		}
+
+		startStr := splits[len(splits)-1]
+
+		// the data just contains date and no time which seems to work
+		start, err := getUTCTime(startStr, ptr.Val(event.GetStart().GetTimeZone()))
```
ryanfkeepers, line 532: not a big deal in this PR, but it would be nice to keep centralizing date and time parsers/formatters in `dttm`.

ryanfkeepers (approved):
lots of updates needed to clues wrapping, plus some other nits.  otherwise looks fine.

src/internal/converters/ics/ics.go
ashmrtn: do we really need it in that package? If it's not used outside of this package it seems like just adding a level of indirection to look things up and adding more to that namespace

src/internal/converters/ics/ics.go
```diff
@@ -192,108 +199,148 @@ func getRecurrencePattern(
...
+			if err != nil {
+				return "", clues.Wrap(err, "marshalling exception instance")
+			}
+
+			exception, err := api.BytesToEventable(exBody)
+			if err != nil {
+				return "", clues.Wrap(err, "converting to eventable")
+			}
+
+			exICalEvent := cal.AddEvent(id)
```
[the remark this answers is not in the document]
meain: We don't have to. Only cancellations need the original one to be updated.

src/internal/converters/ics/ics.go
meain: Yeah, these are specific to ics export. They will not be used outside of that.

src/internal/converters/ics/ics.go
meain: I've updated this and all other instances where we have context available to use `WC` versions of clues functions.

src/internal/converters/ics/ics.go
meain: This felt a bit specific to this usecase and not something we would use anywhere else.

PR #4997, opened 2024-01-09
Fix handling of recurring events with end date

meain:
Previously we were not accurately handling the end date of recurring
events. This was because we treated end date to be start of the day
instead of the end of day.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/converters/ics/ics.go
+++ b/src/internal/converters/ics/ics.go
@@ -106,7 +106,6 @@ func getUTCTime(ts, tz string) (time.Time, error) {
 // https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10
 // https://learn.microsoft.com/en-us/graph/api/resources/patternedrecurrence?view=graph-rest-1.0
 // Ref: https://github.com/closeio/sync-engine/pull/381/files
-// FIXME: When we have daily repeating task the last one is not getting added (due to timezone differences)
 func getRecurrencePattern(
 	ctx context.Context,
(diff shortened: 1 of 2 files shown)


src/internal/converters/ics/ics.go
```diff
@@ -175,10 +174,19 @@ func getRecurrencePattern(
 		case models.ENDDATE_RECURRENCERANGETYPE:
 			end := rrange.GetEndDate()
 			if end != nil {
-				// NOTE: We convert just a date into date+time in a
-				// different timezone which will cause it to not be just
-				// a date anymore.
-				endTime, err := getUTCTime(end.String(), ptr.Val(rrange.GetRecurrenceTimeZone()))
+				parsedTime, err := dttm.ParseTime(end.String())
```
zackrossman, line 177: IIUC parsedTime the start time of the end date? Which is why we need to add 23 hours 59 minutes 59 seconds in the lines below?

zackrossman (approved):

src/internal/converters/ics/ics.go
meain: This is recurrence end date, which should be treated as the last minute of the day so that any events on that day is also counted.

PR #4999, opened 2024-01-09
select list by name while export and restore

HiteshRepo:
selects list by name while export and restore

- [x] :no_entry: No

#### Type of change

#### Issue(s)
#4754 
#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/cli/utils/sharepoint.go
```diff
@@ -148,7 +146,7 @@ func IncludeSharePointRestoreDataSelectors(ctx context.Context, opts SharePointO
 
 	lfp, lfn := len(opts.FolderPath), len(opts.FileName)
 	ls, lwu := len(opts.SiteID), len(opts.WebURL)
-	slp, sli := len(opts.ListFolder), len(opts.ListItem)
+	sl := len(opts.Lists)
```
pandeyabs, line 149: Semi-related to this PR but could you rename these variables to follow the same scheme as in groups (around line 182 above)? It'll make it less bug-prone.

src/cli/utils/sharepoint.go
```diff
@@ -179,21 +177,10 @@ func IncludeSharePointRestoreDataSelectors(ctx context.Context, opts SharePointO
...
-
-		if len(containsFolders) > 0 {
-			sel.Include(sel.ListItems(containsFolders, opts.ListItem))
-		}
-
-		if len(prefixFolders) > 0 {
-			sel.Include(sel.ListItems(prefixFolders, opts.ListItem, selectors.PrefixMatch()))
-		}
+	if sl > 0 {
+		opts.Lists = trimFolderSlash(opts.Lists)
```
pandeyabs, line 181: This could be moved to a common helper which can be used by both groups and sharepoint.

src/pkg/selectors/sharepoint.go
```diff
@@ -640,7 +646,7 @@ func (s SharePointScope) set(cat sharePointCategory, v []string, opts ...option)
 	os := []option{}
 
 	switch cat {
-	case SharePointLibraryFolder, SharePointList, SharePointPage:
+	case SharePointLibraryFolder, SharePointPage:
```
pandeyabs, line 649: Could you please explain why list was removed from here?

HiteshRepo: @pandeyabs 
for lists, 
- there is no nested folders -> there cannot be lists within other lists
- list itself is the item -> so container and item are the same

so, since there is no path involved here, thought to remove it from here as we do not need any `path` filters. Plus wanted to enforce `StrictEqual` check to select lists provided using `--list` flag.

src/cli/utils/sharepoint.go
HiteshRepo: @pandeyabs 
we connot generalise this I guess (w/o using generics):
- `opts` type is different for groups (`GroupsOpts`) and sharepoint (`SharePointOpts`)
- `sel` type is different for groups (`GroupsRestore`) and sharepoint (`SharePointRestore`)

src/pkg/selectors/sharepoint.go
pandeyabs: Thanks for clarifying! Could you please add this note in this function for future reference?

src/cli/utils/sharepoint.go
pandeyabs: Thanks! Doesn't seem worthwhile generalizing then.

src/pkg/selectors/sharepoint.go
HiteshRepo: @pandeyabs Added

src/cli/utils/sharepoint.go
HiteshRepo: @pandeyabs Had to make them common along with validation of common time flags as well. As I was running into SonarQube duolication fails.

HiteshRepo: yeah, but SonarQube does not agree :)

HiteshRepo: is there any way to not let SonarQube fail for this duplication?

pandeyabs: It's okay to ignore sonar failure here. We try not to duplicate prod code but sometimes it can happen. This looks fine to me.

pandeyabs (approved):
🚀

PR #5004, opened 2024-01-10
Few more fixes to ics export

meain:
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/converters/ics/ics.go
+++ b/src/internal/converters/ics/ics.go
@@ -448,6 +448,12 @@ func updateEventProperties(ctx context.Context, event models.Eventable, iCalEven
 		iCalEvent.SetLocation(location)
 	}
 
+	// X-MICROSOFT-LOCATIONDISPLAYNAME (Outlook seems to use this)
+	locationDisplayName := ptr.Val(event.GetLocation().GetDisplayName())
+	if len(locationDisplayName) > 0 {
+		iCalEvent.AddProperty("X-MICROSOFT-LOCATIONDISPLAYNAME", locationDisplayName)
+	}
+
 	meeting := event.GetOnlineMeeting()
 	if meeting != nil {
 		url := ptr.Val(meeting.GetJoinUrl())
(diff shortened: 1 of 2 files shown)


src/internal/converters/ics/ics.go
```diff
@@ -258,18 +250,20 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
 }
 
 func updateEventProperties(ctx context.Context, event models.Eventable, iCalEvent *ics.VEvent) error {
+	// CREATED - https://www.rfc-editor.org/rfc/rfc5545#section-3.8.7.1
```
ryanfkeepers, line 253: I really appreciate these references.  Thanks!

src/internal/converters/ics/ics.go
```diff
@@ -443,12 +438,50 @@ func updateEventProperties(ctx context.Context, event models.Eventable, iCalEven
...
 	if len(location) > 0 {
 		iCalEvent.SetLocation(location)
 	}
 
-	// TODO Handle different attachment type (file, item and reference)
+	// X-MICROSOFT-LOCATIONDISPLAYNAME (Outlook seems to use this)
+	loc := event.GetLocation()
+	if loc != nil {
+		locationDisplayName := ptr.Val(event.GetLocation().GetDisplayName())
+		if len(locationDisplayName) > 0 {
```
ryanfkeepers, line 451: aside: i feel like we should modify `ptr.ValOK` to return OK only if the value was non-nil and non-zero.  We never really use it for non-nil only checks, but we have a lot of non-zero checks.  Thoughts?

ryanfkeepers (approved):

src/internal/converters/ics/ics.go
meain: Hmm, interesting thought. I personally almost always have used `Val` and not much of `ValOK`. That said, I'm not sure if adding an empty check will help much as empty can also be technically "valid".

PR #5005, opened 2024-01-10
adds error for list name exists

HiteshRepo:
adds error for list name exists

- [x] :no_entry: No

#### Type of change

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/errors.go
+++ b/src/pkg/services/m365/api/graph/errors.go
@@ -243,6 +243,11 @@ func IsErrItemAlreadyExistsConflict(err error) bool {
 		parseODataErr(err).hasErrorCode(err, nameAlreadyExists)
 }
 
+func IsErrNameAlreadyExistsConflict(err error) bool {
+	return errors.Is(err, ErrNameAlreadyExistsConflict) ||
+		parseODataErr(err).hasErrorCode(err, nameAlreadyExists)
+}
+
 // LabelStatus transforms the provided statusCode into
 // a standard label that can be attached to a clues error
 // and later reviewed when checking error statuses.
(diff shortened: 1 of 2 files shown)


src/pkg/services/m365/api/graph/errors.go
```diff
@@ -241,6 +243,11 @@ func IsErrItemAlreadyExistsConflict(err error) bool {
 		parseODataErr(err).hasErrorCode(err, nameAlreadyExists)
 }
 
+func IsErrNameAlreadyExistsConflict(err error) bool {
```
pandeyabs, line 246: Can we just use `IsErrItemAlreadyExistsConflict`? Why do we need this new check? Also, I don't see `ErrNameAlreadyExistsConflict` sentinel being used anywhere. Are you planning on using it in a future PR?

HiteshRepo: yes @pandeyabs, will use this for `In-place` restore when there is a `list name` collision

pandeyabs: Thanks! Why not use existing `IsErrItemAlreadyExistsConflict` function? It's also checking against `nameAlreadyExists` odata error. You can reuse the ` graph.ErrItemAlreadyExistsConflict` sentinel as well like the restore code for other services. Let me know if you disagree.

HiteshRepo: yeah I can use `IsErrItemAlreadyExistsConflict` itself. Closing this PR

PR #5009, opened 2024-01-10
get lists with collision key

HiteshRepo:
gets lists with collision key

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/api/lists.go
```diff
@@ -579,3 +579,15 @@ func ListToSPInfo(lst models.Listable) *details.SharePointInfo {
...
+func listCollisionKeyProps() []string {
+	return idAnd("displayName")
+}
+
+func ListCollisionKey(list models.Listable) string {
+	if list == nil {
+		return ""
+	}
+
+	return ptr.Val(list.GetDisplayName())
```
pandeyabs, line 592: Is this a unique enough collision key? Wondering if we have converged on this as the collision key during design discussions.

HiteshRepo: for lists,
- we cannot create 2 lists with same name --> hence going by the name itself as the collision key
- also we cannot set `name` attrbute of the list, we can only set `displayName`. `name ` is set based on display name.

pandeyabs: Thanks for clarifying! Could you please leave this note on top of this function?

src/pkg/services/m365/api/lists_pager.go
```diff
@@ -77,6 +78,26 @@ func (c Lists) GetLists(
...
 
+func (c Lists) GetListsByCollisionKey(
+	ctx context.Context,
+	siteID string,
+) (map[string]string, error) {
+	pager := c.NewListsPager(siteID, CallConfig{Select: listCollisionKeyProps()})
+
+	items, err := pagers.BatchEnumerateItems[models.Listable](ctx, pager)
+	if err != nil {
+		return nil, graph.Wrap(ctx, err, "enumerating lists")
```
pandeyabs, line 89:
```suggestion
		return nil, clues.WrapWC(ctx, err, "enumerating lists")
```

We are moving away from `graph.Wrap`, see https://github.com/alcionai/corso/pull/5018/files

pandeyabs (approved):

PR #4884, opened 2023-12-19
Additional backup config cli tests

ashmrtn:
Further derisk moving to a different way of passing backup config values by adding tests ensuring the default (i.e. no flag given) values of the new backup config and the old backup config are the same.

While this code is temporary and can be removed once the switch is made we should probably create tests checking default values at some point so we don't accidentally change them without meaning to

---

#### Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [x] :robot: Supportability/Tests
- [ ] :computer: CI/Deployment
- [ ] :broom: Tech Debt/Cleanup

#### Issues

- #4896

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #4861, opened 2023-12-14
Update backup op API to take backup options struct

ashmrtn:
Update the existing backup operation API to take a the new backup
options struct instead of passing in the repo options internally.
This effectively separates init/connect config from backup config

This PR doesn't update lower portions of the stack to use backup
ops, it simply translates the incoming config values to the old
options struct and continues using that. Later PRs will update the
lower portion of the stack

Since this PR has many small changes to update tests it may be
easiest to review by commit

This PR does break the current API that SDK users may rely on so
SDK consumers will need to make changes to use the new options
struct

---

#### Does this PR need a docs update or release note?

- [ ] :white_check_mark: Yes, it's included
- [ ] :clock1: Yes, but in a later PR
- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix
- [ ] :world_map: Documentation
- [ ] :robot: Supportability/Tests
- [ ] :computer: CI/Deployment
- [x] :broom: Tech Debt/Cleanup

#### Issues

- #4896

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #5010, opened 2024-01-10
Pass some backup specific options in a separate struct

ashmrtn:
Going with this alternative to #4861 and #4884. While those are roughly
where we'd like to end up eventually, this is much easier to get merged
and provides a better developer experience because config values are only
defined in one exported struct

This moves the incremental and backup preview config values to a new
control struct and passes that struct down through the stack with the
original control struct in backup operations. Overall, it allows passing
those config values when the backup operation is created instead of when
the repo is connected to

Manually verified that disable incrementals flags are passed via CLI as
expected

May be easier to review by commit

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/cli/backup/backup.go
+++ b/src/cli/backup/backup.go
@@ -176,6 +176,7 @@ func genericCreateCommand(
 	serviceName string,
 	selectorSet []selectors.Selector,
 	ins idname.Cacher,
+	backupConfig control.BackupConfig,
 ) error {
 	var (
 		bIDs []string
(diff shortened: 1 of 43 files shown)


ashmrtn:
closing as there's some larger design things that should be addressed first

PR #5016, opened 2024-01-11
minor logging and cli output tweaks

ryanfkeepers:
just touchups for supportability and tracking, plus a ux tweak here and there.  No real logic changes.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/observe/observe.go
+++ b/src/internal/observe/observe.go
@@ -22,7 +22,7 @@ import (
 const (
 	hideProgressBarsFN   = "hide-progress"
 	retainProgressBarsFN = "retain-progress"
-	progressBarWidth     = 40
+	progressBarWidth     = 32
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


src/internal/observe/observe.go
```diff
@@ -117,6 +117,10 @@ func (o *observer) resetWriter(ctx context.Context) {
 	// 	opts = append(opts, mpb.WithAutoRefresh())
 	// }
 
+	if o.mp != nil {
+		logger.Ctx(ctx).Infow("resetting progress bars", "bar_count", o.mp.BarCount())
+	}
+
 	o.mp = mpb.NewWithContext(ctx, opts...)
```
HiteshRepo, line 124: nit: should we move `o.mp = mpb.NewWithContext(ctx, opts...)` into `if o.mp != nil`  block? 

or is the intention to override `o.mp` at line#124?

HiteshRepo (approved):

src/internal/observe/observe.go
ryanfkeepers: Good question: the intention is always to override.

PR #5018, opened 2024-01-11
replace graph Wrap and Stack with clues

ryanfkeepers:
Now that graph errors are always transformed as part of the graph client wrapper or http_wrapper, we don't need to call graph.Stack or graph.Wrap outside of those inner helpers any longer. This PR swaps all those graph calls with equivalent clues funcs as a cleanup.

Should not contain any logical changes.

---

- [x] :no_entry: No

#### Type of change

#### Issue(s)

* #4685 

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/service/exchange/enabled_test.go
```diff
@@ -70,7 +70,7 @@ func (suite *EnabledUnitSuite) TestIsServiceEnabled() {
 				odErr := graphTD.ODataErrWithMsg(string(graph.ResourceNotFound), "message")
 
 				return mockGMB{
-					mailboxErr: graph.Stack(ctx, odErr),
+					mailboxErr: clues.Stack(odErr),
```
ashmrtn, line 73: I realize these are tests, but should it be `clues.StackWC` instead? Other tests seem to be using that

src/internal/m365/service/sharepoint/api/pages.go
```diff
@@ -68,9 +67,13 @@ func GetSitePages(
...
 
-			page, err = serv.Client().SitesById(siteID).PagesById(pageID).Get(ctx, opts)
+			page, err = serv.
+				Client().
+				SitesById(siteID).
+				PagesById(pageID).
+				Get(ctx, opts)
 			if err != nil {
-				el.AddRecoverable(ctx, graph.Wrap(ctx, err, "fetching page"))
+				el.AddRecoverable(ctx, clues.Wrap(err, "fetching page"))
```
ashmrtn, line 76: Could you please remind me what the expectation is here for clues injection? Are we expecting the middleware layers to add that? If not then pretty much all of the calls like this in the files in this PR will need to add clues themselves

src/internal/m365/service/sharepoint/enabled_test.go
```diff
@@ -58,7 +58,6 @@ func (suite *EnabledUnitSuite) TestIsServiceEnabled() {
 			name: "no sharepoint license",
 			mock: func(ctx context.Context) getSiteRooter {
 				odErr := graphTD.ODataErrWithMsg("code", string(graph.NoSPLicense))
-
 				return mockGSR{nil, graph.Stack(ctx, odErr)}
```
ashmrtn, line 61: should this be using `clues.StackWC(ctx, odErr)` along with the below?

src/pkg/services/m365/api/access.go
```diff
@@ -51,7 +49,7 @@ func (c Access) GetToken(
 
 	resp, err := c.Post(ctx, rawURL, headers, body)
 	if err != nil {
-		return graph.Stack(ctx, err)
+		return clues.Stack(err)
```
ashmrtn, line 52: this may also need to add clues if the middleware doesn't handle it

ashmrtn (approved):

src/internal/m365/service/exchange/enabled_test.go
ryanfkeepers: non-critical, but I can change it.

src/pkg/services/m365/api/access.go
ryanfkeepers: Clients could be expected to add clues themselves in these cases.

src/internal/m365/service/sharepoint/api/pages.go
ryanfkeepers: It's handled with the graph service client itself.  So long as the creation of the client uses our standard adapter creation func (it may not here, but pages aren't supported either), we're good.

src/pkg/services/m365/api/access.go
ashmrtn: what's the rule for adding clues then? So far it seems to be "if the caller takes ctx and is a corso function assume it adds clues." This makes it seem like every dev is expected to read the called function to figure out if they need to add clues or not?

ryanfkeepers: > This makes it seem like every dev is expected to read the called function to figure out if they need to add clues or not?

I prefer to defer that to the writer, not the reader.  As in, every dev is expected to add clues if the func/internal package they're adding accepts a ctx, and the func they call does not.  If we uphold that, implementers don't need to go looking at the next func.  They can evaluate purely based on whether they're calling an internal or external package/client.

Issue #3748, opened 2023-07-03
Backup List should display and/or filter with backup data categories


ryanfkeepers:
## Issue

For services with multiple data categories (such as Exchange), each backup isolates its categories to the data that was present in the backup.  For example:
```
create backup exchange                                # => b_1
create backup exchange --data contacts    # => b_2
```
In this example, `b_1` contains all categories of data, while `b_2` only contains contacts.  This restriction is opaque to the end user.  It's only discoverable through a a `backup details` call investigating the contents of each backup.  And even then, only indirectly, since the user would have to read the contents of the backup or do a categorical selection.

Exposing the data categories within the backup list would provide much better transparency and help users select appropriate backup ids for details and restoration.

## Acceptance criteria
* Display data categories within a backup in `backup list` entries
* Allow users to list only backups containing a certain data category with the `--data` flag.

ashmrtn:
this goes along well with the idea that we should be tagging backups with the `Reason`s they correspond to

PR #5023, opened 2024-01-12
show data category in backup list

ryanfkeepers:
Adds the backup's data category to the cli output.

before:
```
  ID                                    Started At            Duration    Status     Resource Owner
  09e8d1e3-92d4-426d-9ca1-3cd063ce8bb9  2024-01-12T17:54:33Z  16.103569s  Completed  rfinders@10rqc2.onmicrosoft.com
  893ef934-9403-4775-b667-09f77cd26e83  2024-01-12T17:55:22Z  5.66253s    Completed  rfinders@10rqc2.onmicrosoft.com
```

after:
```
  ID                                    Started at            Duration    Status     Protected resource               Data
  09e8d1e3-92d4-426d-9ca1-3cd063ce8bb9  2024-01-12T17:54:33Z  16.103569s  Completed  rfinders@10rqc2.onmicrosoft.com  Contacts,Emails,Events
  893ef934-9403-4775-b667-09f77cd26e83  2024-01-12T17:55:22Z  5.66253s    Completed  rfinders@10rqc2.onmicrosoft.com  Emails
```

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #3748

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test


ryanfkeepers:
@ashmrtn I added you to this primarily to sanity check me on this being the best way to get Reasons for each backup.

ashmrtn:
> sanity check me on this being the best way to get Reasons for each backup

I think for now yes, it's probably the simplest way to get them. Going forward though I think we're going to want to change that. Unfortunately selectors don't serialize too well (space-wise) and our backup models are consistently > 1KB with most of that being the serialized selector. If we can remove that we can save some memory when loading manifests which could come in handy for larger repos

Before we remove it though we can add a `Reasons` field to the backup model itself, kind of like what was in the backup linear PoC (I don't think I really settled on a serialization format I was happy with, hence why it wasn't merged directly). Removing the selector may also require getting the protected resource name for the backup in a different manner, but we can work that out in another ticket

ashmrtn (approved):
LGTM, this isn't going to cause trouble with sanity tests or anything right? IIRC they grep the output of some commands

ryanfkeepers:
> this isn't going to cause trouble with sanity tests or anything right?

Shouldn't be an issue.

PR #5024, opened 2024-01-12
modifies list restore handler and handles lists advanced restore

HiteshRepo:
- includes GetList in list restore handler
- includes GetListsByCollisionKey in list restore handler
- updates mock list restore handler

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/collection/site/mock/list.go
```diff
@@ -94,21 +95,72 @@ func (lh *ListHandler) Check(t *testing.T, expected []string) {
 }
 
 type ListRestoreHandler struct {
-	List models.Listable
-	Err  error
+	ignoreSubsequentPostListFails bool
```
ryanfkeepers, line 98: this feels a little cumbersome.  something we've done in other handler mocks is to have props like `idx int, errs []error`.  Then if the test expects two calls using the mock, they can add `errs: [err, nil]` (or just [err], and let out of bounds always return nil).  That tends to be a bit more straight forward for handling.

ryanfkeepers (approved):

HiteshRepo:
All individual PRs have been approved. Apologies for a big PR merge but all changes are `Sharepoint Lists` specifc

PR #5026, opened 2024-01-12
handles lists restore collision

HiteshRepo:
handles lists restore collisions

for collection/site/restore_test, along with tests for lists restore collision, changes for the refactoring of existing test setup also got mingled.

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)
#4754 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


ryanfkeepers (commented):
I'm having a difficult time figuring out which code was moved, deleted, or simply changed.  For example, /site/restore.go's `ConsumeRestoreCollections` is deleted, but I can't tell if that was unused code or if it was rerouted elsewhere, or something else.  I'm going to hold on further review and ask that this gets split up into two PRs (one for movement/deletion, one for changes) to make it easier to review.

HiteshRepo:
> I'm having a difficult time figuring out which code was moved, deleted, or simply changed. For example, /site/restore.go's `ConsumeRestoreCollections` is deleted, but I can't tell if that was unused code or if it was rerouted elsewhere, or something else. I'm going to hold on further review and ask that this gets split up into two PRs (one for movement/deletion, one for changes) to make it easier to review.

@ryanfkeepers Please find the PR that removes unused/non-required code: https://github.com/alcionai/corso/pull/5043

src/internal/m365/collection/site/restore_test.go
```diff
@@ -187,21 +196,262 @@ func (suite *SharePointRestoreSuite) TestListCollection_Restore_invalidListTempl
...
+				nil,
+				count.New(),
 				fault.New(false))
 			require.Error(t, err)
 			assert.Contains(t, err.Error(), api.ErrSkippableListTemplate.Error())
 		})
 	}
 }
 
+func (suite *SharePointRestoreSuite) TestListCollection_RestoreInPlace_skip() {
```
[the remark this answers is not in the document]
HiteshRepo: @ryanfkeepers 
No they are not moved from `/sharepoint/restore_test.go`. The test was inconsequential as the core logic is tested in `collection` package.

src/internal/m365/collection/site/restore.go
```diff
@@ -54,25 +62,100 @@ func restoreListItem(
...
+		ctx,
+		ptr.Val(storedList.GetDisplayName()),
+		storedList,
+		errs)
+	if err == nil {
+		ctr.Inc(count.CollisionReplace)
+		return restoredList, nil
+	}
+
+	_, collidedListErr := rh.PostList(
```
ryanfkeepers, line 139: I'm confused by what's going on with the behavior here.  As a general statement, extra comments are always helpful for other maintainers so that we can understand what's going on in a process.  This is one of those places where a little explanation would go a long way.

So, what I figure is happening is that the func does this:

1. look up the current list
2. delete the current list
3. restore the backed-up list
4. if no error occurs when restoring, good, we exit.
5. if an error occurs when restoring, try to restore the list that we just deleted.

I think I have that right.  If so: that works, but it's not my favorite pattern, because it contains a lot more moving parts and points of failure.  I think you could slim it down to the following for a simpler flow and fewer points of failure.  At least, I think this is possible, assuming the newly created list has a different ID than the previous one.

1. Restore the backed-up list.  If an error occurs, return.
2. Delete the current list.  If an error occurs, return.

And if that works, we might not need this extra func at all.  Up on lines ~85-105 you could do this instead:
```go
 restoredList, err = rh.PostList(ctx, newName, storedList, errs)
 if err != nil {
 	return dii, clues.WrapWC(ctx, err, "restoring list")
 }

if collisionPolicy != control.Replace {
  rh.DeletetList(...)
}
```

src/internal/m365/collection/site/restore.go
```diff
@@ -54,25 +62,100 @@ func restoreListItem(
...
-		return dii, graph.Wrap(ctx, err, "restoring list")
+	if id, ok := collisionKeyToItemID[collisionKey]; ok {
+		log := logger.Ctx(ctx).With("collision_key", clues.Hide(collisionKey))
+		log.Debug("item collision")
+
+		if collisionPolicy == control.Skip {
+			ctr.Inc(count.CollisionSkip)
+			log.Debug("skipping item with collision")
+
+			return dii, core.ErrAlreadyExists
```
ryanfkeepers, line 80:
```suggestion
			return dii, clues.Stack(core.ErrAlreadyExists)
```

src/internal/m365/collection/site/restore_test.go
```diff
@@ -187,21 +196,256 @@ func (suite *SharePointRestoreSuite) TestListCollection_Restore_invalidListTempl
...
+		collisionKeyToItemID,
+		cl,
+		fault.New(true))
+	require.Error(t, err, clues.ToCore(err))
+	assert.True(t, errors.Is(core.ErrAlreadyExists, err))
+	assert.Empty(t, deets)
+	assert.Less(t, int64(0), cl.Get(count.CollisionSkip))
+}
+
+func (suite *SharePointRestoreSuite) TestListCollection_RestoreInPlace_copy() {
```
ryanfkeepers, line 254: various nits:

Naming structure doesn't follow standards.  It should be `Test<FunctionUnderTest>_<variation>`
```suggestion
func (suite *SharePointRestoreSuite) TestRestoreListItem_collisions() {
```

All four new tests duplicate a lot of boilerplate declarations and behavior.  This bloat makes the tests more difficult to read and, in the long term, will require more effort to maintain and refactor.  They should be condensed into a single test using a test table.

src/internal/m365/collection/site/restore_test.go
```diff
@@ -234,46 +478,40 @@ func deleteList(
...
-		ItemID: testName,
+		ItemID: listName,
 		Reader: r,
 	}
 
-	return testName, lrh, destName, mockData
+	return mockData
+}
+
+func createList(listTemplate, listDisplayName string) models.Listable {
```
ryanfkeepers, line 508: nit: prefer `stub` to suggest we're mocking out a list.  `create` could be confused with the process of adding it to an external dataset.
```suggestion
func stubList(listTemplate, listDisplayName string) models.Listable {
```

ryanfkeepers (changes requested):
The general behavior looks reasonable, but I'm going to hold off on approval because I think the implementation could use some hefty re-work.  There are a couple choices here with the potential for problems that we'd want to avoid.

src/internal/m365/collection/site/restore.go
```diff
@@ -54,25 +62,100 @@ func restoreListItem(
...
+) (models.Listable, error) {
+	collidedList, _, err := rh.GetList(ctx, collisionID)
+	if err != nil {
+		return nil, clues.WrapWC(ctx, err, "fetching collided list")
+	}
+
+	err = rh.DeleteList(ctx, collisionID)
+	if err != nil {
+		return nil, clues.WrapWC(ctx, err, "deleting collided list")
+	}
```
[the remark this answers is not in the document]
HiteshRepo: @ryanfkeepers 
for `replace` collision, we need to create list with same name. It is not possible to have 2 lists with same name. Hence following the below order:
- Fetch the collided list and store it temporarily
- Delete collided list
- Create stored list
- If Create stored list fails, reinstate the collided list

src/internal/m365/collection/site/restore.go
HiteshRepo: @ryanfkeepers 
you got it right. I would have gone for the pattern you mentioned as well, but in `replace` collision, we need to create backed-ip list with same name as current/collided list. But site does not allow to create two lists with same name.

src/internal/m365/collection/site/restore.go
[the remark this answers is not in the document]
ryanfkeepers: hmm, what do you think about, instead:
1. restore list with a name like `corso_restore_<date>_<originalName>
2. delete old list
3. patch restored list name to `originalName`

It's still fewer steps/calls, and avoids the potential for full data loss.

src/internal/m365/collection/site/restore.go
ryanfkeepers: Thanks for confirming!  I grok the need to work around it.  I pitched a third workflow option in the other comment; let me know what you think.

src/internal/m365/collection/site/restore.go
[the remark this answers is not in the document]
HiteshRepo: @ryanfkeepers 
This change request has been implemneted in below 3 PRs [approved]:
- https://github.com/alcionai/corso/pull/5072
- https://github.com/alcionai/corso/pull/5073
- https://github.com/alcionai/corso/pull/5074

src/internal/m365/collection/site/restore_test.go
HiteshRepo: @ryanfkeepers 
this change has been accommodated in https://github.com/alcionai/corso/pull/5074

HiteshRepo:
@ryanfkeepers moving ahead with merger of this, as change requests for this PR has been addressed and approved in other PRs:
- https://github.com/alcionai/corso/pull/5072
- https://github.com/alcionai/corso/pull/5073
- https://github.com/alcionai/corso/pull/5074

PR #5049, opened 2024-01-17
Export event data for invites

meain:
---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* https://github.com/alcionai/corso/issues/5039

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/converters/eml/eml.go
+++ b/src/internal/converters/eml/eml.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
 	"github.com/alcionai/corso/src/internal/common/str"
+	"github.com/alcionai/corso/src/internal/converters/ics"
 	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
 )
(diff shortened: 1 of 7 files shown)


src/internal/converters/eml/eml.go
```diff
@@ -50,6 +51,92 @@ func formatAddress(entry models.EmailAddressable) string {
 	return fmt.Sprintf(addressFormat, name, email)
 }
 
+// getICalData converts the emails to an event so that ical generation
+// can generate from it.
+func getICalData(ctx context.Context, data models.Messageable) (string, error) {
+	msg := data.(*models.EventMessageRequest)
+	event := models.NewEvent()
```
ashmrtn, line 58: nit: could you please move this after the if-block below just to keep things a little cleaner? As far as I can tell it's not used if we hit the if-block anyway

src/internal/converters/eml/eml.go
```diff
@@ -50,6 +51,92 @@ func formatAddress(entry models.EmailAddressable) string {
...
+	event.SetRecurrence(msg.GetRecurrence())
+	// event.SetIsCancelled()
+	event.SetSubject(msg.GetSubject())
+	event.SetBodyPreview(msg.GetBodyPreview())
+	event.SetBody(msg.GetBody())
+
+	// https://learn.microsoft.com/en-us/graph/api/resources/eventmessage?view=graph-rest-1.0
+	// In addition, Outlook automatically creates an event instance in
+	// the invitee's calendar, with the showAs property as tentative.
+	event.SetShowAs(ptr.To(models.TENTATIVE_FREEBUSYSTATUS))
```
ashmrtn, line 97: not sure if this is worth pursuing, but given the comment above is this going to set the creator of the event to tentative as well? Is that what would normally happen for the person who created the event?

src/internal/converters/eml/eml.go
```diff
@@ -50,6 +51,92 @@ func formatAddress(entry models.EmailAddressable) string {
...
+
+	event.SetCategories(msg.GetCategories())
+	event.SetWebLink(msg.GetWebLink())
+	event.SetOrganizer(msg.GetFrom())
+
+	// NOTE: If an event was previously created and we added people to
+	// it, the original list of attendee are not available.
```
ashmrtn, line 104: is this eventually going to translate into a known issues lines in the documentation?

src/internal/converters/eml/eml.go
```diff
@@ -60,10 +147,11 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
...
 	email := mail.NewMSG()
-	email.AllowDuplicateAddress = true // More "correct" conversion
-	email.AddBccToHeader = true        // Don't ignore Bcc
-	email.AllowEmptyAttachments = true // Don't error on empty attachments
-	email.UseProvidedAddress = true    // Don't try to parse the email address
+	email.Encoding = mail.EncodingBase64 // Doing it to be safe for when we have eventMessage (newline issues)
+	email.AllowDuplicateAddress = true   // More "correct" conversion
+	email.AddBccToHeader = true          // Don't ignore Bcc
+	email.AllowEmptyAttachments = true   // Don't error on empty attachments
+	email.UseProvidedAddress = true      // Don't try to parse the email address
```
ashmrtn, line 154: Just curious, does this fix the issue we had earlier where the email didn't have an `@` or something?

src/internal/converters/eml/eml.go
```diff
@@ -183,6 +271,21 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
 		}
 	}
 
+	// TODO "#microsoft.graph.calendarSharingMessage" (https://github.com/alcionai/corso/issues/5041)
+	otype := ptr.Val(data.GetOdataType())
+	if otype == "#microsoft.graph.eventMessageRequest" || // creation
+		otype == "#microsoft.graph.eventMessage" || // cancellation
+		otype == "#microsoft.graph.eventMessageResponse" { // accept
```
ashmrtn, line 278: the deserialization code is actually pretty generic and relies on graph API to properly interpret things. Could we actually use a type-switch here instead of manually checking the odata type? For example

```go
switch data.(type) {
case models.EventMessageRequestable, models.EventMessageResponseable, models.EventMessageable:
  cal, err := getICalData(ctx, data)
  ...
}
```

Since the switch (may) allow listing multiple types on the same line it could be more compact than multiple if-statements. If it doesn't then I don't have too much of a preference though the switch block may read a little cleaner? You'd need the 2-return value type assertion(?) in golang though (i.e. `msg, ok := data.(models.EventMessageable)`

I guess I'm asking cause I would slightly prefer getting a compile error if the types change rather than just silently skipping the if-block if for some reason the odata types we're checking for are no longer used (yay me being paranoid about things heh)

edit: after looking at the graph SDK more, maybe this can just be an interface type check like
```go
if _, ok := data.(models.EventMessageable) {
  cal, err := getICalData(ctx, data)
  ...
}
```

The different EventMessage types all seem to embed that one type and if we have tests that check the code works for all the types then we'd know if the SDK got updated in some way that broke our expectations

src/internal/converters/eml/eml_test.go
```diff
@@ -188,3 +191,137 @@ func (suite *EMLUnitSuite) TestConvert_edge_cases() {
 		})
 	}
 }
+
+func (suite *EMLUnitSuite) TestConvert_eml_ics() {
```
ashmrtn, line 195: both of the input data types in the test seem like they're for the `EventMessageRequestable` interface. Could you please make some basic tests for the other types so we have something to ensure we don't lose that functionality going forward?

ashmrtn (approved):
sorry for the delay! generally looks good, a few nits and would like to see basic test coverage for other types of EventMessages

src/internal/converters/eml/eml.go
```diff
@@ -50,6 +51,92 @@ func formatAddress(entry models.EmailAddressable) string {
...
+// getICalData converts the emails to an event so that ical generation
+// can generate from it.
+func getICalData(ctx context.Context, data models.Messageable) (string, error) {
+	msg := data.(*models.EventMessageRequest)
+	event := models.NewEvent()
+
+	// This method returns nil if data is not pulled using the necessary expand property
+	// .../messages/<message_id>/?expand=Microsoft.Graph.EventMessage/Event
+	// Also works for emails which are a result of someone accepting an invite
+	// If we add this field on a cancellation mail, the request fails.
```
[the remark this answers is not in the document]
meain: Hmm, I see how that can be confusing. It is about fetching emails which are regarding cancellations. Let me rephrase that.

src/internal/converters/eml/eml.go
```diff
@@ -50,6 +51,92 @@ func formatAddress(entry models.EmailAddressable) string {
 	return fmt.Sprintf(addressFormat, name, email)
 }
 
+// getICalData converts the emails to an event so that ical generation
+// can generate from it.
+func getICalData(ctx context.Context, data models.Messageable) (string, error) {
+	msg := data.(*models.EventMessageRequest)
```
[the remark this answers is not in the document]
meain: I've added a switch above which should handle this better.

```go
	switch data.(type) {
	case *models.EventMessageRequest:
		// do nothing, we'll continue below
	case *models.EventMessageResponse, *models.EventMessage:
		// we can't handle this, not enough information
		return "", nil
	default:
		return "", clues.NewWC(ctx, "unknown message type").
			With("interface_type", fmt.Sprintf("%T", data))
	}
```

src/internal/converters/eml/eml.go
meain: Yup, this was added upstream by us to handle that.

src/internal/converters/eml/eml.go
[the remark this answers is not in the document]
meain: Picking it up along with https://github.com/alcionai/corso/pull/5049#discussion_r1460083263

src/internal/converters/eml/eml_test.go
meain: We don't handle the others as the data is incomplete. Let me update the code to reflect that.

src/internal/converters/eml/eml.go
meain: The idea was to follow this up by fetching full event data, but it does not look like it is going to be possible. Tracking in https://github.com/alcionai/corso/issues/5039 . Will update as a known issue after spending a bit more time on it.

src/internal/converters/eml/eml.go
meain: Hmm, that is a tricky question. The only time it would cause an issue is if we export the "Sent Items" folder from the organizer -> convert to pst -> import them back. In this case, we would have incorrect status, but I'm not sure if it is worth worrying about. I don't know if exchange even parses out the emails "Sent Items" in imported pst files and create calendar events from them.


As far as the invitee, ie receiver is concerned, it is always "tentative" and the status of the owner is not affected by their personal cal.

PR #5053, opened 2024-01-17
Fix in-place contacts restore when restore destination is empty

ashmrtn:
Fix bug where we tried to create folders with an empty display name
when the user did a restore with `/` or `""` as the restore destination
for contacts

This PR is safe with the current flow for backing up contact folders
because nested folders aren't backed up at all (at least nesting beyond
the first level). It is unclear, though possible, that this patch will
continue to work if we backup nested contact folders

Also adds basic smoke tests for this issue for all exchange data types

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


ashmrtn:
highly recommend looking at the CHANGELOG diff for this commit to get an overview of behavior changes to the system

src/internal/m365/restore_test.go
```diff
@@ -0,0 +1,216 @@
...
+			[][]string{tconfig.M365AcctCredEnvs}),
+	})
+}
+
+// TestRestoreCollections_HandlesEmptyRestoreLocation checks to make sure that
+// even if the restore location is empty we fallback to using the collection
+// path as the folder, resulting in an in-place restore. It doesn't attempt to
+// retore any items because that would bloat the data set in the test user.
+func (suite *RestoreIntgSuite) TestRestoreCollections_HandlesEmptyRestoreLocation() {
+	t := suite.T()
```
ryanfkeepers, line 43: this isn't getting used, is it?  All the test properties are funcs that take in a testing.T.
```suggestion
```

src/internal/m365/restore_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	acct := tconfig.NewM365Account(t)
+
+	table := []struct {
+		service     path.ServiceType
+		category    path.CategoryType
+		selector    func(*testing.T) selectors.Selector
+		defaultPath func(*testing.T) path.Path
+		otherPath   func(t *testing.T, location string) path.Path
```
ryanfkeepers, line 52: nit: you could deduplicate the boilerplate by moving these funcs to adhoc func declarations instead of test params.  Have the caller pass in the category and folders (`buildPath(t, cat, dirs...)`) and it's otherwise the same across all uses.

ryanfkeepers (approved):
approving to unblock, since there's largely nits.  But I would like to see the suggestion about handler interface design go through, whether in this or a follow-up PR.

src/internal/m365/service/exchange/restore.go
```diff
@@ -77,19 +78,45 @@ func (h *exchangeHandler) ConsumeRestoreCollections(
...
+		// folder. Contacts is weird in that it allows creating sub folders with the
+		// same name as the default contacts folder, but which are technically
+		// nested folders.
+		//
+		// Without this check we'll end up restoring to Contacts/Contacts instead of
+		// Contacts if in-place restore is requested and the root Contacts folder
+		// had items.
+		if category == path.ContactsCategory &&
+			len(dc.FullPath().Folders()) == 1 &&
+			restoreFolderPath.String() == handler.DefaultRootContainer() {
```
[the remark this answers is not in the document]
ashmrtn: oh thanks for the suggestion. I wasn't really thinking about that though I also wasn't real happy with the current implementation

PR #5066, opened 2024-01-19
Skip base finding steps if we don't need bases

ashmrtn:
Minor optimization where we skip trying to find BackupBases if we know
we're going to end up dropping them anyway. This can save us from
loading all the kopia manifests only to discard the information we
pulled from them. Note that this optimization only kicks in for
backups that request no merge bases and no assist bases
("full backup")

Existing test coverage in `operations/manifests_test.go` ensures this
doesn't break base selection altogether

As this is an optimization we don't need to merge this PR if it seems
like it will make code maintenance more difficult going forward

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/operations/manifests.go
```diff
@@ -24,6 +24,17 @@ func produceManifestsAndMetadata(
 	tenantID string,
 	getMetadata, dropAssistBases bool,
 ) (kopia.BackupBases, []data.RestoreCollection, bool, error) {
+	// Just return early if we're going to end up dropping all the bases anyway.
```
pandeyabs, line 27: nice!

pandeyabs (approved):

Issue #5062, opened 2024-01-18
1:1 chats data categories


ryanfkeepers:
# Categories of Data

The chats are their own category in the chats service.  Chats are owned by users.  Therefore: for the chats service, each user resource contains zero or more chat entries to backup.

Delta is not supported.

## Packages

- [ ]  `/src/pkg/m365/chats/collection/chat.go` to handle category-specific processing. (it may be possible to use the `/collection/groups.go` framework here, since the design is so similar to channels.)
    
## Interface & Data Compliance

- [x]  `/src/pkg/path/resource_path.go` - canonical enums for representing each category.
- [x]  `/src/pkg/selectors/chats.go` - scope and const additions for the data category.
- [x]  `/src/pkg/backup/details/chats.go` - details iteminfo representation for the category.
- [x]  `RepoRef` and `LocationRef` layout is completed with examples.

## Enumeration

- [x]  All containers.  (to run at the top of every backup or restore operation)
- [x]  All Items-in-container.
- [x]  Kopia assist-ability requirements when enumerating data.

PR #5087, opened 2024-01-22
boilerplate teamschat collection package

ryanfkeepers:
seems like a lot of code, but this is 95% boilerplate additions copied from groups collections packages, with a find-replace for names.

Some noteworthy differences:
* teamsChats does not handle metadata, so all metadata, delta, and previous path handling was removed
* teamsChats does not produce tombstones
* chats are never deleted, so no "removed" items are tracked
* all chats gets stored at the prefix root, so no "containers" are iterated, and therefore only one collection is ever produced.

This means that, overall, while the boilerplate here is still the same, it's much reduced compared to similar packages.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #5062

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


HiteshRepo (approved):

src/internal/m365/collection/teamsChats/backup.go
```diff
@@ -0,0 +1,166 @@
...
+		qp.ProtectedResource.ID(),
+		path.TeamsChatsService,
+		qp.Category,
+		false)
+	if err != nil {
+		return nil, false, clues.WrapWC(ctx, err, "making metadata path prefix").
+			Label(count.BadPathPrefix)
+	}
+
+	metadataCollection, err := graph.MakeMetadataCollection(
```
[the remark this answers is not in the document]
ryanfkeepers: Possibly, though in the past when I've scroodled around with those it's caused issues.  I can follow up on this at the end of the other changes.

src/internal/m365/collection/teamsChats/backup.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, item := range items {
+		if !bh.includeItem(item, scope) {
+			cl.Inc(count.SkippedItems)
+		} else {
+			includedItems = append(includedItems, item)
+		}
+	}
+
+	cl.Add(count.ItemsAdded, int64(len(includedItems)))
+	cl.Add(count.ItemsRemoved, 0)
```
pandeyabs, line 140: This can be removed.

src/internal/m365/collection/teamsChats/collection.go
```diff
@@ -0,0 +1,418 @@
...
+				err = clues.Wrap(err, "getting channel message data").Label(fault.LabelForceNoBackupCreation)
+				el.AddRecoverable(ctx, err)
+
+				return
+			}
+
+			col.getAndAugment.augmentItemInfo(info, col.contains.container)
+
+			if err := writer.WriteObjectValue("", item); err != nil {
+				err = clues.Wrap(err, "writing channel message to serializer").Label(fault.LabelForceNoBackupCreation)
```
pandeyabs, line 196:
```suggestion
				err = clues.Wrap(err, "writing chat message to serializer").Label(fault.LabelForceNoBackupCreation)
```

src/internal/m365/collection/teamsChats/collection.go
```diff
@@ -0,0 +1,418 @@
...
+
+			writer := kjson.NewJsonSerializationWriter()
+			defer writer.Close()
+
+			item, info, err := col.getAndAugment.getItem(
+				ctx,
+				col.protectedResource,
+				id)
+			if err != nil {
+				err = clues.Wrap(err, "getting channel message data").Label(fault.LabelForceNoBackupCreation)
```
pandeyabs, line 187:
```suggestion
				err = clues.Wrap(err, "getting chat message data").Label(fault.LabelForceNoBackupCreation)
```

src/internal/m365/collection/teamsChats/collection.go
```diff
@@ -0,0 +1,418 @@
...
+			if err := writer.WriteObjectValue("", item); err != nil {
+				err = clues.Wrap(err, "writing channel message to serializer").Label(fault.LabelForceNoBackupCreation)
+				el.AddRecoverable(ctx, err)
+
+				return
+			}
+
+			itemData, err := writer.GetSerializedContent()
+			if err != nil {
+				err = clues.Wrap(err, "serializing channel message").Label(fault.LabelForceNoBackupCreation)
```
pandeyabs, line 204:
```suggestion
				err = clues.Wrap(err, "serializing chat message").Label(fault.LabelForceNoBackupCreation)
```

src/internal/m365/collection/teamsChats/collection_test.go
```diff
@@ -0,0 +1,475 @@
...
+			}
+
+			for item := range col.Items(ctx, errs) {
+				itemCount++
+
+				ok := slices.ContainsFunc(test.items, func(mc models.Chatable) bool {
+					return ptr.Val(mc.GetId()) == item.ID()
+				})
+
+				require.True(t, ok, "item must be either added or removed: %q", item.ID())
```
pandeyabs, line 315: Given lazy reader relies on item modTimes, we should add checks for that. For e.g. we should add modTimes for these items and add asserts against them.

something like

```
assert.Equal(
t,
modTime,
item.(data.ItemModTime).ModTime(),
"item mod time")
```

src/pkg/selectors/testdata/groups.go
```diff
@@ -4,7 +4,10 @@ import (
 	"github.com/alcionai/corso/src/pkg/selectors"
 )
 
-const TestChannelName = "Test"
+const (
+	TestChannelName = "Test"
+	TestChatTopic   = "Test"
```
pandeyabs, line 9: Curious, why is chat stuff in groups.go?

pandeyabs (approved):
Looks good

src/pkg/selectors/testdata/groups.go
ryanfkeepers: because I was too lazy to make a new file 😆

src/internal/m365/collection/teamsChats/collection_test.go
ryanfkeepers: I think we have that covered in other places, but I'll put a pin in this and come back to it later for assurances.

PR #5089, opened 2024-01-23
add teamschat boilerplate to m365/service

ryanfkeepers:
more boilerplate implementation

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Issue(s)

* #5062

#### Test Plan

- [x] :zap: Unit test

### DIFF
--- a/src/internal/m365/collection/teamsChats/handlers.go
+++ b/src/internal/m365/collection/teamsChats/handlers.go
@@ -79,7 +79,7 @@ type includeItemer[I chatsItemer] interface {
 // the given builder.  The tenantID and protectedResourceID are assumed
 // to be stored in the handler already.
 type canonicalPather interface {
-	canonicalPath() (path.Path, error)
+	CanonicalPath() (path.Path, error)
 }
 
 // ---------------------------------------------------------------------------
--- a/src/internal/m365/service/teamschats/backup.go
+++ b/src/internal/m365/service/teamschats/backup.go
@@ -1,0 +1,162 @@
+package groups
+
+import (
+	"context"
+
+	"github.com/alcionai/clues"
+
+	"github.com/alcionai/corso/src/internal/common/idname"
+	"github.com/alcionai/corso/src/internal/common/prefixmatcher"
+	"github.com/alcionai/corso/src/internal/data"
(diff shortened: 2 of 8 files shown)


src/internal/m365/collection/teamsChats/handlers.go
```diff
@@ -79,7 +79,7 @@ type includeItemer[I chatsItemer] interface {
 // the given builder.  The tenantID and protectedResourceID are assumed
 // to be stored in the handler already.
 type canonicalPather interface {
-	canonicalPath() (path.Path, error)
+	CanonicalPath() (path.Path, error)
```
HiteshRepo, line 82: why is this exported? all references for `canonicalPath` in this PR seems to be in same package.

HiteshRepo (approved):

src/internal/m365/collection/teamsChats/handlers.go
ryanfkeepers: ah, probably a typo on my end.  I'll clean that up.

src/internal/m365/service/teamschats/backup.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+		scope.Category().PathType().HumanString())
+	defer close(progressMessage)
+
+	bh := teamschats.NewUsersChatsBackupHandler(
+		bc.creds.AzureTenantID,
+		bc.producerConfig.ProtectedResource.ID(),
+		bc.apiCli.Chats())
+
+	// Always disable lazy reader for channels until #4321 support is added
```
pandeyabs, line 147: Should probably clean this up, it's a comment from channels code. Also, I'm assuming we'd have lazy reader support aka KAI for chats, so we should set `useLazyReader` to true?

pandeyabs (approved):
Looks good other than the lazy reader flag. We should address that.

src/internal/m365/service/teamschats/backup.go
ryanfkeepers: fwiw, this gets cleaned up in a later PR where I remove the non-lazy reader and the useLazyReader flag entirely.

PR #5093, opened 2024-01-23
Use more unique data in the kopia SourceInfo for snapshots

ashmrtn:
Use unique info for the kopia snapshot source information to help
isolate snapshots for different resources from each other. This will
come in handy both for checkpoint snapshots and regular snapshots if
kopia's retention policy happens to look for things.

Integration tests for this probably won't pass until the next PR in
the series is merged into this PR

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

merge after:
* #5091
* #5092

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/kopia/wrapper.go
+++ b/src/internal/kopia/wrapper.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 	"github.com/kopia/kopia/snapshot/snapshotmaintenance"
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 
 	"github.com/alcionai/corso/src/internal/common/prefixmatcher"
 	"github.com/alcionai/corso/src/internal/common/readers"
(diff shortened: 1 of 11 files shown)


src/internal/kopia/wrapper.go
```diff
@@ -228,8 +210,69 @@ func (w Wrapper) ConsumeBackupCollections(
 	return s, progress.deets, progress.toMerge, progress.errs.Failure()
 }
 
+type userAndHost struct {
```
ryanfkeepers, line 213: is `user` the best term here.  Would `resource` be more correct overall?

src/internal/kopia/wrapper.go
```diff
@@ -228,8 +210,69 @@ func (w Wrapper) ConsumeBackupCollections(
...
+	user string
+	host string
```
ryanfkeepers, line 215: `host` is an opaque concept from my perspective.  Could you please add a comment either here on the property or on the struct itself explaining its composition, identity, purpose, etc?  I think that'd help maintainers in the long run.

ryanfkeepers (approved):

PR #5107, opened 2024-01-24
Fix event cancellation status in export

meain:
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/converters/ics/ics.go
```diff
@@ -338,7 +338,7 @@ func updateEventProperties(ctx context.Context, event models.Eventable, iCalEven
 
 	// STATUS - https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.11
 	cancelled := event.GetIsCancelled()
-	if cancelled != nil {
+	if cancelled != nil && ptr.Val(cancelled) {
```
pandeyabs, line 341: nit: Should we make this check a bit more explict?
```suggestion
	if cancelled != nil && ptr.Val(cancelled) == true {
```

pandeyabs (approved):

src/internal/converters/ics/ics.go
meain: I think this should be fine.

Issue #5108, opened 2024-01-24
Handle multiple values for list items for a column


HiteshRepo:
Some columns have the capability to hold multiple values.

_**Person Column**_

**Single value**:
![Person-List](https://github.com/alcionai/corso/assets/48874082/3a6cbc31-5623-4d12-88bc-2a8621170777)

Get response:
```
"PersonsLookupId": "10",
```

Request body to create an item:
```
{
    "fields": {
        "Title": "Group-2",
        "PersonsLookupId": 10
    }
}
```

**Multiple value**:
![Person-List-Multi](https://github.com/alcionai/corso/assets/48874082/1de119e1-e4bd-491e-90a5-acce31d7862b)

Get response:
```
{
  "Persons": [
        {
            "LookupId": 10,
            "LookupValue": "Henrietta Mueller",
            "Email": "HenriettaM@10rqc2.onmicrosoft.com"
        },
        {
            "LookupId": 14,
            "LookupValue": "Adele Vance",
            "Email": "AdeleV@10rqc2.onmicrosoft.com"
        }
    ],
}
```

Request body to create an item:
```
{
    "fields": {
        "Title": "Group-2",
        "PersonsLookupId@odata.type": "Collection(Edm.Int32)",
        "PersonsLookupId": [10,11]
    }
}
```

_**Choice Column**_

**Single value**:
![Choice-List](https://github.com/alcionai/corso/assets/48874082/e1cb32d9-8cf8-49d6-b028-07c3874fd0f6)

Get response:
```
"choice": "good",
```

Request body to create an item:
```
{
    "fields": {
        "Title": "Item-2",
        "choice": "bad"
    }
}
```

**Multiple value**:
![Choice-List-Multi](https://github.com/alcionai/corso/assets/48874082/73014658-e8e8-4790-8915-3a43e54beab0)

Get response:
```
 "choice": [
        "good",
        "ok"
    ],
```

Request body to create an item:
```
{
    "fields": {
        "Title": "Item-2",
        "choice@odata.type": "Collection(Edm.String)",
        "choice": ["bad", "ok"]
    }
}
```

_**Lookup Column**_

**Single value**:
![Lookup-List](https://github.com/alcionai/corso/assets/48874082/8e435bff-e3a5-412e-947b-c18056c38206)

Get response:
```
"LookupLookupId": "1",
```

Request body to create an item:
```
{
    "fields": {
        "Title": "Item-2",
        "LookupLookupId": "2"
    }
}
```

[...]

PR #5109, opened 2024-01-24
provisions to identify fields with multiple value

HiteshRepo:
provisions to identify fields with multiple value.
Some fields can hold multiple values and single value based on a property called `allowMultipleValues`.
For example, `Lookup` column:
```
"lookup": {
    "allowMultipleValues": true,
    "allowUnlimitedLength": false,
    "columnName": "Title",
    "listId": "21b45bf2-e495-4582-b114-839577ff8e4f"
}
```
But `choice` columns, even though allows to set multiple choices/value, does not have that particular field `allowMultipleValues` to indicate.
So in this PR we are trying determine the same by the stored values while restoring.

**Original list with `choice` column in site**:
Choice-List-Multi

**Restored list with `choice` column in site**
Restored-Choice-List-Multi
The color does not come through though

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)
#5108 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (approved):

src/pkg/services/m365/api/lists.go
```diff
@@ -472,15 +490,45 @@ func setAdditionalDataByColumnNames(
...
 }
 
+func setMultipleEnabledByFieldData(val any, colDetails *columnDetails) {
+	// for columns like 'choice', even though it has an option to hold single/multiple values,
+	// the columnDefinition property 'allowMultipleValues' is not available.
+	// Hence we determine single/multiple from the actual field data.
+	if reflect.TypeOf(val).Kind() == reflect.Slice {
+		colDetails.isMultipleEnabled = true
+	}
+}
```
ryanfkeepers, line 513: Preference on general func behavior: rather than setting values a provided object (which causes coupling between the caller and the func), funcs that provide return values and let callers act on them (which provides cohesion without coupling), are easier to maintain in the long term.  In this case, it's a better pattern to have a func whose only job is to check whether the value is a slice.  Ex:
```suggestion
func isSlice(val any) bool {
	return reflect.TypeOf(val).Kind() == reflect.Slice
}
```

ryanfkeepers (commented):
have a nit on func patterns, but otherwise looks good.

Issue #5084, opened 2024-01-22
Change in GRAPH API response for certain columns


HiteshRepo:
**GRAPH Version**: v1.0

**Any change log**: No. https://developer.microsoft.com/en-us/graph/changelog/?filterBy=v1.0,Change,Sites%20and%20lists

**What changed**:

_Metadata column_:

![image](https://github.com/alcionai/corso/assets/48874082/d3ab21d6-7f47-4d8c-b9a5-f582725be248)

API response for the columns (https://graph.microsoft.com/v1.0/sites/{{siteid}}/lists/{{listid}}/columns)
```
{
  "columnGroup": "Custom Columns",
  "description": "",
  "displayName": "metadata",
  "enforceUniqueValues": false,
  "hidden": false,
  "id": "b7d8597e-264e-4685-a2c4-6494147e9e72",
  "indexed": false,
  "name": "metadata",
  "readOnly": false,
  "required": false,
  "defaultValue": {
    "value": ""
  }
}
```
- what changed in the column response for  `metadata column`? --> No indication that is an metadata column.
- is there a copy available for older response: No
- what the column response used to look like: as far as I remember, it used to be:
```
{
  "columnGroup": "Custom Columns",
  "description": "",
  "displayName": "metadata",
  "enforceUniqueValues": false,
  "hidden": false,
  "id": "b7d8597e-264e-4685-a2c4-6494147e9e72",
  "indexed": false,
  "name": "metadata",
  "readOnly": false,
  "required": false,
  "term": {
    "value": ""
  }
}
```

API response for item data: https://graph.microsoft.com/v1.0/sites/{{siteid}}/lists/{{listid}}/items/{{itemid}}/fields
```
{
  ...
  "Title": "item-1",
  "metadata": {
    "Label": "Engineering",
    "TermGuid": "6b5d3ce9-3043-499f-8be6-e92fb57bed96",
    "WssId": 4
  },
  ....
}
```
- what changed in the item data response for  `metadata column`? as far as I remember, it used to be:
```
{
  ...
  "Title": "item-1",
  "metadata": "Engineering",
  ....
}
```
- is there a copy available for older response: No

_Lookup column_:
![image](https://github.com/alcionai/corso/assets/48874082/55143141-807f-4daf-b9e5-d13dfbb5282c)

API response for the columns (https://graph.microsoft.com/v1.0/sites/{{siteid}}/lists/{{listid}}/columns)
```
{
  "columnGroup": "Custom Columns",
[...]

PR #5121, opened 2024-01-25
handles single & multiple values for metadata columns

HiteshRepo:
handles single & multiple values for metadata columns

Similar to `Hyperlink` and `Column` columns, `Metadata` column too is unrecognizable from GRAPH API response. Hence identifying from the field column names.

`Metadata` fields are like tags. A `Metadata` fields can be configured to hold multiple values/tags

**Original List with `Metadata` column (Department) with single value/tag**:
Metadata-List

**Restored List with `Metadata` column with single value/tag**:
Restored-Metadata-List

**Original List with `Metadata` column (Department) with multiple value/tag**:
Metadata-List-Multi

**Restored List with `Metadata` column with multiple value/tag**:
Restored-Multi-Metadata-List

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)
#5084 
#5108 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


pandeyabs (approved):
🚀 

Wow, these columns are a handful!

src/pkg/services/m365/api/lists.go
```diff
@@ -638,6 +642,44 @@ func hasHyperLinkFields(additionalData map[string]any) (map[string]any, string,
 	return nil, "", false
 }
 
+func hasMetadataFields(additionalData map[string]any) ([]map[string]any, string, bool) {
+	for fieldName, value := range additionalData {
+		switch valType := reflect.TypeOf(value).Kind(); valType {
```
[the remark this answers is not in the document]
HiteshRepo: @pandeyabs 
No. that is not possible for a single list item.
Either they hold multiple values or single.
If single -> map
if multiple -> slice

src/pkg/services/m365/api/lists.go
```diff
@@ -658,6 +662,44 @@ func hasHyperLinkFields(additionalData map[string]any) (map[string]any, string,
...
```
ryanfkeepers, line 667: This var assignment feels redundant.  Is this the same result?
```suggestion
		switch reflect.TypeOf(value).Kind() {
```

src/pkg/services/m365/api/lists.go
```diff
@@ -658,6 +662,44 @@ func hasHyperLinkFields(additionalData map[string]any) (map[string]any, string,
...
+				}
+			}
+
+			if len(mmdfs) > 0 {
+				return mmdfs, fieldName, true
+			}
+		}
+	}
+
+	return nil, "", false
```
ryanfkeepers, line 695: would you need a boolean return if you returned an empty slice by default?  Or, alternatively, can the caller check for `len(fieldName) == 0` instead of the bool?

src/pkg/services/m365/api/lists.go
```diff
@@ -703,6 +745,23 @@ func concatenateHyperLinkFields(hyperlinkFields map[string]any) string {
 	return ""
 }
 
+func concatenateMetadataFields(metadataFieldsArr []map[string]any) string {
+	labels := make([]string, 0)
+
+	for _, md := range metadataFieldsArr {
+		mdVal, ok := md[MetadataLabelKey].(*string)
```
ryanfkeepers, line 752: can you please use the `common/str` package func `anyValueToString` here?

src/pkg/services/m365/api/lists.go
```diff
@@ -703,6 +745,23 @@ func concatenateHyperLinkFields(hyperlinkFields map[string]any) string {
...
+		if ok {
+			labels = append(labels, ptr.Val(mdVal))
+		}
+	}
+
+	if len(labels) > 0 {
+		return strings.Join(labels, ",")
+	}
+
+	return ""
```
ryanfkeepers, line 762: this is redundant handling.  Join will return an empty string if the slice has 0 len.
```suggestion
	return strings.Join(labels, ",")
```

ryanfkeepers (commented):
similar nits on implementation

PR #5134, opened 2024-01-26
Centralize metadata suffix definitions

pandeyabs:
Centralizing the suffixes so that group mailbox and future services can also use them.

* Move `.meta`, `.data`, `.dirmeta` definitions from `internal/m365/collection/drive/metadata` pkg to `pkg/services/m365/api/graph/metadata`.
* Adjust package names in misc places.
* No logic change. 
* We should probably keep all metadata definitions and code in one place, i.e. relocate drive metadata definitions to `pkg/services/m365/api/graph/metadata`. This is a bigger change, we can do it later.
---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/pkg/services/m365/api/graph/metadata/consts.go
ashmrtn, line 1: I think that centralizing this type of info a bit more is a good idea so that there's some standardization. However, I'm not sure if the new location of the file is the correct place for it. Since it's now in `pkg/services/api/graph/metadata`, it's available to SDK users which seems a bit odd. The metadata files are purely a corso construct meant to aid the restore logic we have in place. Consumers of the corso SDK (e.x. CLI) neither need to know about these types of metadata files nor interact with them

I realize this PR is just moving this file to keep things together, but perhaps we should reconsider if the `pkg/services/api/graph/metadata` package is in the correct spot with respect to the rest of the code in the project

pandeyabs: Thanks @ashmrt. Can we punt on this? I agree with the sentiment, but I am not able to think of a better location for this right now.
Also, some of the code like `IsMetadataFile` has already existed in this package. I think this should be a part of a wider cleanup of `pkg/services/api/graph`. There are a few things we have in here which shouldn't be available to SDK users.

I can edit my following PRs to make sure to not add more stuff to this package.

ashmrtn: `internal/data` may be a reasonable location for stuff like this. Since packages that import that package are also likely to require file suffixes, it seems likely that it would avoid import cycles (haven't tested myself so can't guarantee that)

ashmrtn (approved):

src/pkg/services/m365/api/graph/metadata/consts.go
pandeyabs: Thanks! Good idea. Let me try it out.

pandeyabs: Won't be able to move `IsMetadataFile` logic to it, since it kind of corrupts the generic `internal/data` pkg with service level code. But I'll move the file suffix stuff in there if there is no import cycle.

pandeyabs: Running into import cycles after moving to `internal/data`. I'll file a tech debt and continue with what I have in this PR.

```
package command-line-arguments
        imports github.com/alcionai/corso/src/cli
	imports github.com/alcionai/corso/src/cli/backup
	imports github.com/alcionai/corso/src/cli/flags
	imports github.com/alcionai/corso/src/pkg/control
	imports github.com/alcionai/corso/src/pkg/extensions
	imports github.com/alcionai/corso/src/pkg/backup/details
	imports github.com/alcionai/corso/src/internal/data
	imports github.com/alcionai/corso/src/pkg/backup/details: import cycle not allowed
	
	
package command-line-arguments
	imports github.com/alcionai/corso/src/cli
	imports github.com/alcionai/corso/src/cli/backup
	imports github.com/alcionai/corso/src/cli/flags
	imports github.com/alcionai/corso/src/pkg/control
	imports github.com/alcionai/corso/src/pkg/extensions
	imports github.com/alcionai/corso/src/pkg/backup/details
	imports github.com/alcionai/corso/src/internal/data
	imports github.com/alcionai/corso/src/pkg/control: import cycle not allowed
```

ashmrtn: thanks for testing that out. Sorry you spent time on it at my suggestion and it didn't work out

pandeyabs: Oh no worries. It was a small change. Didn't take me too much time :) Thanks for the review comment!

PR #5135, opened 2024-01-26
Persist metadata files for group mailboxes

pandeyabs:
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


src/internal/m365/collection/groups/collection.go
```diff
@@ -370,6 +376,39 @@ func (col *lazyFetchCollection[C, I]) streamItems(ctx context.Context, errs *fau
...
+				if err != nil {
+					errs.AddRecoverable(ctx, clues.StackWC(ctx, err))
+
+					return
+				}
+
+				col.stream <- storeItem
+			}
+
+			// TODO(pandeyabs): Persist as .data file for conversations only, not for channels
```
pandeyabs, line 407: This change will be added in a follow up PR.

src/internal/m365/collection/groups/collection.go
```diff
@@ -341,7 +342,12 @@ func (col *lazyFetchCollection[C, I]) streamItems(ctx context.Context, errs *fau
 			defer wg.Done()
 			defer func() { <-semaphoreCh }()
 
-			col.stream <- data.NewDeletedItem(id)
+			// This is a no-op for conversations, as there is no way to detect
+			// deleted items in a conversation. It might be added in the future
+			// if graph supports it, so make sure we put up both .data and .meta
+			// files for deletions.
+			col.stream <- data.NewDeletedItem(id + metadata.DataFileSuffix)
```
pandeyabs, line 349: Likewise, I'll make sure channels don't end up getting impacted from this change in a small follow up PR.

src/internal/m365/collection/groups/collection.go
```diff
@@ -370,6 +378,44 @@ func (col *lazyFetchCollection[C, I]) streamItems(ctx context.Context, errs *fau
...
+			itemMeta, _, err := col.getAndAugment.getItemMetadata(
+				ictx,
+				col.contains.container)
+			if err != nil && !errors.Is(err, errMetadataFilesNotSupported) {
+				errs.AddRecoverable(ctx, clues.StackWC(ctx, err))
+
+				return
+			}
+
+			if err == nil {
```
pandeyabs, line 402: Quick note, this code is slightly refactored in the following PR https://github.com/alcionai/corso/pull/5136/files .  Not doing it here to keep the PR within review limits.

src/internal/m365/collection/groups/collection.go
```diff
@@ -381,7 +427,7 @@ func (col *lazyFetchCollection[C, I]) streamItems(ctx context.Context, errs *fau
 					contains:      col.contains,
 					parentPath:    col.LocationPath().String(),
 				},
-				id,
+				id+metadata.DataFileSuffix,
```
pandeyabs, line 430: Persist as .data file for conversations only, not for channels
i.e. only add the .data suffix for conv backups. 

This is fixed in following PR in this chain. https://github.com/alcionai/corso/pull/5136/files

src/internal/m365/collection/groups/collection.go
```diff
@@ -370,6 +378,44 @@ func (col *lazyFetchCollection[C, I]) streamItems(ctx context.Context, errs *fau
...
+			//
+			// If the data download fails for some reason other than deleted in
+			// flight, we will still end up persisting a .meta file. This is
+			// fine however, since the next backup will overwrite it.
+			//
+			// If item is deleted in flight, we will end up with an orphaned
+			// .meta file. The only impact here is storage bloat, which
+			// is minimal. Other impact could be if we do an in-order restore
+			// using a tree built from .meta files. We may have some .meta
+			// files without corresponding .data files.
```
ashmrtn, line 392: if restore is requesting meta files based on translated paths that were sourced from backup details then I think we'll actually be ok. The reason for that is that if we failed to get the data file the item won't appear in backup details, thus we won't translate the path for the meta file and will essentially act as if it doesn't exist

src/pkg/services/m365/api/graph/metadata/metadata_test.go
```diff
@@ -152,3 +152,108 @@ func (suite *MetadataUnitSuite) TestIsMetadataFile_Directories() {
...
+		},
+		{
+			name:       "lists",
+			service:    path.SharePointService,
+			category:   path.ListsCategory,
+			isMetaFile: true,
+		},
+	}
+
+	for _, test := range table {
```
ashmrtn, line 235: I'm wondering if a bit of refactoring could give better test coverage while maybe making things a little more clear. I was thinking of a few things:
1. define a test input struct outside of the test so it can be used in multiple functions
2. create a `runTest` function (maybe with a better name) that has (most of) the logic that's currently in the for loop
3. create two `[]testInput` variables: one for service/categories that support metadata and another for ones that don't
4. create two `TestX` functions: one for the metadata suffix and one for the data suffix (alternatively: create three `TestX` functions for has metadata, doesn't have metadata, and data)
5. call into the `runTest` function in the `TestX` functions

The test struct could be similar to
```go
type testInput struct {
  service path.ServiceType
  category path.CategoryType
  fileName string // <- should include .data/.meta suffix
  expected bool
}
```

if you define the metadata results as two slices in the file (not a function) then the data suffix test can loop over those and just build the correct input struct like below
```go
func TestDataSuffix() {
  allTests := []testInput{}

  // Need to append to an empty slice so it doesn't modify the first list passed.
  for _, input := range append(append([]testInput{}, isMetaTests...), isNotMetaTests...) {
    allTests = append(allTests, testInput{
      service: input.service,
      category: input.category,
      fileName: "testfile" + dataSuffix,
    })
  }

  // Can switch up this logic to better fit your preference.
  for _, test := range allTests {
    suite.Run(test.service.HumanString() + test.category.HumanString() + "Data", func() {
      runTest(test)
    })
  }
}
```

The above has the advantage of running tests for both meta and data suffixes for every service/category. It also only requires a service/category be defined in a single place: the list that determines if it has a metadata file or not

src/internal/m365/collection/groups/conversation_handler_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestConversationHandlerUnitSuite(t *testing.T) {
+	suite.Run(t, &ConversationHandlerUnitSuite{Suite: tester.NewUnitSuite(t)})
+}
+
+// Basic test to ensure metadata is serialized and deserialized correctly.
+func (suite *ConversationHandlerUnitSuite) TestGetItemMetadata() {
+	var (
+		t  = suite.T()
+		bh = conversationsBackupHandler{
+			resourceEmail: "test@example.com",
```
ashmrtn, line 32: nit: could you please make the topic and email consts in this test and just reference them (especially in the asserts)?

```go
const (
  resourceEmail = "test@example.com"
  ...
)
```

ashmrtn (approved):

PR #5138, opened 2024-01-26
use query params in all handlers

ryanfkeepers:
backup handlers are all re-using the same inputs for tenantID and protected resource ID.  In some cases we're storing those values in the handler, other cases we don't. This pr seeks to normalize backup handler design by expecting a common structure for holding the resource and tenant ids.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


HiteshRepo (approved):
LGTM

PR #5140, opened 2024-01-26
introduce intg and unit test setup packages

ryanfkeepers:
corso has a thousand bespoke approaches to setting the same info in all of its tests.  This is a first step towards minimizing and standardizing the lift around that work. Future PRs will distribute these packages through the repo.

---

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/tester/its/m365.go
```diff
@@ -0,0 +1,209 @@
...
+// Intercepting calls with Gock
+// ---------------------------------------------------------------------------
+
+const graphAPIHostURL = "https://graph.microsoft.com"
+
+func V1APIURLPath(parts ...string) string {
+	return strings.Join(append([]string{"/v1.0"}, parts...), "/")
+}
+
+func InterceptV1Path(pathParts ...string) *gock.Request {
```
ashmrtn, line 57: I'm not convinced this is something we want in this particular package. It hides the fact that it's only actually for GET request types. We may want to test other request types as well

src/internal/tester/its/m365.go
```diff
@@ -0,0 +1,209 @@
...
+// m365IntgTestSetup provides all the common references used in an m365 integration
+// test suite.  Call `its.GetM365()` to get the singleton for your test suite.
+// If you're looking for unit testing setup, use `uts.GetM365()` instead.
+type m365IntgTestSetup struct {
+	Acct     account.Account
+	Creds    account.M365Config
+	TenantID string
+
+	AC     api.Client
+	GockAC api.Client
```
ashmrtn, line 84: I think now that it's possible to mock out requesting credentials we should only have either a mocked client or a regular client. Having both makes it seem like it will make things error prone

src/internal/m365/service/teamschats/backup.go
```diff
@@ -145,9 +145,13 @@ func backupChats(
 		scope.Category().PathType().HumanString())
 	defer close(progressMessage)
 
+	qp := graph.QueryParams{
```
ashmrtn, line 148: this doesn't appear related to the rest of the PR. Could you please move it to a separate PR?

src/internal/tester/its/m365.go
```diff
@@ -0,0 +1,209 @@
+package its
```
ashmrtn, line 1: could you please use a more descriptive package name? Maybe something like `service` or `scaffold` or something. `its` forces people to read through the code in the package because it lacks any meaning on its own

src/internal/tester/uts/m365.go
```diff
@@ -0,0 +1,96 @@
+package uts
```
ashmrtn, line 1: this also needs a more descriptive package name/structure

src/internal/tester/its/m365.go
```diff
@@ -0,0 +1,209 @@
...
+	DriveID           string
+	DriveRootFolderID string
+	TestContainerID   string
+	WebURL            string
+}
+
+// m365IntgTestSetup provides all the common references used in an m365 integration
+// test suite.  Call `its.GetM365()` to get the singleton for your test suite.
+// If you're looking for unit testing setup, use `uts.GetM365()` instead.
+type m365IntgTestSetup struct {
```
ashmrtn, line 78: The unit and integration test setups seem to share a lot of the same fields and data. Why not have the integration test setup embed the unit test setup? That way the fields and tests for the test setup only need to be defined once

src/internal/tester/its/m365.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+// GetM365 returns the populated its.m365 singleton.
+func GetM365(t *testing.T) m365IntgTestSetup {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if singleton != nil {
+		return *singleton
+	}
```
ashmrtn, line 106: is there a particular need for this to be a singleton? Often singletons cause more problems than the solve because it's not always apparent when they're initialized or what they're initialized with. The global concurrency limiter we have is an excellent example of why singletons can be confusing/error prone to use

ashmrtn (commented):
I'd like to see some work towards reducing code duplication between the two test setups and discussion/plans for the singleton prior to merging

src/internal/tester/its/m365.go
ryanfkeepers: On one hand, the option for either one goes against the nature of the singleton here, where the (arbitrary) initial caller is expected to produce a live connection and fulfill all the queries.  On the other hand, I think you're right that we shouldn't need both.  The Gock client stuff should probably move out into its own package.  Also, we should probably treat gock'd tests as `unit` instead of `integration`, which would make it an irrelevant addition to this setup.

I'm happy to make these changes, but in the spirit of the scope of this PR I'm currently moving around architecture that's already in place.  Let me punt it to the end of the PR chain (so that I don't have to mess with conflicts) and I'll get these refactored.  Sound good?

src/internal/m365/service/teamschats/backup.go
ryanfkeepers: Huh, weird.  that _is_ from a separate PR.  Not sure how it got caught up in this one.

src/internal/tester/its/m365.go
ryanfkeepers: not a fan of `service` or `scaffold`, personally (imo they also lack descriptivity).  Otherwise, sure, I'll figure out a non-abbreviated name by the end.  Also going to punt til later to minimize refactoring.

src/internal/tester/its/m365.go
ryanfkeepers: Mostly to keep with "a little bit of duplication is better than a little bit of abstraction".  Honestly, though, I'm not sure I convinced myself of the value of the unit test variation by the time I finished plugging in the integration version.  I'm thinking about cutting it out entirely.  What are your feelings on it?

src/internal/tester/its/m365.go
ryanfkeepers: It's not a necessity, no.  But I'd say this is a fairly safe case since the state of initialization is resolved at the time of accessing the singleton (which is non-optional for usage), and there's no difference in what data is produced since that's all controlled internal to the package.  Our other singletons are more problematic because initialization is optional, and singletons can be initialized with different configurations, neither of which is the case here.

ashmrtn: I think whether it's a safe case or not depends on the context of how it's used. The next PR makes this an exported struct, so we now have an exported singleton where many of it's fields are also exported. If any test decides that the way to pass around non-standard data to later parts of a test is to modify the fields in the struct then all tests could be affected. Thus, this ends up encoding project level standards in places where they may not really be required

src/internal/tester/its/m365.go
ryanfkeepers: Sure.  I'll probably move the gock stuffs out to their own package for better control.  Then we can disassociate it from this setup.

src/internal/tester/its/m365.go
ryanfkeepers: That should be safe to do, since the getter returns a copy of the singleton, not its reference.

ashmrtn: that only really works assuming everything remains basic types instead of more complex things like interfaces, pointers, maps, slices, etc. The fact that we return a copy to this type really makes we wonder why it's a singleton though. It seems like it would be less complex to just make a new set of data all the time

ashmrtn (approved):

src/internal/tester/its/m365.go
ryanfkeepers: mostly trying to minimize the re-calling of existing data.  It's not a critical performance issue.

ashmrtn: I think avoiding repopulating data is a valid goal, but I feel this particular pattern is a little too error prone to outweigh the possible benefits. Although this is code for tests, our heavy reliance on automated testing means that we need to have high confidence that our tests are doing what we think they are even as things change over time. Since misuse or unintended consequences from changes to this code could affect a large chunk of the integration tests in the repo, we should make sure it's difficult to misuse or break with future changes

That said, there are other designs that would allow for fetching data only once in a safe manner which we can explore in parallel to this PR going forward (or after) if you want

PR #5142, opened 2024-01-27
replace more local intgTestSetups with its

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/tester/its/m365.go
```diff
@@ -70,6 +70,13 @@ type IDs struct {
 	DriveRootFolderID string
```
ashmrtn, line 70: it feels a lot like this and some of the other fields in this struct be cleaned up now that there's a `RootSite` substruct. For example, `DriveID`, `DriveRootFolderID`, and `WebURL` all seem like they may be duplicates. There's also no comments or documentation to help developers determine which is the "correct" one to use for a given situation

ashmrtn (approved):

src/internal/tester/its/m365.go
ryanfkeepers: Ah, good call.  I'll add some documentation to clearly delineate.

ashmrtn: can we not just dedupe fields?

PR #5152, opened 2024-01-29
Convert serialized posts to eml

pandeyabs:
* Add the serialized `Postable` to EML converter code for group mailboxes and associated tests.
* We should converge this with `Messageable` code at some point with a small translator function which converts `Postable` to `Messageable`. Reason being that posts are a subset of messages. Although I feel it's a bit premature right now. Once we have tested this in prod a bit, and we know for a fact that posts are indeed a true subset of messages, we can do this convergence. 

---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


meain (approved):
> ... with a small translator function which converts `Postable` to `Messageable` ...

This would be better. I personally would have done this conversion as there are more gotchas in the eml conversion. We could always separate it out in case we find that `Postable` is not a subset.

Looking at the interface, these seem to be the functions that are in `Postable` but not in `Messageable`:

```go
GetConversationThreadId()(*string)
GetInReplyTo()(Postable)
GetNewParticipants()([]Recipientable)
```


I'll however leave that decision up to you. The PR looks good to me.

pandeyabs:
Thanks @meain . I'll punt this for now. Here's a tech debt ticket. https://github.com/alcionai/corso/issues/5155 

I had one question about the message -> EML code [here](https://github.com/alcionai/corso/blob/8a74bc893c39b5fa8271a266f43b052f4547beff/src/internal/converters/eml/eml.go#L203).  

```
	if data.GetSentDateTime() != nil {
		email.SetDate(ptr.Val(data.GetSentDateTime()).Format(dateFormat))
	}
```

Any reason we are using sent date time here and not received date time or last modified date time? 
Asking because posts don't have sent date time field. So I am using received date time field as an alternative.

meain:
The [spec](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.1) mentions it to be the "sent" datetime. The idea is to have the datetime where it was "created".

---

[3.6.1](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.1).  The Origination Date Field

   The origination date field consists of the field name "Date" followed
   by a date-time specification.

   orig-date       =   "Date:" date-time CRLF

   The origination date **specifies the date and time at which the creator
   of the message indicated that the message was complete and ready to
   enter the mail delivery system**.  For instance, this might be the time
   that a **user pushes the "send" or "submit" button** in an application
   program.  In any case, it is specifically not intended to convey the
   time that the message is actually transported, but rather the time at
   which the human or other creator of the message has put the message
   into its final form, ready for transport.  (For example, a portable
   computer user who is not connected to a network might queue a message
   for delivery.  The origination date is intended to contain the date
   and time that the user queued the message, not the time when the user
   connected to the network to send the message.)

pandeyabs:
Thanks Abin! I'll change it to `GetCreatedDateTime()`

PR #5153, opened 2024-01-29
Add group mailbox export

pandeyabs:
* Add EML exports for group mailbox.
* Tested E2E manually along with unit tests added in this PR.
* Will follow it up with a sanity test PR.
---

- [ ] :no_entry: No

#### Type of change

- [x] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/groups/export.go
+++ b/src/internal/m365/collection/groups/export.go
@@ -5,19 +5,24 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/alcionai/clues"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 
 	"github.com/alcionai/corso/src/internal/common/ptr"
+	"github.com/alcionai/corso/src/internal/converters/eml"
 	"github.com/alcionai/corso/src/internal/data"
+	groupMeta "github.com/alcionai/corso/src/internal/m365/collection/groups/metadata"
 	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/export"
 	"github.com/alcionai/corso/src/pkg/fault"
+	"github.com/alcionai/corso/src/pkg/logger"
 	"github.com/alcionai/corso/src/pkg/metrics"
 	"github.com/alcionai/corso/src/pkg/path"
 	"github.com/alcionai/corso/src/pkg/services/m365/api"
+	"github.com/alcionai/corso/src/pkg/services/m365/api/graph/metadata"
 )
 
 func NewExportCollection(
@@ -26,7 +31,13 @@ func NewExportCollection(
 	backupVersion int,
 	cec control.ExportConfig,
 	stats *metrics.ExportStats,
+	cat path.CategoryType,
 ) export.Collectioner {
+	streamItems := streamChannelMessages
+	if cat == path.ConversationPostsCategory {
+		streamItems = streamConversationPosts
+	}
+
 	return export.BaseCollection{
 		BaseDir:           baseDir,
 		BackingCollection: backingCollections,
@@ -37,8 +48,12 @@ func NewExportCollection(
 	}
 }
 
-// streamItems streams the items in the backingCollection into the export stream chan
-func streamItems(
+//-------------------------------------------------------------
+// Channel Messages
+//-------------------------------------------------------------
+
+// streamChannelMessages streams the items in the backingCollection into the export stream chan
+func streamChannelMessages(
 	ctx context.Context,
 	drc []data.RestoreCollection,
 	backupVersion int,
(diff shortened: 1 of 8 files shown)


src/internal/m365/collection/groups/export.go
```diff
@@ -198,3 +213,146 @@ func makeMinimumChannelMesasge(item models.ChatMessageable) minimumChannelMessag
...
+			return groupMeta.ConversationPostMetadata{}, err
+		}
+
+		err = json.Unmarshal(metaraw, &meta)
+		if err != nil {
+			return groupMeta.ConversationPostMetadata{}, err
+		}
+	}
+
+	return meta, nil
```
ryanfkeepers, line 357: nit: avoid indented blocks if possible
```suggestion
	if metaRC == nil {
		return nil, nil
	}
	
	metaraw, err := io.ReadAll(metaRC)
	if err != nil {
		return groupMeta.ConversationPostMetadata{}, err
	}

	err = json.Unmarshal(metaraw, &meta)
	if err != nil {
		return groupMeta.ConversationPostMetadata{}, err
	}

	return meta, nil
```

src/internal/m365/collection/groups/export.go
```diff
@@ -198,3 +213,146 @@ func makeMinimumChannelMesasge(item models.ChatMessageable) minimumChannelMessag
...
+}
+
+func fetchAndReadMetadata(
+	ctx context.Context,
+	itemID string,
+	fin data.FetchItemByNamer,
+) (groupMeta.ConversationPostMetadata, error) {
+	metaName := itemID + metadata.MetaFileSuffix
+
+	ictx := clues.Add(ctx, "meta_file_name", metaName)
```
ryanfkeepers, line 321: `ictx` isn't a necessary cast unless we're inside a loop.  All other clues additions can replace the `ctx`.
```suggestion
	ctx = clues.Add(ctx, "meta_file_name", metaName)
```

src/internal/m365/collection/groups/export.go
```diff
@@ -198,3 +213,146 @@ func makeMinimumChannelMesasge(item models.ChatMessageable) minimumChannelMessag
...
+) {
+	defer close(ch)
+
+	errs := fault.New(false)
+
+	for _, rc := range drc {
+		ictx := clues.Add(ctx, "path_short_ref", rc.FullPath().ShortRef())
+
+		for item := range rc.Items(ctx, errs) {
+			itemCtx := clues.Add(ictx, "stream_item_id", item.ID())
```
ryanfkeepers, line 238: not much point in having both ctx vars, since only one is used.
```suggestion
		for item := range rc.Items(ctx, errs) {
			ictx := clues.Add(
				ctx,
				"path_short_ref", rc.FullPath().ShortRef(),
				"stream_item_id", item.ID())
```

src/internal/m365/collection/groups/export.go
```diff
@@ -26,7 +31,13 @@ func NewExportCollection(
 	backupVersion int,
 	cec control.ExportConfig,
 	stats *metrics.ExportStats,
+	cat path.CategoryType,
 ) export.Collectioner {
+	streamItems := streamChannelMessages
+	if cat == path.ConversationPostsCategory {
+		streamItems = streamConversationPosts
+	}
```
ryanfkeepers, line 39: might want a switch with channels, posts, default (err) isntead of this pattern, since groups can handle more categories than these two.  Worst case scenario it could prevent or identify an accidental usage of this func.

ryanfkeepers (approved):
some nits; generally lgtm.

src/internal/converters/eml/eml.go
```diff
@@ -303,3 +304,137 @@ func FromJSON(ctx context.Context, body []byte) (string, error) {
...
 }
+
+//-------------------------------------------------------------
+// Postable -> EML
+//-------------------------------------------------------------
+
+// FromJSONPostToEML converts a postable (as json) to .eml format.
+// TODO(pandeyabs): This is a stripped down copy of messageable to
+// eml conversion, it can be folded into one function by having a post
+// to messageable converter.
```
ashmrtn, line 315: Assuming the function names are the same between the Postable and Messageable interfaces, we could just define our won interface that includes the subset of functions that are accessed in this code. If the function names aren't the same between the graph SDK types then we can make a new type that embeds one of the graph SDK types and implements the proper function names we need. It'll just call the corresponding function on the graph SDK type. I think we did something similar for calendar events which have `GetCalendar()` instead of `GetParentFolderId()`

ashmrtn (commented):
just skimming and thought I'd leave a ntoe

src/internal/converters/eml/eml.go
pandeyabs: Function names aren't the same. We can make a new type like you said, or convert `Postable` to `Messageable`. Either should work out. Tracking this as part of https://github.com/alcionai/corso/issues/5155

PR #5159, opened 2024-01-29
Intg test setup core 4

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/pkg/services/m365/groups_test.go
```diff
@@ -33,16 +35,15 @@ func TestGroupsIntgSuite(t *testing.T) {
 
 func (suite *GroupsIntgSuite) SetupSuite() {
 	t := suite.T()
+	var err error
```
ashmrtn, line 38: nit: could you please move this declaration closer to the actual usage. I has goldfish memory and forget stuff like this when it's not nearby

src/pkg/services/m365/sites_test.go
```diff
@@ -36,16 +38,15 @@ func TestSiteIntegrationSuite(t *testing.T) {
 
 func (suite *siteIntegrationSuite) SetupSuite() {
 	t := suite.T()
+	var err error
```
ashmrtn, line 41: same here, could you please move this closer to the actual usage of the variable?

ashmrtn (approved):

src/pkg/services/m365/groups_test.go
ryanfkeepers: goldfish memory

PR #5160, opened 2024-01-29
more its implementation

ryanfkeepers:

- [x] :no_entry: No

#### Type of change

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/onedrive_test.go
```diff
@@ -45,118 +40,11 @@ var (
...
-}
-
-func (si suiteInfoImpl) ResourceOwner() string {
-	return si.resourceOwner
-}
-
-func (si suiteInfoImpl) Service() path.ServiceType {
-	return si.service
-}
+// type suiteInfoImpl struct {
```
ashmrtn, line 43: could you please remove this as well if it's not going to be revived?

src/internal/m365/onedrive_test.go
```diff
@@ -313,70 +167,44 @@ func TestOneDriveNightlySuite(t *testing.T) {
...
 	// No reason why it couldn't work with previous versions, but this is when it got introduced.
-	testRestoreFolderNamedFolderRegression(suite, version.All8MigrateUserPNToID)
+	testRestoreFolderNamedFolderRegression(suite, suite.m365, suite.rs, version.All8MigrateUserPNToID)
 }
 
 func testRestoreAndBackupMultipleFilesAndFoldersNoPermissions(
-	suite oneDriveSuite,
+	suite tester.Suite,
+	m365 its.M365IntgTestSetup,
+	irs its.ResourceServicer,
```
ashmrtn, line 203: nit: could you please use a more descriptive name for this parameter? Ditto for other functions that are similar to this one

src/internal/m365/onedrive_test.go
```diff
@@ -179,65 +68,47 @@ func TestSharePointIntegrationSuite(t *testing.T) {
...
 }
 
 // ---------------------------------------------------------------------------
 // OneDrive most recent backup version
 // ---------------------------------------------------------------------------
 type OneDriveIntegrationSuite struct {
 	tester.Suite
-	suiteInfo
+	m365 its.M365IntgTestSetup
+	rs   its.ResourceServicer
```
ashmrtn, line 111: nit: could you please use a more descriptive name for this field?

src/internal/tester/its/resourceService.go
```diff
@@ -0,0 +1,25 @@
...
+import "github.com/alcionai/corso/src/pkg/path"
+
+type ResourceServicer interface {
+	Resource() IDs
+	Service() path.ServiceType
+}
+
+type resourceAndService struct {
+	ProtectedResource IDs
+	ServiceType       path.ServiceType
```
ashmrtn, line 12: could you please make these unexported fields? There's an interface for getting them so there shouldn't be a need to export them to other packages

src/internal/tester/its/resourceService.go
```diff
@@ -0,0 +1,25 @@
+package its
+
+import "github.com/alcionai/corso/src/pkg/path"
+
+type ResourceServicer interface {
```
ashmrtn, line 5: Adding here since I can't comment directly on the `M365IntegTestSetup` struct definition... Given that we now have an interface type that allows getting the `IDs` and the service type (i.e. a useful but still minimal superset of data in `M365IntegTestSetup`) it seems like `M365IntegTestSetup` should actually be returning these for the different things it has instead of IDs. That may avoid the problem of having to setup these things separately in tests as they'd all be sourced from the `M365IntegTestSetup` struct

src/internal/tester/its/resourceService.go
```diff
@@ -0,0 +1,25 @@
...
+	Resource() IDs
+	Service() path.ServiceType
+}
+
+type resourceAndService struct {
```
ashmrtn, line 10: could you please add an assignment that ensures continued interface compliance for this struct?
```suggestion
var _ ResourceServicer = resourceAndService{}

type resourceAndService struct {
```

src/internal/tester/its/resourceService.go
```diff
@@ -0,0 +1,25 @@
...
+func (ras resourceAndService) Resource() IDs {
+	return ras.ProtectedResource
+}
+
+func (ras resourceAndService) Service() path.ServiceType {
+	return ras.ServiceType
+}
+
+func NewResourceService(r IDs, s path.ServiceType) ResourceServicer {
+	return &resourceAndService{r, s}
```
ashmrtn, line 24: nit: please use struct field names. That way if more things are added to the struct the dev making that change doesn't have to find all the initialization sites and add a lot of stuff to them

```suggestion
	return &resourceAndService{
		protectedResource: r,
		serviceType: s,
	}
```

ashmrtn (approved):
could use some polishing around interface/struct design

src/internal/m365/onedrive_test.go
ryanfkeepers: whoops, yup, didn't mean for those to hang around

src/internal/tester/its/resourceService.go
ryanfkeepers: good catch, yes!

src/internal/tester/its/resourceService.go
ryanfkeepers: Will make a note and follow-up on this later.

Issue #5167, opened 2024-01-30
[Bug]: mailboxInfo swallows user-not-found errors


ryanfkeepers:
### What happened?

this older handling (https://github.com/alcionai/corso/blame/940d6e82451ca2a0d759be8e615369524d59d5cc/src/pkg/services/m365/api/users.go#L197) was changed when we collapsed resourceNotFound and notFound into a single error.  We need to find a way to keep the two separate.

### Corso Version?

Corso v0.17.0

### Where are you running Corso?

any

PR #5169, opened 2024-01-30
fix mailbox user not found handling

ryanfkeepers:
re-splits the difference between a user not found error and a non-user not found error.

---

#### Type of change

- [x] :bug: Bugfix

#### Issue(s)

* #5167

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E


src/internal/m365/service/exchange/enabled.go
```diff
@@ -23,7 +23,8 @@ func IsServiceEnabled(
 ) (bool, error) {
 	_, err := gmi.GetMailInbox(ctx, resource)
 	if err != nil {
-		if err := api.EvaluateMailboxError(err); err != nil {
+		ignorable := api.IsMailboxErrorIgnorable(err)
```
zackrossman, line 26: should we include the ignorable error in the info log on line 32? same question for all instances of this pattern

src/pkg/services/m365/api/users.go
```diff
@@ -185,26 +185,18 @@ func appendIfErr(errs []error, err error) []error {
 	return append(errs, err)
 }
 
-// EvaluateMailboxError checks whether the provided error can be interpreted
+// IsMailboxErrorIgnorable checks whether the provided error can be interpreted
 // as "user does not have a mailbox", or whether it is some other error.  If
 // the former (no mailbox), returns nil, otherwise returns an error.
```
zackrossman, line 190: nit: does this comment need to be updated to mention UserNotFound scenario

zackrossman (approved):
LGTM but will want a Corso team review as well

src/internal/m365/service/exchange/enabled.go
ryanfkeepers: good idea

src/pkg/services/m365/api/users.go
ryanfkeepers: whoops, good catch.

HiteshRepo (approved):

Issue #5080, opened 2024-01-22
Imrpoving CI run time


meain:
The CI is currently taking up to 1h at time.  
Run analyzed: https://github.com/alcionai/corso/actions/runs/7579723408

Notes:
- Most time is taken up by tests that need network calls
- Currently test caching does not seem to be working

Possilble ways to speed up CI:
- Cache test restults (we can run without cache in `main`)
- Mock out grpah results (we can call actual graph apis in `main`)

Resource on test caching (https://www.airplane.dev/blog/caching-golang-tests-in-ci)

---

- Export data from `Test-Suite-Trusted` run
    
    ```diff
    diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
    index 69aaff1e1..2db3763fa 100644
    --- a/.github/workflows/ci.yml
    +++ b/.github/workflows/ci.yml
    @@ -7,7 +7,7 @@ on:
       pull_request:
     
       push:
    -    branches: [main]
    +    branches: [main, ci-speedup]
         tags: ["v*.*.*"]
     
       repository_dispatch:
    @@ -130,7 +130,7 @@ jobs:
         environment: Testing
         runs-on: ubuntu-latest
         timeout-minutes: 120
    -    if: (startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main') || (needs.precheck.outputs.srcfileschanged == 'true' && github.event.pull_request.head.repo.full_name == github.repository)
    +    # if: (startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main') || (needs.precheck.outputs.srcfileschanged == 'true' && github.event.pull_request.head.repo.full_name == github.repository)
         defaults:
           run:
             working-directory: src
    @@ -188,7 +188,7 @@ jobs:
     
           # Upload the original go test output as an artifact for later review.
           - name: Upload test log
    -        if: failure()
    +        if: always()
             uses: actions/upload-artifact@v4
             with:
               name: ci-test-log
    @@ -302,6 +302,7 @@ jobs:
               CORSO_SECONDARY_M365_TEST_USER_ID: 'foo'
             run: |
               set -euo pipefail
    +          export GODEBUG=gocachehash=1
               go test       \
               -tags testing \
               -json         \
    @@ -310,11 +311,11 @@ jobs:
               -p 1          \
               -timeout 20m  \
               ./... \
    -          2>&1 | tee ./testlog/gotest-unit.log | gotestfmt -hide successful-tests
[...]

PR #5182, opened 2024-02-05
move M365 creds rotation logic to code

HiteshRepo:
moves M365 creds rotation logic to code

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)
#5080 

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,8 +177,8 @@ jobs:
       # run the tests
       - name: Integration Tests
         env:
-          AZURE_CLIENT_ID: ${{ secrets[env.AZURE_CLIENT_ID_NAME] }}
-          AZURE_CLIENT_SECRET: ${{ secrets[env.AZURE_CLIENT_SECRET_NAME] }}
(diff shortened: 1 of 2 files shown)


ryanfkeepers:
This is not a rotation, it's a randomization.  We currently provide consistency between runs (theoretically) by binding each run to an account determined by the run ID, here: https://github.com/alcionai/corso/blob/main/.github/workflows/accSelector.yaml

Are we sure we want to drop this behavior to use a randomizer instead?

HiteshRepo:
> This is not a rotation, it's a randomization. We currently provide consistency between runs (theoretically) by binding each run to an account determined by the run ID, here: https://github.com/alcionai/corso/blob/main/.github/workflows/accSelector.yaml
> 
> Are we sure we want to drop this behavior to use a randomizer instead?

No @ryanfkeepers I am not sure about the solution. The real intention is to not change environment variables across test runs. Because changing env-vars are leading to tests not being cached

PR #5184, opened 2024-02-05
add by-resp-code catcher to err transformer

ryanfkeepers:
Start catching graph responses by broad response code, in addition to the response body details.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/errors.go
+++ b/src/pkg/services/m365/api/graph/errors.go
@@ -117,6 +117,7 @@ func stackWithCoreErr(ctx context.Context, err error, traceDepth int) error {
 	}
 
 	ode := parseODataErr(err)
+	labels := []string{}
 
 	switch {
 	case isErrBadJWTToken(ode, err):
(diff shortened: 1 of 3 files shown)


src/pkg/services/m365/api/graph/errors.go
```diff
@@ -133,16 +133,38 @@ func stackWithCoreErr(ctx context.Context, err error, traceDepth int) error {
 		err = clues.Stack(core.ErrAlreadyExists, err)
 	case isErrNotFound(ode, err):
 		err = clues.Stack(core.ErrNotFound, err)
+	default:
+		err = toErrByRespCode(ode, err)
```
zackrossman, line 137: I like the logical flow of this method -- attempt to match to well-known error code (highly specific but brittle) and if no match, fall back to error code matching (less specific understanding of the error but still gives us a general sense of the root cause)

src/pkg/services/m365/api/graph/errors.go
```diff
@@ -133,16 +133,38 @@ func stackWithCoreErr(ctx context.Context, err error, traceDepth int) error {
...
 	return stackWithDepth(ctx, err, 1+traceDepth)
 }
 
 // unexported categorizers, for use with stackWithCoreErr
 
+// for categorizing errors by their response code and no other response propery
+func toErrByRespCode(ode oDataErr, err error) error {
+	fiveXX := []int{}
+	for i := 0; i < 5; i++ {
+		fiveXX = append(fiveXX, 500+i)
```
pandeyabs, line 149: nit: Having `fiveXX` as a const would be better here. Avoids new allocations.

src/pkg/services/m365/api/graph/errors.go
```diff
@@ -133,16 +133,38 @@ func stackWithCoreErr(ctx context.Context, err error, traceDepth int) error {
...
+	}
```
zackrossman, line 150: What about 
509 –Bandwidth Limit Exceeded
511 – Network Authentication Required

Rather than maintaining this list I'd suggest mapping all 5xx error codes to `core.ErrDownstreamServerError`. Would be a shame if we were to add this graceful handling only to miss one.

src/pkg/services/m365/api/graph/errors_test.go
```diff
@@ -991,3 +993,92 @@ func (suite *GraphErrorsUnitSuite) TestIsErrSharingDisabled() {
...
+			expectNoStack: false,
+			expectIs:      core.ErrInsufficientAuthorization,
+		},
+		{
+			name:          "403",
+			err:           graphTD.ODataErrWithStatus(http.StatusForbidden, "err"),
+			expectNoStack: false,
+			expectIs:      core.ErrInsufficientAuthorization,
+		},
+		{
```
pandeyabs, line 1040: nit: Add a test for 429 too, for completeness?

pandeyabs (approved):
Looks good!

zackrossman (approved):
One suggestion otherwise LGTM

src/pkg/services/m365/api/graph/errors.go
ryanfkeepers: I suppose this gets back to our need to root cause generic errors sometimes.  Fair enough.

src/pkg/services/m365/api/graph/errors_test.go
ryanfkeepers: good catch, done

src/pkg/services/m365/api/graph/errors.go
ryanfkeepers: Per recommendation, I ditched the slice and am now catching all 5xx errors.

PR #5194, opened 2024-02-08
Retry events item enumeration with smaller page size for some cases

ashmrtn:
Update the events item enumerator to switch to a smaller page size for
the delta pager if we fail enumeration with a 503 error and no content.
We've found that that situation is indicative of the Graph server
being slow and a smaller page size allows us to make progress still

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test (in another PR)
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/middleware.go
+++ b/src/pkg/services/m365/api/graph/middleware.go
@@ -164,7 +164,15 @@ func (mw RetryMiddleware) Intercept(
 		mw.isRetriableRespCode(ctx, resp)
 
 	if !retriable {
-		return resp, stackReq(ctx, req, resp, err).OrNil()
+		// Returning a response and an error causes output from either some part of
+		// the middleware/graph/golang or the mocking library used for testing.
+		// Return one or the other to avoid this.
+		err := stackReq(ctx, req, resp, err).OrNil()
+		if err != nil {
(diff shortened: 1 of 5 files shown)


src/pkg/services/m365/api/graph/middleware.go
```diff
@@ -164,7 +164,15 @@ func (mw RetryMiddleware) Intercept(
 		mw.isRetriableRespCode(ctx, resp)
 
 	if !retriable {
-		return resp, stackReq(ctx, req, resp, err).OrNil()
+		// Returning a response and an error causes output from either some part of
+		// the middleware/graph/golang or the mocking library used for testing.
+		// Return one or the other to avoid this.
+		err := stackReq(ctx, req, resp, err).OrNil()
+		if err != nil {
```
pandeyabs, line 171: Unrelated to this PR, we can simplify this function a bit. For successful requests(`err == nil`), we can return early after line 160. Currently we are using `!retriable` as a way to also catch successful requests which is not super ideal.

pandeyabs (approved):

PR #5195, opened 2024-02-08
Don't clobber middleware in mock client

ashmrtn:
The http mocking library we're using replaces the middleware of the
current transport if we just intercept the client. This switches things
to create a custom client with the proper set of middlewares instead so
that we get both mocking and the middleware

* ensure client has default parameters so things like default page size is set
* create a custom client so that we can both use the mock and have access to the middleware we normally add to the mock

More tuning of the custom client is necessary, for example, reducing the retries/time between retries for our middleware and the kiota middleware.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/service_mock.go
+++ b/src/pkg/services/m365/api/graph/service_mock.go
@@ -6,6 +6,7 @@ import (
 	"github.com/alcionai/clues"
 	"github.com/h2non/gock"
 	abstractions "github.com/microsoft/kiota-abstractions-go"
+	khttp "github.com/microsoft/kiota-http-go"
 	msgraphsdkgo "github.com/microsoftgraph/msgraph-sdk-go"
 
 	"github.com/alcionai/corso/src/pkg/account"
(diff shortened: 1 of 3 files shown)


src/pkg/services/m365/api/graph/service_mock.go
```diff
@@ -54,11 +58,27 @@ func CreateGockAdapter(
 	counter *count.Bus,
 	opts ...Option,
 ) (abstractions.RequestAdapter, error) {
-	httpClient, cc := KiotaHTTPClient(counter, opts...)
+	cc := populateConfig(opts...)
+	// We need to manufacture our own graph client since gock.InterceptClient
+	// replaces the transport which replaces all the middleware we add. Start with
```
pandeyabs, line 63: Oof

pandeyabs (approved):

PR #5198, opened 2024-02-09
add authentication to requester

ryanfkeepers:
the graph requester for large item downloads now includes the option to authenticate requests.  The option is configured at the time of creating the requester, therefore all requests using that servier are either authenticatd or not. In our case, we're opting to authenticate all requests, since we do not use this requester for non-graph api calls, and even if we did the addition of auth headers is likely benign.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :green_heart: E2E


src/pkg/services/m365/api/client.go
```diff
@@ -123,12 +123,18 @@ func newLargeItemService(
...
-		return nil, clues.Wrap(err, "generating no-timeout graph adapter")
+		return nil, clues.Wrap(err, "generating azure authorizer")
 	}
 
-	return a, nil
+	a, err := NewService(
+		creds,
+		counter,
+		graph.NoTimeout(),
+		graph.AuthorizeRequester(azureAuth))
```
pandeyabs, line 135: Wondering why we need this change for the large item. This is a graph client so it should already have auth built in.

pandeyabs (approved):
Looks awesome!

src/pkg/services/m365/api/client.go
pandeyabs: @ryanfkeepers Could you please address this comment? Thanks!

pandeyabs: Just noticed that this change was removed in the next PR in the chain. https://github.com/alcionai/corso/pull/5200/files#diff-ea424cd36ba63e8261c19fa218f00c5e42bc30e874b2dd5a5eb09d8feb20574bR131 

So we are good.

PR #5199, opened 2024-02-09
allow eml export when attachments have no name

ryanfkeepers:

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/converters/eml/eml_test.go
+++ b/src/internal/converters/eml/eml_test.go
@@ -162,6 +162,18 @@ func (suite *EMLUnitSuite) TestConvert_edge_cases() {
 				require.NoError(suite.T(), err, "setting attachment content")
 			},
 		},
+		{
+			name: "attachment without name",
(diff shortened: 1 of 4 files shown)


pandeyabs (approved):

src/internal/converters/eml/eml_test.go
```diff
@@ -162,6 +162,18 @@ func (suite *EMLUnitSuite) TestConvert_edge_cases() {
 				require.NoError(suite.T(), err, "setting attachment content")
 			},
 		},
+		{
+			name: "attachment without name",
```
ashmrtn, line 166: to gain extra confidence could we expand the tests here a little bit to cover:
* attachment with `GetName() == nil`
* multiple attachments with no name

ryanfkeepers: sure, I can do that.

PR #5200, opened 2024-02-09
use contentt endopint for files larger than 50gb

ryanfkeepers:
avoid the url cache and get the url directly from graph when downloading very large files to avoid failing on token timeouts from downloads that take longer than an hour.

---

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test

### DIFF
--- a/src/internal/m365/collection/drive/item.go
+++ b/src/internal/m365/collection/drive/item.go
@@ -24,7 +24,7 @@ const (
 	acceptHeaderKey        = "Accept"
 	acceptHeaderValue      = "*/*"
 	gigabyte               = 1024 * 1024 * 1024
-	largeFileDownloadLimit = 50 * gigabyte
+	largeFileDownloadLimit = 15 * gigabyte
 )
 
 // downloadUrlKeys is used to find the download URL in a DriveItem response.
@@ -48,7 +48,7 @@ func downloadItem(
 		// the download could take longer than the lifespan of the download token in the cached
 		// url, which will cause us to timeout on every download request, even if we refresh the
 		// download url right before the query.
-		url    = "https://graph.microsoft.com/v1.0/" + driveID + "/items/" + ptr.Val(item.GetId()) + "/content"
+		url    = "https://graph.microsoft.com/v1.0/drives/" + driveID + "/items/" + ptr.Val(item.GetId()) + "/content"
 		reader io.ReadCloser
 		err    error
 	)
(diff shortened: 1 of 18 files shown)


src/internal/m365/collection/drive/item.go
```diff
@@ -33,39 +35,39 @@ var downloadURLKeys = []string{
...
+	}
 
-		rc, err = downloadFile(ctx, ag, url)
-		if err != nil {
-			return nil, clues.Stack(err)
-		}
+	// smaller files will maintain our current behavior (prefetching the download url with the
+	// url cache).  That pattern works for us in general, and we only need to deviate for very
+	// large file sizes.
+	if ptr.Val(item.GetSize()) > largeFileDownloadLimit {
```
vkamra, line 64: Should this be < `largeFileDownloadLimit`?

src/internal/m365/collection/drive/item.go
```diff
@@ -21,8 +21,10 @@ import (
 )
 
 const (
-	acceptHeaderKey   = "Accept"
-	acceptHeaderValue = "*/*"
+	acceptHeaderKey        = "Accept"
+	acceptHeaderValue      = "*/*"
+	gigabyte               = 1024 * 1024 * 1024
+	largeFileDownloadLimit = 50 * gigabyte
```
vkamra, line 27: We should lower this to ~15GB

pandeyabs (commented):
Changes look good other than what Vaibhav has pointed out. 

Side note, I am a little wary about sending auth token for every request that our requester client makes. We only need this for `/content` request.

If we could have a conditional way to do this, say with an argument to `Request(..., authenticate bool)` function, that would be better. 

Mostly I am worried about the fact that the auth token is not being sent to the intended graph endpoint. I don't have a good understanding of azure auth tokens, so I can't say with a guarantee that these tokens won't be misused or replayed or such. 

What do folks think?

src/internal/m365/collection/drive/item.go
ryanfkeepers: whoops, yes, mixed that up from undoing manual testing.

src/internal/m365/collection/drive/item.go
ryanfkeepers: can do

pandeyabs (approved):
Looks good! Should we add a changelog?

ryanfkeepers:
> Should we add a changelog

Oops!  I'll follow up with that.

PR #5202, opened 2024-02-10
Retry more errors in graph adapter wrapper

pandeyabs:
* We have started seeing `io.ErrUnexpectedEOF` and `read: connection timed out` in the last 2 days for exchange backups. Retry those.
* Also increase retry count from 3 to 6. This is more of a hail mary to retry `InvalidAuthenticationToken` errors. We have observed that retries do help. But for a small set of requests, we end up exhausting retries and eventually fail with `InvalidAuthenticationToken` error. Hoping that bumping this to 6 will get us some relief. This fix may be removed if we find the rootcause/pattern behind this. 
---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [ ] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/service.go
+++ b/src/pkg/services/m365/api/graph/service.go
@@ -382,6 +382,7 @@ func wrapAdapter(gra *msgraphsdkgo.GraphRequestAdapter, cc *clientConfig) *adapt
 	}
 }
 
+// Graph may abruptly close connections, which we should retry.
 var connectionEnded = filters.In([]string{
 	"connection reset by peer",
 	"client connection force closed",
--- a/src/pkg/services/m365/api/graph/service_test.go
+++ b/src/pkg/services/m365/api/graph/service_test.go
@@ -236,13 +236,13 @@ func (suite *GraphIntgSuite) TestAdapterWrap_retriesConnectionInterruptions() {
 		},
 		{
 			name:             "connection reset by peer",
-			providedErr:      clues.New("connection reset by peer"),
+			providedErr:      clues.New("connection reset by peer what is love"),
 			expectRetryCount: 7,
 			expectErr:        assert.Error,
 		},
 		{
 			name:             "read: connection timed out",
-			providedErr:      clues.New("read: connection timed out"),
+			providedErr:      clues.New("read: connection timed out baby don't hurt me"),
 			expectRetryCount: 7,
 			expectErr:        assert.Error,
 		},
@@ -260,7 +260,7 @@ func (suite *GraphIntgSuite) TestAdapterWrap_retriesConnectionInterruptions() {
 		},
 		{
 			name:             "non retriable error",
-			providedErr:      clues.New("non retriable error"),
+			providedErr:      clues.New("no more"),
 			expectRetryCount: 1,
 			expectErr:        assert.Error,
 		},


vkamra (approved):

src/pkg/services/m365/api/graph/service.go
```diff
@@ -416,7 +420,9 @@ func (aw *adapterWrap) Send(
 		err = stackWithCoreErr(ictx, err, 1)
 		e = err
 
-		if IsErrConnectionReset(err) || connectionEnded.Compare(err.Error()) {
+		if IsErrConnectionReset(err) ||
+			connectionEnded.Compare(err.Error()) ||
+			errors.Is(err, io.ErrUnexpectedEOF) {
```
ashmrtn, line 425: just double checking, this was an unexpected EOF error when sending the request, not when reading item data in kopia right? Not sure if you recall, but the latter would need something like the retry reader wrapper that we have

ashmrtn (approved):

src/pkg/services/m365/api/graph/service.go
pandeyabs: This is while sending the request. e.g. we are seeing these errors while doing `GET /messages$expand=microsoft.graph.itemattachment%2Fitem`. Similarly seen for `getting mail attachment by id`.

Also, since we don't have KAI for exchange, all data reads are prefetched.

ashmrtn: >  since we don't have KAI for exchange, all data reads are prefetched.

🤦 excellent point!

src/pkg/services/m365/api/graph/service.go
```diff
@@ -376,6 +379,7 @@ func wrapAdapter(gra *msgraphsdkgo.GraphRequestAdapter, cc *clientConfig) *adapt
 var connectionEnded = filters.Contains([]string{
```
ryanfkeepers, line 379: Wouldn't mind a comment on this var explaining its purpose.

src/pkg/services/m365/api/graph/service.go
```diff
@@ -416,7 +420,9 @@ func (aw *adapterWrap) Send(
 		err = stackWithCoreErr(ictx, err, 1)
 		e = err
 
-		if IsErrConnectionReset(err) || connectionEnded.Compare(err.Error()) {
+		if IsErrConnectionReset(err) ||
+			connectionEnded.Compare(err.Error()) ||
```
ryanfkeepers, line 424: hmm, graph/errors.go should have a pattern to construct these tests internally.  Can you add a todo that we need to migrate them there?  it feels a little hacky with the current implementation.

ryanfkeepers (approved):

src/pkg/services/m365/api/graph/service_test.go
```diff
@@ -221,43 +221,92 @@ func (suite *GraphIntgSuite) TestAdapterWrap_catchesPanic() {
...
+			expectErr:        assert.Error,
+		},
+		{
+			name:             "connection reset by peer",
+			providedErr:      clues.New("connection reset by peer"),
+			expectRetryCount: 7,
+			expectErr:        assert.Error,
+		},
+		{
+			name:             "read: connection timed out",
```
ashmrtn, line 244: could you please make another test like this but with additional input in the error message? I'm hoping it'll help catch accidental misuse of the filters. For example `clues.New("read: connection timed out: host foo")` or something

ashmrtn (approved):

src/pkg/services/m365/api/graph/service_test.go
pandeyabs: yes! haha, i was just typing this out. The failed request appends a bunch of extra text to it anyway, but better to have explicit test!

This is how i found out that `filters.In` is what we need instead of `filters.Compare`.

PR #5204, opened 2024-02-10
add an authRequried flag for requester requests

ryanfkeepers:
this allows us to specify auth necessity on a per-call basis, instead of flooding every call with auth creds.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/drive/item.go
+++ b/src/internal/m365/collection/drive/item.go
@@ -65,14 +65,15 @@ func downloadItem(
 		url = str.FirstIn(item.GetAdditionalData(), downloadURLKeys...)
 	}
 
-	reader, err = downloadFile(ctx, getter, url)
+	reader, err = downloadFile(ctx, getter, url, ptr.Val(item.GetSize()) < largeFileDownloadLimit)
 
 	return reader, clues.StackWC(ctx, err).OrNil()
 }
 
 type downloadWithRetries struct {
-	getter api.Getter
-	url    string
+	getter      api.Getter
+	requireAuth bool
+	url         string
 }
 
 func (dg *downloadWithRetries) SupportsRange() bool {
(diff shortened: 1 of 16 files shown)


src/internal/m365/collection/drive/item.go
```diff
@@ -65,14 +65,15 @@ func downloadItem(
 		url = str.FirstIn(item.GetAdditionalData(), downloadURLKeys...)
 	}
 
-	reader, err = downloadFile(ctx, getter, url)
+	reader, err = downloadFile(ctx, getter, url, ptr.Val(item.GetSize()) < largeFileDownloadLimit)
```
pandeyabs, line 68: Can we add a regression test with a mocked getter or similar which tests expected behavior i.e.
1. Small files( < `largeFileDownloadLimit`)  are downloaded using the download url. Large files use `/content` endpoint. 
2. Auth flag is specified for large files, and not for small files. 

Otherwise, it'd be hard to detect regressions until we hit them in prod, since we don't test with such large files typically.

src/pkg/services/m365/api/graph/http_wrapper.go
```diff
@@ -111,7 +113,11 @@ func (hw httpWrapper) Request(
 	// See https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-decorate-your-http-traffic
 	req.Header.Set("User-Agent", "ISV|Alcion|Corso/"+version.Version)
 
-	if hw.config.requesterAuth != nil {
+	if requireAuth {
+		if hw.config.requesterAuth == nil {
+			return nil, clues.Wrap(err, "http wrapper misconfigured: missing required authorization middleware")
```
pandeyabs, line 118: middleware?

pandeyabs (approved):
Approving to unblock. We should add a regression test.

src/internal/m365/collection/drive/item.go
ryanfkeepers: not a bad idea.  added that.

PR #5207, opened 2024-02-10
Do not backup shared calendars

vkamra:
Skip backup of shared calendars. These will be backed up with the resource that owns the calendar.

---

- [ ] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Issue(s)

* #<issue>

#### Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/container_resolver_test.go
+++ b/src/internal/m365/collection/exchange/container_resolver_test.go
@@ -3,11 +3,13 @@ package exchange
 import (
 	"context"
 	"fmt"
+	"hash/crc32"
 	stdpath "path"
 	"testing"
 
 	"github.com/alcionai/clues"
 	"github.com/google/uuid"
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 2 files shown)


vkamra:
Test plan: I have verified this manually. Will follow up with a mocked container cache test. We don't have one for the event container cache - so this will take me a bit to cobble up.

src/internal/m365/collection/exchange/events_container_cache.go
```diff
@@ -89,11 +90,33 @@ func (ecc *eventContainerCache) Populate(
 		return clues.WrapWC(ctx, err, "enumerating containers")
 	}
 
+	var defaultCalendarOwner string
+
+	// Determine the owner for the default calendar. We'll use this to detect and
+	// skip shared calendars that are not owned by this user.
+	for _, c := range containers {
+		if ptr.Val(c.GetIsDefaultCalendar()) {
+			defaultCalendarOwner = ptr.Val(c.GetOwner().GetAddress())
```
pandeyabs, line 99: We can break out of the loop if we found the `defaultCalendarOwner`.

pandeyabs (commented):
Changes look good!

src/internal/m365/collection/exchange/container_resolver_test.go
```diff
@@ -1017,6 +1019,210 @@ func (suite *ConfiguredFolderCacheUnitSuite) TestAddToCache() {
...
+	}
+
+	table := []struct {
+		name                string
+		enumerateContainers func() []models.Calendarable
+		expectErr           assert.ErrorAssertionFunc
+		assertFunc          func(t *testing.T, ecc *eventContainerCache)
+	}{
+		{
+			name: "one default calendar, one shared",
```
ashmrtn, line 1122: excellent set of test cases! To slightly expand this, could you please add two other tests with the following:
* default calendar has owner info set to nil and there's also a calendar the user owns and a shared calendar
* default calendar has owner info but fields in it are empty and there's also a calendar the user owns and a shared calendar

These tests will make sure we fallback correctly if there's some missing info in the default (I think it should also be ok to backup the shared calendars in this case since we're missing info). It'll also help make sure we don't have NPEs or anything as the code changes over time. Since this set of tests always sets up things to return a properly populated default calendar it may be easiest to put them in a new Test function

ashmrtn (approved):

ashmrtn:
oh, also could you please add a line to the changelog stating shared calendars will only be backed up with the user who created the calendar?

src/internal/m365/collection/exchange/container_resolver_test.go
pandeyabs: Thanks @ashmrtn! I'll add these extra tests in a followup pr. issue tracking this https://github.com/alcionai/corso/issues/5208

PR #5217, opened 2024-02-12
Check for and retry 404s with no content

ashmrtn:
We've started seeing 404 errors with no content being returned. Check for these in the http wrapper we use and retry them.

While graph SDK returns an error for this sort of situation it's a very basic error since it normally expects to parse info out of the response body. Therefore it should be safe to inject our own error that we can check for.

---

- [x] :no_entry: No

#### Type of change

- [ ] :sunflower: Feature
- [x] :bug: Bugfix

#### Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/src/pkg/services/m365/api/graph/service.go
+++ b/src/pkg/services/m365/api/graph/service.go
@@ -442,6 +442,13 @@ func (aw *adapterWrap) Send(
 			// to limit the scope of this fix.
 			logger.Ctx(ictx).Debug("invalid request")
 			events.Inc(events.APICall, "invalidgetrequest")
+		} else if errors.Is(err, ErrNotFoundEmptyResp) {
+			// We've started seeing 404s with no content being returned for messages
+			// message attachments, and events. Attempting to manually fetch the items
+			// succeeds. Therefore we want to retry these to see if we can work around
+			// the problem.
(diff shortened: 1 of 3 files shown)


pandeyabs (approved):
🚢

ryanfkeepers (approved):

src/pkg/services/m365/api/graph/service.go
```diff
@@ -442,6 +442,13 @@ func (aw *adapterWrap) Send(
 			// to limit the scope of this fix.
 			logger.Ctx(ictx).Debug("invalid request")
 			events.Inc(events.APICall, "invalidgetrequest")
+		} else if errors.Is(err, ErrNotFoundEmptyResp) {
```
pandeyabs, line 445: Sorry, this occurred to me just now. Should we restrict the retries to `GET` requests only. This will ensure we don't retry on `POST` 404s with empty content., as those can be non-idempotent. 

```suggestion
		} else if requestInfo.Method.String() == http.MethodGet && errors.Is(err, ErrNotFoundEmptyResp) {
```

ashmrtn: good callout!

PR #5225, opened 2024-02-13
allow skipping error cases in exchange backup

ryanfkeepers:
utilize the previously introduced canSkipItemFailure interface to create skip records for items during backup.

---

- [x] :no_entry: No

#### Type of change

- [x] :bug: Bugfix

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/collection.go
+++ b/src/internal/m365/collection/exchange/collection.go
@@ -19,6 +19,7 @@ import (
 	"github.com/alcionai/corso/src/internal/m365/support"
 	"github.com/alcionai/corso/src/internal/observe"
 	"github.com/alcionai/corso/src/pkg/backup/details"
+	"github.com/alcionai/corso/src/pkg/control"
 	"github.com/alcionai/corso/src/pkg/count"
 	"github.com/alcionai/corso/src/pkg/errs/core"
 	"github.com/alcionai/corso/src/pkg/fault"
@@ -68,21 +69,21 @@ func getItemAndInfo(
 	ctx context.Context,
 	getter itemGetterSerializer,
 	userID string,
-	id string,
+	itemID string,
 	useImmutableIDs bool,
 	parentPath string,
 ) ([]byte, *details.ExchangeInfo, error) {
 	item, info, err := getter.GetItem(
 		ctx,
 		userID,
-		id,
+		itemID,
 		fault.New(true)) // temporary way to force a failFast error
 	if err != nil {
 		return nil, nil, clues.WrapWC(ctx, err, "fetching item").
 			Label(fault.LabelForceNoBackupCreation)
 	}
 
-	itemData, err := getter.Serialize(ctx, item, userID, id)
+	itemData, err := getter.Serialize(ctx, item, userID, itemID)
 	if err != nil {
 		return nil, nil, clues.WrapWC(ctx, err, "serializing item")
 	}
@@ -108,6 +109,7 @@ func NewCollection(
 	bc data.BaseCollection,
 	user string,
 	items itemGetterSerializer,
+	canSkipFailChecker canSkipItemFailurer,
 	origAdded map[string]time.Time,
 	origRemoved []string,
 	validModTimes bool,
@@ -140,6 +142,7 @@ func NewCollection(
 			added:          added,
 			removed:        removed,
 			getter:         items,
+			skipChecker:    canSkipFailChecker,
 			statusUpdater:  statusUpdater,
 		}
 	}
@@ -150,6 +153,7 @@ func NewCollection(
 		added:          added,
 		removed:        removed,
 		getter:         items,
+		skipChecker:    canSkipFailChecker,
 		statusUpdater:  statusUpdater,
 		counter:        counter,
 	}
@@ -167,7 +171,8 @@ type prefetchCollection struct {
 	// removed is a list of item IDs that were deleted from, or moved out, of a container
 	removed map[string]struct{}
 
-	getter itemGetterSerializer
+	getter      itemGetterSerializer
+	skipChecker canSkipItemFailurer
 
 	statusUpdater support.StatusUpdater
 }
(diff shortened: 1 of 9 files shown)


src/internal/m365/collection/exchange/collection.go
```diff
@@ -459,10 +481,13 @@ func (col *lazyFetchCollection) streamItems(
 			&lazyItemGetter{
 				userID:       user,
 				itemID:       id,
+				category:     col.FullPath().Category(),
```
ashmrtn, line 484: nit: please use `col.Category()` since it handles the case where `FullPath() == nil`

src/internal/m365/collection/exchange/collection.go
```diff
@@ -277,8 +283,24 @@ func (col *prefetchCollection) streamItems(
 				col.Opts().ToggleFeatures.ExchangeImmutableIDs,
 				parentPath)
 			if err != nil {
+				// pulled outside the switch due to multiple return values.
+				cause, canSkip := col.skipChecker.CanSkipItemFailure(
+					err,
+					user,
+					id,
+					col.BaseCollection.Opts())
```
ashmrtn, line 291: just `col.Opts()` should work
```suggestion
					col.Opts())
```

src/internal/m365/collection/exchange/collection.go
```diff
@@ -459,10 +481,13 @@ func (col *lazyFetchCollection) streamItems(
 			&lazyItemGetter{
 				userID:       user,
 				itemID:       id,
+				category:     col.FullPath().Category(),
 				getter:       col.getter,
 				modTime:      modTime,
 				immutableIDs: col.Opts().ToggleFeatures.ExchangeImmutableIDs,
 				parentPath:   parentPath,
+				skipChecker:  col.skipChecker,
+				opts:         col.BaseCollection.Opts(),
```
ashmrtn, line 490: could you please simplify this as well
```suggestion
				opts:         col.Opts(),
```

src/internal/m365/collection/exchange/collection_test.go
```diff
@@ -589,6 +713,154 @@ func (suite *CollectionUnitSuite) TestLazyFetchCollection_Items_LazyFetch() {
...
+		},
+		{
+			name: "only added items",
+			added: map[string]time.Time{
+				"fisher":    start.Add(time.Minute),
+				"flannigan": start.Add(2 * time.Minute),
+				"fitzbog":   start.Add(3 * time.Minute),
+			},
+			expectItemCount:    3,
+			expectSkippedCount: 2,
```
ashmrtn, line 748: could you please explain why only two items skipped in this test if all of them look like they're going to have serialization errors?

src/internal/m365/collection/exchange/collection.go
```diff
@@ -108,6 +109,7 @@ func NewCollection(
 	bc data.BaseCollection,
 	user string,
 	items itemGetterSerializer,
+	canSkipFailChecker canSkipItemFailurer,
```
ashmrtn, line 112: looking at how this is used I'm not really a fan of passing in two separate interfaces, especially when the same thing implements both interfaces for production code. It seems like it would be more natural to compose the structs to implement a single interface that has everything. This patch already contains mocks that could be placed inside a backup handler. Adding an interface would also allow using a common struct for things that never skip items

The interface would look a bit like the following:
```go
type backupHandlerImpl struct {
  client api.Client
  skipChecker canSkipItemFailurer
}
```

Normally this field would be populated with one of the checkers that's implemented in the previous PRs. Tests could override this by replacing the value in that field

ashmrtn (approved):

ashmrtn:
please also add at least one test case for each data category with the production error skip configuration showing that skips are used when expected. That will help ensure that things are configured properly for prod. Otherwise it's possible to prove that each component works but prod breaks because the configuration is incorrect in some way

src/internal/m365/collection/exchange/collection.go
ashmrtn: edit: actually embedding the skip checker (or having it in the backup handler struct) probably isn't required since there also needs to be tests with the configuration that's used in releases. Since we need the release config there shouldn't be a need to override the checker which means the mock checkers can probably be removed

ryanfkeepers: Yar.  That's how we have the other handler interfaces designed.  The exchange version was more of our proof of concept, and so it does some weird things with providing the api.Client to fulfill interfaces instead of fulfilling the interfaces itself by using the api.client under the hood.  It'd be nice to bring that handler design in line with the others, but that's out of scope for these changes.  And, honestly, low priority.  I'm in agreement that the multiple interfaces are clunky, but refactoring the handlers to satisfy that concern isn't worth the lift right now.

src/internal/m365/collection/exchange/collection_test.go
ryanfkeepers: I genuinely don't know.  But this is a copy of the other test, and that's the case there as well.  So this is stable...?

ryanfkeepers: Oh wait, figured it out.  It's because the "expected reads" slice defines which items get pushed through the lookup mock (and thus which ones produce an error).  And for some reason we're only reading 2 of 3?  Weird test.  I added the third one to it.

src/internal/m365/collection/exchange/collection_test.go
```diff
@@ -333,6 +336,221 @@ func (suite *CollectionUnitSuite) TestPrefetchCollection_Items() {
...
+		name               string
+		category           path.CategoryType
+		handler            backupHandler
+		added              map[string]time.Time
+		removed            map[string]struct{}
+		expectItemCount    int
+		expectSkippedCount int
+	}{
+		{
+			name:     "no items",
```
ashmrtn, line 355: nit: since other tests are prefixed with the service name could you please prefix this one as well?
```suggestion
			name:     "events no items",
```

ashmrtn (approved):
Thanks for updating tests!

PR #5226, opened 2024-02-14
isolate the actual error from this query

ryanfkeepers:
I couldn't the exact error, so this is the best approximation i could come up with.

---

- [x] :no_entry: No

#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/events_backup.go
+++ b/src/internal/m365/collection/exchange/events_backup.go
@@ -74,6 +74,11 @@ func (h eventBackupHandler) CanSkipItemFailure(
 		return "", false
 	}
 
+	// this is a bit overly cautious.  we do know that we get 503s with empty response bodies
+	// due to fauilures when getting too many instances.  We don't know for sure if we get
+	// generic, well formed 503s.  But since we're working with specific resources and item
+	// IDs in the first place, that extra caution will help make sure an unexpected error dosn't
+	// slip through the cracks on us.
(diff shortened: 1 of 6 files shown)


src/internal/m365/collection/exchange/events_backup.go
```diff
@@ -60,13 +61,21 @@ func (h eventBackupHandler) NewContainerCache(
...
 	opts control.Options,
 ) (fault.SkipCause, bool) {
-	// yes, this is intentionally a todo.  I'll get back to it.
-	if !errors.Is(err, clues.New("todo fix me")) {
+	if err == nil {
+		return "", false
+	}
+
+	if !errors.Is(err, graph.ErrServiceUnavailableEmptyResp) &&
+		!clues.HasLabel(err, graph.LabelStatus(http.StatusServiceUnavailable)) {
```
ashmrtn, line 78: do we want to match on labels as well as no content errors? Can we look at the info we have to determine if these are different failure modes?

ashmrtn (approved):
some discussion needed around if we should skip on regular 503s as well as no content 503s

src/internal/m365/collection/exchange/events_backup.go
ryanfkeepers: settled in sync: we're going to keep with the both error checks (no-content and also broad 503) since appropriate mechanisms should be in place for identification and vetting.

PR #5228, opened 2024-02-14
only pass in map to skip checker

ryanfkeepers:
#### Type of change

- [x] :sunflower: Feature

#### Test Plan

- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/src/internal/m365/collection/exchange/contacts_backup_test.go
+++ b/src/internal/m365/collection/exchange/contacts_backup_test.go
@@ -41,15 +41,15 @@ func (suite *ContactsBackupHandlerUnitSuite) TestHandler_CanSkipItemFailure() {
 			expect: assert.False,
 		},
 		{
-			name: "empty skip on 503",
-			err:  nil,
+			name: "false when map is empty",
+			err:  assert.AnError,
 			opts: control.Options{
 				SkipTheseEventsOnInstance503: map[string][]string{},
 			},
 			expect: assert.False,
 		},
 		{
-			name: "nil error",
+			name: "false on nil error",
 			err:  nil,
 			opts: control.Options{
 				SkipTheseEventsOnInstance503: map[string][]string{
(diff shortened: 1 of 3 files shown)


src/internal/m365/collection/exchange/collection.go
```diff
@@ -288,7 +288,7 @@ func (col *prefetchCollection) streamItems(
 					err,
 					user,
 					id,
-					col.Opts())
+					col.Opts().SkipTheseEventsOnInstance503)
```
ashmrtn, line 291: 🤔 looking at this, I'm not certain this is the pattern we want either. It does decouple the options struct from checking for the error, but does cause create another problem in the sense that we're unable to switch between which set of items to check. This means that if we do extend the skip check to skip another data category like contacts we'll require a larger set of changes. It also requires even more test validation because we need to be absolutely certain that the error check functions don't just return true if the error matches and the item is in the input map (since we no longer scope the input map by service) (also, those tests for contacts and mail should be updated to have the 503 error so that we're actually checking that)

If we want to remove passing `control.Options` as a parameter for the skip checker given the current design of the system, I think perhaps a less error prone implementation pattern would be to extract the skip set from options when the backup handler is created and store it in the handler (i.e. it's the backup handler's own version of a config). For mail and contacts that input would be an empty map or nil. For events it would be `SkipTheseEventsOnInstance503`

The above is a larger refactor though so I'm not sure where it falls on the range of things. I'd say either passing in the options struct like it was originally (even though it's not the best either) or something similar to the proposal in this comment so we try to avoid unexpected behavior down the road

ryanfkeepers: > This means that if we do extend the skip check to skip another data category like contacts we'll require a larger set of changes

Yeah, that was why I originally passed in the options.  But future potential is not now, and right now we're only using the one case.

> The above is a larger refactor though so I'm not sure where it falls on the range of things

I agree that the above would be better.  But that's beginning to optimize design for a future which makes heavier leverage of this current behavior.  That's a future I'd like to avoid, if possible.  The refactor is still a two-way-door here, and we can return to it if necessary.  For now I'll revert it to passing in options.

